PR #9, opened 2016-04-29
Add listener for new blocks that doesn't add to stacks

tmickel:


src/index.js
```diff
@@ -57,6 +57,27 @@ function VirtualMachine () {
             break;
         }
     };
+
+    instance.flyoutBlockListener = function (e) {
```
thisandagain, line 61: Lots of boilerplate, but ok for now.

PR #12, opened 2016-05-02
Allow the Sequencer to execute a JS function for each block

cwillisf:
There are two stub block packages, 'wedo2' and 'scratch3', which together cover all of the blocks currently available in the LLK/scratch-android prototype.


.eslintrc
```diff
@@ -7,7 +7,8 @@
         "linebreak-style": [2, "unix"],
         "max-len": [2, 80, 4],
         "semi": [2, "always"],
-        "strict": [2, "never"]
+        "strict": [2, "never"],
+        "no-console": [2, {"allow": ["log", "warn", "error"]}]
```
cwillisf, line 11: Is everyone OK with this change?
@tmickel @thisandagain

thisandagain, line 11: I think that's fine, we'll need to polyfill for older IE (which we'll likely need to do anyway).

PR #19, opened 2016-05-03
Stacks management, stack frames, repeat, forever

tmickel:
@thisandagain 

https://www.dropbox.com/s/kr9kef2g8hadwdm/repeats.mov?dl=0


src/engine/runtime.js
```diff
@@ -24,6 +24,12 @@ function Runtime () {
     this.blocks = {};
 
     /**
+     * Primitive-accessible execution metadata for each block.
+     * @type {Object.<string, Object>}
+     */
+    this.blockExecutionData = {};
```
tmickel: ah - need to take this out

src/engine/sequencer.js
```diff
@@ -59,8 +59,18 @@ Sequencer.prototype.stepThreads = function (threads) {
...
             if (activeThread.nextBlock === null &&
                 activeThread.status === Thread.STATUS_DONE) {
-                // Finished with this thread - tell the runtime to clean it up.
-                inactiveThreads.push(activeThread);
+                // First attempt to pop from the stack
+                if (activeThread.stack.length > 0
+                       && activeThread.nextBlock === null) {
+                    activeThread.nextBlock = activeThread.stack.pop();
+                    // Don't pop stack frame - we need the data.
+                    // A new one won't be created when we execute.
```
cwillisf: Could you explain in more detail why you don't pop the stack frame?

cwillisf: I see the part below about pushing a new frame unless we just popped, but that makes me nervous about stale data.
Is this a way to make data persist despite a loop getting pushed & popped each time the thread takes a step?

tmickel: yes, that's the exact issue (the loop getting pushed & popped each time the thread takes a step). This way the loopCounter can persist...perhaps we should push to the stack before we enter stepThreads, but that also seems messy. I think Scratch does it by not pushing to the stack unless we actually are entering a substack, which I'd also possibly be fine with?

vm.js
```diff
@@ -1623,8 +1632,16 @@
 	            }
 	            if (activeThread.nextBlock === null &&
 	                activeThread.status === Thread.STATUS_DONE) {
-	                // Finished with this thread - tell the runtime to clean it up.
-	                inactiveThreads.push(activeThread);
+	                // First attempt to pop from the stack
+	                if (activeThread.stack.length > 0
+	                       && activeThread.nextBlock === null) {
```
cwillisf: Either this `activeThread.nextBlock === null` check is redundant, or the one above is.
Actually, it looks to me like a pop here doesn't cause the thread to be pushed into `newThreads`... shouldn't it?

tmickel: yes, both good points - thank you!

PR #36, opened 2016-05-05
Switch LED block to use WeDo color index mode

thisandagain:
Depends on https://github.com/LLK/scratch-android/pull/39


src/blocks/wedo2.js
```diff
@@ -155,32 +106,35 @@ WeDo2Blocks.prototype.motorSpeed = function(argValues) {
...
+        'magenta': 1,
+        'purple': 2,
+        'blue': 3,
+        'green': 6,
+        'white': 10
+    };
+
     if (colorName == 'mystery') {
-        return this._HSVToRGB(Math.random() * 360, 1, 1);
+        return Math.floor((Math.random() * 10) + 1);
```
tmickel, line 128: Just would note that this doesn't return a selection from the above colors (the above colors don't include 4 or 5). I'm ok with this if those colors are working.

thisandagain, line 128: Yup. Those are valid colors just not colors in our palette (namely a red/orange and a blue/green).

PR #48, opened 2016-06-01
Add framework for a demo/dev playground; show auto-updating block representation.

tmickel:
Screenshot:
screen shot 2016-06-01 at 10 20 48 am

I think this will prove to be a nice tool as you can immediately see the representation change as you're snapping blocks together, apart, etc. Eventually it would be good to add more tools to this page to show the running threads, data in memory, etc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   },
   "devDependencies": {
     "eslint": "2.7.0",
-    "highlightjs": "^8.7.0",
+    "highlightjs": "8.7.0",
     "json-loader": "0.5.4",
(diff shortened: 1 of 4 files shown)


tmickel:
Added an update to highlight the JSON representation using highlight.js:
screen shot 2016-06-01 at 10 47 52 am

package.json
```diff
@@ -19,7 +19,9 @@
   },
   "devDependencies": {
     "eslint": "2.7.0",
+    "highlightjs": "^8.7.0",
```
thisandagain: Generally we use absolute version numbers for all dependencies.

thisandagain:
This looks great. One minor nitpick but otherwise 👍

PR #53, opened 2016-06-01
Add stack highlighting to playground blocks.

tmickel:
Hooks up the VM's emitted stack glow events to the playground Scratch Blocks instance's blocks. Lets us see which stacks are running currently in the VM :)


thisandagain:
Simple enough. Thanks @tmickel

Issue #1, opened 2016-04-26
Create Event - Handle blocks with a "next" connection


thisandagain:
#### Use Case
- Assemble a stack of multiple blocks
- Drag to trash
- Undo

tmickel:
The current behavior when doing the steps above is a TypeError:

> vm.js:2361 Uncaught TypeError: Cannot read property 'attribs' of undefined

Issue #51, opened 2016-06-01
Blocks are sometimes not destroyed properly from representation tree 


tmickel:
This seems to happen mostly when you drag a reporter into a shadow block slot.

To reproduce:
-Create a "repeat" block.
-Create an addition block.
-Drag the addition block into the repeat block's shadow space.
-Delete the repeat.

The addition block is still present in the VM's representation.

tmickel:
Related: on step three of this reproduction (dragging addition into repeat's shadow), an error occurs and the tree appears corrupted:

> vm.js:2350 Uncaught TypeError: Cannot read property 'id' of undefined

Issue #50, opened 2016-06-01
Support multiple fields in block representation tree.


tmickel:
Currently, only one field/shadow block is present in the VM's representation. For example, create an addition block. The left field is shown in the tree; the right one is not there.

PR #57, opened 2016-06-06
Update for VM block management

tmickel:
A fairly big update :)
- Proposed restructure of internal block representation for VM. Differences:
  - Separation of fields (which contain values) and inputs (pointers to other blocks) to match more closely with Blockly.
  - Inclusion of a `topLevel` property in the block tree.
  - No more linking blocks to other blocks in the tree. They are referenced by ID.
- A recursive/full implementation of adapter.js, which supports blocks with next blocks, loading multiple blocks in one event, blocks with multiple inputs, etc. I believe/hope that it should now support everything we need for the vertical Scratch grammar.
- A refactor of most block management functions out of Runtime and into a new Blocks.js.
- Updates to sequencer and tests for the new structure.
- Rewrite of some event handling code (MOVE especially) to handle the new cases for inputs.

What's not done:
- Updating block execution for multiple inputs, etc.
- I'd like Blocks.getBlocks and Blocks.getStacks to return immutables (immutable.js?) so we can't easily mess with the block representation when executing blocks.
- Moving block listener handling into Blocks module.
- Would love to have more tests for all of the things this handles now.

What I'd like most eyes on in a review:
- New structure for block representation
- Implementation of adapter
- Implementation of MOVE

Let me know if you have questions/want any walk-through...


tmickel:
Representation for these blocks:
screen shot 2016-06-06 at 4 19 02 pm
is now:
screen shot 2016-06-06 at 4 21 26 pm

tmickel:
(fixes #1, #50, #51)

src/engine/blocks.js
```diff
@@ -0,0 +1,206 @@
...
+ };
+
+ /**
+  * Get the next block for a particular block
+  * @param {?string} id ID of block to get the next block for
+  * @return {?string} ID of next block in the sequence
+  */
+ Blocks.prototype.getNextBlock = function (id) {
+     if (typeof this._blocks[id] === 'undefined') return null;
+     return this._blocks[id].next;
```
RoboErikG, line 47: Drive by suggestion: If you add a .parent field to blocks that points to the block further up the stack you can do traversals in both directions in the stack. This would let you remove the 'stackclick' event and to the traversal to find the top of the stack on the VM side instead.

thisandagain:
This looks great at a high level. Mind if we step through this together tomorrow?

src/engine/blocks.js
tmickel, line 47: That sounds good in the long run. Can't quite do it yet because of https://github.com/google/blockly/issues/336

tmickel:
@thisandagain Yep, any time! I just added some more tests for `adapter` and blocks.js

PR #60, opened 2016-06-08
Instrument VM and playground for better execution debugging

tmickel:
Does a few things:
- Add zoom controls to the scratch-blocks instance in the playground.
- Add a tab to the VM playground which always displays the current value of `vm.runtime.threads`. This value changes quite fast, but the view should be really helpful if we put in a single-stepping mode also.
- Re-add block-by-block glowing to make it easy to follow control flow as it's unfolding.
- Add sequencer-level logging of block execution, including grouped prints of arguments, return values, the stack frame before/after block execution, etc., and a flag to turn it off for production.

@cwillisf 

What to do next:
- Remove YieldTimers altogether, to make things cleaner to develop with.
- Refactor block execution into a separate file and a few separate functions.
- Split scratch3.js blocks into scratch3_control, scratch3_event.
- Re-add execution of arguments using a recursive call to the execution function.
- Add single-stepping debug mode in the VM and the playground.
- Fix implementations for math_number, math_add, control_repeat, control_forever, control_wait, etc.
- Make sure the model for threads is working out, potentially refactor/update it/fix bugs discovered in implementing the above...

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,7 @@
         "max-len": [2, 80, 4],
         "semi": [2, "always"],
         "strict": [2, "never"],
-        "no-console": [2, {"allow": ["log", "warn", "error"]}]
+        "no-console": [2, {"allow": ["log", "warn", "error", "groupCollapsed", "groupEnd"]}]
     },
     "env": {
         "node": true,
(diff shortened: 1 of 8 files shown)


thisandagain:
Thanks for putting this together. Looks good.

PR #67, opened 2016-06-08
Add .eslintrc rule to check JSDoc syntax

tmickel:
And fix the broken syntax.


thisandagain:
Thanks. We should make this rule standard across all our projects.

PR #72, opened 2016-06-09
Refactor for block execution

tmickel:
This code is a pretty big divergence from the old code. It's probably not comparing to the old code - just look at the new code.

Here are notable changes:
- All logic as far as "what block to execute next" is now the responsibility of the thread's stack. I added some helper functions to the Thread module to deal with the stack.

The basic framework is:
1. The current block running is the top block on the stack. The only way `execute` knows how to run a block is by inspecting the top of the stack.
2. To execute a substack or an input block (such as the add block in an argument), push it to the stack and call `execute` again (see execute.js:39).
3. When finished with executing a block, pop the stack, and push the block's next block onto the stack (see `sequencer. proceedThread`). When no "next block" is found, try popping the stack.
- Return values are provided by the return value of the primitive. These are fed into `argValues` in execution, as well as field values. I've implemented math_number, math_add, etc., which you can drag into `wait`, `repeat,` etc., as inputs.
- Changed primitives to take a named arguments object instead of a list of arguments. This matches Blockly more closely, and I find it quite nice, but we could do an ordered list instead...

Some code that I think this is correctly running:

screen shot 2016-06-09 at 5 19 47 pm

I like this structure quite a bit more than the "classic Scratch" structure myself. But I think it could be up for debate. 

I think it's also very important that we get these pieces right, so more than happy have very thorough reviews and to make lots of adjustments/renames/changes.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -13,17 +13,26 @@ function Scratch3OperatorsBlocks(runtime) {
 Scratch3OperatorsBlocks.prototype.getPrimitives = function() {
     return {
         'math_number': this.number,
-        'math_add': this.add
+        'text': this.text,
+        'math_add': this.add,
+        'logic_equals': this.equals
     };
 };
 
-
-Scratch3OperatorsBlocks.prototype.number = function(args) {
+Scratch3OperatorsBlocks.prototype.number = function (args) {
     return Number(args.NUM);
 };
 
-Scratch3OperatorsBlocks.prototype.add = function(args) {
+Scratch3OperatorsBlocks.prototype.text = function (args) {
+    return String(args.TEXT);
+};
+
+Scratch3OperatorsBlocks.prototype.add = function (args) {
     return args.NUM1 + args.NUM2;
+};
+
+Scratch3OperatorsBlocks.prototype.equals = function (args) {
+    return args.VALUE1 == args.VALUE2;
 };
 
 module.exports = Scratch3OperatorsBlocks;
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -1,0 +1,29 @@
+function Scratch3OperatorsBlocks(runtime) {
+    /**
+     * The runtime instantiating this block package.
+     * @type {Runtime}
+     */
+    this.runtime = runtime;
+}
+
+/**
+ * Retrieve the block primitives implemented by this package.
+ * @return {Object.<string, Function>} Mapping of opcode to Function.
+ */
+Scratch3OperatorsBlocks.prototype.getPrimitives = function() {
+    return {
+        'math_number': this.number,
+        'math_add': this.add
+    };
+};
+
+
+Scratch3OperatorsBlocks.prototype.number = function(args) {
+    return Number(args.NUM);
+};
+
+Scratch3OperatorsBlocks.prototype.add = function(args) {
+    return args.NUM1 + args.NUM2;
+};
+
+module.exports = Scratch3OperatorsBlocks;
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -23,6 +23,13 @@ function Blocks () {
 }
 
 /**
+ * Blockly inputs that represent statements/substacks
+ * are prefixed with this string.
+ * @const{string}
+ */
+Blocks.SUBSTACK_INPUT_PREFIX = 'SUBSTACK';
+
+/**
  * Provide an object with metadata for the requested block ID.
  * @param {!string} blockId ID of block we have stored.
  * @return {?Object} Metadata about the block, if it exists.
@@ -60,7 +67,7 @@ Blocks.prototype.getSubstack = function (id, substackNum) {
     if (typeof block === 'undefined') return null;
     if (!substackNum) substackNum = 1;
 
-    var inputName = 'SUBSTACK';
+    var inputName = Blocks.SUBSTACK_INPUT_PREFIX;
     if (substackNum > 1) {
         inputName += substackNum;
     }
@@ -78,6 +85,34 @@ Blocks.prototype.getSubstack = function (id, substackNum) {
 Blocks.prototype.getOpcode = function (id) {
     if (typeof this._blocks[id] === 'undefined') return null;
     return this._blocks[id].opcode;
+};
+
+/**
+ * Get all fields and their values for a block.
+ * @param {?string} id ID of block to query.
+ * @return {!Object} All fields and their values.
+ */
+Blocks.prototype.getFields = function (id) {
+    if (typeof this._blocks[id] === 'undefined') return null;
+    return this._blocks[id].fields;
+};
+
+/**
+ * Get all non-substack inputs for a block.
+ * @param {?string} id ID of block to query.
+ * @return {!Object} All non-substack inputs and their associated blocks.
+ */
+Blocks.prototype.getInputs = function (id) {
+    if (typeof this._blocks[id] === 'undefined') return null;
+    var inputs = {};
+    for (var input in this._blocks[id].inputs) {
+        // Ignore blocks prefixed with substack prefix.
+        if (input.substring(0, Blocks.SUBSTACK_INPUT_PREFIX.length)
+            != Blocks.SUBSTACK_INPUT_PREFIX) {
+            inputs[input] = this._blocks[id].inputs[input];
+        }
+    }
+    return inputs;
 };
 
 // ---------------------------------------------------------------------
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -1,0 +1,154 @@
+var Thread = require('./thread');
+var YieldTimers = require('../util/yieldtimers.js');
+
+var execute = function (sequencer, thread, blockId) {
+    var runtime = sequencer.runtime;
+
+    // Save the yield timer ID, in case a primitive makes a new one
+    // @todo hack - perhaps patch this to allow more than one timer per
+    // primitive, for example...
+    var oldYieldTimerId = YieldTimers.timerId;
+
+    var opcode = runtime.blocks.getOpcode(blockId);
+
+    // Push the current block to the stack
+    thread.stack.push(blockId);
+    // Push an empty stack frame, if we need one.
+    // Might not, if we just popped the stack.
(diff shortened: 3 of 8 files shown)


tmickel:
Just for fun: https://www.dropbox.com/s/xsgp39vq7hyg513/lots-of-code.mov?dl=0 (the blocks at the bottom are too fast to highlight)

tmickel:
Bugs just found:
- Empty loops don't seem to behave quite right (e.g., an repeat loop with no statements inside appears to run forever).
- Dragging a block into an empty loop still doesn't help (it seems permanently stuck).

tmickel:
^ addressed the problems above with the empty substacks - these are behaving as expected now :)

tmickel:
Things I think it would be good to address next:
1. Yielding reporters that return values (e.g., an "HTTP request" extension block).
2. Yielding reporters that block other threads (e.g., "touching color" across a thread boundary).
3. Custom blocks that recurse (how is this done in Scratch 2.0? Maybe it's not so hard, but I haven't even thought about how it works).
4. Custom reporters (this seems like a desired feature - we should think about how it might be implemented sooner rather than later...)

tmickel:
For fun, just added implementations for if and if-else.

src/engine/blocks.js
```diff
@@ -80,6 +87,34 @@ Blocks.prototype.getOpcode = function (id) {
...
+ * @param {?string} id ID of block to query.
+ * @return {!Object} All non-substack inputs and their associated blocks.
+ */
+Blocks.prototype.getInputs = function (id) {
+    if (typeof this._blocks[id] === 'undefined') return null;
+    var inputs = {};
+    for (var input in this._blocks[id].inputs) {
+        // Ignore blocks prefixed with substack prefix.
+        if (input.substring(0, Blocks.SUBSTACK_INPUT_PREFIX.length)
+            != Blocks.SUBSTACK_INPUT_PREFIX) {
```
cwillisf, line 111: Could this check instead be based on some property of `this._blocks[id].inputs[input]`?

src/engine/sequencer.js
```diff
@@ -54,28 +49,18 @@ Sequencer.prototype.stepThreads = function (threads) {
...
                 // Normal-mode thread: step.
-                this.stepThread(activeThread);
+                this.startThread(activeThread);
             } else if (activeThread.status === Thread.STATUS_YIELD) {
                 // Yield-mode thread: check if the time has passed.
-                YieldTimers.resolve(activeThread.yieldTimerId);
-                numYieldingThreads++;
+                if (!YieldTimers.resolve(activeThread.yieldTimerId)) {
+                    numYieldingThreads++;
+                }
```
cwillisf, line 61: At first glance it looks like a single yielding thread could cause `numYieldingThreads` to grow until it's big enough to exceed `threads.length`, as long as there are other threads which want to keep executing and the work time hasn't run out. Is there something here which prevents that?

src/engine/blocks.js
tmickel, line 111: One way to do that would be to separate "statement inputs" from "value inputs" during the construction of the block representation:
https://github.com/LLK/scratch-vm/blob/develop/src/engine/adapter.js#L108 i.e., splitting that into two separate cases that get stored separately.

I want to be careful not to complicate our implementation of the "block move" function (https://github.com/LLK/scratch-vm/blob/develop/src/engine/blocks.js#L186) since Blockly treats both types of inputs the same in this case. One way to do that might just to be adding a boolean flag to the input object for "whether it's a statement." Hmm, thoughts?

src/engine/sequencer.js
tmickel, line 61: That's a pretty good catch I think.

I think we should reset `numYieldingThreads` before the `for` loop, i.e., it should reset every time we are going to iterate across all the threads.

Also...I still think we need to go over this whole yielding flow and how things are treated generally, especially if we intend to implement yielding reporters...

src/engine/blocks.js
cwillisf, line 111: I think the Boolean flag would make sense, since like you suggested it avoids the need for additional complexity in `moveBlock` and maybe other code in the future.

Are there other parts of the code that could benefit from such a flag? On the one hand I'm in favor of explicit data (like an `isStatement` or `inputType` field or whatever) over implicit data (special knowledge about the prefix of some input names), but on the other hand I'm concerned that I'm making a big deal about one substring compare.

tmickel, line 111: Ok, cool. Yes, actually - I think `Blocks.prototype.getSubstack(N)` will also benefit from not dealing with prefixes. Instead we could inspect the actual order the statement inputs come in the block, and not depend on the blocks having particular names.

I'll make an issue for this and fix it up after this PR.

src/engine/sequencer.js
```diff
@@ -94,175 +83,71 @@ Sequencer.prototype.stepThreads = function (threads) {
...
+    if (!substackNum) {
+        substackNum = 1;
     }
+    var currentBlockId = thread.peekStack();
+    var substackId = this.runtime.blocks.getSubstack(
+        currentBlockId,
+        substackNum
+    );
+    if (substackId) {
+        // Push substack ID to the thread's stack.
```
cwillisf, line 123: It may already be too late, but maybe the three of us should discuss terminology to see if there's a good way to avoid pushing a (sub)stack onto the stack :)

tmickel, line 123: Ah, yes...not sure which one to not name stack :)

tmickel, line 123: What do you think of calling it "connected statement input", the way Blockly does, and using "stack" to refer to `thread.stack`?

cwillisf, line 123: I do think I would rather change the term for "block stack" than change the term for "thread stack".

"Connected statement input" seems a little cumber[some](http://i.imgur.com/cCMBamQ.jpg), but on the other hand it's definitely unambiguous. Using terminology similar to Blockly may be a plus, too.

JS calls `{ this; that; }` a "block statement" but obviously "block" already has meaning for us...

Some other languages call it a "compound statement" which is only a little less cumbersome than "connected statement input" especially if you drop the "input" part, and it doesn't get the Blockly bonus points.

I think the word "sequence" might have potential, too.

src/blocks/scratch3_operators.js
```diff
@@ -0,0 +1,38 @@
...
+Scratch3OperatorsBlocks.prototype.number = function (args) {
+    return Number(args.NUM);
+};
+
+Scratch3OperatorsBlocks.prototype.text = function (args) {
+    return String(args.TEXT);
+};
+
+Scratch3OperatorsBlocks.prototype.add = function (args) {
+    return args.NUM1 + args.NUM2;
```
cwillisf, line 31: :tada:

cwillisf:
Overall I think this is heading in the right direction. There are parts of this that I don't feel like I truly "get" yet, but I'm OK with that for now. I propose that we sync up when I'm in town: you can teach me about this stuff and I can teach you about my rendering work.

General thoughts:
- I'm in favor of the named args! I think it's easier to read and has some added benefits besides. I think we should consider storing named args instead of positional args in "SB3" files...
- I'm impressed at the relatively straightforward implementation of blocks like `math_add`.
- I'm a little sad that we need the `util` parameter, but I haven't come up with a better solution yet.
- Implementing "core" blocks as pseudo-extension block packs this way still feels right to me. I don't know if they're going to turn out to be actual extensions or not, but I like how self-contained and insulated from the guts of the VM they are.

src/engine/sequencer.js
thisandagain, line 123: I have a little bit of worry about us changing the terminology for "block stack" as "stack" is generally used throughout the team to refer to a "compound statement" though I appreciate the desire to keep "thread stack" in place as it seems most correct.

I suppose using something that is more idiomatic to Blockly may be a good way to go given how confusing the switch between Blockly <> Scratch naming conventions have been.

src/engine/execute.js
```diff
@@ -0,0 +1,94 @@
...
+var execute = function (sequencer, thread) {
+    var runtime = sequencer.runtime;
+
+    // Current block to execute is the one on the top of the stack.
+    var currentBlockId = thread.peekStack();
+    var currentStackFrame = thread.peekStackFrame();
+
+    // Save the yield timer ID, in case a primitive makes a new one
+    // @todo hack - perhaps patch this to allow more than one timer per
+    // primitive, for example...
```
thisandagain, line 18: Not sure I understand what you are referencing in "for example...".

src/engine/execute.js
```diff
@@ -0,0 +1,94 @@
...
+        return;
+    }
+
+    if (DEBUG_BLOCK_CALLS) {
+        console.groupCollapsed('Executing: ' + opcode);
+        console.log('with arguments: ', argValues);
+        console.log('and stack frame: ', currentStackFrame);
+    }
+    var primitiveReturnValue = null;
+    try {
```
thisandagain, line 61: Are there any performance implications that we should worry about in using try/catch around the execution of each block function?

https://github.com/petkaantonov/bluebird/wiki/Optimization-killers
http://www.html5rocks.com/en/tutorials/speed/v8/

thisandagain:
This looks great. It might be nice to push the latest updates for `scratch-blocks` so we don't have to `npm link` to test this thoroughly. 

I echo @cwillisf 's impressions on the use of named args and how nice the encapsulation of each set of core blocks is as a pseduo-extension. I have a bunch of detailed questions regarding performance, custom blocks, and custom reporters, but I think those are best left when all three of us can step through this code in-person. Really great work @tmickel .

src/engine/execute.js
tmickel, line 18: Will elaborate in more detail with the YieldTimers refactor/rethinking...

src/engine/execute.js
tmickel, line 61: Hmmm, interesting. I do think if our bottleneck is in block execution, we're going to have a lot of problems. On the other hand, I'm re-questioning what the try/catch is offering us during the development phase anyway. I think the intention is so extensions that `throw` won't disrupt other block execution. But it seems fine to me to take it out now, until we actually build something that accepts "proper" extensions.

tmickel:
Ok, thanks for the reviews! I'll merge this in, and make a few new issues for further discussion.

src/engine/execute.js
thisandagain, line 61: 👍

PR #81, opened 2016-06-14
Update playground and block implementations to match scratch-blocks

tmickel:
Changed some naming and added some new (unimplemented) blocks during rendering improvements.


thisandagain:
Thanks for updating this. 👍

Issue #78, opened 2016-06-13
Implement yielding reporters that block all threads


tmickel:
(e.g., "color touching color" across a thread boundary).

In Scratch 1.0/2.0, this is handled by ensuring that all operations that need to block all threads are actually performed synchronously. So, one way would be to make sure that the reporter doesn't give up the JS execution thread until it gets its answer (while loop?). The other way would be to have a flag on `util.yield` that specifically disables running other threads until that reporter returns.

tmickel:
Team decided not to do this.

Issue #77, opened 2016-06-13
Implement yielding reporters that return values


tmickel:
An example of a block that needs this is an "HTTP request" extension block.

How should it work?

Perhaps,
- Mark the thread as yielded, and don't pop the stack to make sure flow control waits on that block.
- Add a `util.report` function that the primitive can execute when it's ready.
- `util.report` somehow bubbles up its value into the caller's arguments object.
- Make sure each reporter is only called once per block execution.

PR #83, opened 2016-06-17
Yielding reporters, blocking yields

tmickel:
Here's a potential implementation of reporters that report after some arbitrary time has passed.

Changes:
- Cleaned up YieldTimers a bit, to move more of the logic into Thread and remove unused functions. With this change, threads can have any number of timers. You could use this, for example, to set something to happen after 500ms, and something else to happen after 750ms.

-Implements an example of a reporter that returns a value after 1s. This is a prototypical "random" that always takes "1s" to return its value.

-When a block calls `util.report`, two things happen:
- the reported value is pushed to the parent's stack frame.
- the same thing that happens when a normal block finishes its yield (compare to `util.done`), returning flow control to the parent block.
- When a block's child reporter yields, it returns immediately. We need to come back to execute it at a later time, and we will, once the stack is popped and the value is present in `stackFrame.reported`. If a value is already in `stackFrame.reported`, the child block is never executed again.
- Moved the old "stack frame" (that the actual block primitive can store data in) to `stackFrame.executionContext`.

I think that's about it. It seems to work pretty well, and can deal with reporters that yield inside reporters that yield, etc. etc. :) Looking forward to your thoughts and reviews.

Screenshot of it working:
screen shot 2016-06-17 at 3 08 34 pm


tmickel:
Should address #77

tmickel:
actually, something is not quite right with reporters-in-reporters - will reopen when I figure it out.

tmickel:
Ok, fixed it up, I think. Keeping the name of the waiting reporter in the stack frame as well, instead of the previous mess, and it seems to fix the bug.

tmickel:
Updates:
- I simplified the logic of how to deal with reported values in general. Previously, if they were returned immediately (if the thread wasn't yielded), I tried to keep it off the stack frame. But, it's become clear to me that it's easier and nicer to just always put return values there. Updated `execute` to deal with reporter values that way, and never actually return values.
- Added a bunch of comments and simplified some code.

tmickel:
Added a blocking yield mode (for #78).
blocking-yield

It was a pretty straightforward implementation.

thisandagain:
@tmickel Mind if we walk through this one in-person?

tmickel:
@thisandagain Sounds good to me!

src/blocks/scratch3_operators.js
```diff
@@ -31,8 +40,54 @@ Scratch3OperatorsBlocks.prototype.add = function (args) {
...
 
+Scratch3OperatorsBlocks.prototype.gt = function (args) {
+    return args.OPERAND1 > args.OPERAND2;
+};
+
+Scratch3OperatorsBlocks.prototype.and = function (args) {
+    if (!args.OPERAND1 || !args.OPERAND2) {
+        return false;
+    }
+    return true;
```
cwillisf: Why not `return args.OPERAND1 && args.OPERAND2`? It seems like that would handle `undefined` and such ok... but maybe there's some odd corner of JS that I'm forgetting :)

tmickel: Hmm, I forget actually. There was something about that that didn't match the way Scratch was working. But, for the purposes of this PR I think you should ignore these operators implementations. I could even take them out - just happened to do them for testing things out. I think we'll need to go through and audit all the weirdness of how Scratch's current operators work and make sure we get these all right.

src/blocks/scratch3_operators.js
```diff
@@ -31,8 +40,54 @@ Scratch3OperatorsBlocks.prototype.add = function (args) {
...
+};
+
+Scratch3OperatorsBlocks.prototype.random = function (args, util) {
+    // As a demo, this implementation of random returns after 1 second of yield.
+    // @todo Match Scratch 2.0 implementation with int-truncation.
+    // See: http://bit.ly/1Qc0GzC
+    util.yieldAndBlock();
+    setTimeout(function() {
+        var randomValue = (Math.random() * (args.TO - args.FROM)) + args.FROM;
+        util.report(randomValue);
```
cwillisf: I wonder if it would make sense to use promises for this. Something like:

``` js
var result = blockFunc(...);
if (result is thenable) {
  result.then(function(actualResult) { util.report(actualResult); });
}
else {
  // handle the result immediately
}
```

I assume the `util.report()` could be merged in / simplified / eliminated in this sort of structure, but you get the idea.
We could further merge the two paths my wrapping everything in `Promise.resolve()` but maybe it's better to keep them separate...?

src/blocks/scratch3_operators.js
tmickel: (Actually it does look like Scratch's is &&. I'll try to reproduce whatever made me switch away from that and make a note of it. :))

tmickel: Perhaps it was string-related... Does 'false' evaluate to false in Scratch vs. JS?

tmickel: In any case there's a better way to write this :)

src/blocks/scratch3_operators.js
tmickel: I like where you're going with that. Will think about it more :)

src/blocks/scratch3_operators.js
tmickel: Ok, here was the reason:
screen shot 2016-06-24 at 10 48 48 am
I think this is made better by just casting to Boolean, though, so I'll update the code to do that. Still maintain that we should do a careful audit :)

tmickel:
I'm realizing that the word "block" is probably not the best choice in the Scratch context...will try to think of alternatives for "this thread makes other threads not precede."

src/blocks/scratch3_operators.js
tmickel: @cwillisf Check out my latest commit - updated it as you suggested. I think right now it doesn't make the implementation much simpler, but I still like it, especially how well it would interact with the renderer, things like https://github.com/request/request-promise, etc.

jmoenig:
you could consider it to be a "modal thread" :-)

tmickel:
@jmoenig not a bad name :)

harbaum:
What happened to this? util.report doesn't seem to exist (anymore?) and just yield'ing a reporter doesn't work as i expect. Is there an example for a reported that needs to do async background stuff?

I am using WebUSB for hardware input and output and would like to block in a reporter until the WebUSB has sent and received a request and reply.

towerofnix:
> Is there an example for a reported that needs to do async background stuff?

Although I'm not certain this is the case nowadays, have you tried simply returning a promise?

```js
myReporterFunc (args, util) {
    return new Promise(resolve => {
        blahCallbackStuff(resolve);
    });
    // or if your helper function already works with promises,
    return blahHelperStuff();
}
```

Do note that this PR is from over two years ago! The code here might not be relevant today at all.

harbaum:
Thanks for the quick response. 

I may not be fully understanding the aspects of promises. But how am i supposed to provide the reporters return value from the promise?

towerofnix:
Of course! I do suggest you take a quick look at how promises work; [Fun Fun Function's video](https://www.youtube.com/watch?v=2d7s3spWAzo) is a good, thorough enough intro.

The basic idea is that when you do `new Promise(someFunction)`, the `someFunction` argument gets passed two values - `resolve` and `reject`. These values are both functions. Resolve sets the value and makes the return value of the promise (and thus the return value of the Scratch reporter block); reject throws an error (which isn't especially relevant in Scratch development).

So, you would do `resolve(80)` to return 80. Likewise, you might do something like this to return a value from a Node.js-style callback function:

```js
return new Promise((resolve, reject) => {
    blahCallbackStuff((err, data) => {
        if (err) {
            reject(err);
        } else {
            resolve(data);
        }
    });
});
```

If your `blahCallbackStuff` function is always successful and simply passes the data through as the first argument, you'd do this:

```js
...
blahCallbackStuff(data => {
   resolve(data);
});
```

Or you could just pass `resolve` in place of the explicit `data => resolve(data)` (since functions are values, we can do that):

```js
...
blahCallbackStuff(resolve);
```

That's what I did in the `myReporterFunc` example. If your helper function doesn't work like that you can use the I have earlier for working with node-style stuff, or you can adapt it to work with whatever kind of code you have. :)

harbaum:
Thanks a lot! This seems to do exactly what i need. Great!

harbaum:
This indeed mostly works as i expect. But I am having trouble getting a hat to work that way. My test code looks like this:

 ```
when_test (args, util) {
	return new Promise(resolve => {
	    if(this.toggle === undefined) this.toggle = true;
	    this.toggle = !this.toggle;
	    setTimeout(resolve.bind(this, this.toggle), 1000);
	});
}	
```

I would expect a hat like this to fire every two seconds. But it doesn't trigger at all and after some time it even stops being called at all. The same code being used for a boolean block works exactly as i'd expect and it toggles  between true and false and delays a second every time.

**Comment by towerofnix** [comment]
@harbaum I'm honestly not sure if or how hat blocks work with promises... I think you might end up having to do something else like this?

```js
when_test (args, util) {
    if (!this.isLoading) {
        this.isLoading = true;
        somePromise().then(() => {
            this.ready = true;
            this.isLoading = false;
        });
        if (this.ready) {
            this.ready = false;
            return true;
        }
    }
    return false;
}
```

Basically, if there's no promise already doing so, this starts one to do some async process (e.g. waiting a second). When that promise resolves, `isLoading` is set to false and `ready` is set to true. Then the hat script sees those conditions and returns true, setting `ready` back to false so that it doesn't return true again the next tick.

That should work regardless of if hat blocks support promises. If they *do*, then there's probably a much simpler way, but I'd have expected your code above to work, and it seems that it doesn't. The code in this post, although complicated, should work. :P

Out of curiosity's sake, what hat block are you making that depends on a promise to resolve to true? There may be a more straightforward way to go depending on your use case!

harbaum:
@towerofnix Thanks again. I'll give it a try asap. I was assuming that hats are supposed to work with promises since otherwise i'd expect errors if i return a promise where a boolean is expected.

The hat block i am making reads a digital input from an ftDuino (http://ftduino.de) connected via USB. You press a button connected there and the hat does its magic. This requires a message to be sent via WebUSB and a message to be received and parsed and imho using promises would result in a nice and clean solution.

harbaum:
@towerofnix Hmm ... your solution is fine if you need to return true every now and then. But i have to deliver a boolean which just takes some time to determine its state. I can modify your example to store my result and return the stored result until a new value is available.

But there's one slightly ugly thing: I don't know what to return until i get the first reply from the hardware. Until that I'd have to return an initial (potential wrong) value which may cause one unwanted initial event at startup if the hardware state differs from this initial value.

Yield'ing btw also does not work. 
```
    when_test (args, util) {
	util.yield();
	return false;
    }
```
This results in permanent hat events at max rate.

towerofnix:
> But there's one slightly ugly thing: I don't know what to return until i get the first reply from the hardware. Until that I'd have to return an initial (potential wrong) value which may cause one unwanted initial event at startup if the hardware state differs from this initial value.

Hmm... I thought my code would deal with that. That is, it'll always return false until the first response. In Scratch terms I think this makes sense - it will do nothing (won't run the hat block's connected scripts) until it is guaranteed the value true from the hardware (i.e. the promise).

harbaum:
My hardware doesn't return a single true if "something happened" (like e.g. a keyboard would). Instead it returns the exact state of its hardware inputs. It returns false as long as the input is open/unconnected and returns true as long as the input is closed/connected.

Hats execute their code whenever their state changes from false to true.

Consider the hardware input is permanently activated (e.g. button pressed). This should not cause the hat to trigger any execution at all since that's supposed to happen at the transition from false to true only and in this case there is no transition.

With your non-blocking version false is being returned until the first reply from the hardware returns true. It will then again return false until the hardware returns the next state which will still be true. So there are permanent transitions from false to true although the hardware input is at a static true. In the scratch logic that hat should never execute when there's a static true level. Using your logic it will permanently fire.

harbaum:
@towerofnix There are more problems with your solution. If you have multiple hats then only one of them will ever be run. The other one will permanently see this.isLoading being true.

It seems i'll just move all the IO logic out of the hats and do the processing somewhere in the background.

towerofnix:
>There are more problems with your solution. If you have multiple hats then only one of them will ever be run. The other one will permanently see this.isLoading being true.

That's true - I didn't think of that. You'd have to set a "was loading this frame" flag somehow, which gets cleared at the end of every frame, though I'm not sure how. And it's rather a lot of effort to mess around with - I doubt anyone could end up with a very clean solution!

PR #85, opened 2016-06-21
Update to allow running in a WebWorker

tmickel:
To the external user, the only difference is including "vm.worker.js" instead of "vm.js". Everything else has been patched to look exactly the same.
- Unify the external VM interface so that everything goes through `vm` (e.g., never touch `vm.runtime`).
- This clears the way to run everything in a worker, optionally.  src/worker.js must mirror src/index.js, but instead be posting messages into the worker thread.
- Update things to use `self` instead of `window` when the latter is not detected.
- Provide a webpack-dev-server since workers won't run on file://

Feedback welcome on both details and overall structure.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,8 @@
     },
     "env": {
         "node": true,
-        "browser": true
+        "browser": true,
+        "worker": true
     },
     "extends": "eslint:recommended"
 }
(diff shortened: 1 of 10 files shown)


thisandagain:
Nice work. This looks like a good start to me.

PR #87, opened 2016-06-29
Straw-man implementation of targets/sprites/clones

tmickel:
Creating this to get the ball rolling on targets and sprites.

"Target" is an abstract code-containing object. In this case the targets are sprite clones which have a shared instance of Blocks.

There are a bunch of decisions/trade-offs on where to put mess here:
- do threads have a link to their associated target, so we can avoid `runtime.targetForThread`?
- Or do we give `runtime` a `runtime.getOpcode`, `runtime.getFields`, etc., which then searches through targets?
- We could alternatively have a single `Blocks` object and just have target IDs stuck in with each block? But that seems potentially messy regarding clones, etc.

Thoughts welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,4 +31,4 @@ coverage:
 
 # ------------------------------------------------------------------------------
 
-.PHONY: build lint test coverage benchmark
+.PHONY: build lint test coverage benchmark serve
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ watch:
 	$(WEBPACK) --watch
 
 serve:
-	$(WEBPACK_DEV_SERVER) --content-base ./
+	$(WEBPACK_DEV_SERVER) --host 0.0.0.0 --content-base ./
 
 # ------------------------------------------------------------------------------
 
--- a/package.json
+++ b/package.json
@@ -16,14 +16,14 @@
   "dependencies": {
     "htmlparser2": "3.9.0",
     "memoizee": "0.3.10",
-    "promise": "7.1.1",
-    "scratch-render-webgl": "git+https://github.com/LLK/scratch-render-webgl.git"
+    "promise": "7.1.1"
   },
   "devDependencies": {
     "eslint": "2.7.0",
     "highlightjs": "8.7.0",
     "json-loader": "0.5.4",
     "scratch-blocks": "git+https://git@github.com/LLK/scratch-blocks.git",
+    "scratch-render": "git+https://git@github.com/LLK/scratch-render.git",
     "tap": "5.7.1",
     "webpack": "1.13.0",
     "webpack-dev-server": "1.14.1"
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
   "dependencies": {
     "htmlparser2": "3.9.0",
     "memoizee": "0.3.10",
-    "promise": "7.1.1"
+    "promise": "7.1.1",
+    "scratch-render-webgl": "git+https://github.com/LLK/scratch-render-webgl.git"
   },
   "devDependencies": {
     "eslint": "2.7.0",
--- a/playground/index.html
+++ b/playground/index.html
@@ -33,259 +33,302 @@
     <div id="blocks"></div>
 
     <xml id="toolbox" style="display: none">
-        <category name="Events">
-          <block type="event_whenflagclicked"></block>
-          <block type="event_whenbroadcastreceived">
-            <value name="BROADCAST_OPTION">
-              <shadow type="event_broadcast_menu"></shadow>
-            </value>
-          </block>
-          <block type="event_broadcast">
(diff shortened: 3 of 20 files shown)


tmickel:
The second commit adds the renderer to the playground and a bit of extra code to implement two example motion blocks. Some clean-up needs to be done, but it should give the idea.

src/engine/runtime.js
```diff
@@ -77,7 +77,7 @@ util.inherits(Runtime, EventEmitter);
 /**
  * How rapidly we try to step threads, in ms.
  */
-Runtime.THREAD_STEP_INTERVAL = 1000 / 30;
+Runtime.THREAD_STEP_INTERVAL = 1000 / 60;
```
towerofnix, line 86: So.. 60 ticks per second, now? Is this temporary..?

thisandagain, line 86: We are still thinking through all of the implications of this (and how to handle backwards compatibility).

tmickel:
@thisandagain @cwillisf Any comments on this? I'd like to move it along...

Makefile
```diff
@@ -13,7 +13,7 @@ watch:
 	$(WEBPACK) --watch
 
 serve:
-	$(WEBPACK_DEV_SERVER) --content-base ./
+	$(WEBPACK_DEV_SERVER) --host 0.0.0.0 --content-base ./
```
thisandagain, line 16: Would be good to add `serve` to the `.PHONY` list.

package.json
```diff
@@ -16,7 +16,8 @@
   "dependencies": {
     "htmlparser2": "3.9.0",
     "memoizee": "0.3.10",
-    "promise": "7.1.1"
+    "promise": "7.1.1",
+    "scratch-render-webgl": "git+https://github.com/LLK/scratch-render-webgl.git"
```
thisandagain: I think this should be a `devDependency`:
`"scratch-render": "git+https://git@github.com/LLK/scratch-render.git"`

tmickel: cool - about to update for the new repo :)

playground/index.html
```diff
@@ -249,6 +733,9 @@
     <script src="../node_modules/scratch-blocks/blockly_compressed_vertical.js"></script>
     <script src="../node_modules/scratch-blocks/blocks_compressed.js"></script>
     <script src="../node_modules/scratch-blocks/blocks_compressed_vertical.js"></script>
+    <script src="../node_modules/scratch-blocks/msg/messages.js"></script>
+    <!-- Renderer -->
+    <script src="../node_modules/scratch-render-webgl/build/render-webgl.js"></script>
```
thisandagain: Due to the repo / name change this should be `../node_modules/scratch-render/render.js`. Sorry for the hassle.

src/blocks/scratch3_operators.js
```diff
@@ -79,16 +86,16 @@ Scratch3OperatorsBlocks.prototype.not = function (args) {
 };
 
 Scratch3OperatorsBlocks.prototype.random = function (args) {
-    // As a demo, this implementation of random returns after 1 second of yield.
-    // @todo Match Scratch 2.0 implementation with int-truncation.
-    // See: http://bit.ly/1Qc0GzC
-    var examplePromise = new Promise(function(resolve) {
```
thisandagain, line 85: The `require('promise')` at the top of this file is now throwing a lint error (unused).

src/index.js
```diff
@@ -81,17 +89,28 @@ VirtualMachine.prototype.stopAll = function () {
...
+};
+
 /*
  * Worker handlers: for all public methods available above,
  * we must also provide a message handler in case the VM is run
  * from a worker environment.
  */
 if (ENV_WORKER) {
+    self.importScripts(
+        './node_modules/scratch-render-webgl/build/render-webgl-worker.js'
```
thisandagain: Same here:
`./node_modules/scratch-render/render-worker.js`

Makefile
tmickel, line 16: got it!

tmickel:
@thisandagain Hmm, looks like there's still an issue with one test, and `self` not being defined. We're using this as a kind of global in the Clone object...any thoughts on how to resolve, or I can hack on it.

thisandagain:
@tmickel Hmm... no quick fix comes to mind.

thisandagain:
Once that last test issue is resolved I think this is 👍

tmickel:
@thisandagain Cool. I pushed another commit - this one checks if `self` and `self.renderer` are defined when the Clone tries to use it. Seems to fix the tests and also works in the demo, even if it's not the most beautiful approach. Let me know if you have thoughts - excited to get this in!

thisandagain:
That seems fine for now. Thanks.

tmickel:
Thanks to you @thisandagain !

PR #99, opened 2016-08-10
Project wide rename substack -> branch

### LINKED ISSUE #74: Naming of "stack", "substack", "thread.stack", etc.
Some of these names are not like the other. Notably, the "thread stack" is a very different thing from the "block stack." This naming situation has led to awkward situations, like pushing the stack to the stack.

We couldn't immediately decide on what better name to use (original discussion: https://github.com/LLK/scratch-vm/pull/72/files#r66675817) - continuing the discussion here.

tmickel:
Should fix #74.


vm.worker.js
```diff
@@ -109,6 +109,10 @@
 	    this.vmWorker.postMessage({method: 'stopAll'});
 	};
 
+	VirtualMachine.prototype.animationFrame = function () {
```
thisandagain, line 112: Is this a tag-along from another bugfix?

thisandagain:
Looks good, just one question.

vm.worker.js
tmickel, line 112: hmmm... yeah, it's just appearing for the first time in the compiled version. that's already in the code.

PR #100, opened 2016-08-11
Rename `stacks` to `scripts`

tmickel:
We were also using the name "stacks" to refer to stacks of blocks. I'd like to call these "scripts" to remove the ambiguity with "thread stack." This PR makes that change.


thisandagain:
Great. This also makes it consistent with how we refer to "scripts" in our research. Thanks.

PR #101, opened 2016-08-11
Filling out operators and fixing up casting to look like Scratch

tmickel:
The first commit finishes implementing the remaining operators, including string join, letter of, string length, mod, round, and the math functions.

The second commit adds a set of utilities for casting and comparing values the way Scratch does. E.g., it will properly coerce numbers, booleans, and strings to each other, when asked. I modified the operator blocks to use these utility functions, and combining them should more or less give the same execution result as Scratch 2.0, minus unknown quirks.

I've tried to document what's going on. Some of them are a little undocumented/mysterious in the 2.0 codebase. E.g., the 2.0 implementation of `compare` also includes (what I think is) a cache of lowercased values, not sure if we actually need this or if it was an unneeded optimization. Similarly, 2.0 simplifies `isNaN` using the `n != n` trick. I took this out for now but I imagine we could add it back in if profiling says isNaN is very slow.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -102,10 +102,7 @@ Scratch3OperatorsBlocks.prototype.random = function (args) {
 };
 
 Scratch3OperatorsBlocks.prototype.join = function (args) {
-    return Cast.toString(
-        Cast.toString(args.STRING1) +
-        Cast.toString(args.STRING2)
-    );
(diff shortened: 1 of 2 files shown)


thisandagain:
Looks good. These would probably be good candidates for unit testing. I'll file a bug as a follow-on to this.

tmickel:
I was thinking the same. Sounds good to me, thanks!

PR #102, opened 2016-08-12
Add stats.js meter to playground

tmickel:
@ericrosenbaum's request.

screen shot 2016-08-11 at 10 54 14 pm

I put the stats end/begin inside the rAF handler, which I think should correspond to how much time it's taken the browser to get from frame to frame. It seems to be pretty accurate - matches Chrome's built-in FPS meter pretty exactly.


thisandagain:
Great. Thanks.

PR #104, opened 2016-08-16
VM "I/O devices"; clock, mouse as demo

tmickel:
Not necessarily intended to be a full/finished implementation of these features, but an idea/demo for a new abstraction, and a test of it with two cases.

I wondered how we want to represent and provide sensor/device/I/O data to the blocks/extensions.

This is one way, coded here:
- Every type of "I/O device" (mouse, keyboard, clock/timer, microphone, video?, ...renderer?? ...more?) is registered with the VM.
- VM has a standardized way of getting data in and out for them. For now, you can see that illustrated in `postIOData`, through which the playground sends mouse X, mouse Y, and mouse down state to the Mouse object registered with the VM.
- Blocks get a utility "query" function to make any kind of query/request to any of the I/O devices.

Feels pretty organized, would give blocks a lot of power if we can expose (selected?) I/O devices to extensions etc.

Too much abstraction? Too little? Thoughts?

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -88,6 +88,24 @@ window.onload = function() {
         workspace.reportValue(data.id, data.value);
     });
 
+    // Feed mouse events as VM I/O events.
+    document.addEventListener('mousemove', function (e) {
+        var rect = canvas.getBoundingClientRect();
+        var coordinates = {
+            x: e.clientX - rect.left - rect.width / 2,
+            y: e.clientY - rect.top - rect.height / 2
+        };
+        window.vm.postIOData('mouse', coordinates);
+    });
+    canvas.addEventListener('mousedown', function (e) {
+        window.vm.postIOData('mouse', {isDown: true});
+        e.preventDefault();
+    });
+    canvas.addEventListener('mouseup', function (e) {
+        window.vm.postIOData('mouse', {isDown: false});
+        e.preventDefault();
+    });
+
     // Run threads
     vm.start();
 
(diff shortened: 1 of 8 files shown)


cwillisf:
This seems pretty clean, and I'd be cool with merging it. In the interest of thorough discussion, though, here are a couple of thoughts that occurred to me while I was looking this over:

For one thing, it seems like there's fairly limited benefit to "devices" which can operate entirely within the VM worker, such as your clock example here. In particular it's not as readable compared to just having `new Clock()` somewhere inside the sensing primitives file and calling into that directly. If multiple groups of blocks want to share the same device instance then calling it directly gets a little more complicated, but could still be done.

For the mouse example, I feel like the main advantage is offering a good way to pass data across the worker boundary: it's basically providing a special I/O-oriented namespace for worker messages. I wonder what it would look like if we were to generalize this, and provide a better way for two arbitrary pieces of code to communicate through the one worker message pipe. The solution that I went with for rendering is OK, I guess, but I think it could be a lot better.

tmickel:
@cwillisf Thanks for looking it over!

Re: utility of "devices" that exist entirely within the VM worker. Agreed, I'm not sure how useful these will be. Yeah, I wondered if it's nice to expose them to blocks in general - I guess my main question is, could an extension make use of the shared global project timer for its own operations? And what other things that we traditionally stick inside the block implementations could be useful to expose to extensions?

and yeah, sounds good re: passing data across the boundary. It seems like it might be good to unify this, the renderer messaging, and the other messaging code into some kind of shared system :)

thisandagain:
This looks good to merge IMO. Agreed re: needing a standard interface (or at least shared approach) for message passing across the worker boundary. That might be a good topic for conversation next time we get together.

tmickel:
sounds good - looking forward to the continued experimenting!

PR #107, opened 2016-08-23
Hat blocks

tmickel:
Here's my best first shot at implementing hat blocks in a generic way. It's a little treacherous, but hopefully not too much more treacherous than Scratch 2.0.

Features:
- One function that does most of the work (`runtime.triggerHats`)
- Edge-triggered hats (with predicate) and event-triggered hats (with predicate if you want).
- Predicate evaluation can do arbitrary expressions.
- Ability to have hats restart running threads (e.g., for green flag and broadcasts), or not (e.g., for when key pressed when implemented).
- Field matching, especially for "broadcast." (thought about having this be a predicate evaluation, but that seemed ugly, since "broadcast and wait" needs to track threads it triggered somehow...)
- What I think are working/matching implementations of broadcast, broadcast and wait, green flag, and "when timer > _" blocks. I think with this framework, the rest of the hats (including both kinds of Scratch 2.0 extension hats?) should be possible.

Thoughts welcome especially on how to make it less treacherous.

### DIFF
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -71,14 +71,10 @@ Scratch3EventBlocks.prototype.broadcastAndWait = function (args, util) {
         }
     }
     // We've run before; check if the wait is still going on.
-    var waiting = false;
-    for (var i = 0; i < util.stackFrame.triggeredThreads.length; i++) {
-        var thread = util.stackFrame.triggeredThreads[i];
-        var activeThreads = this.runtime.threads;
-        if (activeThreads.indexOf(thread) > -1) { // @todo: A cleaner way?
-            waiting = true;
-        }
-    }
+    var instance = this;
+    var waiting = util.stackFrame.triggeredThreads.some(function(thread) {
+        return instance.runtime.isActiveThread(thread);
+    });
     if (waiting) {
         util.yieldFrame();
     }
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -1,5 +1,3 @@
-var Thread = require('../engine/thread');
-
 function Scratch3EventBlocks(runtime) {
     /**
      * The runtime instantiating this block package.
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -34,7 +34,7 @@ Scratch3EventBlocks.prototype.getHats = function () {
         },*/
         'event_whengreaterthan': {
             restartExistingThreads: false,
-            edgeTriggered: true
+            edgeActivated: true
         },
         'event_whenbroadcastreceived': {
             restartExistingThreads: true
@@ -51,28 +51,28 @@ Scratch3EventBlocks.prototype.hatGreaterThanPredicate = function (args, util) {
 };
 
 Scratch3EventBlocks.prototype.broadcast = function(args, util) {
-    util.triggerHats('event_whenbroadcastreceived', {
+    util.startHats('event_whenbroadcastreceived', {
         'BROADCAST_OPTION': args.BROADCAST_OPTION
     });
 };
 
 Scratch3EventBlocks.prototype.broadcastAndWait = function (args, util) {
-    // Have we run before, triggering threads?
-    if (!util.stackFrame.triggeredThreads) {
-        // No - trigger hats for this broadcast.
-        util.stackFrame.triggeredThreads = util.triggerHats(
+    // Have we run before, starting threads?
+    if (!util.stackFrame.startedThreads) {
+        // No - start hats for this broadcast.
+        util.stackFrame.startedThreads = util.startHats(
             'event_whenbroadcastreceived', {
                 'BROADCAST_OPTION': args.BROADCAST_OPTION
             }
         );
-        if (util.stackFrame.triggeredThreads.length == 0) {
+        if (util.stackFrame.startedThreads.length == 0) {
             // Nothing was started.
             return;
         }
     }
     // We've run before; check if the wait is still going on.
     var instance = this;
-    var waiting = util.stackFrame.triggeredThreads.some(function(thread) {
+    var waiting = util.stackFrame.startedThreads.some(function(thread) {
         return instance.runtime.isActiveThread(thread);
     });
     if (waiting) {
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -76,7 +76,8 @@ Scratch3EventBlocks.prototype.broadcastAndWait = function (args, util) {
     var waiting = false;
     for (var i = 0; i < util.stackFrame.triggeredThreads.length; i++) {
         var thread = util.stackFrame.triggeredThreads[i];
-        if (thread.status !== Thread.STATUS_DONE) {
+        var activeThreads = this.runtime.threads;
+        if (activeThreads.indexOf(thread) > -1) { // @todo: A cleaner way?
             waiting = true;
         }
     }
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -1,3 +1,5 @@
+var Thread = require('../engine/thread');
+
 function Scratch3EventBlocks(runtime) {
     /**
      * The runtime instantiating this block package.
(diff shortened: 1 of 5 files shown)


src/blocks/scratch3_event.js
```diff
@@ -12,23 +12,76 @@ function Scratch3EventBlocks(runtime) {
...
+    }
+    // We've run before; check if the wait is still going on.
+    var waiting = false;
+    for (var i = 0; i < util.stackFrame.triggeredThreads.length; i++) {
+        var thread = util.stackFrame.triggeredThreads[i];
+        var activeThreads = this.runtime.threads;
+        if (activeThreads.indexOf(thread) > -1) { // @todo: A cleaner way?
+            waiting = true;
+        }
+    }
```
towerofnix: Sure, this is cleaner:

``` js
for (let thread of util.stackFrame.triggeredThreads) {
  if (this.runtime.threads.includes(thread)) {
    waiting = true;
  }
}
```

"For each of the triggered threads, check if that thread is included in the running threads, and if so set waiting to true." Makes sense to me.

It's really the exact same thing as the current code, but easier to read.. (But it would require a new polyfill - probably not worth it, since obviously you can just use `..indexOf(..) >= 0`!)

EDIT: Is this using promises? I think it might be from looking at other code.. If so it'll be _much_ easier.

src/engine/execute.js
```diff
@@ -21,9 +30,28 @@ var execute = function (sequencer, thread) {
...
+
+    // Hats are implemented slightly differently from regular blocks.
+    // If they have an associated block function, it's treated as a predicate;
+    // if not, execution will proceed right through it (as a no-op).
+    if (isHat) {
+        var nextBlock = target.blocks.getNextBlock(currentBlockId);
+        if (!nextBlock) {
+            // Hat with no next block - don't try to evaluate it.
+            sequencer.retireThread(thread);
+            return;
```
towerofnix: If custom reporters were to become a thing, there would most likely be side-effecting custom reporters, e.g. an array pop implementation. That means that something such as this:

scratchblocks

..would never trigger the "pop" block.

Not really a huge issue but probably worth noting!

src/blocks/scratch3_event.js
tmickel: yep, that would be cleaner! you are right though that those are ES6 features. maybe once we do #3.

I would like it though if the block implementations didn't need to access `runtime`. We are thinking of the block implementations as eventually something that could be stuck into Scratch extensions, so trying to avoid giving them too much privileged interpreter info. But, this seems like an especially tricky case.

We are using promises to let particular blocks "pause" their thread's execution while waiting for certain results. E.g., our color-touching-color implementation might require crossing the WebWorker thread boundary, so it can return a promise and give its result when the answer comes back across the thread.

src/engine/execute.js
tmickel: That's a really interesting point. Since in 2.0 already we could have extension reporters that have side-effects (in JavaScript), I might be inclined to take this out immediately. Any thoughts from you, @cwillisf?

tmickel: Updated this accordingly.

src/blocks/scratch3_event.js
towerofnix: I'm not totally sure if this is possible, but here's how I'd do it:

``` js
if (/* do this once */) {
  // Wait for *all* of the broadcast-triggered threads to be finished.
  thread.setStatus(STATUS_YIELD)
  Promise.all(/* get promises of */ util.stackFrame.triggeredThreads).then(function() {
    thread.setStatus(STATUS_RUNNING)
  })
}
```

"Pause, and then when all of the triggered threads are finished, continue." (or "Wait until all the triggered threads are finished.")

As you can see it doesn't require access to `runtime` either.

tmickel: Could work if we generated a promise for every triggered thread, although that's not how that's currently done :)

tmickel: not sure if that would be useful in any other scenarios, or just this one. If it's just this one, might be simpler to just provide a method in `runtime` for the special case...

src/engine/execute.js
cwillisf: I definitely think @liam4's example should query the `pop` reporter, but I don't have a problem with it doing so ~30 times per second (or more) if that's the convenient implementation.

Hardware extension reporters frequently have side effects, at least at a small scale. For example, at the hardware level the PicoBoard only reports its sensor data (slider position, etc.) in response to a "get sensor data" command. One could implement a PicoBoard extension where each reporter is actually a requester which sends a "get sensor data" command, waits for the reply, processes the reply, and calls Scratch back with the result. If this were called frequently on a reasonably fast computer, the serial port's receive and maybe even transmit buffer could overflow, disrupting communication with the PicoBoard.

I think it's up to the extension author to avoid problems with calling their reporters frequently. In the PicoBoard example, we use `setInterval` to send the "get sensor data" command at a regular interval and cache the response. That way we can implement plain reporters (instead of requesters) in terms of that cache and we don't have to worry about a traffic jam at the serial port regardless of how frequently Scratch is querying the PicoBoard extension's reporters.

src/engine/runtime.js
```diff
@@ -215,9 +342,16 @@ Runtime.prototype.stopAll = function () {
...
+        if (hat.edgeTriggered) {
+            this.triggerHats(hatType);
+        }
+    }
     var inactiveThreads = this.sequencer.stepThreads(this.threads);
-    for (var i = 0; i < inactiveThreads.length; i++) {
-        this._removeThread(inactiveThreads[i]);
+    for (var j = 0; j < inactiveThreads.length; j++) {
+        this._removeThread(inactiveThreads[j]);
     }
```
dekrain, line 366: Why j instead of i ?

tmickel: Accidental change- thanks!

src/blocks/scratch3_event.js
cwillisf: A few disconnected thoughts:
- I'm OK with control structures acting as a special class of blocks which get a little extra access to internals. Here I'm counting hat blocks as "control structures" in addition to if, repeat, etc.
- If we do want extensions to be able to do this sort of thing, I'd feel better about exposing methods on `runtime` to extensions than the `threads` array itself. Maybe something like `runtime.isActiveThread(thread)` which returns a Boolean?
- I've seen some examples of `for (var i = 0; arrayOfObjects[i]; ++i) { doSomethingWith(arrayOfObjects[i]); }` in a few places where the array is known to contain non-false-like things at each index until the end. I think of this as "the Google way" but I can't remember why. Anyway, allegedly it can be faster than querying `length` each time, but it's also pinging my "premature optimization" detector a bit. What do you think?
- I think you can actually reduce this even further using something like the below. Actually, now that I've written it this way, I think this might be more readable too.

``` js
waiting = util.stackFrame.triggeredThreads.some(function(thread) {
    return this.runtime.threads.indexOf(thread) >= 0;
};
// or
waiting = util.stackFrame.triggeredThreads.some(function(thread) {
    return this.runtime.isActiveThread(thread);
};
```

For reference: [`Array.prototype.some()`](http://kangax.github.io/compat-table/es5/#test-Array_methods_Array.prototype.some) (you may have to expand the "Array methods" section).

cwillisf: To clarify: I don't have a problem with the way you calculate `waiting` right now... I was just having some fun thinking about other ways to phrase it :)

src/engine/runtime.js
```diff
@@ -133,14 +147,56 @@ Runtime.prototype.getOpcodeFunction = function (opcode) {
 // -----------------------------------------------------------------------------
 
 /**
+ * Return whether an opcode represents a hat block.
+ * @param {!string} opcode The opcode to look up.
+ * @return {Boolean} True if the op is known to be a hat.
+ */
+Runtime.prototype.getIsHat = function (opcode) {
+    return opcode in this._hats;
```
cwillisf: Should these use `hasOwnProperty` instead of `in`?

src/engine/runtime.js
```diff
@@ -172,29 +228,100 @@ Runtime.prototype.toggleScript = function (topBlockId) {
 };
 
 /**
- * Green flag, which stops currently running threads
- * and adds all top-level scripts that start with the green flag
+ * Run a function `f` for all scripts in a workspace.
+ * `f` will be called with two parameters:
```
cwillisf, line 242: [Three, sir!](https://en.wikipedia.org/wiki/Rabbit_of_Caerbannog#Holy_Hand_Grenade_of_Antioch)

src/engine/runtime.js
```diff
@@ -172,29 +228,100 @@ Runtime.prototype.toggleScript = function (topBlockId) {
...
         var scripts = target.blocks.getScripts();
         for (var j = 0; j < scripts.length; j++) {
             var topBlock = scripts[j];
-            if (target.blocks.getBlock(topBlock).opcode ===
-                'event_whenflagclicked') {
-                this._pushThread(scripts[j]);
-            }
+            var topOpcode = target.blocks.getBlock(topBlock).opcode;
+            var topFields = target.blocks.getFields(topBlock);
+            f(topBlock, topOpcode, topFields);
```
cwillisf: I'd really like this code to be a bit simpler, but that's probably for a separate change or a separate discussion. In my ideal world, this would be something like:

``` js
var topBlockId = scripts[j];
var topBlock = target.blocks.getBlock(topBlockId);
f(topBlock);
```

...and then `f` could grab the block ID, its fields, etc. if necessary.

src/engine/runtime.js
```diff
@@ -172,29 +228,100 @@ Runtime.prototype.toggleScript = function (topBlockId) {
...
+        if (opt_matchFields) {
+            for (var matchField in opt_matchFields) {
+                if (topFields[matchField].value !==
+                    opt_matchFields[matchField]) {
+                    // Field mismatch.
+                    return;
+                }
+            }
+        }
+        if (instance._hats.hasOwnProperty(topOpcode)) {
```
cwillisf: At first glance I thought `topOpcode` would be the first block under the hat rather than the hat itself, and it wasn't until this line that I realized I must be wrong, though I was quite confused about what the `if (topOpcode !== requestedHat)` above could mean!
Anyway, this might just be my brain slipping gears, but I do wonder if renaming the variable would make it more clear. Maybe something like `hatOpcode`? Somewhat related: `requestedHat` doesn't sound like an opcode to me (and your `@param` calls it a "name")... maybe `requestedHatOpcode` or something?

cwillisf:
LGTM! I commented on a few things but they're all fairly minor. Nice work!

src/blocks/scratch3_event.js
tmickel: `runtime.isActiveThread` sounds nice to me, and `.some` is nice style :) I think the "Google style" for-loop is nice also for style, although I'd guess a premature optimization as far as performance...

Yet another way would be to have the actual hat-starting function return a promise that resolves exactly when its threads are finished up. maybe worth a try someday :)

tmickel: Implemented the `isActiveThread` version - feels nice to me.

src/engine/runtime.js
tmickel: no doubt :)

src/engine/runtime.js
tmickel: Agreed. I think it may make most sense to pass both the block ID and the `target` itself, actually. Otherwise we don't know in which target the block is hosted (without doing another search). Will try that and see how it looks.

src/engine/runtime.js
tmickel: Hmmmmm, yeah. I think a better name for `topOpcode` might be `potentialHatOpcode` or something. If a non-hat starts a script, it's not necessarily a hatOpcode (thus the first check in the function). 

`requestedHatOpcode` and `potentialHatOpcode` might work for me.

src/blocks/scratch3_event.js
tmickel: (P.S. our linter currently forbids the "Google style" for-loop).

src/engine/runtime.js
tmickel: now two! :)

src/blocks/scratch3_event.js
```diff
@@ -12,23 +12,72 @@ function Scratch3EventBlocks(runtime) {
...
-        'event_whenflagclicked': this.whenFlagClicked,
-        'event_whenbroadcastreceived': this.whenBroadcastReceived,
-        'event_broadcast': this.broadcast
+        'event_broadcast': this.broadcast,
+        'event_broadcastandwait': this.broadcastAndWait,
+        'event_whengreaterthan': this.hatGreaterThanPredicate
     };
 };
 
+Scratch3EventBlocks.prototype.getHats = function () {
```
thisandagain, line 21: It would be nice to add JSDoc comments to these.

tmickel: Actually we need JSDoc (and maybe tests) for all of the example block implementations, too. I'll add an issue for it.

tmickel:
Thanks @thisandagain !

Issue #96, opened 2016-07-22
2.0 loader


tmickel:
We'd like a 2.0 loader as soon as feasible, so we can start testing things out.
1. Online loader from projects API, including:
2. Load blocks into new blocks structure (including: map old opcodes to new ones), and flow those to scratch-blocks. Flowing to scratch-blocks requires multiple workspaces (multiple sprites).
3. Load sprites into new sprites/targets/clones structure
4. Load SB2s (e.g., JSZip)

tmickel:
It looks like it's going to be fairly sad to translate SB2-format blocks into our current format.

E.g., this:
`["setGraphicEffect:to:", "color", 75]`

Needs to somehow get interpreted into something like:

> ```
> "10D;mdF?1eWv)}i9ON/@": {
>   "id": "10D;mdF?1eWv)}i9ON/@",
>   "opcode": "looks_seteffectto",
>   "inputs": {
>     "EFFECT": {
>       "name": "EFFECT",
>       "block": "YXLu*Gma.D2P5+~;vW)d"
>     },
>     "VALUE": {
>       "name": "VALUE",
>       "block": "*OK-d(i6Z-W:QJz:Ckfp"
>     }
>   },
>   "fields": {},
>   "next": null,
>   "topLevel": true
> },
> "YXLu*Gma.D2P5+~;vW)d": {
>   "id": "YXLu*Gma.D2P5+~;vW)d",
>   "opcode": "looks_effectmenu",
>   "inputs": {},
>   "fields": {
>     "EFFECT": {
>       "name": "EFFECT",
>       "value": "COLOR"
>     }
>   },
>   "next": null,
>   "topLevel": false
> },
> "*OK-d(i6Z-W:QJz:Ckfp": {
>   "id": "*OK-d(i6Z-W:QJz:Ckfp",
>   "opcode": "math_number",
>   "inputs": {},
>   "fields": {
>     "NUM": {
>       "name": "NUM",
>       "value": "75"
>     }
>   },
>   "next": null,
>   "topLevel": false
> }
> ```
> 
>   }

Notably the distinction between fields/inputs and creating all the relevant shadow blocks...

Could try to read `Blockly.Blocks` and do some magic. But even then some of the information is encoded in the toolbox XML (e.g., that a `math_number` goes into the `looks_seteffectto`'s value input.

Suggestions welcome...

tmickel:
[...]

PR #115, opened 2016-08-31
Start of a 2.0 project loader

tmickel:
Pretty exciting, I think. Gets most of the way through #96, minus importing actual files. It loads a bunch of projects from the website (of course, none of them behave as expected).

Demo: https://www.dropbox.com/s/gek6nndums3x6fr/sb2-demo.mov?dl=0

maybe best approached commit-by-commit. I tried to structure them as a logical progression.

Things to look at in particular:
- Blocks.prototype.toXML - using this to generate a Blockly workspace for the loaded blocks.
- Editing targets (the currently selected sprite, for example)
- The entire file "src/import/sb2import.js" including the spec-map approach. It's kind of boilerplate madness™ so maybe there are other approaches. Thoughts about requirements: shouldn't require loading Blockly (so a programmatic solution using `Blockly.Blocks` or toolbox XML wouldn't work), should work offline. If it feels like too much boilerplate, maybe we should consider not changing 2.0 opcodes and trying to use positional args somehow. However, the structure of how it is now does feel like it's pretty nice for some things. And maybe it's not so bad to update the table as we change what 2.0 blocks map to.

cc: @cwillisf


src/engine/blocks.js
```diff
@@ -120,63 +120,57 @@ Blocks.prototype.getInputs = function (id) {
...
-        if (typeof e !== 'object') return;
-        if (typeof e.blockId !== 'string') return;
-
-        // UI event: clicked scripts toggle in the runtime.
-        if (e.element === 'stackclick') {
-            if (opt_runtime) {
-                opt_runtime.toggleScript(e.blockId);
-            }
-            return;
+Blocks.prototype.blocklyListen = function (e, isFlyout, opt_runtime) {
```
tmickel, line 128: (note nothing changed here, other than no longer generating a new function)

TheBrokenRail:
Awesome! Variables don't work though. when a variable is in a blocks it changes the input it's in to no input dark square. When notr in block it turns to unknown block

tmickel:
@TheBrokenRail Hmm, yeah, thanks. I think to avoid ballooning this PR I'd like to address that separately.

src/index.js
```diff
@@ -101,6 +101,65 @@ VirtualMachine.prototype.postIOData = function (device, data) {
...
+/**
+ * Set an editing target. An editor UI can use this function to switch
+ * between editing different targets, sprites, etc.
+ * After switching the editing target, the VM may emit updates
+ * to the list of targets and any attached workspace blocks
+ * (see `emitTargetsUpdate` and `emitWorkspaceUpdate`).
+ * @param {string} targetId Id of target to set as editing.
+ */
+VirtualMachine.prototype.setEditingTarget = function (targetId) {
+    if (targetId == this.editingTarget.id) { // No change.
```
thisandagain: Total nit: not sure about the structure of the comments here. 

``` js
// No change
if (foo === bar) return;

// Do things
// blah
```

playground/playground.js
```diff
@@ -71,6 +90,38 @@ window.onload = function() {
         updateBlockExplorer(data.blocks);
     });
 
+    vm.on('workspaceUpdate', function (data) {
```
thisandagain, line 101: Might be good to add some documentation for each vm event handler.

TheBrokenRail:
I think it's works well enough to be merged.

thisandagain:
This is really exceptional work. Thanks @tmickel . I think this is good to land, though I think it might be good to add issues to track 1.4 support as well as unit testing for the parser / loader.

dekrain:
I'm ready to update my server :smile:

tmickel:
Thanks @thisandagain ! Sounds good, and I'm sure there are bugs here that we'll discover, too.

PR #117, opened 2016-09-01
Keyboard events, "key is down" reporter

tmickel:
Just playing with the keyboard and came up with this quick:

kbd

saving the hat until after #107


src/io/keyboard.js
```diff
@@ -0,0 +1,82 @@
...
+Keyboard.prototype.postData = function (data) {
+    var key = this._keyCodeToScratchKey(data.keyCode);
+    if (key) {
+        this._keyMap[key] = data.isDown;
+    }
+};
+
+Keyboard.prototype.getKeyIsDown = function (key) {
+    if (key == 'any') {
+        for (var k in this._keyMap) {
```
towerofnix: So importantly this wouldn't work with things such as shift, special characters, and all other non-menu keys as it does in Scratch 2.0?

tmickel: Are those keys are only accessed by hacking the JSON?

jeandrek: ~~Yes~~

towerofnix: As well as `key ANY pressed?`.

ed-cooper:
Being honest, why don't we just add straight support for keys like enter, backspace and shift in Scratch 3?
It would make it easier for less experienced scratchers to understand how to get those keys (as opposed to being really confused when more advanced scratchers do it)

The only real issue is the ~~incredibly stupidly designed and annoying~~ backspace shortcut some browsers have which acts as a back button. In theory though, if the editor has focus and accepts the key presses, that shouldn't be an issue.

src/io/keyboard.js
tmickel: Didn't realize about ANY, actually. Thanks for pointing that out!

tmickel:
@chooper100:

I haven't fully investigated an implementation for something like that. But I suspect there will be a couple of challenges/questions:
1. A possible technical challenge: browsers can be really inconsistent with reporting key codes: http://www.quirksmode.org/js/keys.html. Whatever keys we officially support, we'd like them to work consistently for all users. That said, it might not be impossible, just tricky.
2. Not all keyboards have the same keys. E.g., do we support the "option" key that Macs have?
3. As you mention, default behavior for these keys in browsers. E.g., if we support the "tab" key, it breaks tabbing between elements on the page, an accessibility problem. Backspace also could be an accessibility problem...

All of that said, I'm not necessarily opposed to it.

Another thing I'd like to consider is supporting a typeable dropdown for these:
screen shot 2016-09-02 at 9 02 37 am
So non-English characters could be typed, like ñ, etc.

For now, I'm going to work on a refactor so we produce the right behavior for ANY, and I think what will come with it is the possibility for things like the above.

ed-cooper:
Ok, it really sucks that browsers always try to implement things in their own way rather than consistently to other browsers/standards :disappointed:

towerofnix:
:+1: for the typeable dropdown input. Extra special keys such as shift could be part of the dropdown, which I _think_ we can assume is supported by anything running Scratch? (And I don't think it really means anything for accessibility, not on its own anyways.)

TheBrokenRail:
The Option key is the mac name for alt it even says alt at the top of the key, same with the command key is the control key. I do think alt/option, shift, and backspace sould be supported

tmickel:
Ok, this should now properly handle "any" now to include all keys on the keyboard. I expect it should also deal with the hacked block properly, and is ready for any change to the blocks that allows typing (e.g. for ñ).

TheBrokenRail:
It would also be nice to add backspace and stuff to when key pressed hat block

thisandagain:
Thanks for updating the `keyboard` module. This is nice and straightforward. 👍

PR #120, opened 2016-09-02
Implement 'current' block

towerofnix:
Input on implementing a "milliseconds" option? Wouldn't be too hard (though it would require a pull request to scratch-blocks as well, to add the option to the menu).

EDIT: [scratch-flash implementation](https://github.com/LLK/scratch-flash/blob/3c1671b38e692637974c676064d140bc04957e36/src/scratch/ScratchRuntime.as#L987-L1000)

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -16,7 +16,9 @@ Scratch3SensingBlocks.prototype.getPrimitives = function() {
         'sensing_resettimer': this.resetTimer,
         'sensing_mousex': this.getMouseX,
         'sensing_mousey': this.getMouseY,
-        'sensing_mousedown': this.getMouseDown
+        'sensing_mousedown': this.getMouseDown,
+        'sensing_current': this.current,
+        'sensing_currentmenu': this.currentMenu
     };
 };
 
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_sensing.js
```diff
@@ -40,4 +42,21 @@ Scratch3SensingBlocks.prototype.getMouseDown = function (args, util) {
     return util.ioQuery('mouse', 'getIsDown');
 };
 
+Scratch3SensingBlocks.prototype.current = function (args) {
+    var date = new Date();
+    switch (args.CURRENTMENU) {
+    case 'year': return date.getFullYear();
+    case 'month': return date.getMonth() + 1; // getMonth is zero-based
+    case 'date': return date.getDate();
+    case 'dayofweek': return date.getDay();
```
tmickel: Scratch 2.0 uses `day + 1`: Sunday is 1. Testing your implementation, I get 5 for Friday, in Scratch I get 6.

tmickel:
Looks pretty good, thanks!

Regarding milliseconds: not sure whether or not we'd like this to appear in Scratch. I think it could possibly be interesting for projects that need a precise time, or a fast-moving count.

I think we would accept a PR that puts in milliseconds (here and in scratch-blocks), but I can't guarantee that it won't be stripped out for the 3.0 release. So if you'd like to put it in to play with it, sounds good - just be aware it might not make it in the end :)

tmickel:
Thanks!

PR #131, opened 2016-09-02
Variable catagory like scratch-blocks (Code From Scratch-blocks) (Create Variable Button)

TheBrokenRail:
Adds the create variable button from scratch-blocks NOTE: just like scratch 1, 2, and scratch-blocks no variable blocks will show up until a variable is created

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -397,43 +397,7 @@
               </value>
     		    </block>
           </category>
-          <category name="Data" colour="#FF8C1A">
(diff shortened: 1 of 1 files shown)


tmickel:
Thanks! Do note that variables basically don't work yet in the playground. We need to make sure they're loaded in when a project is loaded, redesign the creation dialog (or other experience), and of course, implement the blocks. But this does get us closer!

Issue #122, opened 2016-09-02
Better handling of shadow blocks


tmickel:
I think it might help us if we stored shadow blocks in the block representation in a more explicit way.

For example, `inputs: {name: "x", block: "real_block" or null, shadow: "shadow_block"}`
and then the interpretation can select the shadow block when the real block isn't present.

Basically, without storing the shadow block even when the real block is present, encoding the workspace in a good way becomes difficult/impossible. Blockly emits a delete event when a shadow is obscured, and we're currently deleting and replacing the shadow block with the new input block. This is great until we need to clear and reload the workspace from XML. We lose the information of what was there previously (since Blockly stores it in a hidden way), but without regenerating the shadow block from XML, we get something like this:
<img width="132" alt="screen shot 2016-09-02 at 10 04 25 am" src="https://cloud.githubusercontent.com/assets/120403/18206417/a8341a02-70f4-11e6-827c-49252c3a903d.png">
So, maybe:
- [x] - Store shadow blocks in inputs explicitly as above.
- [x] - When Blockly tells us to delete a shadow block, ignore it.
- [x] - Update `execute` to look for shadow blocks when non-shadow blocks aren't present - actually decided to accomplish this by letting `input.block` alias `input.shadow`.
- [x] - Update XML encoding to generate shadow blocks even when a non-shadow block is present.
- [x] - Update SB2 importer to generate shadow blocks even when a non-shadow block is present.

PR #135, opened 2016-09-02
Shadow improvements

tmickel:
Here's an implementation of #122.

With these:
- Obscured shadows are not lost on serializing/reloading workspaces. The values are also preserved across switching sprites.
- SB2s generate obscured shadows as expected (and they're filled in with Scratch 2.0-like default values, e.g., "10", "", etc.), so no more "blank spaces" under a block.

shadow-demo

And a couple of small tests.


thisandagain:
Thanks for updating the fixtures / tests.

PR #137, opened 2016-09-02
Tweaks from SB2 importer audit

tmickel:
Added:
- some null checks
- extra math_\* types
- converts colors to hex on import
- various JSON tweaks to make consistent with scratch-blocks

### DIFF
--- a/src/util/math-util.js
+++ b/src/util/math-util.js
@@ -45,15 +45,4 @@ MathUtil.wrapClamp = function (n, min, max) {
     return n - Math.floor((n - min) / range) * range;
 };
 
-/**
- * Convert a Scratch color number to a hex string, #RRGGBB.
- * @param {number} color RGB color as a decimal.
- * @return {string} RGB color as #RRGGBB hex string.
- */
-MathUtil.scratchColorToHex = function (color) {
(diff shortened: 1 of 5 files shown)


TheBrokenRail:
The Variable fix should make a lot of imported projects more understandable! Hooray!

src/util/math-util.js
```diff
@@ -45,4 +45,15 @@ MathUtil.wrapClamp = function (n, min, max) {
     return n - Math.floor((n - min) / range) * range;
 };
 
+/**
+ * Convert a Scratch color number to a hex string, #RRGGBB.
+ * @param {number} color RGB color as a decimal.
+ * @return {string} RGB color as #RRGGBB hex string.
+ */
+MathUtil.scratchColorToHex = function (color) {
```
thisandagain: I wonder if we might want to move this into a dedicated `color` utility module eventually. This should also have test coverage.

thisandagain:
Looks good. One minor comment.

src/util/math-util.js
tmickel: sounds good - seems like we'll want more in the future.

PR #138, opened 2016-09-02
Offline Support [READY TO MERGE?] [ALONG WITH THE SCRATCH-RENDER PR]

TheBrokenRail:
When the 2.0 loader cannot load a project it tries to load another online project. If you couldn't access scratch servers that would mean you couldn't play at all. This PR fixes that by instead of fall backing to another online project it fallbacks to a offline blank project. NOTE: The NPM Test Failed Because Of Newline Characters, So Nothing To Worry About


TheBrokenRail:
no it worked as a input. (In normal html input doses not need to be closed) the ploblemn was since it was a input not a textarea it only allows one liners

playground/playground.js
```diff
@@ -1,11 +1,13 @@
...
     var url = 'https://projects.scratch.mit.edu/internalapi/project/' +
         id + '/get/';
     var r = new XMLHttpRequest();
-    r.addEventListener('load', function() {
-        window.vm.loadProject(this.responseText);
-    });
+    r.onreadystatechange = function() {
+        if (this.readyState == 4) {
+            window.vm.loadProject(this.responseText);
+        };
```
tmickel: This semicolon probably shouldn't be here.

src/import/sb3new.js
```diff
@@ -0,0 +1,16 @@
+var Sprite = require('../sprites/sprite');
+var Blocks = require('../engine/blocks');
+
+function sb3new (runtime) {
```
tmickel: sb3new is not really an "import" function, so I don't think it should live with the /import/ files. Rather it be a function of the VM's index.js, called e.g., `createNewProject`.

src/import/sb3new.js
```diff
@@ -0,0 +1,16 @@
...
+    var blocks = new Blocks();
+    var sprite = new Sprite(blocks);
+    sprite.name = 'Sprite1';
+    var target = sprite.createClone();
+    runtime.targets.push(target);
+    target.x = 0;
+    target.y = 0;
+    target.direction = 90;
+    target.size = 100;
+    target.visible = true;
```
tmickel: If we could have the target's initial costume be the Scratch cat, that would be good :)

src/index.js
```diff
@@ -107,6 +108,13 @@ VirtualMachine.prototype.postIOData = function (device, data) {
  * @param {string} json JSON string representing the project.
  */
 VirtualMachine.prototype.loadProject = function (json) {
+    if (!json) {
+        sb3new(this.runtime);
+        this.editingTarget = this.runtime.targets[0];
+        this.emitTargetsUpdate();
+        this.emitWorkspaceUpdate();
+        return;
```
tmickel, line 152: I think this should all happen instead of `loadProject`, not inside it.

tmickel:
Looks like also we're failing lint because of Windows-style line breaks. Your text editor needs to be configured to produce the standard unix kind.

Thanks for the contribution! I'm a little hesitant to integrate this. The VM playground is intended to be a demo of various VM features, so it actually feels like it could be helpful to have a project loaded by default. It's not intended to be the Scratch 3.0 project editor, where we will definitely make it easy to start with a new project. But maybe I could be convinced. Any thoughts?

luxaritas:
@tmickel Not sure that this is the best place to ask, but since you mentioned it, can you share any info on when we'll see the editor start on GitHub? If it's still under wraps at this point that's fine.

tmickel:
@LFP6 It's not under wraps as much as not quite yet underway! :) It's been in the design/sketching/planning phase. We hope to be getting started on the implementation very soon, and shortly after it's started, it will probably show up on Github.

TheBrokenRail:
I think it sould be here so you can test offline.

TheBrokenRail:
I have it inside of loadProject so it can act as a fallback like if you entered a invalid project id.

TheBrokenRail:
I moved sb3new.js out of import and into src and renamed it to newProject.js. Ready to merge? I can't figure out how to change the line ending on windows

src/newProject.js
```diff
@@ -1,7 +1,7 @@
-var Sprite = require('../sprites/sprite');
-var Blocks = require('../engine/blocks');
+var Sprite = require('sprites/sprite');
+var Blocks = require('engine/blocks');
```
dekrain: prefix with './'

TheBrokenRail:
Thanks!

dekrain:
TravisCi log tells you use bad line end

TheBrokenRail:
Fixed Newline

TheBrokenRail:
Ready To Merge! (I Couldn't make it do scratch cat by default causes I couldn't find the chatter squirrel svg file to replace with scratch cat.)

dekrain:
I think scratch cat is in scratch-render
Look at: https://github.com/LLK/scratch-render/blob/develop/src/Drawable.js#L115
(I post this before, but in bad topic)

TheBrokenRail:
Ready To Merge! Scratch Cat By Default!

playground/index.html
```diff
@@ -32,8 +32,8 @@
             <pre id="blockexplorer"></pre>
         </div>
         <div id="tab-importexport">
-            Import/Export
-            <input id="projectId" value="119615668" />
+            Import/Export<br />
+            Project ID: <input id="projectId" value="" />
```
tmickel, line 37: Could we leave in 119615668 as the default value, to make it easy to load an example project?

playground/playground.js
tmickel: Is there a reason for that? ^

src/index.js
```diff
@@ -107,6 +108,13 @@ VirtualMachine.prototype.postIOData = function (device, data) {
  * @param {string} json JSON string representing the project.
  */
 VirtualMachine.prototype.loadProject = function (json) {
+    if (!json) {
```
tmickel, line 113: I'd still prefer it if this were outside of `loadProject`, before/after it's called. The reason being that the description of this function - "Load a project from a Scratch 2.0 JSON representation." - says nothing about loading a fallback.

src/newProject.js
```diff
@@ -0,0 +1,21 @@
+var Sprite = require('./sprites/sprite');
+var Blocks = require('./engine/blocks');
+
+function newProject (runtime) {
```
tmickel: Rather than a separate file, it still feels preferable to me to have it as a method on VirtualMachine.prototype (index.js).

tmickel:
Thanks for working on this - makes sense to me to have an offline fallback. Apologies for all the picky comments!

TheBrokenRail:
Ready to Merge?

src/index.js
```diff
@@ -106,6 +106,7 @@ VirtualMachine.prototype.postIOData = function (device, data) {
 /**
  * Load a project from a Scratch 2.0 JSON representation.
  * @param {string} json JSON string representing the project.
```
dekrain: use `@param{?string}`
edit: `@param {?string}`

towerofnix: I'm a nitpick.. wouldn't it be `@param {?string}`

dekrain: Yes, fixed

tmickel:
@TheBrokenRail - to merge a PR, we do need to have all the questions/comments addressed or agreed upon. While I appreciate the excitement to get changes in, we have to balance that with making sure the code is consistent within the project.
1. It would be useful if the project ID in the text input was left to a real value (119615668).
2. `newProject` should be a part of the `VirtualMachine` file, not its own file.
3. Addressing the small issues like the semicolon on playground.js:9 and and what @dekrain and @liam4 mentioned above (since the json string is now nullable).

If you can make those changes, we will hopefully be ready to merge.

TheBrokenRail:
I'm going to need to make a pull request to scratch-render before this will be truly offline. (the svg file is still online)

TheBrokenRail:
I made the pull request for scratch-render

TheBrokenRail:
Once the pull request on scratch-render is merged this should be ready to merge?

tmickel:
@TheBrokenRail - Ok, we'll look again once we get a chance to look on scratch-render :) Thanks for all your work on this.

TheBrokenRail:
'target2.isStage = true;' is so its compatible with pull request #149

TheBrokenRail:
Added Stage Creation

TheBrokenRail:
Ready to merge when scratch-render is merged! (This time with Stage)

TheBrokenRail:
Have You Looked At Scratch-Render Merge Or No Merge? @tmickel

tmickel:
@TheBrokenRail I'm going to merge this and then make some adjustments.

For any future PRs, we won't be able to accept this type of thing (CSS changes) if it's not the main project of the PR (offline support).

PR #143, opened 2016-09-03
More hats

dekrain:


tmickel:
Doesn't seem to quite work for me.

Because these aren't edge-activated hats (i.e., they're started when a key comes in), I think somewhere we need a call to `startHats` (see: https://github.com/LLK/scratch-vm/blob/2b84c8d0fec6e8d31f11bb5177dfc28aad2a290b/src/engine/runtime.js#L266). This needs to be provided somehow to the virtual I/O devices, but we haven't quite put that piece of infrastructure in yet.

dekrain:
Edge activate isn't work so good for now

src/blocks/scratch3_event.js
```diff
@@ -25,7 +25,8 @@ Scratch3EventBlocks.prototype.getHats = function () {
             restartExistingThreads: true
         },
         'event_whenkeypressed': {
-            restartExistingThreads: false
+            restartExistingThreads: false,
+            edgeActivated: true
```
tmickel: But this shouldn't be an edge-activated hat - there's no need to check the predicate on every frame.

dekrain:
Now it working. (Except numbers)
(Phew...)

dekrain:
I forgot to sync it :/

dekrain:
I think scratch cat is in scratch-render

src/io/keyboard.js
```diff
@@ -1,11 +1,16 @@
...
     /**
      * List of currently pressed keys.
      * @type{Array.<number>}
      */
     this._keysPressed = [];
+    /**
+     * For activate hat
+     * @type{Runtime}
+     */
+    this._rt = rt;
```
tmickel: Let's make the argument `runtime` and `this.runtime` to avoid introducing new names.

src/io/keyboard.js
```diff
@@ -1,11 +1,16 @@
...
 
-function Keyboard () {
+function Keyboard (rt) {
     /**
      * List of currently pressed keys.
      * @type{Array.<number>}
      */
     this._keysPressed = [];
+    /**
+     * For activate hat
```
tmickel: "Reference to the owning Runtime. Can be used, for example, to activate hats."

src/io/keyboard.js
```diff
@@ -1,11 +1,16 @@
...
+     * @type{Runtime}
```
tmickel: `@type{!Runtime}`. Runtime can't be null - if it is, it would throw errors when you try to activate hats.

tmickel:
This looks like a good start. Made a couple more style suggestions, other than those I'd be happy to get this in as a first shot. Note also that in Scratch 2.0, if you hold down a keyboard key, the hats is activated over and over (although never if the thread is already running). To do this, I think maybe we could introduce a `step` method on I/O devices that would do the activation there. But we can merge this and treat that as a separate compatibility issue.

src/io/keyboard.js
```diff
@@ -1,11 +1,16 @@
...
 
-function Keyboard () {
+function Keyboard (runtime) {
     /**
      * List of currently pressed keys.
      * @type{Array.<number>}
      */
     this._keysPressed = [];
+    /**
+     * Reference to the owning Runtime. Can be used, for example, to activate hats.
```
tmickel: Looks like the test is actually failing because this line is too long. If we split it into two, it'll be good.

P.S. you may want to set up ESLint in your text editor so it tells you immediately.

src/blocks/scratch3_event.js
```diff
@@ -23,12 +23,13 @@ Scratch3EventBlocks.prototype.getHats = function () {
         'event_whenflagclicked': {
             restartExistingThreads: true
         },
-        /*'event_whenkeypressed': {
+        'event_whenkeypressed': {
             restartExistingThreads: false
         },
         'event_whenthisspriteclicked': {
```
tmickel, line 29: one more thing...this isn't implemented, let's keep it commented?

dekrain:
> P.S. you may want to set up ESLint in your text editor so it tells you immediately.

Thanks.

PR #148, opened 2016-09-06
Create a new input if one doesn't exist

tmickel:
Originally changed this in #135 to fix obscuring a pre-existing shadow block. But, `moveBlock` potentially needs to create a new input object if one doesn't exist. Instead of trying to set properties on a (possibly non-existent) input, this will always copy the old shadow property.

Also adds two tests to make sure the behavior is right.


tmickel:
This should fix the problem Eric was seeing today.

dekrain:
~~When @tmickel/costumes ?~~

tmickel:
@dekrain That's a separate branch from this fix.

thisandagain:
🎉

PR #149, opened 2016-09-06
Stage, costumes, backdrops

tmickel:
Distinguish the stage from other clones (`isStage`) and set it in the SB2 importer. Add a method to runtime to get the stage. Disables some properties from being set on the stage (position, rotation, size).

Implements these blocks:
screen shot 2016-09-06 at 2 16 21 pm


thisandagain:
This reminds me: do we have a issue outstanding for changing the toolbox / palette listing depending on stage / sprite?

thisandagain:
Related: https://github.com/LLK/scratch-blocks/issues/119

PR #150, opened 2016-09-06
Clones

tmickel:
A couple of quirks here to make this work:
- Moving `target` to be a property of `Thread`. Why? Previously `targetForThread` worked by searching all targets' blocks to see if the thread's top block was inside that target. But, now multiple targets may include the same blocks (if a clone is made). So, we can just associate the target with the thread instead, and two threads can be running on the same blocks (but different target clones).
- Added a property used to indicate if a target is "original", i.e., not a clone of an existing clone. That way, you can do things like, when you hit the stop button, remove all clones (but not the original sprites). Or, don't show the clones in the editor's list of sprites.
- I ended up passing `runtime` to Sprite and Clone. This is a little unfortunate. The reason being is we can't start new clones' hats ("when this clone started") until the drawable is initialized. Otherwise, you could do something like when this clone started -> delete this clone (which needs to wait for the drawable to finish being created in order to destroy it)...thoughts on this would be great as I don't love this solution.

Demo:
sep-06-2016 16-26-33
(The squirrel shows up because the renderer momentarily renders a fallback sprite while the real one is possibly coming from the server - something to fix over there, probably).

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -16,7 +16,7 @@ var isPromise = function (value) {
  */
 var execute = function (sequencer, thread) {
     var runtime = sequencer.runtime;
-    var target = runtime.targetForThread(thread);
+    var target = thread.target;
 
     // Current block to execute is the one on the top of the stack.
     var currentBlockId = thread.peekStack();
--- a/src/sprites/clone.js
+++ b/src/sprites/clone.js
@@ -31,8 +31,6 @@ function Clone(sprite, runtime) {
      * @type {?Number}
      */
     this.drawableID = null;
-
-    this.initDrawable();
 }
 util.inherits(Clone, Target);
 
@@ -324,7 +322,6 @@ Clone.prototype.makeClone = function () {
     }
     this.runtime.changeCloneCounter(1);
     var newClone = this.sprite.createClone();
-    newClone.isOriginal = false;
     newClone.x = this.x;
     newClone.y = this.y;
     newClone.direction = this.direction;
@@ -332,6 +329,8 @@ Clone.prototype.makeClone = function () {
     newClone.size = this.size;
     newClone.currentCostume = this.currentCostume;
     newClone.effects = JSON.parse(JSON.stringify(this.effects));
+    newClone.initDrawable();
+    newClone.updateAllDrawableProperties();
     return newClone;
 };
 
--- a/src/sprites/clone.js
+++ b/src/sprites/clone.js
@@ -5,10 +5,12 @@ var Target = require('../engine/target');
 /**
  * Clone (instance) of a sprite.
  * @param {!Sprite} sprite Reference to the sprite.
+ * @param {Runtime} runtime Reference to the runtime.
  * @constructor
  */
-function Clone(sprite) {
+function Clone(sprite, runtime) {
     Target.call(this, sprite.blocks);
+    this.runtime = runtime;
     /**
      * Reference to the sprite that this is a clone of.
      * @type {!Sprite}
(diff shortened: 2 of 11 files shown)


tmickel:
This also doesn't have the clone limit we have in 2.0 - I imagine we'll want to put that in somehow.

thisandagain:
We should probably talk about loading in general sometime soon. I'll add to the 3.0 agenda.

thisandagain:
Filed bug to track clone limit here: https://github.com/LLK/scratch-vm/issues/159

thisandagain:
@cwillisf Any thoughts on the `runtime` / drawable initialization issue here?

tmickel:
Just added the clone limit (#159)

cwillisf:
Hey @tmickel, remember the other day when we were theorizing about whether there would be a situation where the renderer needs to return a `Promise` even when the VM & renderer are on the same thread? I think you may have found it with this change set...

I think when the renderer is asked create a new `Drawable` it should promise a `Drawable` ID in return. That `Promise` should get resolved only once the `Drawable` is ready to render, including SVG rasterization if necessary. We may also want new `Drawable`s to be hidden by default.

I think this model, combined with `createDrawableWithProperties` (LLK/scratch-render#30) would efficiently resolve the temporary squirrel. It might help with the hat triggering trickiness as well: `when I start as a clone` hats could be [tipped](https://goo.gl/LoZuLS) when the `Drawable` ID promise resolves.

src/engine/execute.js
```diff
@@ -16,7 +16,7 @@ var isPromise = function (value) {
  */
 var execute = function (sequencer, thread) {
     var runtime = sequencer.runtime;
-    var target = runtime.targetForThread(thread);
+    var target = thread.target;
```
cwillisf, line 19: This change makes me happy

src/sprites/clone.js
```diff
@@ -235,4 +250,39 @@ Clone.prototype.getName = function () {
     return this.sprite.name;
 };
 
+/**
+ * Create a clone of this clone, copying any run-time properties.
+ * If we've hit the global clone limit, returns null.
+ * @return {?Clone} New clone object, if successful, or null.
+ */
+Clone.prototype.cloneClone = function () {
```
cwillisf: Maybe `makeClone`? OTOH, `cloneClone` does make me smile... ;)

tmickel:
@cwillisf That makes sense to me.

But - why not do the whole thing synchronously, i.e., the renderer should take its time to return from `createDrawable`? Presumably, we will want to download any assets and rasterize costumes before starting any threads at all (when the project loads); or maybe rasterize right on `createDrawable`/`createDrawableWithProperties` (in which case, why not have those synchronous, too)?

cwillisf (approved):
LGTM overall :+1:

cwillisf:
Hmm... that might be possible. I feel like we'll end up in situations where the renderer can't guarantee that a Drawable is ready synchronously. We certainly can't right now since there's no asset sharing in the renderer, but if we implement asset sharing (which we want anyway) and insist that a clone is always created from something that was already render-ready then we might be able to make the guarantee.

If it turns out that we can reasonably make it synchronous then I certainly don't object to doing so.

cwillisf:
(marked as "needs work" due to merge conflicts)

src/engine/execute.js
tmickel: :)

src/sprites/clone.js
tmickel: `clone.cloneClone`

tmickel:
Yeah, that makes sense. I'd like to think more carefully about execution order for clones' scripts if we're starting them asynchronously, too...I'm not sure but I suspect this sort of thing happens synchronously in 2.0. However seems like it could also be not too big of a deal if clones take a few extra frames to initialize sometimes.

For now I think it's pretty easy to switch back and forth between a Promise-based `createDrawable`, so I think I'm happy with it myself.

Issue #15, opened 2016-05-02
Glow behavior when moving around stacks


tmickel:
Scratch-blocks is very simple about what stacks are glowed - you just tell it which ones to do and it will do it. We need to build in some smart behavior to the VM for when you are moving around executing blocks, which stacks should be glowed. For example, if you start executing a stack, let the execution point go to the bottom of the stack, and then pull off the bottom, the bottom will still be executing, when the top half is the highlighted one :(

Scratch 2.0 has a bunch of bugs here also so it may be tricky to get right, but important, I think.

PR #152, opened 2016-09-07
Better glows

### LINKED ISSUE #116: Stack glow issues
After merging #107 and #115, a couple of stack glow issues will crop up:
1. VM tries to glow stacks even if they're not in the currently selected target.
2. VM doesn't reglow stacks when the selected target is changed.
3. Edge-activated hats do a lot of flashing as their predicates are evaluated, even if they're not running blocks.

I'd like to resolve these so that stack glowing feels stable.

tmickel:
This should fix #116 and also possibly #15, an improvement over Scratch 2.0.

Features:
- Appropriately glow only scripts that are on the current editing target (i.e., that scratch-blocks knows about), and update script glows when target is switched.
- Tracks and refreshes glows on a per-frame basis. Should guarantee that a desired glow is visible for at least one frame (e.g., clicking "move 10 steps" will always glow for a frame - previously it would just send glow/unglow at the same time and rely on the browser's slowness).
- Doesn't glow edge-triggered hat blocks unless that script actually executed a command block during the frame, preventing the "flashiness."
- The improvement over Scratch 2.0 and 1.4: glowing follows the script of execution. Did this by tracking the parent in the blocks representation and looking it up when determining glows.
  glows

This prevents the weirdness here, where even though the "forever" isn't highlighted, the cat is spinning:
screen shot 2016-09-07 at 10 03 28 am

### DIFF
--- a/src/engine/adapter.js
+++ b/src/engine/adapter.js
@@ -31,7 +31,7 @@ function domToBlocks (blocksDOM) {
         }
         var tagName = block.name.toLowerCase();
         if (tagName == 'block' || tagName == 'shadow') {
-            domToBlock(block, blocks, true);
+            domToBlock(block, blocks, true, null);
         }
     }
     // Flatten blocks object into a list.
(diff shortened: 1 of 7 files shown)


thisandagain:
This is great. Thanks for fixing up the stack / script language as well.

Issue #89, opened 2016-07-05
Implement `interp.redraw` equivalent


tmickel:
In Scratch 2.0, there's a global interpreter flag that tracks whether a block has triggered a sprite's drawn representation to become out-of-date. E.g, a "move 10 steps" block would trigger redraw, but a "set var to 20" wouldn't. The value of this flag determines whether or not a loop yields at the end of a frame (if the interpreter isn't in turbo mode). We need an equivalent flag in scratch-vm, and we can also use something similar to determine whether we should call the renderer's `draw` method.

https://github.com/LLK/scratch-flash/blob/master/src/interpreter/Interpreter.as#L84

PR #154, opened 2016-09-07
Implement glide

tmickel:
glide

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -1,4 +1,6 @@
+var Cast = require('../util/cast');
 var MathUtil = require('../util/math-util');
+var Timer = require('../util/timer');
 
 function Scratch3MotionBlocks(runtime) {
     /**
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_motion.js
```diff
@@ -52,6 +55,41 @@ Scratch3MotionBlocks.prototype.pointInDirection = function (args, util) {
...
+        util.stackFrame.startX = util.target.x;
+        util.stackFrame.startY = util.target.y;
+        util.stackFrame.endX = Cast.toNumber(args.X);
+        util.stackFrame.endY = Cast.toNumber(args.Y);
+        if (util.stackFrame.duration <= 0) {
+            // Duration too short to glide.
+            util.target.setXY(util.stackFrame.endX, util.stackFrame.endY);
+            return;
+        }
+        util.yieldFrame();
```
towerofnix, line 73: Note the yielding is _after_.. so I guess "glide 0 secs" won't yield?

tmickel: Hmmm, thanks. I think that's a mistake, although we also need to address #89 of course as a broader yielding issue..

tmickel: Actually, after looking at it further, I think this is right:
https://scratch.mit.edu/projects/120528894/#editor

`yieldFrame` here means pause the thread's execution for a frame and come back to the block on the next frame (_not_ the same as `interp.redraw` in 2.0).

I think if we moved it up to even yield when glide 0s, the behavior would be different from 2.0. In the above project, 3 frames would pass before the turn.

PR #155, opened 2016-09-07
More color utilities, touching color, color touching color

tmickel:
Demo (https://scratch.mit.edu/projects/120343713/):
color-touching

This is mostly a bunch of utilities to convert colors from one format to another. Since we are keeping things in either hex format (scratch-blocks) or an RGB number (if the user puts in a math block). For completeness/future convenience I added the rest of the color utilities to go between RGB, hex, and decimal colors. Then it's just calls to the renderer.

@cwillisf

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -1,3 +1,5 @@
+var Cast = require('../util/cast');
+
 function Scratch3SensingBlocks(runtime) {
     /**
      * The runtime instantiating this block package.
@@ -12,6 +14,9 @@ function Scratch3SensingBlocks(runtime) {
  */
 Scratch3SensingBlocks.prototype.getPrimitives = function() {
     return {
+        'colour_picker': this.colorPicker,
+        'sensing_touchingcolor': this.touchingColor,
+        'sensing_coloristouchingcolor': this.colorTouchingColor,
         'sensing_timer': this.getTimer,
         'sensing_resettimer': this.resetTimer,
         'sensing_mousex': this.getMouseX,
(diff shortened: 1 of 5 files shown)


src/util/color.js
```diff
@@ -1,17 +1,79 @@
...
+ * CC-BY-SA
+ * https://stackoverflow.com/questions/29241442/
+ * decimal-to-rgb-in-javascript-and-php
+ * @param {number} decimal RGB color as decimal.
+ * @returns {Object} {r: R, g: G, b: B}, values between 0-255
+ */
+Color.decimalToRgb = function (decimal) {
+    var r = Math.floor(decimal / (256 * 256));
+    var g = Math.floor(decimal / 256) % 256;
+    var b = decimal % 256;
```
cwillisf: Is there an advantage to doing it this way instead of with bitwise operators? I usually do something like this:

``` js
var r = (decimal >> 16) & 0xFF;
var g = (decimal >> 8) & 0xFF;
var b = decimal & 0xFF;
```

The one potential downside I can think of is that the bitwise operators always result in integers, preventing a result like `{r: 123.45, g: 2, b: 4}`, but if that were a problem then `Math.floor` would be undesirable as well.

cwillisf:
LGTM. We can always change the renderer's API to be more "friendly" toward the color values we have at runtime, though IMO sanitization of the colors should remain outside the renderer.

src/util/color.js
tmickel: No reason - because I wasn't thinking too carefully about it and just grabbed a snippet :) I prefer your version for sure. The decimal number shouldn't have more than 24 bits anyway so I think there should never be fractional results.

thisandagain:
Agreed that validation / sanitization of the colors should happen within the VM. Setting this to `needs-work` given @cwillisf 's feedback.

tmickel:
Updated!

src/blocks/scratch3_sensing.js
```diff
@@ -24,6 +29,21 @@ Scratch3SensingBlocks.prototype.getPrimitives = function() {
     };
 };
 
+Scratch3SensingBlocks.prototype.colorPicker = function (args) {
+    return args.COLOUR;
+};
```
tmickel, line 34: will remove after merging #160

PR #156, opened 2016-09-07
When clicked hats

tmickel:
Since the renderer uses canvas coordinates for `pick` (rather than Scratch coordinates), this required passing in canvas coordinates to the VM. I might rather change the renderer to use Scratch coordinates for `pick`, esp. when we consider doing things like the "touching mouse pointer" query. Thoughts @cwillisf?

### DIFF
--- a/src/io/mouse.js
+++ b/src/io/mouse.js
@@ -1,20 +1,47 @@
 var MathUtil = require('../util/math-util');
 
-function Mouse () {
+function Mouse (runtime) {
     this._x = 0;
     this._y = 0;
     this._isDown = false;
+    /**
+     * Reference to the owning Runtime.
+     * Can be used, for example, to activate hats.
+     * @type{!Runtime}
+     */
+    this.runtime = runtime;
 }
 
 Mouse.prototype.postData = function(data) {
     if (data.x) {
-        this._x = data.x;
+        this._x = data.x - data.canvasWidth / 2;
     }
     if (data.y) {
-        this._y = data.y;
+        this._y = data.y - data.canvasHeight / 2;
     }
     if (typeof data.isDown !== 'undefined') {
         this._isDown = data.isDown;
+        if (this._isDown) {
+            this._activateClickHats(data.x, data.y);
+        }
+    }
+};
+
+Mouse.prototype._activateClickHats = function (x, y) {
+    if (self.renderer) {
+        var pickPromise = self.renderer.pick(x, y);
+        var instance = this;
+        pickPromise.then(function(drawableID) {
+            for (var i = 0; i < instance.runtime.targets.length; i++) {
+                var target = instance.runtime.targets[i];
+                if (target.hasOwnProperty('drawableID') &&
+                    target.drawableID == drawableID) {
+                    instance.runtime.startHats('event_whenthisspriteclicked',
+                        null, target);
+                    return;
+                }
+            }
+        });
     }
 };
 
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -152,17 +152,35 @@ window.onload = function() {
     document.addEventListener('mousemove', function (e) {
         var rect = canvas.getBoundingClientRect();
         var coordinates = {
-            x: e.clientX - rect.left - rect.width / 2,
-            y: e.clientY - rect.top - rect.height / 2
+            x: e.clientX - rect.left,
+            y: e.clientY - rect.top,
+            canvasWidth: rect.width,
+            canvasHeight: rect.height
         };
         window.vm.postIOData('mouse', coordinates);
     });
     canvas.addEventListener('mousedown', function (e) {
-        window.vm.postIOData('mouse', {isDown: true});
+        var rect = canvas.getBoundingClientRect();
+        var data = {
+            isDown: true,
+            x: e.clientX - rect.left,
+            y: e.clientY - rect.top,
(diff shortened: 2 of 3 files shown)


cwillisf:
Yeah, I think you have a point: it would be less confusing to keep everything in the same coordinate system. As long as we don't truncate to integer I don't think there's any significant downside to expressing the `pick` coordinates in Scratch space instead of canvas space.

src/io/mouse.js
```diff
@@ -1,20 +1,47 @@
...
     }
     if (typeof data.isDown !== 'undefined') {
         this._isDown = data.isDown;
+        if (this._isDown) {
+            this._activateClickHats(data.x, data.y);
+        }
+    }
+};
+
+Mouse.prototype._activateClickHats = function (x, y) {
```
cwillisf, line 30: I feel like this code should be at a higher level. Maybe `Mouse` reports a click event to the VM and the VM responds by doing this. In theory we could have other IO Devices which generate click events, and I would want those click events to trigger click hats as well.

tmickel: Do you think the I/O devices should just be event emitters for this (and possible other?) purposes?

cwillisf: Hmm... that seems like a good idea at first glance. Can you think of any downsides?

Of course, thinking about that makes me wonder:
There's not a whole lot of difference between the object being sent using `postMessage` and the object being emitted by an `EventEmitter` (or similar). Would it be oversimplifying things to replace the VM side of the I/O device paradigm with something that translates certain posted messages into events emitted by a central "I/O Devices" event emitter?

I think that would work well for mouse and keyboard, but not for the clock. Are there other I/O devices that wouldn't work well with this model? If it works for everything except the clock then maybe the clock needs to be a different type of thing. On the other hand, I don't want to remove meaningful power or flexibility just because it makes the system simpler...

tmickel: To me it's a question of, how useful do you think the "I/O devices" abstraction is?

Here are some other cases to think about:
- Video sensing: where do you want to do this processing? Specifically something is going to need to receive frames (or some preprocessed data) from outside the worker and map it into some data accessible by block primitives (global video motion for when motion > _, and on-time reporter values).
- Microphone sensing: I think it's the same situation, more or less.
- Variables: we need a place to put variables and lists, which can be global to the project or local to the clone, or cloud variables which require I/O.
- The renderer, the audio engine: as we're discovering with Eric, it would seem good to have a centralized message passing system for these (and presumably cloud variables, etc.)

Regarding your original question: what other things, other than the mouse, are going to produce click hats? I could imagine touch events. If we want to support those, would it be better just to rename Mouse to Pointer or something, and have it handle touch events too? Or just duplicate the click-hat event code in a Touch device? Or should we have one top-level object handling all the logic for creating all these events, as you might suggest?

cwillisf: I think I'm starting to come around to what I suspect is your point of view. I think I had a different "meaning" for I/O device than you had, and I got confused when you followed your mental model instead of mine ;)

I had been thinking of the I/O devices as a low-level data source, and that concept gets pretty weak when you introduce a central message dispatcher. That's where my "I feel like this code should be at a higher level" comment came from, as well as my thoughts about translating messages into events.

However, if you think of an I/O device class as a place to handle data related to a particular device, then it makes sense to have this hat block code in there. The message passing is there sort of as a consequence of the fact that we don't currently have another way to handle that. If we add a central message dispatcher or even stop using a worker (#165) for the VM, then this class is still valuable as a place to put mouse- (or pointer-) specific work.

I think I still like the idea of centralizing message-to-event conversion if it turns out to make sense, but I now think that would just clarify the purpose of `src/io/*.js` instead of replacing them.

Regarding the cases you mentioned:
- Video sensing: I suspect we'll want to detect motion on the main thread in order to avoid passing the whole video stream as messages. Once the data gets to the VM side, it could be caught by a "video I/O" device which then dispatches Scratch events if appropriate.
- Microphone sensing: Similar, as you suggested.
- Variables: `src/io/cloud.js`? It feels weird initially but really it's doing the same sort of thing. It makes me feel like "I/O device" might not be the right term for these.
- Audio & stage rendering: perhaps the "remote control" (`RenderWebGLWorker.js` for example) should be an I/O device, or should be wrapped by a thin I/O device? I could see that making sense.

Sorry for the disconnect. Does this seem like we're closer to being on the same page?

tmickel: I think that all makes sense to me. Yeah, I was always thinking of it as a place for "raw data" from a particular device/domain to get processed into "Scratch-consumable-data" and events. Probably I/O device is not the best name here, I'll try to think of a new one sometime...

And agreed re: message passing - it will be good when we resolve a central solution for that or remove the worker.

Video sensing: agreed, passing that across with messages will likely be ugly.
Variables: yeah, the name feels weird there. Almost makes sense if you think of variables as memory/RAM/a hard disk/a Dropbox full of cloud data/etc.
Audio/stage: Yep, not sure. I think the need for that becomes less so if we strip away the worker. With the worker, it seems like we need a place for it.

Also sorry for the disconnect, it's all pretty abstract...

But yeah all in all I agree. Now regarding this PR...any steps forward? :P

tmickel: also thanks for thinking this through in such detail!!

cwillisf:
Looks good. Thanks for your patience with the discussion :)

PR #158, opened 2016-09-07
Key hat repeating and I/O device stepping

tmickel:
Demo project: https://scratch.mit.edu/projects/120369399/

Note that in Scratch 2.0 it goes something like this:
- Key pressed hat is fired once
- There's a small delay (250-300ms?)
- Key pressed hat is fired over and over as long as it's still pressed.

In this implementation, it's just fired over and over as long as it's pressed. Do we need to build the delay in, or is it better this way?

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -355,6 +355,13 @@ Runtime.prototype.stopAll = function () {
  * inactive threads after each iteration.
  */
 Runtime.prototype._step = function () {
+    // Allow I/O devices to step.
+    for (var i in this.ioDevices) {
+        var ioDevice = this.ioDevices[i];
+        if (ioDevice.step) {
+            ioDevice.step();
+        }
+    }
     // Find all edge-activated hats, and add them to threads to be evaluated.
     for (var hatType in this._hats) {
         var hat = this._hats[hatType];
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -356,8 +356,8 @@ Runtime.prototype.stopAll = function () {
  */
 Runtime.prototype._step = function () {
     // Allow I/O devices to step.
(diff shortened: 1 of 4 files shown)


thisandagain:
I think we likely want to keep the delay and mimic the behavior of 2.0 here. The key hat is often used in games. The throttling of the keypress event implicitly enforces the speed of interaction (e.g. shooting lasers, throwing bananas, swinging a pillow, etc.) which is sometimes used as part of the mechanics of the game (don't let the user shoot ∞ lasers). 

@cwillisf @carljbowman Any thoughts?

tmickel:
Actually, this seems to me to be affected by the system settings:
screen shot 2016-09-12 at 1 18 05 pm
I.e., it's not Scratch implementing this delay, but some interaction of typing keys and Flash...not sure if we can get that info in JavaScript or not?

thisandagain:
If it's the OS doing the throttling and not the `scratch-flash` code then it seems fine to be consistent here. 👍

tmickel:
Hmmm, looking at it even further, it looks like the browser may trigger keydown events according to the OS's key repeat settings, just like Flash. In that case I think we should trigger the hats when the event comes in, not on every step. So I might rewrite this to keep the behavior the same... :P

cwillisf:
I think Scratch 2.0 is hooking a different event. Rather than checking if the key is down, it catches the key repeat event which the OS controls. It looks like this is called "keypress" in JavaScript, as opposed to "keydown"

tmickel:
@cwillisf Yep. keydown also repeats, actually, and I think is preferable because it includes non-character keys (shift, backspace, etc.) Just need to start hats every time the event is fired, instead of the original which only did when it became pressed.

cwillisf:
Ah, good to know! Thanks :)

Issue #146, opened 2016-09-06
Simplify evaluation of shadow blocks


tmickel:
Currently, block implementations have several places with code like this:

> ```
> Scratch3LooksBlocks.prototype.effectMenu = function (args) {
>     return args.EFFECT.toLowerCase();
> };
> ```

That's because the shadow block (menu, or number/string input, etc.) needs to be evaluated by the interpreter. Above, we're treating it just like any block implementation. But I think it would reduce code complexity to do:
- During execution, if we encounter a block with no op function,
- If the block has a single field, and no inputs,
- Immediately return the value of the field.

I think that would be pretty safe, can be overridden simply (by providing an op function), and would let us code a bunch of duplicated boilerplate code.

PR #160, opened 2016-09-08
Better shadow evaluation

tmickel:
An implementation of #146. Allows us to cut out meaningless primitives for the shadow blocks.

Required moving a bit of code to keep using `handleReport` at the top of the function. But the main change is here: https://github.com/LLK/scratch-vm/compare/develop...tmickel:feature/shadow-eval?expand=1#diff-6175aeaffc8aee750c94d0db41b34718R75


tmickel:
Since there was a lot of implicit casting work being done by these block implementations, I've updated the primitives to do it in each block implementation.

thisandagain:
Nice. This looks good.

Issue #91, opened 2016-07-13
Deleting a script while it's running is broken


tmickel:
Currently it basically crashes the entire VM until you stop it manually by clicking the button.

TheBrokenRail:
If it is fixed shouldn't this issue be closed

tmickel:
Yep, missed closing the issue - thanks @TheBrokenRail !

PR #162, opened 2016-09-08
Cleanly handle deleting running scripts

tmickel:
Fix for #91.

Two parts:
1. When a block disappears (i.e., it no longer exists in the VM representation) but a thread is still trying to work on it (i.e., because it's in the thread's stack), retire the thread.
2. When a top-level block (script) gets deleted, remove it from the runtime's glow tracking, so we don't emit an extraneous unglow event.


thisandagain (approved):
Looks good. Sorry for the slow review ... I wanted to make sure I had a good understanding of how the "glow" stuff worked here and in the blocks.

Issue #56, opened 2016-06-01
Use more accurate timing with performance.now()


tmickel:
See: https://developer.mozilla.org/en-US/docs/Web/API/Performance/now

Should be polyfilled when not available.

thisandagain:
https://github.com/nicjansma/usertiming.js appears to be a decent / maintained polyfill.

Also: https://www.npmjs.com/package/perfnow

PR #163, opened 2016-09-08
Augment Timer with documentation and `self.performance.now`

tmickel:
Resolving #56 and making it more clear what each method of Timer does.

### DIFF
--- a/src/util/timer.js
+++ b/src/util/timer.js
@@ -1,20 +1,69 @@
 /**
- * Constructor
+ * @fileoverview
+ * A utility for accurately measuring time.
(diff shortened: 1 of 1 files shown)


thisandagain:
This looks good though it doesn't hande situations where `performance.now` is available, but with a vendor prefix (see: https://github.com/daniellmb/perfnow.js/blob/master/perfnow.js#L13).

tmickel:
@thisandagain Yep - although I think all of those should be under our target browser versions, and this has a fallback too. If we want we can add it.

PR #168, opened 2016-09-12
Key repeats, with OS delay

tmickel:
Here's actually a much simpler version, too. Instead of needing a `step`, since the keydown event is actually repeated, we can just make sure to start the hats every time the event comes in. Works just like Scratch 2.0.


thisandagain:
🎉

PR #171, opened 2016-09-12
Added 'New Project' Button

TheBrokenRail:
Allows you to create a empty project with out messing with the URL or turning your internet off

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -30,10 +30,10 @@ window.onload = function() {
         document.location = '#' + document.getElementById('projectId').value;
         location.reload();
     };
-    document.getElementById('createEmptyProject').onclick = function () {
+    document.getElementById('createEmptyProject').addEventListener('click', function() {
         document.location = '#' + 'createEmptyProject';
         location.reload();
-    };
+    });
     loadProject();
 
     // Instantiate the renderer and connect it to the VM.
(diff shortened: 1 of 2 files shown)


tmickel:
Given the point of the empty project being offline VM development, what do we need this for?

TheBrokenRail:
It could allow you to test with a blank project without creating a blank project on scratch.mit.edu

playground/playground.js
```diff
@@ -30,6 +30,10 @@ window.onload = function() {
         document.location = '#' + document.getElementById('projectId').value;
         location.reload();
     };
+    document.getElementById('createEmptyProject').onclick = function () {
```
towerofnix: Looks like `addEventListener` is [generally](https://github.com/TheBrokenRail/scratch-vm/blob/736cfb9145d9a959930e99d7d706a0abfe474ba4/playground/playground.js#L165) [used](https://github.com/TheBrokenRail/scratch-vm/blob/736cfb9145d9a959930e99d7d706a0abfe474ba4/playground/playground.js#L183) [everywhere](https://github.com/TheBrokenRail/scratch-vm/blob/736cfb9145d9a959930e99d7d706a0abfe474ba4/playground/playground.js#L194) [else](https://github.com/TheBrokenRail/scratch-vm/blob/736cfb9145d9a959930e99d7d706a0abfe474ba4/playground/playground.js#L194) (rather than directly setting the `on<event>` method).. maybe it should be used here too?

tmickel:
Yep - can we switch to `addEventListener` for consistency? Also for consistency, can we move the button to the Import/Export tab?

PR #175, opened 2016-09-13
Drop worker support

tmickel:
This was pretty easy - just had to fix one issue with the Blockly event listener (which was previously only defined for the worker).

The "VM lock problem project" now runs correctly, since touching queries become synchronous:
sep-13-2016 14-29-21

cc: @cwillisf @ericrosenbaum


cwillisf:
LGTM... glad this was so simple!

thisandagain:
That was quick! Thanks @tmickel this looks great.

PR #176, opened 2016-09-13
Better handling of float values (fix #172)

### LINKED ISSUE #172: Numbers: float vs int
Scratch 2.0 and the current VM are handling this case differently:

screen shot 2016-09-12 at 5 20 48 pm

In 2.0, the presence of any decimal point - "1.", "1.0", ".0", etc.- will cause the interpreter to label the number as a non-integer. "Pick random" then picks floats or rounded ints accordingly.

The current VM implementation is based on if there's an actual value, so:
screen shot 2016-09-12 at 5 22 39 pm
is treated as "give me an int from 1 to 10."

The sane way to deal with this case is probably to provide a method in `Cast` that's something like `Cast.isInt` that checks the string for a decimal point. That would replace this (too basic) check here: https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_operators.js#L84

tmickel:
Adds `Cast.isInt` which can be applied to uncasted arguments to determine if it's an int. I think I captured all the current Scratch behavior, but not 100% sure.

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -109,16 +109,16 @@ Cast.compare = function (v1, v2) {
  */
 Cast.isInt = function (val) {
     // Values that are already numbers.
-    if (typeof val == 'number') {
+    if (typeof val === 'number') {
         if (isNaN(val)) { // NaN is considered an integer.
             return true;
         }
         // True if it's "round" (e.g., 2.0 and 2).
         return val == parseInt(val);
-    } else if (typeof val == 'boolean') {
+    } else if (typeof val === 'boolean') {
         // `True` and `false` always represent integer after Scratch cast.
         return true;
-    } else if (typeof val == 'string') {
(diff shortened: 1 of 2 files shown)


src/util/cast.js
```diff
@@ -102,4 +102,27 @@ Cast.compare = function (v1, v2) {
...
 };
 
+/**
+ * Determine if a Scratch argument number represents a round integer.
+ * @param {*} val Value to check.
+ * @return {boolean} True if number looks like an integer.
+ */
+Cast.isInt = function (val) {
+    // Values that are already numbers.
+    if (typeof val == 'number') {
```
thisandagain: Strict equality for `typeof` checks?

thisandagain:
One minor comment, otherwise this looks good. This would be another place where unit testing would be helpful.

Issue #173, opened 2016-09-13
We need to do some kind of escaping when blocks are output


tmickel:
To make a problem:
- Create a "say" block.
- Type "<block>" in the text input field.
- Switch to a different sprite.
- Switch back to the original sprite.

The blocks will be broken, because invalid XML was output from the VM.

PR #177, opened 2016-09-13
Escape field values for XML-unsafe characters

tmickel:
A fix for #173.

Now should handle things like this just fine:
screen shot 2016-09-13 at 3 48 53 pm


thisandagain:
Looks good. Mind adding an issue for test coverage?

PR #186, opened 2016-09-15
Add Saving And Loading To JSON (The JSON Inside The Compressed File)

TheBrokenRail:
Works With Multiple Sprites


TheBrokenRail:
Any comments? Ready to merge?

tmickel:
I'd like to hear @thisandagain's thoughts. But I'm pretty sure the Scratch team will be designing an SB3 format within the coming months, so I'm not sure if continued effort on this is likely to be merged right now. We have many concerns other than just importing/exporting projects - such as compatibility with old project files, our scratch.mit.edu infrastructure, and existing tools for handing projects there. In addition, an SB3 format will likely need to include some kind of compression (e.g., in SB2 they are .zip files that make the data smaller), and a means to store assets (including sounds).

TheBrokenRail:
Currently this is just the json part that goes into the compressed file

tmickel:
Yes, but even the format of the JSON file itself is up for design. That part potentially needs to be compatible with tools and infrastructure, too.

TheBrokenRail:
Do you think it is compatible?

thisandagain:
Thanks @TheBrokenRail but I'd prefer we discuss this as a group before charging forward with an implementation. I've created an issue to track this here:
https://github.com/LLK/scratch-vm/issues/194

PR #187, opened 2016-09-16
Variables and lists

tmickel:
Here's a start of variables and lists.

I think it should support everything Scratch 2.0 did, minus cloud variables and bugs (and creating/deleting/renaming variables from the blocks).

There were some quirks in how the list blocks in 2.0 works, e.g.:
https://github.com/LLK/scratch-vm/compare/develop...tmickel:feature/variables?expand=1#diff-95b1e86eb11427aa2b46e69618351f62R83 because of https://github.com/LLK/scratch-flash/blob/master/src/primitives/ListPrims.as#L88 for delete and https://github.com/LLK/scratch-flash/blob/master/src/primitives/ListPrims.as#L181 for the rest of them. But I think I captured most/all of the quirks.

Variables and lists are stored on the clone itself, just like scratch-flash. But I could be argued into some central variable store, too, if anyone would prefer that.


src/sprites/clone.js
```diff
@@ -31,6 +33,18 @@ function Clone(sprite, runtime) {
...
+     * Key is the variable name.
+     * @type {Object.<string,*>}
+     */
+    this.variables = {};
+    /**
+     * Dictionary of lists and their contents for this clone.
+     * Key is the list name.
+     * @type {Object.<string,*>}
+     */
+    this.lists = {};
```
tmickel: Maybe these and `lookupOrCreateVariable`, `lookupOrCreateList` should actually be provided on the superclass, `Target`?

src/blocks/scratch3_data.js
```diff
@@ -0,0 +1,180 @@
...
+ * Compute a 1-based index into a list, based on a Scratch argument.
+ * Two special cases may be returned:
+ * LIST_ALL: if the block is referring to all of the items in the list.
+ * LIST_INVALID: if the index was invalid in any way.
+ * @param {*} index Scratch arg, including 1-based numbers or special cases.
+ * @param {number} length Length of the list.
+ * @param {boolean} useRound If set, Math.round (not Math.floor for 2.0 compat).
+ * @return {(number|string)} 1-based index for list, LIST_ALL, or LIST_INVALID.
+ */
+Scratch3DataBlocks.prototype._computeIndex = function (
```
tmickel: Could provide this on `List` instead...

src/blocks/scratch3_data.js
```diff
@@ -0,0 +1,180 @@
...
+Scratch3DataBlocks.prototype.setVariableTo = function (args, util) {
+    var variable = util.target.lookupOrCreateVariable(args.VARIABLE);
+    variable.value = args.VALUE;
+};
+
+Scratch3DataBlocks.prototype.changeVariableBy = function (args, util) {
+    var variable = util.target.lookupOrCreateVariable(args.VARIABLE);
+    var castedValue = Cast.toNumber(variable.value);
+    var dValue = Cast.toNumber(args.VALUE);
+    variable.value = castedValue + dValue;
```
tmickel, line 45: Are we comfortable with setting these properties on `Variable` (and `list.contents`) vs. providing methods somewhere? if not, where?

src/sprites/clone.js
tmickel: Decided to go for that - completely imaginable that we could have unrendered objects that have variables and lists as well as blocks.

src/blocks/scratch3_data.js
tmickel: Decided it actually makes most sense to me with `Cast`.

src/blocks/scratch3_data.js
cwillisf: This seems fine to me

src/blocks/scratch3_data.js
```diff
@@ -0,0 +1,180 @@
...
+};
+
+Scratch3DataBlocks.prototype.getList = function (args, util) {
```
cwillisf: I think I would call this `getListContents` or `joinList` or `getListAsString` or... I'm not sure what :/
The name `getList` sounds to me like it would retrieve the list itself, though. Like it would be `getVariable` but for lists, which it kinda-sorta is, but I think I would expect it to return an array.

src/import/sb2import.js
```diff
@@ -90,8 +113,8 @@ function parseScratchObject (object, runtime, topLevel) {
     target.isStage = topLevel;
     // The stage will have child objects; recursively process them.
     if (object.children) {
-        for (var j = 0; j < object.children.length; j++) {
-            parseScratchObject(object.children[j], runtime, false);
+        for (var m = 0; m < object.children.length; m++) {
+            parseScratchObject(object.children[m], runtime, false);
```
cwillisf, line 118: This sort of thing makes me want `let`

src/util/cast.js
```diff
@@ -125,4 +125,48 @@ Cast.isInt = function (val) {
...
+            return Cast.LIST_ALL;
+        }
+        if (index == 'last') {
+            if (length > 0) {
+                return length;
+            }
+            return Cast.LIST_INVALID;
+        } else if (index == 'random' || index == 'any') {
+            if (length > 0) {
+                return 1 + Math.floor(Math.random() * length);
```
cwillisf, line 153: 1-based indexing messes with my head. I had to read this a few times before I convinced myself that it's correct. 😵

src/util/cast.js
```diff
@@ -125,4 +125,48 @@ Cast.isInt = function (val) {
...
+            }
+            return Cast.LIST_INVALID;
+        }
+        if (isNaN(Number(index))) {
```
cwillisf: Should this be based on `Cast.toNumber(index)` instead of `Number(index)`?

cwillisf (changes requested):
There should be a "request discussion" option rather than "request changes" :)

src/blocks/scratch3_data.js
tmickel: Agreed - I think getListAsString might be preferable. The current way was matching the block opcode, `data_list` - but maybe that should be changed as well?

src/util/cast.js
tmickel: Yeah, agreed. My basic approach was reproducing https://github.com/LLK/scratch-flash/blob/master/src/primitives/ListPrims.as#L174 but with some minor improvements, and then checking that it works as expected in the blocks. If you have ideas about how to write it more clearly, I think that would be welcome :P

src/util/cast.js
tmickel: Only if this should return "thing", it currently doesn't:
screen shot 2016-09-19 at 5 16 55 pm

tmickel: Maybe...`Cast.toNumber` will coerce NaN to 0, which is one potential problem. But I'm struggling to come up with a case where this would cause things to execute differently.

tmickel: Actually I'm thinking this entire check should be unnecessary.

If it's a non-number castable value (e.g., "hi"), in the code below, index will get put to 0. And then `Cast.LIST_INVALID` will be returned because it's out of bounds...thoughts?

src/blocks/scratch3_data.js
tmickel: Updated to `getListAsString`

cwillisf: Ah, that makes sense. The 2.0 opcode is `contentsOfList:` so it might make sense to go with something involving "contents" for both the opcode and the function name.

Edit: Oops, I didn't refresh to see your `getListAsString` comment before I said the above. Sorry!

tmickel: Hmm, ok :P Maybe we should think about in general at some point what 3.0 opcodes and prim function names should be.

src/util/cast.js
cwillisf: I think it's fine... you explicitly mention 1-based indexing in the function header comment, which I think should serve as fair warning to those looking at these expressions :)

src/util/cast.js
cwillisf: I did some experimenting with 2.0 and I think you're right that this check is unnecessary. I was concerned about values like '42foo' but testing seems to suggest that my concerns were unfounded.

tmickel: Yep, I think I've convinced myself. I think the isNaN(Number()) test is actually necessary in 2.0 because the value is never coerced to a number there (i.e., no equivalent of `Cast.toNumber` ever happens). Because we now do `Cast.toNumber` below, and that will force any non-number-like values to 0, which then gets returned as Cast.LIST_INVALID... I think it ends up being equivalent. I'll take it out.

src/blocks/scratch3_data.js
```diff
@@ -0,0 +1,136 @@
...
+Scratch3DataBlocks.prototype.setVariableTo = function (args, util) {
+    var variable = util.target.lookupOrCreateVariable(args.VARIABLE);
+    variable.value = args.VALUE;
+};
+
+Scratch3DataBlocks.prototype.changeVariableBy = function (args, util) {
+    var variable = util.target.lookupOrCreateVariable(args.VARIABLE);
+    var castedValue = Cast.toNumber(variable.value);
+    var dValue = Cast.toNumber(args.VALUE);
+    variable.value = castedValue + dValue;
```
tmickel, line 45: Just noticed another casting weirdness here in 2.0:
https://github.com/LLK/scratch-flash/blob/master/src/interpreter/Interpreter.as#L715

Because it uses `Number` and not `numarg` (equivalent to our `Cast.toNumber`), this script:
screen shot 2016-09-19 at 5 54 25 pm
Sets `lol` to NaN (not to 1). In the implementation I gave, I think `lol` would be 1.... should we reproduce the 2.0 way?

tmickel: FWIW, in 1.4 the value would be 1. Maybe we should consider it a 2.0 bug.

src/util/cast.js
```diff
@@ -125,4 +125,45 @@ Cast.isInt = function (val) {
...
+            }
+            return Cast.LIST_INVALID;
+        } else if (index == 'random' || index == 'any') {
+            if (length > 0) {
+                return 1 + Math.floor(Math.random() * length);
+            }
+            return Cast.LIST_INVALID;
+        }
+    }
+    if (useRound) {
```
tmickel, line 158: Also, in 1.4 the behavior of all list primitives is to use `floor` for the list index, not `round`. Maybe we could ignore that 2.0 uses `round` for delete?

src/blocks/scratch3_data.js
cwillisf: Ah, good catch. Hmm... this is a tough one. I think my vote is to diverge from Scratch 2.0 here.

My justifications:
- I don't think very many existing projects will be affected by such a change. That is, I doubt very many working projects meaningfully use a `NaN` generated in this way.
- I'm inclined to consider it a bug that these two stacks produce different results: 
  image

src/util/cast.js
cwillisf: That's tempting, but I'm less confident about compatibility fallout than I am about the "Change {var} by" NaN thing above. I also wonder if switching to `round` was an intentional change. I'd suggest more investigation on this one -- maybe it should be filed as an "enhancement" bug?

src/import/sb2import.js
towerofnix: 👍

#3, eventually..

src/util/cast.js
tmickel: Fair enough. I'll make an issue to investigate.

cwillisf (approved):

src/import/sb2import.js
jeandrek: Why don't you use `let`?

tmickel: @Jonathan50   Browser support - Safari 9, IE < 11, Android Browser 4.4, etc. all don't support `let`. But we hope to in the future...

thisandagain:
This looks great. Follow-up on the inconsistencies and questions raised here are probably best to continue in individual issues like #202 . 👍  to land.

PR #188, opened 2016-09-17
Fix Bug With Create New Project And Make Scratch Embedable

TheBrokenRail:
The Bug Was If You Load scratch-vm NOT in the root of a website it would be unable to load assets as it would be looking from root/assets and not root/scratch-vm/assets/


tmickel:
We're really intending the playground to be used to develop the scratch-vm, not to make Scratch projects. We'll be coming out with a repository for Scratch's main GUI soon, as well as a system for loading/saving assets in general between the scratch-vm, scratch-render, and scratch-gui components. Once those are in progress, your contributions would be very welcome in those areas! But while I appreciate your effort on this, I think this PR probably doesn't make sense to merge right now.

PR #189, opened 2016-09-17
No Resize Textarea

TheBrokenRail:


tmickel:
Hmm, I'm not sure why? It doesn't seem like a problem to me to have it resizable.

PR #196, opened 2016-09-19
Fix up property imports from SB2

### LINKED ISSUE #185: Sprite hidden status doesn't seem to load from SB2
Test project: https://scratch.mit.edu/projects/121401203/
Playground link: http://localhost:8080/playground/#121401203

tmickel:
Should fix #185.

There were two problems:
1. Drawable properties weren't being refreshed on a target's creation.
2. If any property was "false"-like (e.g., visible: false), it wasn't copied to the target.


thisandagain (approved):
Looks good!

PR #198, opened 2016-09-19
Fix/improve playground threads display

tmickel:
Implement #174, and filter out a circular data reference that was preventing JSON serialization.


thisandagain (approved):
Nice use of the `stringify` replacer argument. Thanks.

PR #200, opened 2016-09-20
Scope renderer to instance

rschamp:
This allows usage without global scope to attach a renderer to the VM. It also provides the ability to have multiple VMs/renderers to be used at once.


src/engine/runtime.js
```diff
@@ -19,14 +19,21 @@ var defaultBlockPackages = {
...
     // Bind event emitter
     EventEmitter.call(this);
 
     // State for the runtime
 
     /**
+     * Renderer
+     * @type {!RenderWebGL}
+     */
+    this.renderer = renderer;
```
rschamp: All the things that needed access to the renderer had access to the runtime, so I decided to make it a property of the runtime.  It seems like an ok thing for the runtime to have since it manages the "stuff", but I am open to suggestions about a better place to put this.

src/index.js
```diff
@@ -10,16 +10,17 @@ var Blocks = require('./engine/blocks');
  * Handles connections between blocks, stage, and extensions.
  *
  * @author Andrew Sliwinski <ascii@media.mit.edu>
+ * @param {!RenderWebGL} renderer Renderer for the VM
  */
-function VirtualMachine () {
+function VirtualMachine (renderer) {
```
tmickel: Would maybe prefer it I think if this were an optional parameter (`opt_renderer`, `@param {RenderWebGL=}`) or if it was set after instantiation (`vm.attachRenderer`)? I'm imagining we might sometimes want a VM with no renderer.

src/engine/runtime.js
tmickel: Having it a property of runtime seems fine to me. But I also think it should be optional/possible to be missing.

rschamp:
@tmickel I went the `attachRenderer` route which resolved both issues.

I also took out the global scope for the VM to get rid of the remaining globals.  The package that webpack builds still exposes `VirtualMachine` to `window`, but if you import `scratch-vm` nothing gets attached to `window`.

PR #212, opened 2016-09-22
Mutations in block representation; an unfeatured procedure call

### LINKED ISSUE #211: Keep mutations in the block representation
This seems to be necessary to implement things like the "stop" block and procedures. We'll need to keep mutation field in the block representation and either the XML, or some object conversion of it.

tmickel:
This allows us to import some procedures from SB2s.

Procedures with parameters don't import very well, because right now there's no way to represent them in Blockly/scratch-blocks. You'd get something like this:
screen shot 2016-09-22 at 2 28 10 pm

But for a project like this, with simple named procedures, https://scratch.mit.edu/projects/10015060/ - it kind of works.

screen shot 2016-09-22 at 3 20 49 pm

Things that are not even touched includes:
- Parameters
- Yielding for recursive procedures
- Run without screen refresh...

Fixes #211

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -259,6 +259,7 @@ Blocks.prototype.changeBlock = function (args) {
 
     if (args.element == 'field') {
         // Update block value
+        if (!this._blocks[args.id].fields[args.name]) return;
         this._blocks[args.id].fields[args.name].value = args.value;
     } else if (args.element == 'mutation') {
         this._blocks[args.id].mutation = mutationAdapter(args.value);
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -1,4 +1,5 @@
 var adapter = require('./adapter');
+var mutationAdapter = require('./mutation-adapter');
 var xmlEscape = require('../util/xml-escape');
 
 /**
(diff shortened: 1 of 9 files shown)


TheBrokenRail:
I created a pull request for your pull request that implements custom reporters.

tmickel:
@TheBrokenRail Responded over there!

src/engine/blocks.js
```diff
@@ -389,6 +425,25 @@ Blocks.prototype.blockToXML = function (blockId) {
     return xmlString;
 };
 
+/**
+ * Recursively encode a mutation object to XML.
+ * @param {!Object} mutation Object representing a mutation.
+ * @return {string} XML string representing a mutation.
+ */
+Blocks.prototype.mutationToXML = function (mutation) {
```
thisandagain, line 433: I'm wondering if we should move these types of XML formatters into the `xml` utility module?

thisandagain (approved):
This looks great. Just one minor question in the line notes.

src/engine/blocks.js
tmickel: Sounds like a good idea in some form or another. I'll make an issue for it!

PR #213, opened 2016-09-22
Drop `useRound` option from `Cast.toListIndex`

### LINKED ISSUE #202: Drop `useRound` from Cast.toListIndex
In Scratch 2.0, this block:
screen shot 2016-09-20 at 10 55 23 am
Will delete the first item of the list. Because 0.5 is rounded to 1.

However,
screen shot 2016-09-20 at 10 55 57 am
will not do anything. Because it floors 0.5 to 0. (insert, item of also work this way).

Was this an intentional change in 2.0? In 1.4, delete works like "floor." Can we drop the round behavior and switch it to always floor without affecting compatibility of projects?

See: #187 for original discussion.

tmickel:
Fix #202

### DIFF
--- a/src/blocks/scratch3_data.js
+++ b/src/blocks/scratch3_data.js
@@ -73,7 +73,7 @@ Scratch3DataBlocks.prototype.addToList = function (args, util) {
 
 Scratch3DataBlocks.prototype.deleteOfList = function (args, util) {
     var list = util.target.lookupOrCreateList(args.LIST);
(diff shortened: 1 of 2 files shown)


thisandagain:
Looks good. Added compat note to Wiki here:
https://github.com/LLK/scratch-vm/wiki/Compatibility#consistent-rounding-method-for-list-indexes

thisandagain (approved):

PR #217, opened 2016-09-25
Corrected the URL we're cloning from

IAP-Reloaded:
I updated the URL so it doesn't give you any errors about permission.


tmickel:
Thanks!

PR #218, opened 2016-09-25
Publish to NPM and Github Pages

rschamp:
Make the playground rely only on files within its directory so that it can be published standalone to gh-pages.

Publish to NPM by creating an .npmrc file with an authentication token pointing to the encrypted `NPM_TOKEN` environment variable stored in Travis settings.

Remove the checked-in built assets, as they're now built before it's published to NPM. When installing from git, they will be built on npm install.

To publish to gh-pages, there is an environment variable `GH_TOKEN`, which stores a Personal Access Token I generated. If this token stops working, it can be updated in the Settings panel for the repo in Travis.


webpack.config.js
```diff
@@ -23,23 +22,80 @@ var base = {
...
+            library: 'VirtualMachine',
+            libraryTarget: 'commonjs2',
+            path: __dirname,
+            filename: '[name].js'
+        }
+    }),
+    // Playground
+    defaultsDeep({}, base, {
+        entry: {
+            'playground/vm': './src/index.js',
```
tmickel: nice!

tmickel (approved):

package.json
```diff
@@ -11,24 +11,27 @@
...
     "promise": "7.1.1",
-    "webpack": "1.13.0"
-  },
-  "devDependencies": {
-    "eslint": "2.7.0",
-    "highlightjs": "8.7.0",
-    "scratch-blocks": "git+https://git@github.com/LLK/scratch-blocks.git#develop",
-    "scratch-render": "git+https://git@github.com/LLK/scratch-render.git#develop",
+    "scratch-blocks": "0.1.0-80849",
+    "scratch-render": "0.1.0-9f26a",
```
tmickel: btw...fixing these to a published version number is interesting. we can still use `npm link` for development, but people will get an outdated version by default if they just clone?

rschamp: They will... although I think if we changed this to `^0.1.0` or just `*`, then we could have the most recent version on npm install.

rschamp:
Will need to resolve #219 before merging:

> http://llk.github.io/scratch-vm/
> The green-flag and repeat-arrow icons are showing for me as the "image missing" icon
> screen shot 2016-09-25 at 5 13 19 pm

rschamp:
The error on the playground `Uncaught TypeError: Cannot read property 'isConnected' of null` exists in the current develop branch playground.

rschamp:
@tmickel @thisandagain I updated one last thing — now that the playground doesn't rely on paths like `../node_modules/...`, we don't need to serve the entire project directory. So now the playground is served from `./playground` and is available at just `http://localhost:8080`.

tmickel:
Sounds good to me!

package.json
tmickel: Yeah - I might prefer that for new downloaders but don't strongly care, since we should be doing `npm link` anyway.

rschamp: Done

thisandagain:
This looks great. 👍

PR #221, opened 2016-09-28
Round costume values

tmickel:
On `Clone.setCostume` and on initialization from SB2. Values are set directly in the SB2 loader so `updateDrawableProperties` is done as a batch.

Sort of fixes project 96920096.


rschamp (approved):
Looks great, thanks!

Issue #103, opened 2016-08-12
Test Coverage - Blocks & Utils


thisandagain:
## Blocks
- [x] ./src/blocks/scratch3_operators.js
## Utilities
- [x] ./src/util/cast.js
- [x] ./src/util/color.js
- [x] ./src/util/math-util.js
- [x] ./src/util/timer.js
- [x] ./src/util/uid.js
- [x] ./src/util/xml-escape.js

PR #228, opened 2016-10-03
Add unit tests for all utility modules

thisandagain:
Implements one half of GH-103. Will handle blocks (as well as bugs exposed by this PR) in separate PRs.
## Before

image
image
## After

image
image


test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+    t.equal(cast.isInt(0), true);
+    t.equal(cast.isInt(1), true);
+    t.equal(cast.isInt(0.0), true);
+    t.equal(cast.isInt(3.14), false);
+    t.equal(cast.isInt(NaN), true);
+
+    // String
+    t.equal(cast.isInt('0'), true);
+    t.equal(cast.isInt('1'), true);
+    t.equal(cast.isInt('0.0'), false);      // @todo This should be true
```
thisandagain: Will file bug and follow-up in separate PR.

test/unit/util_math.js
```diff
@@ -0,0 +1,37 @@
+var test = require('tap').test;
+var math = require('../../src/util/math-util');
+
+test('degToRad', function (t) {
+    // @todo This is incorrect
```
thisandagain, line 5: Will file bug and follow-up in separate PR.

test/unit/util_color.js
```diff
@@ -0,0 +1,62 @@
+var test = require('tap').test;
+var color = require('../../src/util/color');
+
+test('decimalToHex', function (t) {
+    t.equal(color.decimalToHex(0), '#000000');
+    t.equal(color.decimalToHex(1), '#000001');
+    t.equal(color.decimalToHex(16777215), '#ffffff');
+    t.equal(color.decimalToHex(-16777215), '#000001');                  // @todo
```
thisandagain: I'm not sure if this behavior matters to us or if this seems correct.

thisandagain: /cc @cwillisf

test/unit/util_cast.js
tmickel: Actually I think this might be the right behavior, but I'll write a note in the issue!

test/unit/util_color.js
tmickel: Hmmm... I think all our decimal colors are represented by positive integers? So while not technically correct, I'd venture this may be reasonable.

test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+    // Boolean
+    t.equal(cast.toNumber(true), 1);
+    t.equal(cast.toNumber(false), 0);
+    t.equal(cast.toNumber('true'), 0);
+    t.equal(cast.toNumber('false'), 0);
+
+    // Undefined & object
+    t.equal(cast.toNumber(undefined), 0);
+    t.equal(cast.toNumber({}), 0);
+    t.end();
```
tmickel, line 27: May want to do `t.equal(cast.toNumber(NaN), 0);` since that's specifically dealt with in the function.

test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+    t.equal(cast.toBoolean(undefined), false);
+    t.equal(cast.toBoolean({}), true);
+    t.end();
+});
+
+test('toString', function (t) {
+    // Numeric
+    t.equal(cast.toString(0), '0');
+    t.equal(cast.toString(1), '1');
+    t.equal(cast.toString(3.14), '3.14');
```
tmickel: Btw, I'm assuming `t.equal` does `===` and not `==`?

test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+    t.equal(cast.compare(0, 0), 0);
+    t.equal(cast.compare(1, 0), 1);
+    t.equal(cast.compare(0, 1), -1);
+    t.equal(cast.compare(1, 1), 0);
+
+    // String
+    t.equal(cast.compare('0', '0'), 0);
+    t.equal(cast.compare('0.1e10', '1000000000'), 0);
+    t.equal(cast.compare('foobar', 'FOOBAR'), 0);
+    t.equal(cast.compare('dog', 'cat'), 1);
```
tmickel: seems good although the docstring for cast.compare does allow for values >= 1 or <= -1. maybe more correct to do `t.ok(cast.compare('dog', 'cat') > 0`, for example, but I don't strongly care :)

test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+test('toListIndex', function (t) {
+    var list = [0,1,2,3,4,5];
+    var empty = [];
+
+    // Valid
+    t.equal(cast.toListIndex(1, list.length), 1);
+    t.equal(cast.toListIndex(6, list.length), 6);
+
+    // Invalid
+    t.equal(cast.toListIndex(-1, list.length), 'INVALID');
```
tmickel: Maybe using the constants here (`Cast.LIST_INVALID`, `Cast.LIST_ALL`)?

test/unit/util_cast.js
```diff
@@ -0,0 +1,178 @@
...
+    t.ok(random <= list.length);
+    t.ok(random > 0);
+    t.equal(cast.toListIndex('random', empty.length), 'INVALID');
+
+    // "any" (alias for "random")
+    var any = cast.toListIndex('any', list.length);
+    t.ok(any <= list.length);
+    t.ok(any > 0);
+    t.equal(cast.toListIndex('any', empty.length), 'INVALID');
+    t.end();
```
tmickel, line 178: a valid case as well?

test/unit/util_color.js
```diff
@@ -0,0 +1,62 @@
...
+    t.end();
+});
+
+test('hexToDecimal', function (t) {
+    t.deepEqual(color.hexToDecimal('#000'), 0);
+    t.deepEqual(color.hexToDecimal('#000000'), 0);
+    t.deepEqual(color.hexToDecimal('#fff'), 16777215);
+    t.deepEqual(color.hexToDecimal('#ffffff'), 16777215);
+    t.deepEqual(color.hexToDecimal('#0fa'), 65450);
+    t.deepEqual(color.hexToDecimal('#00ffaa'), 65450);
```
tmickel: could/should these be `t.equal` instead of `t.deepEqual`?

test/unit/util_color.js
cwillisf: The feature allowing an integer input to a `set color` block is a bit controversial, but as long as we're keeping it I think we want this behavior (including -16777215 => `#000001`). While people might not intentionally put negative numbers into that block very often, it's easy for something like this to trigger such behavior:
image

tmickel:
@thisandagain This is really awesome!!!! I've made a few tiny/nit comments, but on the whole I think this is great to have.

test/unit/util_color.js
thisandagain: Sounds good. I'll pull the `@todo` notes from these.

test/unit/util_cast.js
thisandagain: Actually it does not. I'll fix that here.

test/unit/util_color.js
thisandagain: Good catch.

thisandagain (commented):
@tmickel Thanks for all the feedback. Addressed and ready for re-R:
https://github.com/LLK/scratch-vm/pull/228/files

tmickel (approved):
Looks good! Exciting

PR #235, opened 2016-10-04
Implement the point-towards block

towerofnix:
Test project ID: [124055382](https://scratch.mit.edu/projects/124055382/)

Both pointing towards a sprite and the mouse-pointer work (assuming that the `TOWARDS` argument being `_mouse_` means to point towards the mouse pointer!).

Based on [scratch-flash's implementation](https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/primitives/MotionAndPenPrims.as#L112-L121).

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -71,6 +71,7 @@ Scratch3MotionBlocks.prototype.pointTowards = function (args, util) {
         targetY = util.ioQuery('mouse', 'getY');
     } else {
         var pointTarget = this.runtime.getSpriteTargetByName(args.TOWARDS);
+        if (!pointTarget) return;
         targetX = pointTarget.x;
         targetY = pointTarget.y;
     }
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -21,6 +21,7 @@ Scratch3MotionBlocks.prototype.getPrimitives = function() {
         'motion_turnright': this.turnRight,
         'motion_turnleft': this.turnLeft,
         'motion_pointindirection': this.pointInDirection,
+        'motion_pointtowards': this.pointTowards,
         'motion_glidesecstoxy': this.glide,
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_motion.js
```diff
@@ -62,6 +63,24 @@ Scratch3MotionBlocks.prototype.pointInDirection = function (args, util) {
...
 };
 
+Scratch3MotionBlocks.prototype.pointTowards = function (args, util) {
+    var targetX = 0;
+    var targetY = 0;
+    if (args.TOWARDS === '_mouse_') {
+        targetX = util.ioQuery('mouse', 'getX');
+        targetY = util.ioQuery('mouse', 'getY');
+    } else {
+        var pointTarget = this.runtime.getSpriteTargetByName(args.TOWARDS);
```
tmickel, line 73: What happens if `pointTarget` is null or undefined? `this.runtime. getSpriteTargetByName` might not return a target if the block looks like this, for example:
screen shot 2016-10-04 at 3 00 15 pm

I think we will want to do:
`if (!pointTarget) return;`
or similar.

tmickel:
Thanks @liam4! This should be nice to have - one note and then it should be good!

src/blocks/scratch3_motion.js
towerofnix: Good point, fixed!

towerofnix:
(Small bonus issue.. for some reason the glass cup displays on the far left side of the screen, instead of where it should be, like in Scratch 2.0.. I don't know anything about that though! [Image](https://cloud.githubusercontent.com/assets/9948030/19088506/d9a9e5e6-8a4c-11e6-940c-97a0b30260e7.png))

tmickel:
Thanks for the quick patch!

Re: the glass cup - not sure about that either. Could be a SVG rendering issue or some other VM bug.

PR #233, opened 2016-10-04
Add unit test coverage for operator blocks

thisandagain:
Implements minimal test coverage for the `scratch3_operators` block module. Resolves GH-103
## Before

image
image
## After

image
image

PR #236, opened 2016-10-04
Add secure environment var for coveralls. Re GH-103

thisandagain:
Follow-up from https://github.com/LLK/scratch-vm/pull/233

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,15 @@
 language: node_js
 node_js:
-- "4"
-- "stable"
+- '4'
+- stable
 sudo: false
 cache:
   directories:
   - node_modules
+env:
+  global:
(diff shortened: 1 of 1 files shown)


rschamp:
I have begun storing these in the Travis settings panel for the repo because it allows you to adjust them and rerun without any new commits. So if you need to change an API key, you can still rerun old tests with the new key, for instance.  It also keeps junk out of the travis config — it's hard to keep track of them, especially since the Travis CLI removes comments every time you add a new encrypted variable.  In the settings panel you can see the names of the variables (just not the values).

What you lose is the ability to run tests requiring these variables in a new repo before setting up the same variables in that repo's settings.  Given our workflow, I don't think this is a big deal though since we always end up building in the LLK Travis when we make PRs.

thisandagain:
Handled manually via Travis-CI repo settings.

rschamp:
HAHAHA

Issue #79, opened 2016-06-13
Implement custom reporters


tmickel:
We can create a couple of stub blocks in scratch-blocks and start working on this right away, I think.

towerofnix:
Title is slightly confusing.. @tjvr noted this: https://scratch.mit.edu/discuss/post/2044070/

> You realise that could be parsed as “Implement ((custom blocks) and (reporters))”, right? :P

tmickel:
Sorry - I definitely don't mean for this to be a certain confirmation. We have a lot of work to do to determine if it will be included in Scratch 3.0.

But, we want to get started playing with the idea in scratch-vm to make sure we at least have the right code structure in its early stages to support things for this (I did mean "custom reporters").

I appreciate the excitement though :)

tmickel:
A couple of notes about implementation of procedures in 2.0:

Behavior is to yield if a call is detected to be recursive:
https://github.com/LLK/scratch-flash/blob/9c0728bfbe2f2ed1cff1c9f3af8d2a55552b99a5/src/interpreter/Interpreter.as#L662

Detecting if a call is recursive seems to do a 5-layer thread stack search: https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/interpreter/Thread.as#L98

To me this looks like it can cause weird behavior, e.g.: https://scratch.mit.edu/projects/119356687/

Seems preferable to do an N-layer stack search (with whatever fast implementation) unless there was some reason for the 5-limit.

Procedures affect the "stop this script" block, which calls this: https://github.com/LLK/scratch-flash/blob/9c0728bfbe2f2ed1cff1c9f3af8d2a55552b99a5/src/interpreter/Interpreter.as#L672

The other interesting feature is the "warp thread", which:
- is turned on by setting the procedure as "run without screen refresh"
- prevents all enclosed blocks from yielding
- has an extended timer (`warpMSecs`)

Params are dealt with separately from general Scratch variables (`primVarGet` vs. `primGetParam`). This of course makes sense - the values for the params come from `activeThread.args` which is basically kept in a stack.

mrjacobbloom:
~~In Scratch 2, arguments aren't separate for each thread, afaik there's one copy of each argument in memory no matter how many levels of recursion are happening. Would it be possible to fix this in Scratch 3?~~

Edit: it actually looks like this issue was resolved. Never mind!

TheBrokenRail:
[...]

PR #237, opened 2016-10-04
Implement custom reporters

dekrain:


dekrain:
This not work very good. Try to put `if` block to definition and it break

TheBrokenRail:
Try Using startHats(() and copying code from "broadcast and wait" block

TheBrokenRail:
I made PR For startHats

tmickel:
Hey @dekrain and @TheBrokenRail! Cool to be thinking about this problem.

I'm worried about storing values on the `Scratch3ProcedureBlocks` object. What happens when you run two custom reporters in parallel? I could imagine that if both report at the same time, something weird would happen.

Traditionally programming languages implement this sort of thing (returning values from a procedure) using a stack:
https://en.wikipedia.org/wiki/Call_stack

We have most of the infrastructure needed in `thread.stack` and `thread.stackFrame` - the former which tracks descent into the block stack, and the latter of which is intended for storing arbitrary data, like parameters for custom procedures and (potentially) reported values.

A correct implementation of custom blocks and custom reporters will need to use a stack associated with the thread in some form or another. Otherwise, crazy things seem likely to happen when two blocks execute in parallel or one thread yields for a frame.

tmickel:
P.s., I checked out this PR and tried to run it, but I get this error:
`vm.js:5206 Uncaught TypeError: Cannot read property 'length' of undefined`

dekrain:
Ok, I've read scratch-vm code many times and I have not see any problems. I'm annoyed.

src/blocks/scratch3_procedures.js
```diff
@@ -4,6 +4,8 @@ function Scratch3ProcedureBlocks(runtime) {
      * @type {Runtime}
      */
     this.runtime = runtime;
+    this.REPORT = {};
+    this.currentCall = null;
```
tmickel: We can't store data on the `Scratch3ProcedureBlocks` object, as I previously mentioned. Any data that's relevant for procedure calls needs to be associated with the thread.

src/blocks/scratch3_procedures.js
```diff
@@ -13,14 +15,34 @@ function Scratch3ProcedureBlocks(runtime) {
...
+    };
+};
+
+Scratch3ProcedureBlocks.prototype.getHats = function() {
+    return {
+        'procedures_defnoreturn': {
+            restartExistingThreads: false
+        },
+        'procedures_defreturn': {
+            restartExistingThreads: false
```
tmickel, line 31: These aren't hat blocks, and the mechanism for starting hats isn't intended for starting procedures.

src/blocks/scratch3_procedures.js
```diff
@@ -13,14 +15,34 @@ function Scratch3ProcedureBlocks(runtime) {
...
     };
 };
 
 Scratch3ProcedureBlocks.prototype.defNoReturn = function () {
     // No-op: execute the blocks.
 };
 
+Scratch3ProcedureBlocks.prototype.defReturn = function (args) {
+    var name = args.NAME;
+    this.REPORT[name] = ((typeof this.REPORT[name] === 'undefined') ?
```
tmickel: What if you run two of the same custom reporter in parallel?

src/blocks/scratch3_procedures.js
```diff
@@ -29,4 +51,48 @@ Scratch3ProcedureBlocks.prototype.callNoReturn = function (args, util) {
...
+            delete this.REPORT[procedureName];
+            this.currentCall = this.currentCall.parent;
+            return rep;
+        }
+    }
+};
+
+Scratch3ProcedureBlocks.prototype.report = function (args) {
+    if (this.currentCall) {
+        this.REPORT[this.currentCall.name] = args.VALUE;
```
tmickel: Same comment as above

tmickel:
@dekrain I'm not sure how fruitful it's going to be to work on this without digging into the details of the VM. The Scratch team was planning to do an implementation of this sometime in the next several weeks/months.

thisandagain:
@dekrain Thanks for taking a first crack at this. 

For big feature implementations like this we would prefer to talk about the approach before generating a bunch of code and a PR. This helps reduce the amount of work for you (and for us) and makes sure that we are all on the same page before charging into a solution. As @tmickel mentioned we are looking to produce an implementation for this over the coming weeks/months, but in the meantime I think it would be most fruitful to continue this discussion here:
https://github.com/LLK/scratch-vm/issues/79

PR #238, opened 2016-10-04
Implement "go to" block

towerofnix:
Test project ID: [124084337](https://scratch.mit.edu/projects/124084337/)

As you'd expect, it works with both the mouse and sprites. Also, it won't do anything if the target sprite doesn't exist.

It (probably) does _not_ work with the pen as the pen hasn't been implemented at all yet, as far as I know - it'll probably need to be changed then. (Maybe a more generic `Clone.prototype.goTo`, sort of like [scratch-flash does](https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/primitives/MotionAndPenPrims.as#L312-L320)?)

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -55,6 +55,9 @@ Scratch3MotionBlocks.prototype.goTo = function (args, util) {
     if (args.TO === '_mouse_') {
         targetX = util.ioQuery('mouse', 'getX');
         targetY = util.ioQuery('mouse', 'getY');
+    } else if (args.TO === '_random_') {
+        targetX = this.runtime.constructor.STAGE_WIDTH * (Math.random() - 0.5)
+        targetY = this.runtime.constructor.STAGE_HEIGHT * (Math.random() - 0.5)
     } else {
         var goToTarget = this.runtime.getSpriteTargetByName(args.TO);
         if (!goToTarget) return;
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -18,6 +18,7 @@ Scratch3MotionBlocks.prototype.getPrimitives = function() {
     return {
         'motion_movesteps': this.moveSteps,
         'motion_gotoxy': this.goToXY,
+        'motion_goto': this.goTo,
         'motion_turnright': this.turnRight,
         'motion_turnleft': this.turnLeft,
         'motion_pointindirection': this.pointInDirection,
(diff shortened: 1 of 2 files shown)


tmickel:
Thanks! This looks good to me, but what about go to "random"?

Don't worry about the pen - `clone.setXY` will probably have that built in since the pen needs to act anytime the clone moves around.

towerofnix:
The features of Scratch you never use, so you forget are there.. :)

tmickel:
P.s. a note about implementation if you choose to do random - it would be good to pick the number based on a constant/adjustable stage width/height, perhaps on `runtime`, e.g., `Runtime.STAGE_WIDTH = 480; Runtime.STAGE_HEIGHT = 360;`

towerofnix:
Thanks - I've been a bit busy but yeah, I was looking for a way to use the constant in the code earlier.

EDIT: I guess I'll need to make that constant, it doesn't exist right now AFAIK.

tmickel:
^ yep, doesn't exist yet :)

src/blocks/scratch3_motion.js
```diff
@@ -56,8 +56,10 @@ Scratch3MotionBlocks.prototype.goTo = function (args, util) {
         targetX = util.ioQuery('mouse', 'getX');
         targetY = util.ioQuery('mouse', 'getY');
     } else if (args.TO === '_random_') {
-        targetX = this.runtime.constructor.STAGE_WIDTH * (Math.random() - 0.5)
-        targetY = this.runtime.constructor.STAGE_HEIGHT * (Math.random() - 0.5)
+        var stageWidth = this.runtime.constructor.STAGE_WIDTH;
+        var stageHeight = this.runtime.constructor.STAGE_HEIGHT;
```
towerofnix, line 60: `this.runtime.constructor` feels a bit hacky. But requiring `../engine/runtime.js` just returns an empty object ([because it's a circular require](http://stackoverflow.com/a/23875299/4633828)), so this is all I could come up with.

I'm not sure if `.constructor` is part of ES5 though.

EDIT: [Never mind; we're very good on `.constructor` support, lol.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/constructor#Specifications)

tmickel: I'm fine with this, alternatively if you're worried about it I guess we could add `runtime.getStageDimensions()`

tmickel:
Just added Math.round to be consistent with 2.0. thanks again!

PR #239, opened 2016-10-04
Implement "distance to" block

towerofnix:
Test project ID: [124104690](https://scratch.mit.edu/projects/124104690/)

In theory "distance to _random_" would do.. something.. but I couldn't get it to work in Scratch 2.0, so I haven't bothered implementing it here.

Based on [scratch-flash's implementation](https://github.com/LLK/scratch-flash/blob/33f784a9253b3d469b02cd8a069a9b6983d0fc6c/src/primitives/SensingPrims.as#L242-L249).

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -16,6 +16,7 @@ Scratch3SensingBlocks.prototype.getPrimitives = function() {
     return {
         'sensing_touchingcolor': this.touchingColor,
         'sensing_coloristouchingcolor': this.colorTouchingColor,
+        'sensing_distanceto': this.distanceTo,
         'sensing_timer': this.getTimer,
         'sensing_resettimer': this.resetTimer,
         'sensing_mousex': this.getMouseX,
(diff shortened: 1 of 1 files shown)


tmickel:
Thanks for this!

Distance to _random_ ...definitely makes no sense!

One thing looking at scratch-flash:
`if ((s == null) || (p == null))`
It's strange to thing `s` would ever be null. But in fact I think most notably `interp.targetSprite` returns null when the target is the stage. That should almost never happen - but you _can_ drag the distance-to block to the stage, and there in 2.0 it appears to almost always return 10,000.

I think in your current implementation, this situation would be treated as if the stage is at (0,0). That seems fine and logical to me, but maybe we should maintain its returning 10,000 for compatibility...

To do that, you could check `util.target.isStage`. What do you think?

towerofnix:
Sounds good to me.

EDIT: But.. what's the philosophy behind 10000 being an "error" value? :)

tmickel:
Looks good, thanks again! Regarding 10000: I'm not sure about the history, although it could make sense as to mean "very far away"...

mrjacobbloom:
Then why not return Infinity? I'd still prefer distance to the origin though.

PR #242, opened 2016-10-05
Add NPM Script 'build'

TheBrokenRail:
### Proposed changes

_Changes NPM Start to something that dose not require make (For Windows Users) And Adds NPM Build To Easily Build scratch-vm without running the dev server_

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "prepublish": "./node_modules/.bin/webpack --bail",
     "start": "webpack-dev-server",
-    "start": "webpack --colors --progress",
+    "build": "webpack --colors --progress",
     "test": "make test",
     "version": "./node_modules/.bin/json -f package.json -I -e \"this.repository.sha = '$(git log -n1 --pretty=format:%H)'\""
   },
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   "main": "./dist.js",
   "scripts": {
     "prepublish": "./node_modules/.bin/webpack --bail",
-    "start": "make serve",
+    "start": "webpack-dev-server",
+    "start": "webpack --colors --progress",
     "test": "make test",
     "version": "./node_modules/.bin/json -f package.json -I -e \"this.repository.sha = '$(git log -n1 --pretty=format:%H)'\""
   },


package.json
```diff
@@ -12,7 +12,8 @@
   "main": "./dist.js",
   "scripts": {
     "prepublish": "./node_modules/.bin/webpack --bail",
-    "start": "make serve",
+    "start": "webpack-dev-server",
+    "start": "webpack --colors --progress",
```
towerofnix: _Two_ `start` properties?

rschamp:
Hi @TheBrokenRail — this PR is a bit confusing to me.  Would you please fill out the Pull Request template I added to the description?  Thanks!

TheBrokenRail:
Changed Description

TheBrokenRail:
Any Comments?

package.json
dekrain: I think it should be `start` and `build`

rschamp: Like @dekrain said, this doesn't add the script `build`, it adds a second `start` script which makes this invalid JSON. Please update to `build`.

rschamp (changes requested):
It seems that the goal of this PR is not only to add the script build, but to make `package.json` usable under Windows.

@cwillisf In other repos that use Makefiles, e.g. https://github.com/LLK/scratch-render how do you run them?

TheBrokenRail:
Fixed Second Start Bug

package.json
TheBrokenRail: Fixed!

dekrain (approved):

rschamp (approved):
@TheBrokenRail Looks good, thanks! Thanks for bringing up the issues with running `make` on Windows.

cwillisf:
@rschamp I usually use Cygwin to deal with Scratch-related repositories, and I make sure my installation of Cygwin has GNU Make and all the other tools that our repositories require.

There are other options which might be more common, though. You can install Node.js by itself and end up without Bash, let alone Make. You can also install [Git for Windows](https://git-for-windows.github.io/) which gets you Bash but not Make. It's also worth noting that Git for Windows uses MSYS rather than Cygwin, so there are a few differences including path conventions. You can install GNU Make on MSYS, but those who are just interested in Node-style development usually don't.

Long story short, I do think it would be more friendly to Windows users if we worked on requiring fewer pre-installed tools. Ideally we could get it down to requiring only `npm` and things installed by `npm`, like Webpack.

rschamp:
@cwillisf Thanks for the background! We took a crack at using npm only, and no Makefile in https://github.com/LLK/scratch-audioengine.

I suppose the commands that use `&&` might still be a problem on Windows?

PR #243, opened 2016-10-05
Add issue and PR templates for contributors

rschamp:
### Proposed changes

This adds two Github-specific files which will provide starter templates for new Issues and Pull Requests
### Reason for changes

Internally we have good communication about issues in our projects, and changes that should be made. When we receive contributions from the community, they often come without context. These templates should help provide some of that missing context.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,15 @@
+### Expected behavior
+
+_Please describe what should happen_
+
+### Actual behavior
+
+_Describe what actually happens_
+
+### Steps to reproduce
+
+_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
+
+### Operating system and browser
+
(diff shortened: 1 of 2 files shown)


tmickel:
Good with me as a general suggestion, as long as we're generally ok with deleting/replacement with free-form.

Here's also a CONTRIBUTORS.md I wrote up for scratch-blocks. If that's good over there, could we pull it here?
https://github.com/LLK/scratch-blocks/pull/626

tmickel (approved):

rschamp:
I was thinking this would serve mostly for things that come in without any context, so I think it would be fine to continue having freeform PRs/issues, so long as they include enough information to be useful.  Having this here we could at least ask "please resubmit following the template" if there isn't enough information.

The CONTRIBUTING file you linked and made a PR for looks good to me!

thisandagain:
Works for me. Thanks for putting this together @rschamp. It would be nice to add this to our repo boilerplate and get this added across all of our open repos.

PR #248, opened 2016-10-08
Change Color Scheme to Be More Like 2.0

TheBrokenRail:
### Proposed changes

_Changes The Color Scheme To Look More Like 2.0_
### Reason for changes

_So It Looks More Like 2.0._

### DIFF
--- a/playground/playground.css
+++ b/playground/playground.css
@@ -1,5 +1,5 @@
 body {
-    background: rgb(36,36,36);
+    background: #DDDEDE;
 }
 a {
     color: rgb(217,217,217);
@@ -16,7 +16,7 @@ h2 {
     font-family: "Helvetica Neue", Helvetica, sans-serif;
 }
 #vm-devtools {
-    color: rgb(217,217,217);
+    color: #DDDEDE;
     position: absolute;
     left: 1%;
     right: 60%;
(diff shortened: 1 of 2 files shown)


CosmicWebServices:
This kind of goes against the design and color scheme that the scratch team is looking to use in scratch 3.0 (https://github.com/LLK/scratch-blocks/wiki/Design) also this pr should be in [scratch-blocks](https://github.com/LLK/scratch-blocks) not scratch-vm.

TheBrokenRail:
No, I don't think it doses it doesn't even change the blocks at all, I love the flat blocks design, this pr just changes the backgrounds and stuff.

CosmicWebServices:
ok

mrjacobbloom:
Yeah I'd leave the design stuff up to the ST. And the VM playground isn't the final design for the editor, it's just what the ST is using to test what everything looks like hooked together.

You could try userstyles to customize the CSS on your end

TheBrokenRail:
No this just changes the backround color of the workspace

tmickel:
We're actively working on the Scratch GUI, including redesigning some aspects. Although these are the colors and formatting for 2.0, they are likely to look different in 3.0. The VM playground is also not necessarily going to look like the Scratch GUI, as it's meant to be a development tool.

PR #256, opened 2016-10-12
Fix new project creation

dekrain:
### Proposed changes

I want to update new project creation
### Reason for changes

For now, on new project, any Sprite doesn't shown, b/c it hasn't pointer for runtime

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -1,8 +1,8 @@
-const newProjectStr = 'createEmptyProject';
+var NEW_PROJECT_HASH = 'createEmptyProject';
 
 var loadProject = function () {
     var id = location.hash.substring(1);
-    if (id === newProjectStr) {
+    if (id === NEW_PROJECT_HASH) {
         window.vm.createEmptyProject();
         return;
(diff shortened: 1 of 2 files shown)


playground/playground.js
```diff
@@ -1,6 +1,12 @@
+const newProjectStr = 'createEmptyProject';
```
tmickel: For compatibility with older browsers, I don't think we can support `const`. how about `var NEW_PROJECT_HASH`?

playground/playground.js
```diff
@@ -1,6 +1,12 @@
...
+
 var loadProject = function () {
     var id = location.hash.substring(1);
-    if (id.length < 1) {
+    if (id === newProjectStr) {
+        window.vm.createEmptyProject();
+        return;
+    }
```
tmickel, line 8: Cool - this seems like it's the right behavior.

tmickel (changes requested):

tmickel:
Just one request but otherwise seems good.

PR #257, opened 2016-10-12
Updated blocks for new renderer features

tmickel:
- Add layer blocks implementation (go to front, go back N layers)
- Add "touching" block implementation, including "touching mouse pointer" (using pick), "touching edge" (using bounds), and "touching sprite" (using touching-drawables).
- Add "if on edge, bounce" implementation. I didn't try to pull apart how this works exactly, was just reproducing it as a test. Had to adjust the coordinates to be in Scratch coordinates instead of top-left (0,0) that the original implementation uses. I also refactored the "keep on stage" logic into its own method, `keepInFence`, which can take an arbitrary bounding box to keep the clone inside. I figure this will be useful when we try to reproduce fence mode. Original for reference: https://github.com/LLK/scratch-flash/blob/master/src/primitives/MotionAndPenPrims.as#L202

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -173,8 +173,7 @@ Scratch3MotionBlocks.prototype.ifOnEdgeBounce = function (args, util) {
         return; // Not touching any edge.
     }
     // Point away from the nearest edge.
-    // @todo: Fix after GH-229 is resolved.
-    var radians = MathUtil.degToRad(util.target.direction);
+    var radians = MathUtil.degToRad(90 - util.target.direction);
     var dx = Math.cos(radians);
     var dy = -Math.sin(radians);
     if (nearestEdge == 'left') {
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -141,6 +141,9 @@ Scratch3MotionBlocks.prototype.glide = function (args, util) {
 
 Scratch3MotionBlocks.prototype.ifOnEdgeBounce = function (args, util) {
     var bounds = util.target.getBounds();
+    if (!bounds) {
+        return;
+    }
     // Measure distance to edges.
     // Values are positive when the sprite is far away,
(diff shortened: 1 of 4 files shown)


src/sprites/clone.js
```diff
@@ -352,6 +352,77 @@ Clone.prototype.getName = function () {
 };
 
 /**
+ * Return the clone's tight bounding box.
+ * @param {number} x X coordinate of test point.
+ * @param {number} y Y coordinate of test point.
+ * @return {Boolean} True iff the clone is touching the point.
```
cwillisf: Incorrect `@param` & `@return`

src/blocks/scratch3_motion.js
```diff
@@ -138,6 +139,59 @@ Scratch3MotionBlocks.prototype.glide = function (args, util) {
...
 };
 
+Scratch3MotionBlocks.prototype.ifOnEdgeBounce = function (args, util) {
+    var bounds = util.target.getBounds();
+    // Measure distance to edges.
+    // Values are positive when the sprite is far away,
+    // and clamped to zero when the sprite is beyond.
+    var stageWidth = this.runtime.constructor.STAGE_WIDTH;
+    var stageHeight = this.runtime.constructor.STAGE_HEIGHT;
+    var distLeft = Math.max(0, stageWidth / 2 + bounds.left);
```
cwillisf, line 152: If there's no renderer then this will throw due to `getBounds()` having returned `null`. Maybe the function should just have an `if (!bounds) { return; }` or similar near the top?

cwillisf (changes requested):
Looks good overall -- just a couple small tweaks.

src/blocks/scratch3_motion.js
tmickel: Agreed - thanks!

src/sprites/clone.js
tmickel: That was silly - thanks

cwillisf (approved):
👍

PR #258, opened 2016-10-12
Custom Reporters (Untested)

TheBrokenRail:
### Proposed changes

_Implements Custom Reporters_
### Notes

_Doesn't Support Multiple Custom Reporters Running At Once, Instead It Waits Tell None Are Running To Run._


tmickel:
Limiting execution to one at a time is interesting, but I think it would have many unintended consequences. For example, it means we can't support a simple case of recursion (here is an example that works in Snap):
screen shot 2016-10-12 at 8 25 06 pm
If you have ideas would also be happy to discuss in #79 instead of PRs.

PR #259, opened 2016-10-13
Copy Assets To Playground With Webpack

TheBrokenRail:
### Proposed changes

_(See Title)_
### Reason for changes

_So There Are Sprites On The Playground._

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -110,7 +110,8 @@ module.exports = [
             }, {
                 from: 'node_modules/highlightjs/styles/zenburn.css'
             }, {
-                from: 'assets'
+                from: 'assets',
+                to: 'assets'
             }])
         ])
     })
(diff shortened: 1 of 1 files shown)


dekrain (approved):

webpack.config.js
```diff
@@ -108,6 +109,8 @@ module.exports = [
                 to: 'media'
             }, {
                 from: 'node_modules/highlightjs/styles/zenburn.css'
+            }, {
+                from: 'assets'
```
rschamp: This will copy the contents of assets to the root of the playground. I.e., assets/stage.png will be available at `localhost:8080/stage.png` not `localhost:8080/assets/stage.png`.  As far as I can see, they need to be available under `/assets/`, so you need to add `to: 'assets'` here.

rschamp (changes requested):

TheBrokenRail:
Fixed!

webpack.config.js
TheBrokenRail: Fixed

TheBrokenRail:
@rschamp

rschamp (approved):

Issue #249, opened 2016-10-08
Add "stretch" graphic effect


TheBrokenRail:
This would be really easy to implement cause it would just be changing the scale when rendering and adding a few blocks. (For Example Of Base Code See #261)
### Mockup

![capture](https://cloud.githubusercontent.com/assets/17478432/19403308/424e5476-9234-11e6-8759-b232201ea851.PNG)

![capture2](https://cloud.githubusercontent.com/assets/17478432/19410996/d6118178-92c6-11e6-8bfe-0ecdd5e39fa9.PNG)

TheBrokenRail:
Should I make A PR on scratch-blocks that adds the blocks?

tmickel:
@TheBrokenRail We can't take any PRs that add new feature-related blocks at this time without significant design discussion first.

TheBrokenRail:
Any Discussion or updates about design and stuff?

TheBrokenRail:
I added a mockup!

CosmicWebServices:
[offtopic] @TheBrokenRail  what is your scratch username

TheBrokenRail:
Comments, Questions, Design Questions, What do you think of my mockup (really just a fork of scratch-blocks)?

TheBrokenRail:
Any Comments?

CosmicWebServices:
Let me look for a minute or two...

CosmicWebServices:
Does it stretch the whole x or the whole y up and down or just the top of the image or the bottom of the image... If that makes sense?

CosmicWebServices:
Also, I think the mockup looks great and is understandable but could it be considered a motion block?

TheBrokenRail:
The Whole Axis Stretches and I don't think it would be a motion block because it doesn't change position or rotation and it is a cosmetic effect

TheBrokenRail:
@tmickel

griffpatch:
You could even consider this an effects block named squeeze that multiplies the width. 100% being normal relative to current size, 50% being half width and 100% height still. The only reason for doing it like this is to try to reduce the number of new scratch blocks ;). You still want it to apply 'before' the rotation.

griffpatch:
Maybe that's a silly idea... You wouldn't want the width to be reset when doing a reset all effects?

TheBrokenRail:
[...]

PR #261, opened 2016-10-13
Add Basic Stretch Functions

TheBrokenRail:
### Proposed changes

_Add Functions In The Clone Object To Be Stretched_
### Reason for changes

_To Allow For Implementation of Stretch Blocks (If you decide to add them, just so there there)_

### DIFF
--- a/src/sprites/clone.js
+++ b/src/sprites/clone.js
@@ -259,6 +259,44 @@ Clone.prototype.setSize = function (size) {
 };
 
 /**
+ * Set stretch X
+ * @param {!number}
+ */
+Clone.prototype.setStretchX = function (stretchX) {
+    if (this.isStage) {
+        return;
+    }
+    // Keep size between 5% and 535%.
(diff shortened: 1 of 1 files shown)


src/sprites/clone.js
```diff
@@ -106,6 +106,18 @@ Clone.prototype.visible = true;
 Clone.prototype.size = 100;
 
 /**
+ * How Far It Is Stretched On The X Axis
```
towerofnix, line 109: Might want to adjust this to fit more with the other JSdoc comments - "it" is not implied ("whether THE CLONE is currently visible") and The Capitalization Could Use Some Adjusting. (Also it's missing a period at the end of the line.)

Same goes for "How Far It Is Stretched On The Y Axis".

thisandagain:
This requires design discussion before we can consider landing features to support it. See discussion in https://github.com/LLK/scratch-vm/issues/249

PR #263, opened 2016-10-13
Fix relative localization for assets

dekrain:
For upcoming chage: on new project creation, assets directory is in main directory of server


rschamp:
Thanks!

PR #267, opened 2016-10-13
Executing blocks in the toolbox

tmickel:
Blocks in the toolbox don't really belong to any given target. So, instead of storing their blocks in `editingTarget`, this stores them in a separate blocks object, `runtime.flyoutBlocks`. These also won't be part of block searches for things like the green flag (as desired) since those search target-attached blocks.

However, during execution, we need to make sure these blocks can be found (even though they're not associated with the execution target). If a block isn't found on a target, now execution will search for the block in the flyout before assuming it's deleted.

Also adds a new method, `flyoutBlockListener`, which VM consumers can attach to the scratch-blocks flyout.

### DIFF
--- a/playground/playground.js
+++ b/playground/playground.js
@@ -73,8 +73,9 @@ window.onload = function() {
     window.workspace = workspace;
 
     // Attach scratch-blocks events to VM.
-    // @todo: Re-enable flyout listening after fixing GH-69.
     workspace.addChangeListener(vm.blockListener);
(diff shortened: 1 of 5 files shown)


dekrain:
But..... Where is toolbox???

tmickel:
@dekrain I think something with the scratch-blocks build happened out of order, because the toolbox XML has changed location.

PR #276, opened 2016-10-17
Replace Makefile with npm scripts

### LINKED ISSUE #262: Remove Makefile
On Windows, `make` doesn't exist. Move everything in the Makefile to npm scripts for interoperability.

rschamp:
### Proposed changes

Removes the Makefile and moves what it does to npm scripts.

I also updated the lint step to just lint everything that isn't ignored so nothing gets missed as we add files/directories.
### Reason for changes

Fixes #262. Should allow Windows users to run all the commands currently in the Makefile (untested).

Assigning @cwillisf to test my assumption that these work on Windows. @TheBrokenRail please try it out as well.


package.json
```diff
@@ -11,10 +11,16 @@
...
+    "coverage": "./node_modules/.bin/tap ./test/{unit,integration}/*.js --coverage --coverage-report=lcov",
+    "lint": "./node_modules/.bin/eslint .",
+    "prepublish": "npm run build",
+    "prepublish-watch": "npm run watch",
+    "start": "./node_modules/.bin/webpack-dev-server",
+    "tap-integration": "./node_modules/.bin/tap ./test/integration/*.js",
+    "tap-unit": "./node_modules/.bin/tap ./test/unit/*.js",
+    "test": "npm run lint && npm run tap-unit && npm run tap-integration",
+    "watch": "./node_modules/.bin/webpack --progress --colors --watch",
     "version": "./node_modules/.bin/json -f package.json -I -e \"this.repository.sha = '$(git log -n1 --pretty=format:%H)'\""
```
cwillisf, line 24: NPM actually puts `./node_modules/.bin/` or an equivalent onto the PATH while running scripts, so these entries could be simplified to:

``` json
    "build": "webpack --progress --colors --bail",
    "coverage": "tap ./test/{unit,integration}/*.js --coverage --coverage-report=lcov",
    "lint": "eslint .",
```

...and so on. They work as-is too, though.

cwillisf (approved):
👍 thanks!

Tested successfully from a command prompt (not Bash!) on Windows 10.

Note that linting will currently fail due to CRLF complaints on Windows clients using the default git configuration. Adding a `.gitattributes` can fix that, but I would do so as a separate change.

package.json
rschamp: Ah right — recently I hit some issues with the path, so that's why I used the explicit path: https://github.com/LLK/scratch-blocks/commit/1831890a7d3803592c2386be50ab00139172c1b6.  It would fail to find the .bin when run as a dependency of another project.  I think the issue was raised when it ran postinstall as an `npm link`ed dependency — it wouldn't be able to find webpack or it would use the incorrect config?

rschamp:
I'll copy the .gitattributes from scratch-render since this is intending to fix the project for Windows more than just remove the Makefile. Thanks for pointing that out.

Issue #229, opened 2016-10-03
Math - Fix `degToRad` method


thisandagain:
The `degToRad` method is currently written as `(Math.PI * (90 - deg)) / 180` which is incorrect. `(90 - deg)` is being used by the `moveSteps` block but should be factored out as to maintain the universality (and correctness) of the `degToRad` method.

From https://github.com/LLK/scratch-vm/pull/228

PR #279, opened 2016-10-17
Fix `degToRad` function definition

thisandagain:
Resolves GH-229
- Modify `degToRad` utility function to work as expected
- Update `moveSteps` block definition to reflect change
- Update unit tests


tmickel:
nice, thanks!

Issue #65, opened 2016-06-08
Single-stepping debug mode


tmickel:
It would be great to have a mode of execution where things are stepped by a button in the VM playground. The button would essentially call stepThreads() with a tiny WORK_TIME, to proceed one block per thread every time the button is pushed.

In the long run, it would be good to have single-stepping mode a la Scratch 1.4

thisandagain:
👍

NoMod-Programming:
If I may add some insight into this, it might be a good idea to add a slider for speed in a "single-stepping" mode, that way, rather than having to use a button, which can be very cumbersome for large projects, or using a predetermined speed, which may not be the speed needed for debugging a part of a project, the user can choose his/her own speed.

PR #280, opened 2016-10-17
Interpreter fixes, enhancements, features

### LINKED ISSUE #207: Pause mode
Would like a pause mode that lets us inspect state at any point the button is pressed.

tmickel:
I spent a bunch of time reworking and refactoring parts of the sequencer, control blocks, and stepping mechanisms to ensure 2.0 compatibility and add a bunch of things we wanted.

---
## Fixed our stepping mechanism to work like 2.0.

In playing with custom blocks, I noticed that scratch-vm executed simple recursive procedures like "calculate factorial(n)" about 100-1000x slower than Scratch 2.0, even with any kind of yielding turned off. That was pretty shocking, so I started digging into the reason. Basically, we were switching threads way too often.

Consider this set:
screen shot 2016-10-17 at 12 51 51 pm

In Scratch 2.0 and 1.4, and after this PR, both of those scripts are guaranteed to execute atomically. So, as long as the left script executes before the right script, the sprite will _always_ turn.

In the previous scratch-vm implementation, the thread would switch as soon as executing one block. I.e., it would do each green flag, then a single set, then check the if, etc. Not only is this slow, it's just not how Scratch currently executes things.

I believe I've fixed this by switching our previous `startThread` to a `stepThreads` that looks like Scratch 2.0's.

---
## Add warp-mode thread

Since I was already digging into thread stepping, I added the warp-mode thread for "run procedure without screen refresh." This is again modeled on 2.0's implementation.

---
## Add `interp.redraw` equivalent and correct behavior for yielding at the end of loops.

This is pretty closely modeled on 2.0, including when `stepThread` and `stepThreads` returns.

---
## Add turbo mode support

With the previous implemented, it was very easy to add turbo mode.

---
## Add 2.0 compatibility 30TPS mode

I thought it would be helpful to have this to do a closer project comparison. Basically, it switches the stepping interval to 30TPS.

---
## Add pause mode

This suspends the stepping timer and also tells the project timer to pause itself as appropriate, so the "timer" block is also paused.

---
## Add single-stepping mode and a sliding speed adjustment

Getting into the weeds of stepping, this also made sense to address and was fun to do. Basically, it throws "running" status threads into a single-tick yield after executing. Otherwise, it adjusts the stepping interval according to the given speed. This also involved updating the "glow update" method, to handle highlighting the blocks in single-stepping mode.

screen shot 2016-10-17 at 1 00 43 pm

screen shot 2016-10-17 at 1 04 59 pm

---

I also improved JSDoc and documentation in several places. If merged, I think this resolves #207, #89, and #65.

cc: @rschamp FYI - not sure if/when we want to use any of these options in the GUI!

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -15,6 +15,12 @@
             <button id="greenflag">Green flag</button>
             <button id="stopall">Stop</button>
         </div>
+        <div>
+            Turbo: <input id='turbomode' type='checkbox' />
+        </div>
+        <div>
+            Pause: <input id='pausemode' type='checkbox' />
+        </div>
         <br />
         <ul id="playgroundLinks">
             <li><a id="renderexplorer-link" href="#">Renderer</a></li>
--- a/playground/index.html
+++ b/playground/index.html
@@ -26,6 +26,7 @@
         </div>
         <div>
             Single stepping:  <input id='singlestepmode' type='checkbox' />
+            <input id='singlestepspeed' type='range' min='1' max='20' value='10' />
         </div>
         <br />
         <ul id="playgroundLinks">
--- a/playground/index.html
+++ b/playground/index.html
@@ -21,6 +21,9 @@
         <div>
             Pause: <input id='pausemode' type='checkbox' />
         </div>
+        <div>
+            Compatibility (30 TPS):  <input id='compatmode' type='checkbox' />
+        </div>
         <br />
         <ul id="playgroundLinks">
             <li><a id="renderexplorer-link" href="#">Renderer</a></li>
--- a/playground/index.html
+++ b/playground/index.html
@@ -24,6 +24,9 @@
         <div>
             Compatibility (30 TPS):  <input id='compatmode' type='checkbox' />
         </div>
+        <div>
+            Single stepping:  <input id='singlestepmode' type='checkbox' />
+        </div>
         <br />
         <ul id="playgroundLinks">
             <li><a id="renderexplorer-link" href="#">Renderer</a></li>
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -488,11 +488,8 @@ Runtime.prototype._step = function () {
         }
     }
     this.redrawRequested = false;
-    var inactiveThreads = this.sequencer.stepThreads(this.threads);
+    this.sequencer.stepThreads();
     this._updateScriptGlows();
-    for (var i = 0; i < inactiveThreads.length; i++) {
-        this._removeThread(inactiveThreads[i]);
-    }
 };
 
 Runtime.prototype.setEditingTarget = function (editingTarget) {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -70,6 +70,18 @@ function Runtime () {
      * @type {number}
      */
     this._cloneCounter = 0;
+
+    /**
+     * Whether the project is in "turbo mode."
+     * @type {Boolean}
+     */
+    this.turboMode = false;
+
+    /**
+     * Whether the project is in "pause mode."
+     * @type {Boolean}
+     */
+    this.pauseMode = false;
 }
 
 /**
@@ -451,6 +463,10 @@ Runtime.prototype.stopAll = function () {
  * inactive threads after each iteration.
  */
 Runtime.prototype._step = function () {
+    if (this.pauseMode) {
+        // Don't do any execution while in pause mode.
+        return;
+    }
     // Find all edge-activated hats, and add them to threads to be evaluated.
     for (var hatType in this._hats) {
         var hat = this._hats[hatType];
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -644,7 +644,7 @@ Runtime.prototype._updateScriptGlows = function (opt_extraThreads) {
         var thread = searchThreads[i];
         var target = thread.target;
         if (target == this._editingTarget) {
-            var blockForThread = thread.peekStack() || thread.topBlock;
+            var blockForThread = thread.peekStack();
             if (thread.requestScriptGlowInFrame) {
                 var script = target.blocks.getTopLevelScript(blockForThread);
                 if (!script) {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -65,6 +65,8 @@ function Runtime () {
 
     this._scriptGlowsPreviousFrame = [];
     this._editingTarget = null;
+    this._steppingInterval = null;
+    this.currentStepTime = null;
     /**
      * Currently known number of clones.
      * @type {number}
@@ -82,6 +84,12 @@ function Runtime () {
      * @type {Boolean}
      */
     this.pauseMode = false;
+
+    /**
+     * Whether the project is in "compatibility mode" (30 TPS).
+     * @type {Boolean}
+     */
+    this.compatibilityMode = false;
 }
 
 /**
@@ -135,6 +143,11 @@ util.inherits(Runtime, EventEmitter);
  * How rapidly we try to step threads, in ms.
  */
 Runtime.THREAD_STEP_INTERVAL = 1000 / 60;
+
+/**
+ * In compatibility mode, how rapidly we try to step threads, in ms.
+ */
+Runtime.THREAD_STEP_INTERVAL_COMPATIBILITY = 1000 / 30;
 
 /**
  * How many clones can be created at a time.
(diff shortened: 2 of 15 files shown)


CosmicWebServices:
Whoa nice

src/engine/runtime.js
```diff
@@ -610,21 +806,37 @@ Runtime.prototype.getTargetForStage = function () {
...
+Runtime.prototype.requestRedraw = function () {
+    this.redrawRequested = true;
+};
+
+/**
  * Handle an animation frame from the main thread.
  */
 Runtime.prototype.animationFrame = function () {
     if (this.renderer) {
+        // @todo: Only render when this.redrawRequested or clones rendered.
```
thisandagain, line 829: Mind adding a GH issue for this?

src/engine/sequencer.js
```diff
@@ -115,68 +186,39 @@ Sequencer.prototype.stepToBranch = function (thread, branchNum) {
...
-    if (nextBlockId) {
-        thread.pushStack(nextBlockId);
+    // Check if the call is recursive. If so, yield.
+    if (thread.isRecursiveCall(procedureCode)) {
+        thread.status = Thread.STATUS_YIELD;
     }
-    // If we can't find a next block to run, mark the thread as done.
-    if (!thread.peekStack()) {
-        thread.setStatus(Thread.STATUS_DONE);
+    thread.pushStack(definition);
```
thisandagain, line 222: This section could probably use some additional exposition.

thisandagain (approved):
The work on the sequencer here looks great. A couple really minor notes otherwise this is 👍

thisandagain:
@cwillisf Mind giving this a look as well?

src/engine/runtime.js
tmickel: Done!

playground/index.html
```diff
@@ -15,6 +15,19 @@
...
             <button id="stopall">Stop</button>
         </div>
+        <div>
+            Turbo: <input id='turbomode' type='checkbox' />
+        </div>
+        <div>
+            Pause: <input id='pausemode' type='checkbox' />
+        </div>
+        <div>
+            Compatibility (30 TPS):  <input id='compatmode' type='checkbox' />
```
towerofnix, line 25: Two spaces here?

playground/index.html
```diff
@@ -15,6 +15,19 @@
...
+        </div>
+        <div>
+            Single stepping:  <input id='singlestepmode' type='checkbox' />
```
towerofnix, line 28: Also two spaces here?

Just nitpicking. It's not like HTML generally cares about multiple whitespace, but still..

tmickel: Thanks, fixed. Haven't been too worried about the HTML in the playground, since it's just a testing/rapid prototype environment.

tmickel:
Dealing with _command blocks_ that return Promises is now a little broken. Reporters work fine. This has been pretty tricky...might like to fix the command-block-promises in a separate PR.

src/engine/runtime.js
```diff
@@ -449,44 +542,124 @@ Runtime.prototype.stopAll = function () {
...
+        }
+    }
+    this.pauseMode = pauseModeOn;
+};
+
+/**
+ * Set whether we are in 30 TPS compatibility mode.
+ * @param {boolean} compatibilityModeOn True iff in compatibility mode.
+ */
+Runtime.prototype.setCompatibilityMode = function (compatibilityModeOn) {
```
cwillisf, line 593: Potential future change: what do you think about expressing this in terms of setting a target frame rate or frame time instead of a Boolean?

src/engine/runtime.js
```diff
@@ -509,7 +682,30 @@ Runtime.prototype._updateScriptGlows = function () {
             finalScriptGlows.push(currentFrameGlow);
         }
     }
+    for (var m = 0; m < this._blockGlowsPreviousFrame.length; m++) {
+        var previousBlockGlow = this._blockGlowsPreviousFrame[m];
+        if (requestedBlockGlowsThisFrame.indexOf(previousBlockGlow) < 0) {
+            // Glow turned off.
+            try {
+                this.glowBlock(previousBlockGlow, false);
```
cwillisf, line 690: Is there a reasonable way to predict when this will throw and instead avoid the call in those situations? This type of "catch and suppress" situation has hidden bugs in Scratch 2.0 frequently enough that I get nervous when I see an empty `catch` clause...

src/engine/sequencer.js
```diff
@@ -115,68 +192,51 @@ Sequencer.prototype.stepToBranch = function (thread, branchNum) {
...
-        // Push null, so we come back to the current block.
         thread.pushStack(null);
     }
 };
 
 /**
  * Step a procedure.
  * @param {!Thread} thread Thread object to step to procedure.
- * @param {!string} procedureName Name of procedure defined in this target.
+ * @param {!string} procedureCode Proceduer code of procedure to step to.
```
cwillisf: Typo in "Procedure"

src/engine/thread.js
```diff
@@ -80,7 +99,14 @@ Thread.prototype.pushStack = function (blockId) {
...
     // Might not, if we just popped the stack.
     if (this.stack.length > this.stackFrames.length) {
+        // Copy warp mode from any higher level.
+        var warpMode = false;
+        if (this.stackFrames[this.stackFrames.length - 1]) {
+            warpMode = this.stackFrames[this.stackFrames.length - 1].warpMode;
+        }
         this.stackFrames.push({
+            isLoop: false, // Whether this level of the stack is a loop.
+            warpMode: warpMode, // Whether this level is in warp mode.
```
cwillisf, line 109: Potential future change: it seems like storing `warpMode` in the stack frame gets complicated, in that there's more than one place (here, `goToNextBlock`, more?) which need to remember to copy the flag to a callee's stack frame.

Is there a safer way to track the flag, less prone to human error? Maybe as a function a argument? I don't want mile-long argument lists either, though. Just something to think about...

src/index.js
```diff
@@ -67,6 +67,53 @@ VirtualMachine.prototype.greenFlag = function () {
 };
 
 /**
+ * Set whether the VM is in "turbo mode."
+ * When true, loops don't yield to redraw.
+ * @param {Boolean} turboModeOn Whether turbo mode should be set.
+ */
+VirtualMachine.prototype.setTurboMode = function (turboModeOn) {
+    this.runtime.turboMode = (turboModeOn == true);
```
cwillisf: Just curious: why `(turboModeOn == true)` rather than, say, `!!turboModeOn`? Not that it should matter :)

cwillisf (approved):
Awesome work! 👍

src/engine/runtime.js
tmickel: Hmm, I'm not sure. I guess it's both a question from the GUI/design perspective and from the API perspective...

My instinct was to do it this way in case we needed to use compatibility mode for more than just frame rate. E.g., back in our WebWorker days we had talked about doing locking behavior/whatever specifically for old projects in compatibility mode. Not sure if there are any of those cases anymore, but there could be.

src/engine/runtime.js
tmickel: That's a good point - I wasn't super happy writing this code.

I have a couple of ideas, will think about it more.

src/engine/thread.js
tmickel: Yeah, it would be nice to reduce this complexity. A function argument might work, but I'm a little nervous about making sure we always call with the right warpMode value...it would be worth trying to implement it that way but I'm not confident it'll be simpler. Also, when the thread runs out of warp time, we need to make sure when we resume execution of it, warp mode continues. Which might be annoying if we're not storing the value on the stack - we'll need to walk up to the proc definition again and look up the flag.

Hmmm...

src/index.js
tmickel: No reason - happy to switch to `!!` if that's more idiomatic!

src/engine/sequencer.js
tmickel: Thanks for catching that!

tmickel:
Ok, will merge for now and hope to continue those discussions in issues! Thanks for the quick review @cwillisf and @thisandagain !!

src/engine/thread.js
tjvr: FWIW, (sorry to butt in!), I believe Phosphorus stores this as a thread property, but copies it to an interpreter global while executing the thread. :-)

tmickel: @tjvr Thanks - that's useful! I suspect Phosphorus is working similar to Scratch 2.0, where things are loaded from the stack if necessary into an interpreter global. I'll check it out.

alexwenbj:
@tmickel Hi,I check the latest version of VM,but can't find the setPauseMode() in the source code. Was it deleted or replaced by something?Thank you.

tmickel:
@alexwenbj This was just a prototype a long time ago, it never made it into the production version! :)

alexwenbj:
@tmickel Oh,Thank you.But in fact it is a very useful feature for users.

PR #282, opened 2016-10-17
Add VM.clear method

### LINKED ISSUE #274: VM clear method notes
For the GUI, we'd like to be able to unload projects from the VM without deleting the whole VM object. Then new projects could be loaded at any time.

Roughly, `clear` method should dispose all the current targets, using `runtime.disposeTarget`. https://github.com/LLK/scratch-vm/blob/3bfd755e6005fcf528a544548e517dce6122c8b0/src/engine/runtime.js#L375 We could actually just do something like "for each target in `runtime.targets` (dispose target)`. 

Note that we should also update `clone.dispose`, to _allow_ disposing of original sprites:
https://github.com/LLK/scratch-vm/blob/3bfd755e6005fcf528a544548e517dce6122c8b0/src/sprites/clone.js#L420
And move that check to the "delete this clone" block instead.

We also have to stop all the currently running threads - I recommend `runtime.stopAll`.

rschamp:
Use it before loading projects so targets don't accumulate when multiple projects are loaded on the same instance.

Move check to see if the clone is the original clone onto the block implementation so all clones can be removed.

Fixes #274

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -378,7 +378,7 @@ Runtime.prototype.dispose = function () {
 
 /**
  * Dispose of a target.
- * @param {!Target} target Target to dispose of.
+ * @param {!Target} disposingTarget Target to dispose of.
  */
 Runtime.prototype.disposeTarget = function (disposingTarget) {
     this.targets = this.targets.filter(function (target) {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -369,17 +369,25 @@ Runtime.prototype.startHats = function (requestedHatOpcode,
 };
 
 /**
+ * Dispose all targets. Return to clean state.
+ */
+Runtime.prototype.dispose = function () {
+    this.stopAll();
+    this.targets.map(this.disposeTarget, this);
+};
+
+/**
  * Dispose of a target.
  * @param {!Target} target Target to dispose of.
  */
-Runtime.prototype.disposeTarget = function (target) {
-    // Allow target to do dispose actions.
-    target.dispose();
-    // Remove from list of targets.
-    var index = this.targets.indexOf(target);
-    if (index > -1) {
-        this.targets.splice(index, 1);
-    }
+Runtime.prototype.disposeTarget = function (disposingTarget) {
+    this.targets = this.targets.filter(function (target) {
+        if (disposingTarget !== target) return true;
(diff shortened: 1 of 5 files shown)


src/engine/runtime.js
```diff
@@ -369,17 +369,25 @@ Runtime.prototype.startHats = function (requestedHatOpcode,
...
-Runtime.prototype.disposeTarget = function (target) {
-    // Allow target to do dispose actions.
-    target.dispose();
-    // Remove from list of targets.
-    var index = this.targets.indexOf(target);
-    if (index > -1) {
-        this.targets.splice(index, 1);
-    }
+Runtime.prototype.disposeTarget = function (disposingTarget) {
+    this.targets = this.targets.filter(function (target) {
```
rschamp, line 384: I needed to treat `this.targets` as immutable because otherwise the list changes under the feet of `Runtime.dispose` and not all targets are disposed.

src/index.js
```diff
@@ -74,6 +74,15 @@ VirtualMachine.prototype.stopAll = function () {
 };
 
 /**
+ * Clear out current running project data.
+ */
+VirtualMachine.prototype.clear = function () {
+    this.runtime.dispose();
+    this.editingTarget = null;
+    this.emitTargetsUpdate();
```
rschamp, line 82: I could be convinced that this is the responsibility of the `vm.clear()` caller.

src/index.js
```diff
@@ -237,7 +247,7 @@ VirtualMachine.prototype.emitTargetsUpdate = function () {
             return [target.id, target.getName()];
         }),
         // Currently editing target id.
-        editingTarget: this.editingTarget.id
+        editingTarget: this.editingTarget ? this.editingTarget.id : null
```
rschamp, line 250: Due to the above, `editingTarget` is sometimes `null` when `emitTargetsUpdate` is called.

src/index.js
tmickel: I like this way - a VM consumer could call `vm.clear()` and automatically get the target updates as expected.

src/engine/runtime.js
tmickel: Makes sense!

tmickel:
lgtm minus the lint/JSDoc update. thanks Ray!

rschamp:
@tmickel Thank you for the detailed description!

PR #283, opened 2016-10-17
Fix parent setting in sb2 importer

### LINKED ISSUE #205: Something's up with stack highlights.
From @picklesrus:
"I replaced the repeat block in Champ99 with a repeat forever block, leaving the old repeat floating the in the workspace.  When I run it, the floating repeat block highlights along with the stack of blocks actually attached to the start block."
Seems bad to me - like something is having the wrong parent set in the block representation?

tmickel:
Previously, the SB2 importer would set the parent of all blocks in an inner block list to the outer block. But, the expected behavior is that blocks after the 0th have their parent set to the previous block. This was leading to weird behavior where splitting the blocks out of the input could lead to old non-connected blocks to be highlighted when the block was running. I believe this fixes #205.

### DIFF
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -292,8 +292,14 @@ function parseBlock (sb2block) {
                     // Single block occupies the input.
                     innerBlocks = [parseBlock(providedArg)];
                 }
+                var previousBlock = null;
                 for (var j = 0; j < innerBlocks.length; j++) {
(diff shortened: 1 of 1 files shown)


thisandagain:
Ah. Good catch ... I should have seen that when writing the tests. 👍

MooShoe:
>

PR #291, opened 2016-10-19
Functions to add Sprite2/costumes/backdrops

tmickel:
These are intended to be used by the GUI and pair with a PR over there.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -175,12 +175,49 @@ VirtualMachine.prototype.loadProject = function (json) {
     this.clear();
     // @todo: Handle other formats, e.g., Scratch 1.4, Scratch 3.0.
     sb2import(json, this.runtime);
-    // Select the first target for editing, e.g., the stage.
-    this.editingTarget = this.runtime.targets[0];
+    // Select the first target for editing, e.g., the first sprite.
+    this.editingTarget = this.runtime.targets[1];
     // Update the VM user's knowledge of targets and blocks on the workspace.
     this.emitTargetsUpdate();
     this.emitWorkspaceUpdate();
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -175,15 +175,52 @@ VirtualMachine.prototype.loadProject = function (json) {
...
+};
+
+/**
+ * Add a single sprite from the "Sprite2" (i.e., SB2 sprite) format.
+ * @param {?string} json JSON string representing the sprite.
+ */
+VirtualMachine.prototype.addSprite2 = function (json) {
+    sb2import(json, this.runtime, true);
+    // Select new sprite.
+    this.editingTarget = this.runtime.targets[this.runtime.targets.length - 1];
```
rschamp: Maybe `sb2import` should return the target that was just imported? Then this wouldn't need to know how things should be ordered.

tmickel: Hmmm, it could. Although for a sb2 (not sprite2), it's actually just imported _many_ targets. It could return the top-level target (stage) for that situation, though?

rschamp: I think that would make sense. I also think it could return `null` there to indicate there's no one target to set as the editing target. In that case it's up to the thing that imported to decide which target to use?  I think the stage idea might make more sense though.

tmickel: Ok, done!

tmickel: (with the stage idea)

rschamp: Great, thanks! Makes total sense to return the stage seeing as that's what `target` ends up being in that case anyway.

rschamp (approved):
Looks great, thanks!

PR #294, opened 2016-10-20
Drop single-stepping and pause modes

tmickel:
After a design discussion yesterday, it seems that both of these would need to be reimplemented in a different way. Also, I realized that this implementation doesn't match Scratch 1.0 single-stepping. In that implementation, execution proceeds to step through threads in the same order it would in normal mode. But, in this implementation, it allows all threads to be stepped in a single stepThreads, which actually changes the execution behavior. If we want a correct implementation, it would need to be redone.

### DIFF
--- a/playground/index.html
+++ b/playground/index.html
@@ -19,14 +19,7 @@
             Turbo: <input id='turbomode' type='checkbox' />
         </div>
         <div>
-            Pause: <input id='pausemode' type='checkbox' />
-        </div>
-        <div>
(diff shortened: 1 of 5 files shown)


CosmicWebServices:
Noooooo!!!!!

tmickel:
@CosmicWebServices Haha, I'm sorry. I hope something like them will come back eventually...

CosmicWebServices:
😭

PR #298, opened 2016-10-24
Use shared lint config: eslint-config-scratch

rschamp:
Towards more consistent style across our JS projects, I've added a centralized ESLint configuration at https://github.com/LLK/eslint-config-scratch.  This PR installs the config and updates the syntax to match its expectations.

Since it's a centralized config to be used in all our projects, the rules should be a collective decision so I assigned a bunch of people who might care.  If any of these changes annoy you, let's talk about it.  I just created it based on what I thought made sense, and then removed rules as I tried to lint this repo.

A few of the rules that required most of the changes:
- [`'func-style': ['error', 'expression']`](http://eslint.org/docs/rules/func-style): Functions should be written in the form `var functionName = function () {}`. This ensures that functions are defined before they're used, which makes code easier to understand as it's read. It also forces the code to be written as it will be executed. The declaration style `function functionName () {}` relies on functions to be automatically hoisted to the top of the file.
- [`'space-before-function-paren': ['error', 'always']`](http://eslint.org/docs/rules/space-before-function-paren): This is the style we've been using elsewhere. Thankfully you can automatically apply this rule to existing code by running `eslint` with the `--fix` argument.
- [`'camelcase': ['error']`](http://eslint.org/docs/rules/camelcase): All variables should be written in "camelCase" style, without underscores. This improves consistency of naming.  There is a convention in this codebase to name options with a mix of underscore and camelcase, like `opt_optionName`.  These have been renamed `optOptionName`.  Unfortunately there's no way to configure a pattern for allowed variable names to include the `opt_` prefix.
- [`'eqeqeq': ['error']`](http://eslint.org/docs/rules/eqeqeq): Always using `===` or `!==` leads to fewer ["wat?"](https://www.destroyallsoftware.com/talks/wat) errors
- [`'no-undefined': ['error']`](http://eslint.org/docs/rules/no-undefined): Don't use `undefined`. This rule is mostly to ensure consistency in syntax across the projects in the way we use `undefined`, but also avoids similar errors to those avoided by the `===` rule.


src/engine/execute.js
```diff
@@ -197,16 +196,16 @@ var execute = function (sequencer, thread) {
...
-                runtime.startHats(requestedHat, opt_matchFields, opt_target)
+                runtime.startHats(requestedHat, optMatchFields, optTarget)
             );
         },
         ioQuery: function (device, func, args) {
             // Find the I/O device and execute the query/function call.
             if (runtime.ioDevices[device] && runtime.ioDevices[device][func]) {
                 var devObject = runtime.ioDevices[device];
-                return devObject[func].call(devObject, args);
+                return devObject[func](args);
```
rschamp: http://eslint.org/docs/rules/no-useless-call

src/blocks/scratch3_motion.js
```diff
@@ -149,10 +149,10 @@ Scratch3MotionBlocks.prototype.ifOnEdgeBounce = function (args, util) {
...
     var stageWidth = this.runtime.constructor.STAGE_WIDTH;
     var stageHeight = this.runtime.constructor.STAGE_HEIGHT;
-    var distLeft = Math.max(0, stageWidth / 2 + bounds.left);
-    var distTop = Math.max(0, stageHeight / 2 - bounds.top);
-    var distRight = Math.max(0, stageWidth / 2 - bounds.right);
-    var distBottom = Math.max(0, stageHeight / 2 + bounds.bottom);
+    var distLeft = Math.max(0, (stageWidth / 2) + bounds.left);
+    var distTop = Math.max(0, (stageHeight / 2) - bounds.top);
+    var distRight = Math.max(0, (stageWidth / 2) - bounds.right);
+    var distBottom = Math.max(0, (stageHeight / 2) + bounds.bottom);
```
rschamp, line 155: http://eslint.org/docs/rules/no-mixed-operators

src/blocks/scratch3_operators.js
```diff
@@ -79,10 +79,10 @@ Scratch3OperatorsBlocks.prototype.random = function (args) {
     var nTo = Cast.toNumber(args.TO);
     var low = nFrom <= nTo ? nFrom : nTo;
     var high = nFrom <= nTo ? nTo : nFrom;
-    if (low == high) return low;
+    if (low === high) return low;
     // If both arguments are ints, truncate the result to an int.
     if (Cast.isInt(args.FROM) && Cast.isInt(args.TO)) {
-        return low + parseInt(Math.random() * ((high + 1) - low));
+        return low + parseInt(Math.random() * ((high + 1) - low), 10);
```
rschamp, line 85: http://eslint.org/docs/rules/radix

src/engine/adapter.js
```diff
@@ -41,7 +27,21 @@ function domToBlocks (blocksDOM) {
...
+ * @param {Object} e `Blockly.events.create`
+ * @return {Array.<Object>} List of blocks from this CREATE event.
+ */
+module.exports = function (e) {
+    // Validate input
+    if (typeof e !== 'object') return;
+    if (typeof e.xml !== 'object') return;
+
+    return domToBlocks(html.parseDOM(e.xml.outerHTML));
+};
```
rschamp, line 44: http://eslint.org/docs/rules/no-use-before-define

src/engine/execute.js
```diff
@@ -76,12 +77,10 @@ var execute = function (sequencer, thread) {
...
                     sequencer.retireThread(thread);
                 }
-            } else {
+            } else if (!resolvedValue) {
                 // Not an edge-activated hat: retire the thread
                 // if predicate was false.
-                if (!resolvedValue) {
-                    sequencer.retireThread(thread);
-                }
+                sequencer.retireThread(thread);
```
rschamp: http://eslint.org/docs/rules/no-lonely-if

src/engine/execute.js
```diff
@@ -133,8 +132,8 @@ var execute = function (sequencer, thread) {
         var input = inputs[inputName];
         var inputBlockId = input.block;
         // Is there no value for this input waiting in the stack frame?
-        if (typeof currentStackFrame.reported[inputName] === 'undefined'
-            && inputBlockId) {
+        if (typeof currentStackFrame.reported[inputName] === 'undefined' &&
+            inputBlockId) {
```
rschamp, line 138: http://eslint.org/docs/rules/operator-linebreak

src/import/sb2import.js
```diff
@@ -214,44 +215,44 @@ function flatten (blocks) {
...
-function parseProcedureArgMap (procCode) {
+var parseProcedureArgMap = function (procCode) {
     var argMap = [
         {} // First item in list is op string.
     ];
     var INPUT_PREFIX = 'input';
     var inputCount = 0;
     // Split by %n, %b, %s.
-    var parts = procCode.split(/(?=[^\\]\%[nbs])/);
+    var parts = procCode.split(/(?=[^\\]%[nbs])/);
```
rschamp, line 225: http://eslint.org/docs/rules/no-useless-escape

rschamp (commented):
I tried to explain some of the other fixes required for the new rules.  The last one I left out was using minilog rather than console. This is mostly for consistency, but also gives us the ability to log somewhere other than the console if/when we want to do that, and prefixes the messages with `vm`, which will be useful when we have messages coming from many different projects at once.

.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    extends: ['scratch', 'scratch/node'],
+    env: {
+        node: true
```
thisandagain: The `node: true` environment isn't included as part of `scratch/node`?

src/blocks/scratch3_control.js
```diff
@@ -1,19 +1,19 @@
 var Cast = require('../util/cast');
 var Timer = require('../util/timer');
 
-function Scratch3ControlBlocks(runtime) {
+var Scratch3ControlBlocks = function (runtime) {
```
thisandagain, line 4: I'm not sure I understand the usefulness of this eslint feature.

src/blocks/scratch3_motion.js
thisandagain: I'm not 100% sure about this one given, "Please Excuse My Dear Aunt Sally" but it does make things a little easier to read so 👍

src/blocks/scratch3_control.js
```diff
@@ -36,10 +36,10 @@ Scratch3ControlBlocks.prototype.getHats = function () {
     };
 };
 
-Scratch3ControlBlocks.prototype.repeat = function(args, util) {
+Scratch3ControlBlocks.prototype.repeat = function (args, util) {
     var times = Math.floor(Cast.toNumber(args.TIMES));
     // Initialize loop
-    if (util.stackFrame.loopCounter === undefined) {
+    if (typeof util.stackFrame.loopCounter === 'undefined') {
```
thisandagain, line 42: 👍

src/blocks/scratch3_control.js
```diff
@@ -102,21 +102,21 @@ Scratch3ControlBlocks.prototype.ifElse = function(args, util) {
     }
 };
 
-Scratch3ControlBlocks.prototype.stop = function(args, util) {
+Scratch3ControlBlocks.prototype.stop = function (args, util) {
     var option = args.STOP_OPTION;
-    if (option == 'all') {
+    if (option === 'all') {
```
thisandagain, line 107: 👍

src/engine/blocks.js
```diff
@@ -399,7 +399,7 @@ Blocks.prototype.blockToXML = function (blockId) {
     // Encode properties of this block.
     var tagName = (block.shadow) ? 'shadow' : 'block';
     var xy = (block.topLevel) ?
-        ' x="' + block.x +'"' + ' y="' + block.y +'"' :
+        ' x="' + block.x +'" y="' + block.y +'"' :
```
thisandagain: This is a curious fix. It makes sense, but I'm surprised we aren't enforcing spaces around operators (e.g. `'foo' + 'bar'` vs `'foo'+ 'bar'`) like we are elsewhere (e.g. `function ()`)

src/engine/execute.js
thisandagain: I'm not sure about this one.

src/engine/list.js
```diff
@@ -8,9 +8,7 @@
   * @param {Array} contents Contents of the list, as an array.
   * @constructor
   */
-function List (name, contents) {
+module.exports = function List (name, contents) {
```
thisandagain: The `'func-style': ['error', 'expression']` rule seems to be adding inconsistency to our use of the class pattern here. Before this rule was applied it seemed fairly consistent / clear:

``` js
function Foo () {
    // constructor stuff
}

Foo.prototype.bar = function () {
    // other stuff
};

module.exports = Foo;
```

We now have inconsistency:

``` js
module.exports = function Foo () {
    // ...
};
```

``` js
var Foo = function () {
    // ...
};

module.exports = Foo;
```

test/unit/util_color.js
```diff
@@ -11,25 +11,25 @@ test('decimalToHex', function (t) {
 });
 
 test('decimalToRgb', function (t) {
-    t.deepEqual(color.decimalToRgb(0), {r:0,g:0,b:0});
-    t.deepEqual(color.decimalToRgb(1), {r:0,g:0,b:1});
-    t.deepEqual(color.decimalToRgb(16777215), {r:255,g:255,b:255});
-    t.deepEqual(color.decimalToRgb(-16777215), {r:0,g:0,b:1});
-    t.deepEqual(color.decimalToRgb(99999999), {r:245,g:224,b:255});
+    t.deepEqual(color.decimalToRgb(0), {r: 0,g: 0,b: 0});
```
thisandagain: If we are going to enforce a space after the `:` we should probably also enforce a space after the `,`.

thisandagain (commented):
Thanks for pulling this together @rschamp. Mostly looks good though I left a few notes inline.

test/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
thisandagain, line 1: Are we using `undefined` extensively in the tests?

src/engine/blocks.js
rschamp: Good catch, it looks like there's something missing re: spaces around operators.

src/engine/list.js
rschamp: I don't think this rule is very important since `no-use-before-define` keeps things understandable even with the hoisting from the declaration style.  I think I was going back and forth when I made the changes for `module.exports = function Foo () {}`.
What if it was always consistent e.g.

``` javascript
var Foo = function () {
    // ...
};

module.exports = Foo;
```

test/unit/util_color.js
rschamp: Yeah another example that something is up with operator spacing.

test/.eslintrc.js
rschamp: Yes, e.g. https://github.com/LLK/scratch-vm/blob/84bd42d734a6fac07d3c92b2e9dea97d3a303325/test/unit/util_cast.js

src/engine/execute.js
rschamp: On second look, it actually makes things less readable because you collapse the conceptual block contained in the `else { }` into another line of logic.  Removing this.

.eslintrc.js
rschamp: No... but it should be.

test/.eslintrc.js
thisandagain: Whoops! Forgetting about code I wrote. 😝

src/blocks/scratch3_looks.js
```diff
@@ -82,33 +82,33 @@ Scratch3LooksBlocks.prototype.hide = function (args, util) {
  * Matches the behavior of Scratch 2.0 for different types of arguments.
  * @param {!Target} target Target to set costume/backdrop to.
  * @param {Any} requestedCostume Costume requested, e.g., 0, 'name', etc.
- * @param {boolean=} opt_zeroIndex Set to zero-index the requestedCostume.
+ * @param {boolean=} optZeroIndex Set to zero-index the requestedCostume.
```
tmickel, line 85: note that `opt_` as a prefix seems to be a Closure/Blockly convention - this will just have us diverge from that, which I'm fine with.

src/blocks/scratch3_motion.js
tmickel: Dear before Aunt!

src/engine/execute.js
```diff
@@ -52,7 +53,7 @@ var execute = function (sequencer, thread) {
 
 
     if (!opcode) {
-        console.warn('Could not get opcode for block: ' + currentBlockId);
+        log.warn('Could not get opcode for block: ' + currentBlockId);
```
tmickel, line 56: not sure if I'm a fan of having to import log to do console things, but I guess it makes sense for final linted commits.

src/engine/list.js
tmickel: ^ I like that version myself

src/import/sb2specmap.js
```diff
@@ -22,1367 +22,1367 @@
...
+                'inputOp': 'math_number',
+                'inputName': 'NOTE'
             },
             {
-                'type':'input',
-                'inputOp':'math_number',
-                'inputName':'BEATS'
+                'type': 'input',
+                'inputOp': 'math_number',
+                'inputName': 'BEATS'
```
tmickel: for your sake I hope you had a way to automatically transform this!

tmickel (approved):
Looks sweet to me, thanks for doing all the work to investigate and apply these!!!

src/import/sb2specmap.js
rschamp: Yes, it's kind of cool — you run `npm run lint -- --fix`! Most whitespace-related rules can be automatically resolved.

src/engine/execute.js
rschamp: Reverted this one. I forgot had this argument with myself already and just made this rule a warning instead. This way you at least think about if it should be a new conceptual block or not.  Sometimes the opposite is true and it makes things more confusing than they need to be.

Thoughts on warning vs disabling it?

src/engine/list.js
thisandagain: I'm happy with any of the three patterns as long as we are consistent.

src/blocks/scratch3_motion.js
```diff
@@ -16,23 +16,23 @@ var Scratch3MotionBlocks = function (runtime) {
...
+        motion_movesteps: this.moveSteps,
+        motion_gotoxy: this.goToXY,
+        motion_goto: this.goTo,
+        motion_turnright: this.turnRight,
+        motion_turnleft: this.turnLeft,
+        motion_pointindirection: this.pointInDirection,
+        motion_pointtowards: this.pointTowards,
+        motion_glidesecstoxy: this.glide,
+        motion_ifonedgebounce: this.ifOnEdgeBounce,
+        motion_setrotationstyle: this.setRotationStyle,
```
tmickel, line 28: seems fine as long as we stick to reasonable keys. what if we have a key with non-js-identifier characters in it, like "move:steps" or whatever?

rschamp:
Updated spaces around infix operators and enforced consistent quotes around object properties: if any of the properties need quotes, then they all need quotes. Otherwise none of them should use quotes.  Also fixed the node env config @thisandagain pointed out.

src/blocks/scratch3_motion.js
tmickel: ^ oh ok I saw your comment!

thisandagain:
I think the only question I have left is the `func-style` rule / pattern consistency. Otherwise this looks 👍  to me. Any thoughts @mewtaylor @cwillisf ?

rschamp:
@thisandagain Updated to the `var Foo = function () {}; ... module.exports = Foo` style, this is what we mostly use elsewhere.

.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+    extends: ['scratch', 'scratch/node']
```
cwillisf, line 2: 👍

src/blocks/scratch3_control.js
cwillisf: I agree... I have mixed feelings about this one. I understand the reasoning that defining a function before it's called can make things more readable, but it's not always possible. If you introduce a situation where `var funcA = function (...) { if (something) funcB(); }` and `var funcB = function (...) { if (something) funcA(); }` then how does this rule react?

mewtaylor (approved):
Thanks for this! With the suggestions made by others, this looks great!

src/engine/sequencer.js
```diff
@@ -219,11 +219,9 @@ Sequencer.prototype.stepToProcedure = function (thread, procedureCode) {
         var doWarp = definitionBlock.mutation.warp;
         if (doWarp) {
             thread.peekStackFrame().warpMode = true;
-        } else {
+        } else if (isRecursive) {
```
cwillisf: 👍

src/engine/execute.js
```diff
@@ -197,16 +198,21 @@ var execute = function (sequencer, thread) {
...
             // Find the I/O device and execute the query/function call.
             if (runtime.ioDevices[device] && runtime.ioDevices[device][func]) {
                 var devObject = runtime.ioDevices[device];
-                return devObject[func].call(devObject, args);
+                // @todo Figure out why eslint complains about no-useless-call
+                // no-useless-call can't tell if the call is useless for dynamic
+                // expressions... or something. Not exactly sure why it
+                // complains here.
+                // eslint-disable-next-line no-useless-call
+                return devObject[func].call(devObject[func], args);
```
cwillisf: This is different from the old code.
Also, I wonder if this would make the code easier to understand and/or suppress the `eslint` complaint:

``` js
var realFunc = devObject[func];
realFunc.call(devObject, args);
```

(Please double-check that my suggestion is equivalent to the original code...!)

cwillisf (changes requested):
Overall 👍 but I'm unsure about the `var` vs. `function` change and I think there's a typo in the latest version of the `devObject[func]` stuff.

Thanks for doing all this work! I look forward to improved consistency across our Scratch 3.0-related repositories :D

src/blocks/scratch3_control.js
rschamp: For a circular reference like this, you would need make an exception for the `no-use-before-define` rule, but the function definition style wouldn't make any difference since by the time the function was called, it would be defined.

src/engine/execute.js
rschamp: Thanks, fixed

thisandagain (approved):
Thanks for wrangling all this @rschamp

cwillisf (approved):

PR #309, opened 2016-10-24
Divide rotation center by bitmap resolution

tmickel:
I noticed the values were doubled when bmp resolution = 2. This seems to fix a lot of projects (+ SVG quirk mode).


tmickel:
(actually...should this be done by the renderer? hmm)

dekrain (approved):

PR #310, opened 2016-10-24
Improve Test Coverage

thisandagain:
## Overview
- Add test coverage for `blocks/scratch3_control` module
- Add test coverage for `io/clock` module
- Add test coverage for `io/keyboard` module
- Add test coverage for `io/mouse` module
- Add JSDoc for `io/keyboard` module
- Add JSDoc for `io/mouse` module
- Add `./coverage` directory to `.eslintignore`
- Merge `tap-unit` and `tap-integration` NPM scripts to resolve issue with code coverage reporting to [coveralls](https://coveralls.io/github/LLK/scratch-vm?branch=develop)
## Before

image
## After

image


src/io/keyboard.js
```diff
@@ -52,6 +59,11 @@ Keyboard.prototype._keyCodeToScratchKey = function (keyCode) {
     return null;
 };
 
+/**
+ * Keyboard DOM event handler.
+ * @param  {object} data DOM event.
```
tmickel: Currently the things passed to `postData` are not necessarily DOM events: https://github.com/LLK/scratch-vm/blob/develop/playground/playground.js#L171

src/io/keyboard.js
```diff
@@ -52,6 +59,11 @@ Keyboard.prototype._keyCodeToScratchKey = function (keyCode) {
...
+ * @return {void}
```
tmickel: Just wondering, why `@return {void}` instead of just excluding it?

src/io/keyboard.js
```diff
@@ -74,6 +86,11 @@ Keyboard.prototype.postData = function (data) {
     }
 };
 
+/**
+ * Get key down state for a specified Scratch key name.
+ * @param  {Any} key Scratch key argument.
```
tmickel, line 90: I think this should be `@param {string}`

tmickel: Never mind - I see that the casting is handled elsewhere.

test/unit/blocks_control.js
```diff
@@ -0,0 +1,130 @@
...
+    var c = new Control(rt);
+
+    // Test harness (mocks `util`)
+    var i = 0;
+    var repeat = 10;
+    var util = {
+        stackFrame: Object.create(null),
+        startBranch: function () {
+            i++;
+            c.repeat({TIMES: repeat}, util);
```
tmickel, line 23: This makes sense that it works but feels weird that `startBranch` is going to call `c.repeat` again. In reality `startBranch` would set up the sequencer so that we get something more like `for (var i = 0; i < 10; i++) { c.repeat{TIMES: 10}, util);}` if that makes any sense. However I'm fine with this as a unit test, I suppose...

test/unit/blocks_control.js
```diff
@@ -0,0 +1,130 @@
...
+
+    // Execute test
+    c.stop({STOP_OPTION: 'all'}, util);
+    c.stop({STOP_OPTION: 'other scripts in sprite'}, util);
+    c.stop({STOP_OPTION: 'other scripts in stage'}, util);
+    c.stop({STOP_OPTION: 'this script'}, util);
+    t.strictEqual(state.stopAll, 1);
+    t.strictEqual(state.stopOtherTargetThreads, 2);
+    t.strictEqual(state.stopThread, 1);
+    t.end();
```
tmickel, line 129: Overall these are pretty nice!

test/unit/io_clock.js
```diff
@@ -0,0 +1,34 @@
...
+    t.type(c.resume, 'function');
+    t.type(c.resetProjectTimer, 'function');
+    t.end();
+});
+
+test('cycle', function (t) {
+    var rt = new Runtime();
+    var c = new Clock(rt);
+
+    t.ok(c.projectTimer() <= 0.001);
```
tmickel: is it possible the test could fail if we take a lot of time between instantiation of `c` and this line here? Seems exceedingly unlikely, but could it happen in a slow-CPU environment like travis?

test/unit/io_mouse.js
```diff
@@ -0,0 +1,49 @@
...
+
+test('mouseUp', function (t) {
+    var rt = new Runtime();
+    var m = new Mouse(rt);
+
+    m.postData({
+        x: 1,
+        y: 10,
+        isDown: false,
+        canvasWidth: 480,
```
tmickel, line 25: note that we are planning to refactor what data needs to be posted after https://github.com/LLK/scratch-render/issues/29

src/io/mouse.js
```diff
@@ -12,21 +12,13 @@ function Mouse (runtime) {
...
-        if (this._isDown) {
-            this._activateClickHats(data.x, data.y);
-        }
-    }
-};
-
+/**
+ * Activate "event_whenthisspriteclicked" hats if needed.
+ * @param  {int} x X position to be sent to the renderer.
+ * @param  {int} y Y position to be sent to the renderer.
```
tmickel: Another JSDoc curiosity - I've seen `int` and `float` in some places, although these aren't JavaScript types as far as I know. what is our general approach?

tmickel (approved):
Just a bunch of questions/notes, but looks pretty good to me. thanks!

package.json
```diff
@@ -18,9 +18,8 @@
     "prepublish": "npm run build",
     "prepublish-watch": "npm run watch",
     "start": "./node_modules/.bin/webpack-dev-server",
-    "tap-integration": "./node_modules/.bin/tap ./test/integration/*.js",
-    "tap-unit": "./node_modules/.bin/tap ./test/unit/*.js",
-    "test": "npm run lint && npm run tap-unit && npm run tap-integration",
+    "tap": "./node_modules/.bin/tap ./test/{unit,integration}/*.js",
```
thisandagain, line 21: Unfortunately coveralls requires all tests to be run at once in order to be able to generate a coverage report. This should resolve the issue where coveralls is reporting our coverage rate as far too low.

src/io/keyboard.js
thisandagain: Just a old habit. I can remove these.

test/unit/blocks_control.js
thisandagain: I agree that it's a little odd. I was looking to minimize the test harness as much as possible.

test/unit/io_clock.js
thisandagain: It could which is why this is not `0`. These worst case scenario I was able to get was `0.001` but it may make more sense to give it a margin of error closer to 100ms (`0.1`).

test/unit/io_mouse.js
thisandagain: 👍

package.json
rschamp: I had broken this out in case the `{unit,integration}` syntax breaks on Windows. Not sure if that's the case or not.

src/io/mouse.js
thisandagain: Another good catch. I'll clean that up to use `Number`.

package.json
thisandagain: That string is handled by [node-glob](https://github.com/isaacs/node-glob)

test/unit/io_clock.js
```diff
@@ -0,0 +1,34 @@
...
+        c.resetProjectTimer();
+        setTimeout(function () {
+            t.ok(c.projectTimer() > 0);
+            c.pause();
+            t.ok(c.projectTimer() > 0);
+            c.resume();
+            t.ok(c.projectTimer() > 0);
+            t.end();
+        }, 100);
+    }, 100);
```
rschamp, line 33: It's not that long, but given the resolution of the timer, could these timeouts be smaller for what's being tested?
I am wondering if it would be useful to flesh the test out a bit more so that it can tell if the pause and resume methods are functioning? It would pass if the timer was paused for the entirety of the inner `setTimeout`, or if it never paused even after pause was called.  Is this the place for that test?

rschamp (approved):
Sorry, I just killed your ✅  when I merged the lint PR.  I think this looks great overall — the timer tests don't really need changes from the standpoint of coverage.

rschamp:
Assigning to @thisandagain for conflicts/possible lint issues.

thisandagain:
Some crazy-how I didn't hit any lint issues. 💥

PR #319, opened 2016-10-26
Rename and delete sprites

### LINKED ISSUE #144: Add "new sprite", "delete sprite", "rename sprite" mechanisms

tmickel:
Adds rename and delete sprite methods, which take a target ID. Resolves #144.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -203,6 +203,37 @@ VirtualMachine.prototype.renameSprite = function (targetId, newName) {
 };
 
 /**
+ * Delete a sprite and all its clones.
+ * @param {string} targetId ID of a target whose sprite to delete.
+ */
+VirtualMachine.prototype.deleteSprite = function (targetId) {
+    var target = this.runtime.getTargetById(targetId);
+    if (target) {
+        if (target.isStage) {
+            throw new Error('Cannot delete stage targets.');
+        }
+        var sprite = target.sprite;
+        if (!sprite) {
+            throw new Error('No sprite associated with this target.');
+        }
+        var currentEditingTarget = this.editingTarget;
+        for (var i = 0; i < sprite.clones.length; i++) {
+            var clone = sprite.clones[i];
+            this.runtime.stopForTarget(sprite.clones[i]);
+            this.runtime.disposeTarget(sprite.clones[i]);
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -184,6 +184,56 @@ VirtualMachine.prototype.addBackdrop = function (backdropObject) {
...
+/**
+ * Delete a sprite and all its clones.
+ * @param {string} targetId ID of a target whose sprite to delete.
+ */
+VirtualMachine.prototype.deleteSprite = function (targetId) {
+    var target = this.runtime.getTargetById(targetId);
+    if (target) {
+        if (target.isStage) {
+            throw new Error('Cannot delete stage targets.');
+        }
```
rschamp, line 217: We should probably carry this check to the rename method as well, unless you intended to make renaming the stage possible?  If you do carry it over, maybe both of these checks should be put onto `RenderedTarget` as a method like `isSprite`? Or something less confusing than that.

src/index.js
```diff
@@ -184,6 +184,56 @@ VirtualMachine.prototype.addBackdrop = function (backdropObject) {
...
+            throw new Error('No sprite associated with this target.');
+        }
+        var currentEditingTarget = this.editingTarget;
+        for (var i = 0; i < sprite.clones.length; i++) {
+            var clone = sprite.clones[i];
+            this.runtime.stopForTarget(sprite.clones[i]);
+            this.runtime.disposeTarget(sprite.clones[i]);
+            // Ensure editing target is switched if we are deleting it.
+            if (clone === currentEditingTarget) {
+                this.setEditingTarget(this.runtime.targets[0].id);
```
rschamp, line 229: Is it certain that `targetId !== this.runtime.targets[0].id`?

rschamp (commented):
A couple of questions — overall it looks good to me!

src/index.js
tmickel: Done!

src/index.js
tmickel: Yes, since the target was just disposed - if it was originally targets[0], there is now a different target in targets[0]. The only risk is if there's no more targets - if so we might have a problem here. I've added a check to that effect.

rschamp (approved):

Issue #161, opened 2016-09-08
Emit target info (sprite info) for editingTarget & allow posting target info for editingTarget


tmickel:
Would allow the immediate implementation of the "sprite info" pane.

tmickel:
@rschamp FYI!

tmickel:
If you have ideas about how this should look!

towerofnix:
The sprite info pane is a big topic..

Generally you use it to get some quick info about what the status of a sprite is, but it's _very_ limited:
1. You can only rename a sprite, view its position and direction, change whether it can be dragged in the player, and whether it's shown or not.
2. When getting the info of a sprite, it's actually selected in the scripts pane as well - which, while handy for testing scripts with clones, can cause lag from viewing a large workspace. It also gets rid of the current scrolled area in the previously selected workspace - though I can imagine this will be (or has been) fixed in scratch-blocks?
3. Often other frequently required information, such as variable data, list data, or even other properties such as the size of the sprite can't be seen.

I've got a couple ideas to fix some of these, though I'm sure others will also have ideas..

For a lack of information (1, 3), IMO you should be able to customize what information you're displaying. Just drag in a variable reporter block, and a watcher will appear, for example:

http://i.imgur.com/Y1nBoaK.png
http://i.imgur.com/wGl2jlF.png

That would be super handy for monitoring the state of a sprite/clone.

As for 2 - it probably won't be near as big of a problem in Scratch 3.0 but a context menu item ("view this `sprite/clone`") on the target might be nice.

Specifically how this looks - it doesn't look bad in Scratch 2.0 (though I'm not a designer!).

<sup>Also I'm not 100% sure this is where I should be posting this, there isn't a "design" tagged issue about this..</sup>

tmickel:
@liam4 This issue is intended just for the VM API for providing and setting the data. But you make a good point that what data is get/set could be up for design debate...thanks for the write-up, those should be useful to keep in mind!

@carljbowman ^fyi

PR #320, opened 2016-10-26
Implement sprite info emitting and posting

tmickel:
Includes a demo in the playground:
screen shot 2016-10-26 at 12 25 37 pm

I've included only properties included in Scratch 2.0, but this could be extended later...
#161


rschamp (approved):
Looks great!

One thing that would be useful in addition to this is if the choices for rotation style were exposed on VM so it would be easy to render a selection.  That could be a separate PR though.

PR #322, opened 2016-10-28
Fixes #255

### LINKED ISSUE #255: Min/max scale doesn't match Scratch 2.0
Currently it's set to 5%/500% respectively.

However, it appears to have been set based on a function of costume width/height:

https://github.com/LLK/scratch-flash/blob/a931df6ab30e34a1cb2d6f7288f99df1c8f7f49b/src/scratch/ScratchSprite.as#L253

To pull that off, I think we'll need to provide costume/width height from the renderer.

CosmicWebServices:
I based it off of https://github.com/LLK/scratch-flash/blob/a931df6ab30e34a1cb2d6f7288f99df1c8f7f49b/src/scratch/ScratchSprite.as#L253 so that the min/max is like it was in scratch 2

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -250,7 +250,6 @@ RenderedTarget.prototype.setSize = function (size) {
         return;
     }
     // Keep size between 5% and 535%.
-
     this.size = MathUtil.clamp(size, Math.min(1, Math.max(5 / costume.width, 5 / costume.height)), Math.min((1.5 * 480) / costume.width, (1.5 * 360) / costume.height);
     if (this.renderer) {
         var renderedDirectionScale = this._getRenderedDirectionAndScale();
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -250,8 +250,8 @@ RenderedTarget.prototype.setSize = function (size) {
         return;
     }
     // Keep size between 5% and 535%.
-    //@todo get minimum size
-    this.size = MathUtil.clamp(size, 5, Math.min((stage.height / costume.height) * 100, (stage.width / costume.width) * 100));
+
+    this.size = MathUtil.clamp(size, Math.min(1, Math.max(5 / costume.width, 5 / costume.height)), Math.min((1.5 * 480) / costume.width, (1.5 * 360) / costume.height);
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks! Looks like you likely have a syntax error somewhere which is causing the build to fail. Try doing a clean `rm -rf node_modules && npm install` followed by running the test suite via `npm test` prior to the next commit.

CosmicWebServices:
OK I am sadly on my chromebook atm but I will do it later...

CosmicWebServices:
Ah hah! I forgot an ending parenthesis!

CosmicWebServices:
nope... still not the problem :(

CosmicWebServices:
OK, I know the issue now... I will do it later today...

src/sprites/rendered-target.js
```diff
@@ -250,7 +250,9 @@ RenderedTarget.prototype.setSize = function (size) {
         return;
     }
     // Keep size between 5% and 535%.
-    this.size = MathUtil.clamp(size, 5, 535);
+    var sizeMin = Math.min(1, Math.max(5 / costume.width, 5 / costume.height));
+    var sizeMax = Math.min((1.5 * 480) / costume.width, (1.5 * 360) / costume.height);
```
dekrain: Didn't you mean `this.costume` ?

CosmicWebServices: Yes

CosmicWebServices: Fixed

dekrain:
Now good :smile:

CosmicWebServices:
Yay! Thanks so much @dekrain!
@thisandagain I am ready to be checked again.

dekrain:
@TheBrokenRail ?? Where?

CosmicWebServices:
@dekrain Whoops sorry! I put the wrong mention!

tmickel:
@CosmicWebServices I'm skeptical that this could work, since `this.costume` seems undefined.

tmickel:
Yes - verified that this breaks the resize functionality. "Uncaught TypeError: Cannot read property 'width' of undefined"

CosmicWebServices:
Yeah I will fix this... later

PR #323, opened 2016-10-28
daysSince2000: Do the math before hand.

CosmicWebServices:


src/blocks/scratch3_sensing.js
```diff
@@ -116,7 +116,7 @@ Scratch3SensingBlocks.prototype.getKeyPressed = function (args, util) {
 };
 
 Scratch3SensingBlocks.prototype.daysSince2000 = function () {
-    var msPerDay = 24 * 60 * 60 * 1000;
+    var msPerDay = 86400000; //The amount of milliseconds in a day.
```
dekrain: From Travis: Use `// comment` instead of `//comment`

jeandrek:
Can't optimizy JS compilers do this for you?

tmickel:
Hi @CosmicWebServices:

I find the original version slightly preferable. In modern web browsers the multiplication should take nanoseconds, and will probably be calculated only one time per execution. So I lean toward the more readable expanded version.

PR #332, opened 2016-11-13
Implement Custom Reporters

TheBrokenRail:
### Proposed changes

_Adds Code for custom reporters_

### Notes

_I did not add another definition block because they do the same thing, if there just isn't a return value it just returns 'undefined'_

### DIFF
--- a/src/blocks/scratch3_procedures.js
+++ b/src/blocks/scratch3_procedures.js
@@ -21,7 +21,7 @@ Scratch3ProcedureBlocks.prototype.getPrimitives = function () {
     };
 };
 
-Scratch3ProcedureBlocks.prototype.defNoReturn = function () {
+Scratch3ProcedureBlocks.prototype.defNoReturn = function (args, util) {
     // No-op: execute the blocks.
     util.stackFrame.done = true;
 };
--- a/src/blocks/scratch3_procedures.js
+++ b/src/blocks/scratch3_procedures.js
@@ -31,7 +31,7 @@ Scratch3ProcedureBlocks.prototype.callNoReturn = function (args, util) {
         var procedureCode = args.mutation.proccode;
         var paramNames = util.getProcedureParamNames(procedureCode);
         util.stackFrame.thread = this.thread;
-        this.thread = this.thread + 1;
+        this.thread++;
         for (var i = 0; i < paramNames.length; i++) {
             if (args.hasOwnProperty('input' + i)) {
                 util.pushParam(paramNames[i], args['input' + i]);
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_procedures.js
```diff
@@ -14,25 +15,49 @@ Scratch3ProcedureBlocks.prototype.getPrimitives = function () {
...
+        util.startProcedure(procedureCode, util.stackFrame.thread);
+    }
+};
+
+Scratch3ProcedureBlocks.prototype.callReturn = function (args, util) {
+    if (!util.stackFrame.executed) {
+        var procedureCode = args.mutation.proccode;
+        var paramNames = util.getProcedureParamNames(procedureCode);
+        util.stackFrame.thread = this.thread;
+        this.thread = this.thread + 1;
```
dekrain:
``` JavaScript
this.thread++;
```

src/blocks/scratch3_procedures.js
```diff
@@ -14,25 +15,49 @@ Scratch3ProcedureBlocks.prototype.getPrimitives = function () {
...
+    util.stackFrame.done = true;
 };
 
 Scratch3ProcedureBlocks.prototype.callNoReturn = function (args, util) {
     if (!util.stackFrame.executed) {
         var procedureCode = args.mutation.proccode;
         var paramNames = util.getProcedureParamNames(procedureCode);
+        util.stackFrame.thread = this.thread;
+        this.thread = this.thread + 1;
         for (var i = 0; i < paramNames.length; i++) {
```
dekrain, line 35: ...and here.

src/engine/execute.js
```diff
@@ -198,6 +227,9 @@ var execute = function (sequencer, thread) {
         getParam: function (paramName) {
             return thread.getParam(paramName);
         },
+        setProcedureReturn: function (value) {
+            thread.stackFrames[0].executionContext.return = value;
```
dekrain, line 231:
``` JavaScript
currentStackFrame.executionContext.return = value;
```

dekrain (changes requested):

dekrain:
`src/engine/execute.js#231`:

``` javascript
currentStackFrame.executionContext.return = value;
```

src/engine/execute.js
TheBrokenRail: Wouldn't that cause the value to be set to the report block instead of the top block (hat)

dekrain: So why you added

``` JavaScript
for (; k < this.runtime.threads[k].stackFrames.length; k++)
```

?

TheBrokenRail: That runs from the calling block and not the hat block so it has to "find" the stackFrame

dekrain: Shouldn't it be

``` JavaScript
var stack = this.runtime.threads[k].stackFrames[0];
```

?

CosmicWebServices:
@TheBrokenRail you should finish this

TheBrokenRail:
I can't test this until the scratch blocks PR is merged

towerofnix:
You can probably figure out how to get a local copies of scratch-blocks and scratch-vm to work together, but I don't know if there's any official documentation on that.

TheBrokenRail:
I will close this until a design for implementation is more finalized and testable.

PR #333, opened 2016-11-14
Keep min and max scale in relative bounds as in Scratch 2.0

### LINKED ISSUE #255: Min/max scale doesn't match Scratch 2.0

tmickel:
Use the formula from scratch-flash for calculating the min and max scale for a sprite. Resolves #255. Note that the min and max still don't always match 2.0, because `renderer.getSkinSize` doesn't always report the same width and height that 2.0 is calculating. That is possibly a rendering bug....

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -249,9 +249,18 @@ RenderedTarget.prototype.setSize = function (size) {
     if (this.isStage) {
         return;
     }
-    // Keep size between 5% and 535%.
-    this.size = MathUtil.clamp(size, 5, 535);
     if (this.renderer) {
+        // Clamp to scales relative to costume and stage size.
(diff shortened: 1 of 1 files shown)


NahidaY:
this is not fixed by https://github.com/TurboWarp/scratch-vm/pull/336, the maintainer from what i was told is mad at me for submitting a fix and deleted the corresponding issue, sorry for the note (translated using google)

i was just told about this by a friend after they apparently emailed the maintainer :<

PR #335, opened 2016-11-21
Fix dev port number

rschamp:
### Proposed changes

Update the dev port number from 8080 to 8073

### Reason for changes

scratch-www: 8333
scratch-api: 8491
scratch-gui: 8601

Clearly the port should be 8073 for scratch-vm.

Also because 8080 is one of the ports already used by scratchr2, so this will avoid conflicts if that is already running (which it frequently is for some of us).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ StartServerWindows.bat
 ```
 
 ## Playground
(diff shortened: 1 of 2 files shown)


tmickel (approved):

towerofnix:
Out of curiosity's sake why 8073 and not 8079 or 8081 or 9001 or anything else?

tmickel:
I believe these particular numbers require sideways reading...

PR #346, opened 2016-11-22
Lint

tmickel:
- Addresses all lint warnings from eslint-config-scratch 2.0.0
- Upgrade to eslint-config-scratch 3.0.0, and address those errors and warnings also.


src/engine/execute.js
```diff
@@ -207,12 +203,7 @@ var execute = function (sequencer, thread) {
             // Find the I/O device and execute the query/function call.
             if (runtime.ioDevices[device] && runtime.ioDevices[device][func]) {
                 var devObject = runtime.ioDevices[device];
-                // @todo Figure out why eslint complains about no-useless-call
-                // no-useless-call can't tell if the call is useless for dynamic
-                // expressions... or something. Not exactly sure why it
-                // complains here.
-                // eslint-disable-next-line no-useless-call
```
tmickel, line 214: In this case, the linter was onto something right. `call()` expects `thisArg, arg1, arg2, arg3 ...`, so the line could have been replaced by `devObject[func](args);`. But I think what I was going for was not `call` but `apply`, which is `thisArg, [arg]`. No complaints about apply.

src/blocks/scratch3_control.js
```diff
@@ -73,16 +73,16 @@ Scratch3ControlBlocks.prototype.forever = function (args, util) {
...
+    if (util.stackFrame.timer) {
         var duration = Math.max(0, 1000 * Cast.toNumber(args.DURATION));
         if (util.stackFrame.timer.timeElapsed() < duration) {
             util.yield();
         }
+    } else {
+        util.stackFrame.timer = new Timer();
+        util.stackFrame.timer.start();
+        util.yield();
+        this.runtime.requestRedraw();
```
rschamp, line 85: The rules that emit warnings were ones that I didn't want to disable altogether because I wanted to suggest the more readable style, but also didn't want to enforce completely in cases where it's more readable to break the rule.

E.g., most of the time you want positive conditionals but in this case, I sort of think having the negative conditional first is easier to read since then you have the timer initialization ahead of its usage.  Maybe that use case is enough to just remove the rule and leave it up to reviewers to suggest the other way?

src/blocks/scratch3_event.js
```diff
@@ -47,7 +47,6 @@ Scratch3EventBlocks.prototype.getHats = function () {
 Scratch3EventBlocks.prototype.hatGreaterThanPredicate = function (args, util) {
     var option = Cast.toString(args.WHENGREATERTHANMENU).toLowerCase();
     var value = Cast.toNumber(args.VALUE);
-    // @todo: Other cases :)
```
rschamp, line 50: Same with warnings on TODOs — I mostly made them warnings so that it was easy to see our todos just by linting, not necessarily to suggest that you should remove the comment.

src/engine/execute.js
```diff
@@ -77,12 +77,10 @@ var execute = function (sequencer, thread) {
...
                     sequencer.retireThread(thread);
                 }
-            } else {
+            } else if (!resolvedValue) {
                 // Not an edge-activated hat: retire the thread
                 // if predicate was false.
-                if (!resolvedValue) {
-                    sequencer.retireThread(thread);
-                }
+                sequencer.retireThread(thread);
```
rschamp, line 83: Again on this one... I think the if inside the else might be a little easier to follow. This might always be the case so maybe that rule should be removed?

src/engine/execute.js
rschamp: This was so confusing to me. Without knowing how things were constructed, wouldn't you expect `devObject[func](args)` to be equivalent to `devObject[func].call(devObject[func], args)` not `devObject[func].call(devObject, args)`? You don't know if `devObject[func]` was bound or not, so `this` by default is going to be the function not the object containing the function, no?

src/engine/sequencer.js
```diff
@@ -217,11 +216,9 @@ Sequencer.prototype.stepToProcedure = function (thread, procedureCode) {
         var doWarp = definitionBlock.mutation.warp;
         if (doWarp) {
             thread.peekStackFrame().warpMode = true;
-        } else {
+        } else if (isRecursive) {
             // In normal-mode threads, yield any time we have a recursive call.
-            if (isRecursive) {
-                thread.status = Thread.STATUS_YIELD;
-            }
+            thread.status = Thread.STATUS_YIELD;
```
rschamp, line 221: Another case where I think it's more clear to step into the else and then consider the if, rather than considering both at once. The `isRecursive` just seems like a different concern than `!doWarp`, so considering them together in the same line of logic is more confusing than helpful I think.

rschamp (changes requested):
I think I'd rather leave a few of these as warnings since it's clearer as-is. It will probably make sense to just remove those rules from eslint-config-scratch seeing them be useless here. We can rely on reviews as the linter for those cases.

src/engine/execute.js
tmickel: errm lol, this is also confusing me now...I think I should dig into this again.

src/blocks/scratch3_control.js
tmickel: maybe actually it should be something like `var isFirstTime = !...` or `var needsInitialization = !...`. That might be the most clear and pass the warning :P

src/blocks/scratch3_event.js
tmickel: I see, sorry. I moved them to Github issues which I thought was the intention.

tmickel:
going to give another shot at this sometime, to try to be less heavy-handed with the warning lints...

PR #348, opened 2016-11-23
Add PROJECT_RUN_START/STOP events

rschamp:
These events signal when any threads are running or when all threads have stopped running. This maps to whether the green flag or the stop button should be illuminated in the GUI.

I also cleaned up some inconsistencies in the other glow code since I was in there:
* `Runtime.SCRIPT_GLOW_ON/OFF` now maps to `'SCRIPT_GLOW_ON/OFF'` not `'STACK_GLOW_ON/OFF'`
* The signature of the emits from Runtime are now the same as when they're echoed by VirtualMachine. E.g., `VISUAL_REPORT` now emits an object from Runtime, just like VirtualMachine does.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -84,6 +84,12 @@ var Runtime = function () {
     this._scriptGlowsPreviousFrame = [];
 
     /**
+     * Number of threads running during the previous frame
+     * @type {number}
+     */
+    this._threadCount = 0;
+
+    /**
      * Currently known number of clones, used to enforce clone limit.
      * @type {number}
      */
@@ -178,6 +184,18 @@ Runtime.BLOCK_GLOW_ON = 'BLOCK_GLOW_ON';
  * @const {string}
  */
 Runtime.BLOCK_GLOW_OFF = 'BLOCK_GLOW_OFF';
+
+/**
+ * Event name for glowing the green flag
+ * @const {string}
+ */
+Runtime.PROJECT_RUN_START = 'PROJECT_RUN_START'
+
+/**
+ * Event name for unglowing the green flag
+ * @const {string}
+ */
+Runtime.PROJECT_RUN_STOP = 'PROJECT_RUN_STOP'
 
 /**
  * Event name for visual value report.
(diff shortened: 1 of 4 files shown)


src/engine/runtime.js
```diff
@@ -552,6 +570,7 @@ Runtime.prototype._step = function () {
     this.redrawRequested = false;
     var inactiveThreads = this.sequencer.stepThreads();
     this._updateGlows(inactiveThreads);
+    this._setThreadCount(this.threads.length);
```
tmickel: Note that there is a chance that a thread will both start and stop in the same frame. This usually happens when running a single block for example: the block executes and then gets scrubbed from `this.threads` immediately. For script glows, I addressed this by passing `inactiveThreads` in the line above to the update function; any threads that were just scrubbed from `this.threads` are still counted as "glowing" for this frame because they're reported in `inactiveThreads`.

In Scratch 2.0, it looks like the green flag/stop button glow does respond when running a single block. Which to me indicates that this line should maybe be:

`this._setThreadCount(this.threads.length + inactiveThreads.length);`

Then threads which were just scrubbed out of `this.threads` this frame will still be counted, until a full frame has passed.

P.S. maybe `inactiveThreads` here should be called `scrubbedThreads` or something.

tmickel:
LGTM other than the one comment about thread-counting. Thanks for cleaning up stuff!

src/engine/runtime.js
rschamp: How about `doneThreads`?  Looking at it again, would it work to just count `this.threads.length` ahead of `stepThreads` to get the count for the frame before scrubbing? (While testing it seems to work, but I don't know if there is some kind of dynamic generation of threads in stepThreads that could happen somehow)

tmickel: `doneThreads` seems like a good name!

It is possible that `stepThreads` will cause new threads to be "generated". For example, when you run a "broadcast" block, it could start one new thread for each when-broadcast hat.

tmickel (approved):

PR #351, opened 2016-11-29
Add Dynamic Exporting And Importing From JSON (Not SB3, But Part Of SB3 (Saving VM State To JSON))

TheBrokenRail:
### Proposed changes

_Adds Functionality for dynamically importing and exporting from JSON_

### Reason for changes

_This is better than a fixed implementation of exporting and importing because a fixed implementation has all the variables to be imported and exported hard coded in but a dynamic implementation searches the object except for a exclusion list (runtime, functions, etc) and dynamically imports them making adding variables to them much easier._

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -189,7 +189,7 @@ VirtualMachine.prototype.fromJSON = function (json) {
             if (x === 'sprite') {
                 continue;
             }
-            target[x] = obj.sprites[x];
+            target[x] = obj.sprites[i][x];
         }
     }
     this.editingTarget = this.runtime.targets[1];
--- a/src/index.js
+++ b/src/index.js
@@ -181,6 +181,9 @@ VirtualMachine.prototype.importFromJson = function (json) {
         var target = sprite.createClone();
         this.runtime.targets.push(target);
         for (x in obj.sprites[i]) {
+            if (x == "sprite") {
+                continue;
+            }
             target[x] = obj.sprites[y];
         }
     }
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,7 @@
 var EventEmitter = require('events');
 var util = require('util');
-var Blocks = require('../engine/blocks');
-var RenderedTarget = require('../sprites/rendered-target');
-var Sprite = require('../sprites/sprite');
+var Blocks = require('./engine/blocks');
+var Sprite = require('./sprites/sprite');
 var Runtime = require('./engine/runtime');
 var sb2import = require('./import/sb2import');
 
--- a/src/index.js
+++ b/src/index.js
@@ -163,10 +163,10 @@ VirtualMachine.prototype.exportToJson = function () {
     for (; i < this.runtime.targets; i++) {
         obj.sprites.push(this.runtime.targets[i].export());
     }
-    obj.meta.name = "WIP";
+    obj.meta.name = 'WIP';
     obj.meta.useragent = navigator.userAgent;
     return JSON.stringify(obj);
-}
+};
 
 VirtualMachine.prototype.importFromJson = function (json) {
     var obj = JSON.parse(json);
@@ -174,16 +174,18 @@ VirtualMachine.prototype.importFromJson = function (json) {
     for (; i < obj.sprites.length; i++) {
         var blocks = new Blocks();
         var sprite = new Sprite(blocks, this.runtime);
+        var y = null;
         for (y in obj.sprites[i].sprite) {
             sprite[y] = obj.sprites[i].sprite[y];
         }
         var target = sprite.createClone();
         this.runtime.targets.push(target);
+        var x = null;
         for (x in obj.sprites[i]) {
-            if (x == "sprite") {
+            if (x === 'sprite') {
                 continue;
             }
-            target[x] = obj.sprites[y];
+            target[x] = obj.sprites[x];
         }
     }
     this.editingTarget = this.runtime.targets[1];
@@ -191,7 +193,7 @@ VirtualMachine.prototype.importFromJson = function (json) {
     this.emitTargetsUpdate();
     this.emitWorkspaceUpdate();
     this.runtime.setEditingTarget(this.editingTarget);
-}
+};
 
 /**
  * Add a single sprite from the "Sprite2" (i.e., SB2 sprite) format.
--- a/src/index.js
+++ b/src/index.js
@@ -170,9 +170,16 @@ VirtualMachine.prototype.fromJSON = function (json) {
     var i = 0;
     for (; i < obj.sprites.length; i++) {
         var blocks = new Blocks();
+        var z = null;
+        for (z in obj.sprites[i].sprite.blocks) {
+            blocks[z] = obj.sprites[i].sprite.blocks[z];
+        }
         var sprite = new Sprite(blocks, this.runtime);
         var y = null;
         for (y in obj.sprites[i].sprite) {
+            if (y === 'blocks') {
+                continue;
+            }
             sprite[y] = obj.sprites[i].sprite[y];
         }
         var target = sprite.createClone();
--- a/src/index.js
+++ b/src/index.js
@@ -191,6 +191,7 @@ VirtualMachine.prototype.fromJSON = function (json) {
             }
             target[x] = obj.sprites[i][x];
         }
+        target.updateAllDrawablePropertie();
     }
     this.editingTarget = this.runtime.targets[1];
     // Update the VM user's knowledge of targets and blocks on the workspace.
--- a/src/index.js
+++ b/src/index.js
@@ -157,7 +157,6 @@ VirtualMachine.prototype.loadProject = function (json) {
 
 VirtualMachine.prototype.exportToJson = function () {
     var obj = new Object();
-    var i = 0;
     obj.sprites = this.runtime.targets;
     obj.meta.name = 'WIP';
     obj.meta.useragent = navigator.userAgent;
--- a/src/index.js
+++ b/src/index.js
@@ -158,6 +158,7 @@ VirtualMachine.prototype.loadProject = function (json) {
 VirtualMachine.prototype.exportToJson = function () {
     var obj = new Object();
     obj.sprites = this.runtime.targets;
+    obj.meta = new Object();
     obj.meta.name = 'WIP';
     obj.meta.useragent = navigator.userAgent;
     return JSON.stringify(obj, null, 2);
--- a/src/index.js
+++ b/src/index.js
@@ -156,19 +156,16 @@ VirtualMachine.prototype.loadProject = function (json) {
 };
 
 VirtualMachine.prototype.exportToJson = function () {
-    this.clear();
     var obj = new Object();
     var i = 0;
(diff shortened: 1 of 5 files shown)


rschamp:
@TheBrokenRail This could be useful until we have a spec for SB3.

I think this code could be cleaned up if you used [the `replacer` argument of `JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter) along with the [`reviver` argument to `JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#Using_the_reviver_parameter).

[If an object has a `toJSON` method, it will get called automatically when you run `JSON.stringify` on the object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#toJSON()_behavior), so that would be a better name for your `export` functions.

TheBrokenRail:
Could this be a spec for sb3, it is very expandable?

rschamp:
@TheBrokenRail A spec is a definition of the format that can be referenced by other projects. Since JSON serialization outputs everything, it doesn't serve as a spec. However in the short term it could help serve the purpose of persisting state.

TheBrokenRail:
Why not just save the state instead of making a spec it would make it easier to add attributes to sprites? I changed rendered-target's export() function to toJSON()

rschamp:
@TheBrokenRail We need a specification of the SB3 file format because the VM isn't the only thing that reads projects.  Persisting the VM's internal state is a separate — but related — issue from the SB3 file format.

As you work on this, please add tests that show that `exportToJson` and `importFromJson` work as expected.

I also recommend renaming `VirtualMachine.exportToJson` to `toJSON` as well, so one can just call `JSON.stringify(vm)`.  And if you do that, `fromJSON` for the reverse function would make sense.

TheBrokenRail:
What kinds of tests (screenshot?)?

rschamp:
The sb2import function has tests you can use as an example: https://github.com/LLK/scratch-vm/blob/develop/test/unit/import_sb2.js. Basically from a known VM state, show that the JSON that is exported is what you would expect. And from known JSON, the VM instance that's constructed is the state you would expect.  Since the project is under heavy development, the de/serialization is likely to break, so these tests will warn us of that.

TheBrokenRail:
If you do run JSON.stringify(vm); you would actually get a string **containing** the the json so you would have to parse it twice.

TheBrokenRail:
By the way, if this is able to save the state of the vm, could this be incorporated into the larger spec for saving the vm state?

TheBrokenRail:
### Example JSON For Blank Project

```js
{
  "sprites": [
    {
      "variables": {},
      "lists": {},
      "isOriginal": true,
      "currentCostume": 0,
      "isStage": true,
      "x": 0,
      "y": 0,
      "direction": 90,
      "visible": true,
      "size": 100,
      "rotationStyle": "all around",
      "sprite": {
        "blocks": {
          "_blocks": {},
          "_scripts": []
        },
        "name": "Stage",
        "costumes": [
          {
            "skin": "https://cdn.assets.scratch.mit.edu/internalapi/asset/739b5e2a2435f6e1ec2993791b423146.png/get/",
            "name": "backdrop1",
            "bitmapResolution": 1,
            "rotationCenterX": 240,
            "rotationCenterY": 180
          }
        ]
      }
    },
    {
      "variables": {},
      "lists": {},
      "isOriginal": true,
      "x": 0,
      "y": 0,
      "direction": 90,
      "size": 100,
      "visible": true,
      "currentCostume": 0,
      "rotationStyle": "all around",
      "isStage": false,
      "sprite": {
        "blocks": {
          "_blocks": {},
          "_scripts": []
        },
        "name": "Sprite1",
        "costumes": [
          {
            "skin": "https://cdn.assets.scratch.mit.edu/internalapi/asset/09dc888b0b7df19f70d81588ae73420e.svg/get/",
            "name": "costume1",
            "bitmapResolution": 1,
            "rotationCenterX": 47,
            "rotationCenterY": 55
          },
          {
            "skin": "https://cdn.assets.scratch.mit.edu/internalapi/asset/3696356a03a8d938318876a593572843.svg/get/",
            "name": "costume2",
            "bitmapResolution": 1,
            "rotationCenterX": 47,
            "rotationCenterY": 55
          }
        ]
      }
    }
  ],
  "meta": {
    "name": "WIP",
    "useragent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.75 Safari/537.36"
  }
}
```

TheBrokenRail:
```js
window.vm.toJSON();
```
Returns normal object for serialization.
```js
window.vm.toPrettyJSON();
```
Returns serialized pretty json.

src/index.js
```diff
@@ -155,13 +155,17 @@ VirtualMachine.prototype.loadProject = function (json) {
...
+}
+
 VirtualMachine.prototype.toJSON = function () {
     var obj = new Object();
     obj.sprites = this.runtime.targets;
     obj.meta = new Object();
     obj.meta.name = 'WIP';
     obj.meta.useragent = navigator.userAgent;
-    return JSON.stringify(obj, null, 2);
+    return obj;
```
towerofnix, line 177: I wonder if this method could be renamed to something along the lines of "save state", since that's what it's doing (I think?), *not* converting to JSON (you'd use `toPrettyJSON` for that).

TheBrokenRail: The toJSON method is for JSON.stringify(vm); cause when JSON.stringify tries to serialize a object it checks first if it has a toJSON method and if it does it will stringy it's return value instead.

TheBrokenRail:
I don't think there are any more coments, ready to merge/needs review?

rschamp:
@TheBrokenRail Please add the tests I mentioned. Writing tests often requires you to change your code to be more testable, and exposes bugs you didn't catch. So I wouldn't want to review the code before the tests are there since the code might change.

TheBrokenRail:
It has a test function now that will import and export JSON in debug mode and run a JSON equivalency test. (JS Console required)
```js
window.vm.testJSON();
```

tmickel:
@TheBrokenRail 

I think @rschamp is referring to the automated tests like those here: https://github.com/LLK/scratch-vm/tree/develop/test/unit

You might find this fun reading :)
http://web.mit.edu/6.005/www/fa16/classes/03-testing/

xmeow:
thank you @TheBrokenRail , I made some test on your patch, the function used in JSON import should be
`this.runtime.setEditingTargets(this.editingTarget);` instead of `this.runtime.setEditingTarget(this.editingTarget);`

And It seems that the scripts of sprites didn't load properly, I got an empty workspace after load from json.

TheBrokenRail:
Why the sb2 import function uses this.runtime.setEditingTarget just fine.

TheBrokenRail:
I added a test that automates the testJSON() function without debug messages. (test/unit/save_state.js)

src/index.js
```diff
@@ -226,26 +226,37 @@ VirtualMachine.prototype.fromJSON = function (json, testing) {
...
-    console.log('Exporting to JSON...');
-    var json = this.toPrettyJSON(true);
-    console.log('Importing from JSON...');
-    this.fromJSON(json, true);
-    console.log('Exporting to JSON...');
-    var json2 = this.toPrettyJSON(true);
+VirtualMachine.prototype.testJSON = function (debug) {
+    if (debug == true) {
+        console.log('Exporting to JSON...');
+    }
```
towerofnix, line 232: A shorter way to do this would be something like this:

```js
if (debug) console.log(...)
```

Or you could make a "debug" function:

```js
var deb = function() {
  if (debug == true) console.log.apply(console, arguments)
}

deb(...)
```

src/index.js
```diff
@@ -154,6 +155,110 @@ VirtualMachine.prototype.loadProject = function (json) {
     this.runtime.setEditingTarget(this.editingTarget);
 };
 
+VirtualMachine.prototype.toPrettyJSON = function (testing) {
```
thisandagain, line 158: This should be handled by the client and not the VM. `toPrettyJSON` sounds like a fine thing for the playground or GUI to do.

src/index.js
```diff
@@ -154,6 +155,110 @@ VirtualMachine.prototype.loadProject = function (json) {
...
+    // Update the VM user's knowledge of targets and blocks on the workspace.
+    this.emitTargetsUpdate();
+    this.emitWorkspaceUpdate();
+    this.runtime.setEditingTarget(this.editingTarget);
+    if (testing == true) {
+        console.log('Imported from JSON');
+    }
+};
+
+VirtualMachine.prototype.testJSON = function (debug) {
```
thisandagain, line 229: When @rschamp was requesting tests he was referring to automated unit tests. This should be removed from the VM interface.

src/sprites/rendered-target.js
```diff
@@ -64,6 +64,30 @@ RenderedTarget.prototype.initDrawable = function () {
     }
 };
 
+RenderedTarget.prototype.toJSON = function () {
```
thisandagain, line 67: `RenderedTarget` already has a `toJSON` definition. This is conflicting.

src/sprites/sprite.js
```diff
@@ -54,4 +54,27 @@ Sprite.prototype.createClone = function () {
     return newClone;
 };
 
+Sprite.prototype.export = function () {
```
thisandagain, line 57: Why not use `toJSON` here?

src/sprites/sprite.js
```diff
@@ -54,4 +54,27 @@ Sprite.prototype.createClone = function () {
...
+    var result = new Object();
+    var notSaved = ['clones', 'runtime'];
+    var x = null;
+    for (x in this) {
+        if (typeof this[x] === 'function') {
+            continue;
+            if (this.runtime.targets.testing == true) {
+                console.log('Ignoring ' + x);
```
thisandagain, line 65: The VM should never write to the console directly. See: http://www.linfo.org/rule_of_silence.html

test/unit/save_state.js
```diff
@@ -0,0 +1,25 @@
...
+    t.doesNotThrow(function () {
+        vm.start();
+        vm.clear();
+        vm.setCompatibilityMode(false);
+        vm.setTurboMode(false);
+        vm.loadProject(file);
+        var value = vm.testJSON();
+    });
+
+    t.ok(value);
```
thisandagain, line 22: This is sufficient for an integration test, but for unit testing we expect more coverage. I would recommend running `npm run coverage` and modifying your tests until you have reached 100% coverage for your additions.

thisandagain (changes requested):
Thanks @TheBrokenRail but this still needs quite a bit of work. In addition to my comments in-line, this will export blocks in a very inefficient way. For example, serializing the `demo.json` project results in a JSON string which is 29827 bytes vs. 5695 for the current format. Based on this, I recommend we close this for now until we have some time to discuss #194 at which time this work will be really helpful to bring back into the fold.

PR #357, opened 2016-12-21
Fix "stable" specifier for Travis

rschamp:
This shorthand has become unrecognized, so match the docs suggestion of "node" for the most recent stable release.


rschamp:
For posterity, `"stable"` should also work. https://travis-ci.org/LLK/scratch-vm/jobs/182607139 failed because of a failed download, not this specifier.

PR #358, opened 2016-12-22
give external blocks a chance to load into vm without modify the defaultBlockPackages

xmeow:
### Proposed changes

give external blocks a chance to load into vm without modify the defaultBlockPackages setup.

### Reason for changes

User don't have to add primitive source code to scratch-vm\src\blocks when apply a custom block.
For example: 
```javascript
var runtime = this.props.vm.runtime;

var pluginPackage = {
       "pluginName":pluginmodule
};
runtime._registerBlockPackages(pluginPackage);
```

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -231,10 +231,11 @@ Runtime.MAX_CLONES = 300;
 /**
  * Register default block packages with this runtime.
  * @todo Prefix opcodes with package name.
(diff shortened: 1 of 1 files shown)


thisandagain:
While we are excited to start work on the extension interface, we need to have a design discussion before landing support for hacked blocks / custom blocks / extensions.

PR #361, opened 2016-12-23
Fix hat blocks in horizontal grammar

### LINKED ISSUE #360: Horizontal - Hat blocks with fields do not work
Hat blocks without fields (such as the "green flag") work well in the horizontal grammar. However, hat blocks with fields (such as the WeDo "tilt" block) do not work as the VM is unable to read the value.

Tracking fix in https://github.com/LLK/scratch-vm/commit/3eff0d38083e6ec9e2baa41eb20dfc801b8067b7#diff-38d875e1d50b3158dc96404b65275a54R411

thisandagain:
Resolves GH-360

### Proposed changes

If hat fields are not found when attempting to `startHats`, check inputs before continuing. This resolves the issue where hat blocks with options (in shadow blocks) do not work in the horizontal grammar.

### Reason for changes

Hat blocks with options (such as the WeDo "onTilt" block) should work in the horizontal grammar.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -435,12 +435,24 @@ Runtime.prototype.startHats = function (requestedHatOpcode,
             // Not the right hat.
             return;
         }
+
         // Match any requested fields.
         // For example: ensures that broadcasts match.
         // This needs to happen before the block is evaluated
         // (i.e., before the predicate can be run) because "broadcast and wait"
         // needs to have a precise collection of started threads.
         var hatFields = target.blocks.getFields(topBlockId);
+
+        // If no fields are present, check inputs (horizontal blocks)
+        if (Object.keys(hatFields).length === 0) {
+            var hatInputs = target.blocks.getInputs(topBlockId);
+            for (var input in hatInputs) {
+                var id = hatInputs[input].block;
+                var fields = target.blocks.getFields(id);
+                hatFields = Object.assign(fields, hatFields);
+            }
+        }
+
         if (optMatchFields) {
             for (var matchField in optMatchFields) {
                 if (hatFields[matchField].value !==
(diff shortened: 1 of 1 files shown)


src/engine/runtime.js
```diff
@@ -435,12 +435,24 @@ Runtime.prototype.startHats = function (requestedHatOpcode,
...
         // For example: ensures that broadcasts match.
         // This needs to happen before the block is evaluated
         // (i.e., before the predicate can be run) because "broadcast and wait"
         // needs to have a precise collection of started threads.
         var hatFields = target.blocks.getFields(topBlockId);
+
+        // If no fields are present, check inputs (horizontal blocks)
+        if (Object.keys(hatFields).length === 0) {
+            var hatInputs = target.blocks.getInputs(topBlockId);
+            for (var input in hatInputs) {
```
cwillisf, line 449: I would suggest putting something like this at the top of the loop body:
```js
if (!hatInputs.hasOwnProperty(input)) continue;
```

We're actually missing this sort of check in several places. If you'd like to see the VM explode, try typing something like this into your JS console while the VM is running:
```js
Object.prototype.harmless = false;
```

I'm of the opinion that we should try to fix this, but I'd be happy to discuss if you disagree. Also, even if you agree that we should fix this sort of thing, since it's an issue in several places I don't feel the need to hold this bug-fix up. We can do a `hasOwnProperty` pass at some other time.

cwillisf (approved):
LGTM but see comment

src/engine/runtime.js
thisandagain: Agreed that we should do a `hasOwnProperty` pass. Mind filing an issue?

thisandagain:
Tracking `hasOwnProperty` issues here: #366

PR #362, opened 2016-12-24
Add integration test coverage

thisandagain:
### Proposed changes
- Adds integration tests for the following block categories:
    - motion
    - looks
    - data
    - events
    - control
- Adds a new `extract` fixture which allows SB2 (zip) projects to be imported directly into tests
- Fixes two issues in `engine/runtime` with the scoping of `setInterval` and `clearInterval`
- Increases coverage from ~47% to ~71%

### Testing approach
To make it easy to run and update tests, each Integration tests is simply a Scratch 2.0 project. For example, here is the "event" test:
image

The test itself (e.g. `test/integration/event.js`), simply unzips the SB2 file, extracts the `project.json` and then loads the project into a VM instance. Once loaded, the test triggers the "green flag" hat and then allows the project to execute for two seconds. At the end of two seconds, the VM dumps the current state (`getPlaygroundData`) and exits. Currently the tests are not inspecting the VM state. It provides increased test coverage by executing a large number of block opcodes and edge states, as well as a place for us to start inspecting the state of the VM in the future.

### Before
before

### After
after


test/integration/control.js
```diff
@@ -0,0 +1,35 @@
...
+
+test('control project', function (t) {
+    var vm = new VirtualMachine();
+
+    // Evaluate playground data and exit
+    vm.on('playgroundData', function (e) {
+        var threads = JSON.parse(e.threads);
+        t.ok(threads.length > 0);
+        t.end();
+        process.nextTick(process.exit);
```
cwillisf, line 17: Is this `process.exit` here because the top-level test function will have exited already? Does `tap` run each test in a separate process?

Instead of calling `process.exit` like this I think you can return a `Promise` from your top-level function, then resolve the promise instead of calling `t.end()` once the test is complete. I think that's the "official" way to deal with async tests, anyway. Some of the tests in `scratch-storage` work this way, though they're more complicated than this one would be.

cwillisf (changes requested):
Looks good overall but I'm not sure about the `process.nextTick(process.exit)` pattern that's in a bunch of the tests.

test/integration/control.js
thisandagain: `tap` executes each test file in a separate process, but contrary to what you may expect `t.end()` does not exit the process (regardless of how it is called). Rather the Node.js process will stay open as long as something is keeping it running (e.g. a singleton, an interval timer, etc.). In the case of the VM, once you `vm.start()` an interval timer is created which will prevent the Node.js from exiting. One way to resolve this would be to build up some `vm.halt()` infrastructure, the other is to simply use `process.nextTick(process.exit)` which will force the process to exit regardless of the state of the program. This is fairly normal and something we use in other test suites for some of our other repos.

cwillisf: Ah, that makes sense. Thanks for the explanation!

cwillisf (approved):

PR #369, opened 2017-01-05
Improve integration test coverage

thisandagain:
### Changes
- Adds additional integration tests:
    - `complex` (used to test manipulation of sprites / costumes / backdrops)
    - `procedure`
    - `sensing`
- Improves existing integration tests
- Increases overall test coverage from 71.19% to 80.34%

### Before
image

### After
image


test/integration/complex.js
```diff
@@ -0,0 +1,88 @@
...
+        t.end();
+        process.nextTick(process.exit);
+    });
+
+    // Manipulate each target
+    vm.on('targetsUpdate', function (data) {
+        var targets = data.targetList;
+        for (var i in targets) {
+            if (targets[i].isStage === true) continue;
+            if (targets[i].name === 'test') continue;
```
cwillisf, line 29: Note for the future: at some point `vm.renameSprite` will probably disambiguate two sprites with the same name, at which point this might need to change to check if `targets[i].name` starts with `'test'` or something like that.

That said... it [looks like](https://github.com/LLK/scratch-vm/blob/dc9126e0676d8b6a565982f1c0616203c3e2a447/src/index.js#L308) `targetsUpdate` gives you a frozen JSON copy of the targets, so the rename below shouldn't cause this `continue` to be executed even if the targets move around in `runtime.targets`. Is that what this check was added for, or something else?

test/integration/sensing.js
```diff
@@ -0,0 +1,35 @@
+var path = require('path');
+var test = require('tap').test;
+var extract = require('../fixtures/extract');
+var VirtualMachine = require('../../src/index');
+
+var uri = path.resolve(__dirname, '../fixtures/sensing.sb2');
+var project = extract(uri);
+
+test('sensing', function (t) {
```
cwillisf, line 9: I'm getting the feeling that this duplicated code (below) should be in one shared place so that these basic project tests can `require` the common code. Ideally they could also add tests of their own to be called from the `playgroundData` event handler. It looks to me like the `threads.length` check and the project URI, and maybe the timeout duration, are the only test-specific parts of this right now.

I don't think it's critical to do this at the moment... if you don't object to such a refactor then I can file a low-priority issue for it.

cwillisf (approved):

test/integration/complex.js
```diff
@@ -0,0 +1,88 @@
...
+            rotationCenterX: 480,
+            rotationCenterY: 360
+        });
+    });
+
+    // After two seconds, get playground data and stop
+    setTimeout(function () {
+        vm.getPlaygroundData();
+        vm.stopAll();
+    }, 2000);
```
rschamp, line 87: Two seconds seems a bit long to me — is there a reason you chose that duration? If there were a way to trigger it after _n_ frames, would that work?

rschamp (approved):
This all looks reasonable to me, thanks for increasing our coverage!

test/integration/complex.js
thisandagain: We could probably shorten it to one second or so. In order to test some of the blocks which yield (glide, wait, etc.) for a specific time period you'll need to set the yield times to small decimals in order to perform full tests.

test/integration/sensing.js
thisandagain: Totally agreed. I think it's probably something to follow-up with as a new fixture specifically for integration testing.

PR #373, opened 2017-01-06
Integrate audio engine

ericrosenbaum:
Work in progress on adding the audio engine support to the VM. Provides an API to attach the audio engine and block implementations.


rschamp (approved):
dino_optimized

CosmicWebServices:
Nice gif

PR #375, opened 2017-01-11
Move build outputs into a `dist/` subdirectory

cwillisf:
### Proposed changes

Move build outputs into a `dist/` subdirectory, in addition to other minor build & settings improvements.

### Reason for changes

Currently the build outputs are `/dist.js` and `/vm.js`. It's easy to miss that these are the build outputs and the difference between the two is not obvious from their names. Moving the files into dedicated `dist/node/` and `dist/web/` subdirectories helps with both of these problems.

### Detailed list of changes
- Move Node output: `/dist.js` => `/dist/node/scratch-vm.js`
- Move web output: `/vm{.js,.min.js}` => `/dist/web/scratch-vm{.js,.min.js}`
- Update build output references in `package.json` and the playground's `index.html`
- Move the `VirtualMachine` class out of `index.js` into its own file, referenced by `index.js`. The `VirtualMachine` class is otherwise unchanged.
- Add `.gitattributes` rules for new file types which were added to this repository without specifying their text/binary attributes
- Turn on source maps in webpack and add corresponding `.gitignore` rule


thisandagain (changes requested):
This all looks reasonable to me with the exception of "move the VirtualMachine class out of index.js into its own file, referenced by index.js. The VirtualMachine class is otherwise unchanged". Could you clarify the reasoning for that?

cwillisf:
It comes down to personal preference; here are the reasons I prefer it this way:
- If I'm looking for `class VirtualMachine`, it's obvious which file to open. If I hit my editor's "quick open" hotkey and start typing a class name it'll open the right file. (Actually my preference would be for `VirtualMachine.js` instead of `virtual-machine.js` but I don't think it's worth renaming ALL the files right now...)
- Each file has one job: the job of `virtual-machine.js` is to define the `VirtualMachine` class and the job of `index.js` is to establish the overall module's exports.
- If we want to export more than one thing from the module we can do so without modifying the way `VirtualMachine` looks within the module. See [here](https://github.com/LLK/scratch-storage/blob/develop/src/index.js) for an example in another one of our repositories.

All that said, it's just a preference. I'm happy to change it back if you/we don't find those arguments convincing :)

webpack.config.js
```diff
@@ -9,6 +9,7 @@ var base = {
         host: '0.0.0.0',
         port: process.env.PORT || 8073
     },
+    devtool: 'source-map',
```
rschamp, line 12: Not sure about this change.  [Webpack's documentation](https://webpack.github.io/docs/configuration.html#devtool) on the various devtools is decidedly confusing — I can't tell what the default is to know what this is being changed from.  One thing is clear though, that `source-map` is the slowest option of all.  I haven't had too much of an issue with the current devtool, since at least in Chrome, I'm always able to figure out where the error is occurring.  As far as I can tell from googling and the docs, the recommended choice is `cheap-module-eval-source-map` for development.

rschamp (approved):
I think I agree with the move from `index.js` to `virtual-machine.js` for the arguments mentioned by @cwillisf.

One comment about the `devtool`, which I also am not super concerned about.  Working on scratch-www I found `eval` to cause the fewest problems — sometimes using `source-map`, you  have to do hard cache clearing to have the stack trace make sense, but then you have to re-do the behavior to cause the exception. Not usually a huge problem though.

webpack.config.js
cwillisf: > I haven't had too much of an issue with the current devtool, since at least in Chrome, I'm always able to figure out where the error is occurring.

Wow, really? For me it just doesn't do source mapping at all in Chrome without specifying `devtool`. That's not too bad since `playground.js` isn't minified, but still... I like source mapping. :)

Still, you have a point about `source-map` being the slowest option, and the fact that it requires external files complicates caching and ignoring. I'll experiment with the other options -- `cheap-module-eval-source-map` in particular -- and see how it goes.

cwillisf: I tried a few different options (certainly not the whole set) and here's what I saw:
- omit `devtool` entirely: no source maps, but `scratch-vm.js` is fairly readable.
- `source-map`: debugging and mapping to workspace files both work.
- `cheap-module-eval-source-map`: works well for debugging but appends `?abcd` (four hex digits) to most file names, which seems to prevent Chrome from mapping to workspace files.
- `eval-source-map`: effectively the same as `cheap-module-eval-source-map`.
- `eval`: gives good file names but that's about it. Functions in `scratch-vm.js` are no longer readable. I tried this because one of our other repositories uses it.

A few web searches seemed to suggest that updating webpack to 1.14 or newer would fix the `?abcd` thing. My vote is to stick with `source-map` for the moment and dig a little deeper (including a webpack upgrade) as a separate adventure.

rschamp: Works for me. And when I said "I'm always able to figure out where the error is occurring" I didn't mean because it gives the source mapping, just because if I click to the line in the inspector, it's clear which file I'm looking at from the context since we don't minify in development.

thisandagain (approved):

PR #384, opened 2017-01-19
Implement pen blocks

### LINKED ISSUE #6: Pen support
Thoughts:
- Provide a method to draw a line from (x1, y1) to (x2, y2) at color (RGB), width (px). VM handles color calculations and deciding when to draw lines.
- Stamp method for current drawable and clear method also provided.

I believe scratch-flash's pen draws the equivalent of canvas `lineCap`="round". In doing the HTML5 player, I also remember something about drawing lines at .5px off to prevent blur, but there might be a better way to deal with that.

cwillisf:
### Resolves

Along with LLK/scratch-render#72, this resolves LLK/scratch-render#6

### Proposed Changes

These blocks implement pen features as found in Scratch 2.0

Supporting changes include:
- `Target` is now an event emitter
- `RenderedTarget` now emits an event when it moves
- `Target` can now store additional arbitrary data, called "custom state" in the code, using a `Target`'s `setCustomState` and `getCustomState` methods. This is used to store per-target pen state without requiring `Target` or `RenderedTarget` to know anything about the pen.
- `Cast` can now cast to an RGB color object.
- `Color` now has functions to convert between RGB and HSV, constants for for black & white, and a `mixRgb` function to lerp between two colors.

### Reason for Changes

Implementing pen blocks brings us closer to Scratch 2.0 compatibility.

### Test Coverage

All newly-added `util` functions are covered:
- `toRbgColorObject` from `util/cast.js` is covered in `test/unit/util_cast.js`
- `hsvToRgb`, `rgbToHsv`, and `mixRgb` from `util/color.js` are covered in `test/unit/util_color.js`

All pen blocks are covered by the new `pen.js` integration test. Unfortunately our current integration test setup makes it very difficult (impossible?) to hook to the renderer, so the blocks which actually draw something will skip parts of their functionality during this test. The three affected pen blocks are `clear`, `pen down`, and `stamp`.


src/util/color.js
```diff
@@ -99,28 +99,54 @@ Color.hexToDecimal = function (hex) {
...
+        r = t;
+        g = p;
+        b = v;
+        break;
+    case 5:
+        r = v;
+        g = p;
+        b = q;
+        break;
+    }
```
towerofnix, line 149: Hmph. This is (arguably) less readable than before, part because everything's on separate lines, somewhat away from the `case`s, and part because it's a switch-case, which gets you all the cludge of `case` and `break`...

Nitpicks :)

cwillisf: Yeah... I'm open to ideas if you can think of a way to make it more readable while following the linting rules :)
I have some ideas but I'm not sure they'd work out. I probably won't mess with it right away.

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
+var Cast = require('../util/cast');
+var Color = require('../util/color');
+var MathUtil = require('../util/math-util');
+var RenderedTarget = require('../sprites/rendered-target');
+
+// Place the pen layer in front of the backdrop but behind everything else
+// We should probably handle this somewhere else... somewhere central that knows about pen, backdrop, video, etc.
+// Maybe it should be in the GUI?
+var penOrder = 1;
```
thisandagain: We have generally been handling things like this and the `stateKey` (below) like this:
https://github.com/LLK/scratch-vm/blob/develop/src/engine/sequencer.js#L23

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
...
+};
+
+var Scratch3PenBlocks = function (runtime) {
+    /**
+     * The runtime instantiating this block package.
+     * @type {Runtime}
+     */
+    this.runtime = runtime;
+
+    this._penSkinId = -1;
```
thisandagain: Missing jsdoc.

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
...
+};
+
+/**
+ * Clamp a pen size value to the range allowed by the pen.
+ * @param {number} requestedSize - the requested pen size.
+ * @returns {number} the clamped size.
+ * @private
+ */
+Scratch3PenBlocks.prototype._clampPenSize = function (requestedSize) {
+    return MathUtil.clamp(requestedSize, 1, 255);
```
thisandagain: Should we declare the max pen size at the top of the file?

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
...
+};
+
+Scratch3PenBlocks.prototype._getPenLayerID = function () {
```
thisandagain, line 94: Missing jsdoc.

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
...
+
+/**
+ * @param {Target} target - collect pen state for this target. Probably, but not necessarily, a RenderedTarget.
+ * @returns {PenState} the mutable pen state associated with that target. This will be created if necessary.
+ * @private
+ */
+Scratch3PenBlocks.prototype._getPenState = function (target) {
+    var penState = target.getCustomState(stateKey);
+    if (!penState) {
+        penState = JSON.parse(JSON.stringify(defaultPenState));
```
thisandagain: Maybe we should create a "clone" util to make this easier to read?

src/blocks/scratch3_pen.js
```diff
@@ -0,0 +1,231 @@
...
+        pen_changepenshadeby: this.changePenShadeBy,
+        pen_setpenshadeto: this.setPenShadeToNumber,
+        pen_changepensizeby: this.changePenSizeBy,
+        pen_setpensizeto: this.setPenSizeTo
+    };
+};
+
+Scratch3PenBlocks.prototype.clear = function () {
+    var penSkinId = this._getPenLayerID();
+    this.runtime.renderer.penClear(penSkinId);
```
thisandagain: I think we should always check to make sure the renderer exists. E.g.:
https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L56

src/engine/target.js
```diff
@@ -112,10 +129,20 @@ Target.prototype.lookupOrCreateList = function (name) {
  */
 Target.prototype.postSpriteInfo = function () {};
 
+Target.prototype.getCustomState = function (stateId) {
```
thisandagain, line 137: Missing jsdoc.

src/engine/target.js
```diff
@@ -112,10 +129,20 @@ Target.prototype.lookupOrCreateList = function (name) {
...
+    return this._customState[stateId];
+};
+
+Target.prototype.setCustomState = function (stateId, newValue) {
```
thisandagain, line 146: Missing jsdoc.

thisandagain (changes requested):
This looks great overall. Thanks for expanding the test coverage. Some fairly minor notes for the most part.

thisandagain (approved):

SillyInventor:
For pen clear to work you need to set the skin to dirty:
this.runtime.renderer._allSkins[penSkinId]._canvasDirty=true;

Issue #390, opened 2017-01-22
Clear Block Doesn't Work


SillyInventor:
### Expected Behavior

The clear screen block should remove all stamps and pen trails. Discussed in #384 and #386 

### Actual Behavior

Doesn't remove them.

### Steps to Reproduce

Stamp, move, clear.

### Operating System and Browser

Windows 7, chrome 55.0.2883.87 m

### Suggested solution
Add the following or similar to block code:
this.runtime.renderer._allSkins[penSkinId]._canvasDirty=true;

thisandagain:
@SillyInventor Thanks for reporting! Did you want to take this one?

TheBrokenRail:
has patch (#394)

PR #394, opened 2017-01-23
Clear Fix

SillyInventor:
Added a call to tell the skin canvas that it's dirty.

### Resolves

Fixes clear block, per issue #390

### Proposed Changes

Add a line that signals skin dirty.

### Test Coverage

None added.

### DIFF
--- a/src/blocks/scratch3_pen.js
+++ b/src/blocks/scratch3_pen.js
@@ -189,7 +189,7 @@ Scratch3PenBlocks.prototype.clear = function () {
     if (penSkinId >= 0) {
         this.runtime.renderer.penClear(penSkinId);
         this.runtime.requestRedraw();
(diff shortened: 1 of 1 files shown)


cwillisf (changes requested):
Hi @SillyInventor!

Thanks finding this issue and for submitting a fix for it! I think it would be better to handle this in the renderer, though, so that it matches the way the other pen-related functions work. Would you mind making this change in the [LLK/scratch-render](/LLK/scratch-render) repository, in `src/PenSkin.js`'s `clear` function? You can look at `drawLine` in the same file for reference.

Thanks again!

SillyInventor:
Moved change to scratch-render via @cwillisf 's suggestion in LLK/scratch-render#74

Issue #326, opened 2016-11-01
Importers: adjust block coordinate calculation so blocks overlap less.


tmickel:
Currently, block coordinates from old projects are multiplied by 1.1 for new scratch-blocks size and shape. However this seems to cause a lot of block overlapping. Pretty sure it might be difficult or impossible to put the blocks in the "right" relative position, so we need to play with the multipliers to make it feel good, at least.

2.0:
![pasted image at 2016_11_01 01_17 pm](https://cloud.githubusercontent.com/assets/120403/19905994/799625c0-a04f-11e6-8145-ed5fa8b7b84c.png)

3.0:
![pasted image at 2016_11_01 01_17 pm-1](https://cloud.githubusercontent.com/assets/120403/19906582/b7c93434-a051-11e6-8c76-e53ce3a20fae.png)

(thanks @rachel-fenichel)

https://github.com/LLK/scratch-vm/blob/develop/src/import/sb2import.js#L152

griffpatch:
I think fix for this has already been pulled?

PR #400, opened 2017-01-27
Importers: adjust block coordinate calculation so blocks overlap less. #326

griffpatch:
Better estimate for scratch 2 block import placement.


griffpatch:
Hmm... I only wanted to submit the pull request for the last fix I made as it was a help requested one... I think github desktop submitted them all?

Issue #398, opened 2017-01-26
Runtime does not exit warp mode if at end of loop


griffpatch:
### Expected Behavior

When a custom block (with run without screen refresh) is called as the last block within a loop (that is not running in warp mode) then you would expect the runtime to yield at the end of the loop and the screen to update.

### Actual Behavior

The runtime continues in warp mode around the outer loop and the screen is not updated in line with the project.

### Steps to Reproduce

http://llk.github.io/scratch-vm/#141944442

* Run this project and it hits around 2 fps while just drawing a single pen down.
* Add any new block (like a set var to 0) as the last statement in the loop after the call to the custom block draw and run it again to see 60fps.

I have created a fix here: e849c7576cd0cb7b2bf9ddda62653c6fef6a142f

Here it is running at 60fps on my forked branch: https://griffpatch.github.io/scratch-vm/#141944442

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

thisandagain:
Resolved in #406

PR #406, opened 2017-01-28
Runtime does not exit warp mode if at end of loop #398

griffpatch:
When popping down the stack frame it is assumed that you keep using the previous warp state rather than looking at the warp state of the level you just popped out to. This causes the loss of screen updates if the warping block was the last statement in a loop as the loop does not obey it's 'non warp' status.

### DIFF
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -131,12 +131,17 @@ Sequencer.prototype.stepThread = function (thread) {
         // If no next block has been found at this point, look on the stack.
         while (!thread.peekStack()) {
             thread.popStack();
+
             if (thread.stack.length === 0) {
                 // No more stack to run!
                 thread.status = Thread.STATUS_DONE;
                 return;
             }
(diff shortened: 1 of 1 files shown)


griffpatch:
See Runtime does not exit warp mode if at end of loop #398

src/engine/sequencer.js
```diff
@@ -131,12 +131,17 @@ Sequencer.prototype.stepThread = function (thread) {
...
+
             if (thread.stack.length === 0) {
                 // No more stack to run!
                 thread.status = Thread.STATUS_DONE;
                 return;
             }
-            if (thread.peekStackFrame().isLoop) {
+			
+			var stackFrame = thread.peekStackFrame();
+	        isWarpMode = stackFrame.warpMode;
```
towerofnix: GitHub and your text editor aren't working together.. the indentation here is a little messed up. Should be 12 spaces.

thisandagain (approved):

PR #409, opened 2017-01-28
Optimisation - Remove redunant lookups

griffpatch:
It's surprising how much time it takes to do an object member lookup, so reducing these is a great idea.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -54,8 +54,8 @@ Blocks.prototype.getScripts = function () {
   * @return {?string} ID of next block in the sequence
   */
 Blocks.prototype.getNextBlock = function (id) {
-    if (typeof this._blocks[id] === 'undefined') return null;
-    return this._blocks[id].next;
+    var block = this._blocks[id];
+    return (typeof block === 'undefined') ? null : block.next;
 };
 
 /**
@@ -85,8 +85,8 @@ Blocks.prototype.getBranch = function (id, branchNum) {
  * @return {?string} the opcode corresponding to that block
  */
 Blocks.prototype.getOpcode = function (id) {
-    if (typeof this._blocks[id] === 'undefined') return null;
-    return this._blocks[id].opcode;
+    var block = this._blocks[id];
+    return (typeof block === 'undefined') ? null : block.opcode;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


tmickel:
@griffpatch neat!

On this same piece of code:
https://github.com/LLK/scratch-vm/blob/develop/src/engine/execute.js#L48
The above has a series of function calls to get all information about the same blocks. Possibly might want to just reduce that to a single `getBlock`, depending on the cost of the function call.

Here is a more aggressive needed optimization, for a challenge:
Currently, all blocks are passed through `execute`. This takes the most liberal definition of "block", including all of the "shadow blocks" (a Blockly term for the numerical/string/color/etc. inputs - these are rendered as a block). Basically, when inputs are evaluated here:
https://github.com/LLK/scratch-vm/blob/develop/src/engine/execute.js#L137
Even inputs (like the 10 in "move 10 steps") are put on the stack and then evaluated using `execute`. I'd estimate, given that each block might have an average of 2 inputs/shadow blocks, we could reduce the number of times `execute` is called to 1/3 or so...

Another thing:
Creating this object (https://github.com/LLK/scratch-vm/blob/develop/src/engine/execute.js#L171) with all the function calls for `blockFunction` is really nice for having block implementations be neat and clean. I haven't done measurements, but I would guess though that creating that object and all the closures every time `execute` is run could cause a serious performance hit/loss of JIT capabilities/etc. With some refactoring to get rid of that object (and replace calls in the block implementations directly to calls in `runtime`, `runtime.sequencer`, etc.), and if we were willing to lose the neatness benefit, I imagine there is a speed-up to be had.

griffpatch:
Hmm.. Well I was trying to optimise without changing the existing structure
of things too much. I would be quite happy to go deeper with this if I am
released to, but I don't want to get in the way of development of step on
any one shoes.

tmickel:
totally :) just thought you might find those interesting!

griffpatch:
I do have more optimisations I have yet to commit, one of which includes
the single call to get the three bits of data, but also I removed the
closure and created an object class to pass on to avoid the extra overhead.
My local version does run between 2 to 3 times faster... But I haven't gone
deeper than that yet.

thisandagain (approved):

PR #410, opened 2017-01-28
Optimisation - Only check browser compatability once

griffpatch:
This saves doing the checks everytime the time functions are referenced


thisandagain (changes requested):
Hmm ... this is getting pretty squirrelly / hard to follow. I'm wondering if we might want to move these checks into the constructor which can provide the correct code path during instantiation. Thoughts?

griffpatch:
I've been looking at more optimisations, and I've discovered that the
self.performance.now() function is quite a bit slower than using just
Date.now(). The whole scratch-vm sped up by an extra 33% after removing the
performance references altogether.

Can I ask whether you actually need the extra precision as this could be
quite a saving altogether?

thisandagain:
@griffpatch Great catch. I'd be happy to remove the high precision timer and simply stick with `Date.now()` until such time we feel like we need the extra decimals. Thanks.

griffpatch:
I've changed this around a little more... If you'd still prefer the new var nowObj to be moved into the constructor then we can do that, it just adds a little more overhead when running in turbo mode, etc. But I haven't checked out the performance difference, I guess it might be negligible at the instantiation phase?

thisandagain (approved):

Issue #413, opened 2017-01-30
SB2 Import - Couldn't find SB2 block: Remark: %s


thisandagain:
### Steps to Reproduce

- Load project `https://llk.github.io/scratch-vm/#142330668`
- JS Console
```
minilog.js:17vm Couldn't find SB2 block:  Remark: %s

Uncaught TypeError: Cannot set property 'parent' of undefined
    at parseBlockList (sb2import.js:197)
    at parseScripts (sb2import.js:164)
    at parseScratchObject (sb2import.js:70)
    at parseScratchObject (sb2import.js:130)
    at sb2import (sb2import.js:145)
    at VirtualMachine.loadProject (virtual-machine.js:147)
    at XMLHttpRequest.r.onreadystatechange (playground.js:13)
```

### Operating System and Browser

MacOS 10.12.2
Chrome 55.0.2883.95 (64-bit)

PR #414, opened 2017-01-30
Bugfix - Continue import if opcode is not found during SB2 import

thisandagain:
### Resolves

GH-413

### Proposed Changes

Checks `parsedBlock` to ensure it is defined before pushing it into the block list. This prevents undefined or unimplemented opcodes (such as `remark: %s`) from throwing a fatal exceptions during import.

### DIFF
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -193,6 +193,7 @@ var parseBlockList = function (blockList) {
     for (var i = 0; i < blockList.length; i++) {
         var block = blockList[i];
         var parsedBlock = parseBlock(block);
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
I'm OK with this for now, but I don't think it should be our permanent solution. I'd like to support "undefined" blocks in a way that's similar to Scratch 2.0, where they indicate their broken state and do nothing when executed. The Scratch 2.0-style approach mitigates the impact of extension changes and IMO is more tinker-friendly.

thisandagain:
Yup! I agree. Just trying to un-break some existing projects for testing.

PR #422, opened 2017-02-01
Added new util function that sends tan function infinities correctly

### LINKED ISSUE #420: Tan Function Rounding Error
### Expected Behavior

Tan of 90 returns infinite, NaN, or similar. Tan of 180 should give 0.

### Actual Behavior

Because we're converting from degrees to radians, there is a rounding error. Such that you get to nearly 0, and large numbers, but not the exact number / result desired.

### Steps to Reproduce

Input values into mather operation, set to tan, click.

Ubuntu 16.04, chrome 56.0.2924.76 

I am sure you're already working on this, just adding an issue so I can share our first play test findings / complaints.

---
Moved from LLK/scratch-blocks#779
/cc @SillyInventor

SillyInventor:
### Resolves

Resolves #420 

### Proposed Changes

Added new math op. function that casts to infinity, and correctly gives 0 at multiples of 180
Changed math op. to call new math util tan function
Changed sin & cos to round correctly (to get 0)
Added testing for the new math util function
Added more testing for altered new math op. functions

### Reason for Changes

Returns 0s and +-inf when appropriate for trig functions

### Test Coverage

Added a math util test for new util function, and a few additional tests for math op.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -1,4 +1,5 @@
 var Cast = require('../util/cast.js');
+var MathUtil = require('../util/math-util.js');
 
 var Scratch3OperatorsBlocks = function (runtime) {
     /**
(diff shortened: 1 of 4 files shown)


thisandagain (approved):
Looks good! Thanks for expanding the test coverage.

PR #424, opened 2017-02-01
Refactor sound engine

ericrosenbaum:
### Resolves

This is work in progress and depends on a PR to audio engine: https://github.com/LLK/scratch-audio/pull/14. 

### Proposed Changes

- Move sound-related state information into the vm target using the custom state mechanism. This includes volume, current instrument, and effects settings.

- Improve checks, casting, wrapping and clamping on inputs to sound blocks

- Changes to reflect refactor of sound functionality, separating it into a single-instance AudioEngine and AudioPlayers for each sprite or clone.

### Reason for Changes

Loading sounds in the AudioEngine is a step toward moving to the Scratch storage system in the future.

Moving audio state information into vm targets using custom state is a step toward serializing the audio state for copying into clones and saving projects.

### Test Coverage

No tests yet.


src/sprites/rendered-target.js
```diff
@@ -400,20 +399,6 @@ RenderedTarget.prototype.getCostumeIndexByName = function (costumeName) {
...
- * @return {number} Index of the named sound, or -1 if not present.
- */
-RenderedTarget.prototype.getSoundIndexByName = function (soundName) {
-    for (var i = 0; i < this.sprite.sounds.length; i++) {
-        if (this.sprite.sounds[i].name === soundName) {
-            return i;
-        }
-    }
-    return -1;
-};
```
cwillisf, line 414: 👍

cwillisf (approved):
LGTM!
I would recommend adding doc strings to the sound blocks when you get the chance, but since they were already missing I don't think it's a big deal for this PR.

Issue #435, opened 2017-02-06
Empty or white space strings equal 0


griffpatch:
### Expected Behavior

("   " = 0) should resolve to false.

### Actual Behavior

It resolves to true.

### Steps to Reproduce

Add an equality comparison block to an empty project, enter a 0 on one side and leave the other black. Click it to get the answer 'true'. Try entering an few white spaces on the empty side and repeat... true again. This is because Javascript counts white space as a number. If it helps, here is one possible fix:

```
Cast.compare = function (v1, v2) {
    var n1 = Number(v1);
    var n2 = Number(v2);

    // Check for empty or white space strings if number is 0

    if (n1 === 0 && (v1 == null || typeof v1 === 'string' && v1.trim().length === 0)) {
        n1 = NaN;
    }
    if (n2 === 0 && (v2 == null || typeof v2 === 'string' && v2.trim().length === 0)) {
        n2 = NaN;
    }

    // Now continue as normal

    if (isNaN(n1) || isNaN(n2)) {
        // At least one argument can't be converted to a number.
        // Scratch compares strings as case insensitive.
        var s1 = String(v1).toLowerCase();
        var s2 = String(v2).toLowerCase();
        return s1.localeCompare(s2);
    } else {
        // Compare as numbers.
        return n1 - n2;
    }
};

```

I can create a patch, but my branch is pretty out of sync with your main branch due to my optimisation tests and it will take me a little time to get it in order.

### Operating System and Browser

Any

towerofnix:
I believe this is fixed?

Test project ID: [150822401](https://scratch.mit.edu/projects/150822401/)

Scratch 2.0:

![Scratch 2.0](https://cloud.githubusercontent.com/assets/9948030/24045065/902b289a-0afc-11e7-9014-75e589658fad.png)

Scratch 3.0, on [an old version](https://github.com/LLK/scratch-vm/tree/635de4113d652fa321c45eb04bf6056d6309da86):

![Old Scratch 3.0](https://cloud.githubusercontent.com/assets/9948030/24045038/81976f64-0afc-11e7-8184-2240e9e1a5d0.png)

Scratch 3.0, the [current version](https://llk.github.io/scratch-gui/#150822401):

![Present Scratch 3.0](https://cloud.githubusercontent.com/assets/9948030/24045095/a9b06988-0afc-11e7-8eee-24ab600a429f.png)

<sup>(edit: grammar)</sup>

thisandagain:
Resolved in #438

PR #438, opened 2017-02-08
Fix for the "Empty or white space strings equal 0" bug

griffpatch:
See: Empty or white space strings equal 0 #435

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -100,6 +100,12 @@ Cast.toRgbColorObject = function (value) {
 Cast.compare = function (v1, v2) {
     var n1 = Number(v1);
     var n2 = Number(v2);
+    if (n1 === 0 && (v1 == null || typeof v1 === 'string' && v1.trim().length === 0)) {
+        n1 = NaN;
+    }
+    if (n2 === 0 && (v2 == null || typeof v2 === 'string' && v2.trim().length === 0)) {
+        n2 = NaN;
+    }
     if (isNaN(n1) || isNaN(n2)) {
         // At least one argument can't be converted to a number.
         // Scratch compares strings as case insensitive.
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -91,6 +91,15 @@ Cast.toRgbColorObject = function (value) {
 };
 
 /**
+ * Determine if a Scratch argument is a white space string (or null / empty).
+ * @param {*} val value to check.
+ * @return {boolean} True if the argument is all white spaces or null / empty.
+ */
+Cast.isWhiteSpace = function (val) {
+    return val === null || typeof val === 'string' && val.trim().length === 0;
+};
+
+/**
  * Compare two values, using Scratch cast, case-insensitive string compare, etc.
  * In Scratch 2.0, this is captured by `interp.compare.`
  * @param {*} v1 First value to compare.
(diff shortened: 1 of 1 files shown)


src/util/cast.js
```diff
@@ -100,6 +100,12 @@ Cast.toRgbColorObject = function (value) {
 Cast.compare = function (v1, v2) {
     var n1 = Number(v1);
     var n2 = Number(v2);
+    if (n1 === 0 && (v1 == null || typeof v1 === 'string' && v1.trim().length === 0)) {
```
thisandagain: This string of comparators should probably be moved out into it's own `util/cast` method for the sake of readability, reusability, and testing. Perhaps we could call it simply `isEmpty` and then amend the unit test suite for `compare` and add test coverage for the new method?

thisandagain (changes requested):

griffpatch:
Problem is... function calls are mega expensive when is the core runtime
code... You really want to avoid pulling things out into functions if
possible in my experience. I was very careful with my choice of comparators
here to make it as quick as possible given the key location it is being
used here? I do agree it would be nice to pull out if you could assume
javascript would re-inline it again... but in my experience it hasn't been
doing that :(

You still want me to do it?

griffpatch:
I was wondering, I know you are currently building and running off the
non-minimized javascript for ease of debugging... Are there settings in the
build scripts to auto inline things like this (in the production build)
such that breaking it out into a function will not in the end impact the
speed of execution but instead be exactly the same? *hopeful expression*.
If this was the case it would make a big difference to how I went about
coding in these repositories?

thisandagain:
I do. At least in the case of Chrome (v8) and Firefox (Gecko) we can be fairly certain that the JS runtime will optimize and memoize the function call for something like this. For now I'd rather err towards readability, maintenance, and test coverage than performance optimization though I do certainly appreciate that we need to find a balance between the two.

thisandagain:
> build scripts to auto inline things like this (in the production build)
such that breaking it out into a function will not in the end impact the
speed of execution but instead be exactly the same

Good question. Not that I know of. Most webpack plugins (like UglifyJS) are more about compression than optimization. Most optimization of this type happens within each vendor's JS runtime implementation. That said, [Google's Closure Compiler](http://closure-compiler.appspot.com/home) is capable of inlining and tail call optimzation.

griffpatch:
Just I've noticed when using the profiler in chrome that the most simple
function calls that do pretty much nothing except return values still
execute faster if they are inlined...  I'll do some benchmarking on the
change and see what impact it has before I re-commit... Might have to wait
until tomorrow.

Thanks

thisandagain:
One of my favorite tools for testing things like this is [jsperf.com](https://jsperf.com/)

griffpatch:
Yes, I did use a few examples on there to try things out... but I think
because of the way webpack packages the classes things are not quite as
simple as they appear when taken on their own... However, it's a great way
to test simple things :)

thisandagain:
Yup. This is one of the reasons I've been a little hesitant to move the VM over to ES6 as adding Babel to the build process will add yet another layer of indirection.

griffpatch:
Is Babel just a convertor from the yet to be fully implemented ES6 back to
ES5?

thisandagain:
Yup, but it includes a series of plugins which can modify the source a fair amount and sometimes lead to indirection (particularly when dealing with performance optimization) IME.

griffpatch:
Is the scratch-render using this already? I noted it had the 'const' and
'let' keywords and class declaration... but the debugging of those
javascript files had reverted back to ES5...

thisandagain:
Yes. `scratch-render` uses ES6 / Babel:
https://github.com/LLK/scratch-render/blob/develop/webpack.config.js#L24

griffpatch:
Ok, I've amended this pull request as you requested. The good news is that it doesn't appear to slow things down in the slightest! In fact it's debatably slightly faster which (very very bad timing using project #144580026). But it gives a gist which is useful in these things. Problem is the variance can be huge. Anyhow, it's no big performance issue either way so I'm quite happy with that.

thisandagain (approved):

thisandagain:
Looks good. Thanks.

Issue #434, opened 2017-02-05
Stop this script should not kill thread 


griffpatch:
### Expected Behavior

The stop this script block should exit out of a custom block and resume executing the script that called it.

### Actual Behavior

This block is scripted to terminate the entire thread. I do have a fix for this in my local build (but it's based on my already modified code base)

### Operating System and Browser

Any

thisandagain:
Resolved in #439

PR #439, opened 2017-02-08
Implement "Stop this script" function

griffpatch:
Fixes: Stop this script should not kill thread #434

Existing implementation incorrectly terminates the entire thread.

See: http://llk.github.io/scratch-vm/#144142535


thisandagain (approved):

thisandagain:
Thanks for the straightforward example / test case.

Issue #433, opened 2017-02-05
Broadcasts are case sensitive


griffpatch:
### Expected Behavior

A broadcast in scratch 2 is case insensitive. Some projects (not sure how) broadcast and receive in different case. You would expect these to still trigger. 

### Actual Behavior

They do not trigger. 

### Steps to Reproduce

You need a project with a broadcast and receiver with the same name, but different case.

See: http://llk.github.io/scratch-vm/#144148381

### Operating System and Browser

Any

thisandagain:
Resolved in #440

PR #440, opened 2017-02-08
Bug caseinsensitivebroadcast

griffpatch:
Broadcasts are case sensitive #433

See: http://llk.github.io/scratch-vm/#144148381


src/engine/runtime.js
```diff
@@ -439,6 +439,11 @@ Runtime.prototype.startHats = function (requestedHatOpcode,
     }
     var instance = this;
     var newThreads = [];
+
+    for (var opts in optMatchFields) {
+        optMatchFields[opts] = optMatchFields[opts].toUpperCase();
+    }
```
rschamp, line 445: It feels a little icky to me to overwrite the arguments coming in. Would you please update to just compare `hatFields[matchField].value.toUpperCase() !== optMatchFields[opts].toUpperCase()`?

rschamp (changes requested):

src/engine/runtime.js
griffpatch: I know where you are coming from, but if you do this within the function below then it will repeat the uppercase conversion once for every hat block rather than just once for the triggered event. If you want we can instead create a new array here to hold the uppercase arguments but I was just trying to save allocating more memory. Would you prefer that?

rschamp: Doh — I thought I made sure it wasn't going to have a performance effect to do it differently. Can I blame no coffee? I think it's fine as-is in that case.

rschamp (approved):

Issue #404, opened 2017-01-27
Volume Doesn't Work For Notes / Instruments


SillyInventor:
### Expected Behavior

Volume control should reduce the volume of play note / instrument

### Actual Behavior

Volume only impacts sounds

### Steps to Reproduce

Set volume to 0, play _note __ for __ beats_, you can hear the note per normal.

### Operating System and Browser

Windows 7, 55.0.2883.87 m

### Suggested Solution
You must set the instrument gain node value to change the volume of preexisting instruments. For example, in the set volume command, you could call:
util.target.audioPlayer.instrumentPlayer.outputNode._gainNode.gain.value = value/100; //default value is 1

SillyInventor:
Resolved by LLK/scratch-audio#26

PR #444, opened 2017-02-09
Pass volume to audioengine playnote

ericrosenbaum:
### Resolves

Addresses https://github.com/LLK/scratch-vm/issues/404

Depends on PR to scratch-audio https://github.com/LLK/scratch-audio/pull/26

### Proposed Changes

Pass the volume from the vm in the call to the audio engine to play the note.

### Reason for Changes

This is a way to set the volume of the notes played by the play note block. The current InstrumentPlayer approach will be replaced, but this is a workaround for one of its limitations, which is that it cannot be easily connected to the effects chain for an AudioPlayer, which would control its volume.


cwillisf (approved):
🔊🔉🔈🔉🔊

Issue #366, opened 2017-01-03
Perform hasOwnProperty validation 


thisandagain:
From @cwillisf 

I would suggest putting something like this at the top of the loop body:
```js
if (!hatInputs.hasOwnProperty(input)) continue;
```

We're actually missing this sort of check in several places. If you'd like to see the VM explode, try typing something like this into your JS console while the VM is running:
```js
Object.prototype.harmless = false;
```

I'm of the opinion that we should try to fix this, but I'd be happy to discuss if you disagree. Also, even if you agree that we should fix this sort of thing, since it's an issue in several places I don't feel the need to hold this bug-fix up. We can do a `hasOwnProperty` pass at some other time.

thisandagain:
Resolved in #446

PR #446, opened 2017-02-10
Perform hasOwnProperty validation

ghost:
### Resolves
https://github.com/LLK/scratch-vm/issues/366

### Proposed Changes
add hasOwnProperty check to for loops in vm

### Reason for Changes
without it things like `Object.prototype.harmless = false;` break the app

That still breaks it because there is the same problem in the audio engine

### Test Coverage

I didn't add it but there is this: http://eslint.org/docs/2.0.0/rules/guard-for-in

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -192,6 +192,7 @@ Scratch3SoundBlocks.prototype.changeEffect = function (args, util) {
 Scratch3SoundBlocks.prototype.clearEffects = function (args, util) {
     var soundState = this._getSoundState(util.target);
     for (var effect in soundState.effects) {
+        if (!soundState.effects.hasOwnProperty(effect)) continue;
         soundState.effects[effect] = 0;
     }
     if (util.target.audioPlayer === null) return;
--- a/src/engine/adapter.js
+++ b/src/engine/adapter.js
@@ -24,6 +24,7 @@ var domToBlocks = function (blocksDOM) {
     // Flatten blocks object into a list.
     var blocksList = [];
     for (var b in blocks) {
+        if (!blocks.hasOwnProperty(b)) continue;
         blocksList.push(blocks[b]);
     }
     return blocksList;
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -149,6 +149,7 @@ Blocks.prototype.getTopLevelScript = function (id) {
  */
 Blocks.prototype.getProcedureDefinition = function (name) {
     for (var id in this._blocks) {
+        if (!this._blocks.hasOwnProperty(id)) continue;
         var block = this._blocks[id];
         if ((block.opcode === 'procedures_defnoreturn' ||
             block.opcode === 'procedures_defreturn') &&
@@ -166,6 +167,7 @@ Blocks.prototype.getProcedureDefinition = function (name) {
  */
 Blocks.prototype.getProcedureParamNames = function (name) {
     for (var id in this._blocks) {
+        if (!this._blocks.hasOwnProperty(id)) continue;
         var block = this._blocks[id];
         if ((block.opcode === 'procedures_defnoreturn' ||
             block.opcode === 'procedures_defreturn') &&
@@ -415,6 +417,7 @@ Blocks.prototype.blockToXML = function (blockId) {
     }
     // Add any inputs on this block.
     for (var input in block.inputs) {
+        if (!block.inputs.hasOwnProperty(input)) continue;
         var blockInput = block.inputs[input];
         // Only encode a value tag if the value input is occupied.
         if (blockInput.block || blockInput.shadow) {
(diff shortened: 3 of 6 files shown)


thisandagain:
Thanks @magmaboat .

@rschamp @cwillisf any thoughts on adding the `for in` guard to our eslint config?

rschamp (approved):
I'm +1 on adding `guard-for-in` to our lint config.

rschamp:
I do worry about performance with this though — it looks expensive to add a function call to every loop, especially for some of these which are called on each tick or more.  Perhaps we should just keep an eye on it, and in the most expensive cases, cache and iterate with `Object.keys(objToIterate)` or something?

griffpatch:
I had been thinking just the same thing.

Is there a legitimate reason to worry about the object model being
compremised in this way? I have been actively removing them where possible
when there was no chance of it causing issue in the normal running of
scratch 3 on my performance branch.

On 10 Feb 2017 6:14 p.m., "Ray Schamp" <notifications@github.com> wrote:

I do worry about performance with this though — it looks expensive to add a
function call to every loop, especially for some of these which are called
on each tick or more. Perhaps we should just keep an eye on it, and in the
most expensive cases, cache and iterate with Object.keys(objToIterate) or
something?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/LLK/scratch-vm/pull/446#issuecomment-279021893>, or mute
the thread

tmickel:
Have been enjoying watching these discussions :)

I got ~90% slower with the guard: https://jsperf.com/for-in-hasownproperty-or-not - although I do think it's probably an extreme case, not sure if any loop in the execution path is executing enough times per tick to make it significant.

cwillisf:
My favorite example to explain what `hasOwnProperty` does is to type this into the JS console, and then try to run some JS that isn't protected by `hasOwnProperty` checks:
```js
Object.prototype.harmless = false
```

For example:
```js
var stuff = {a: 1, b: 2}
for (var propertyName in stuff) { console.log(propertyName, '=', stuff[propertyName]) }
```

The output of the above:
```
a = 1
b = 2
harmless = false
```

Even with that demo, though, there are two valid questions being discussed here:

## Is this really something to worry about?

Extending `Object.prototype` is hopefully exceedingly rare in professional code, since it has the potential to disrupt `for-in` loops like this. However, if answers on Stack Overflow are any guide, there seems to be a strong segment of the JavaScript programming audience that finds it convenient to extend native prototypes.

What I really worry about is libraries which extend, say, `Array` with convenience methods. These libraries might be used in extensions -- Scratch extensions or even browser extensions -- and disrupt functionality in fairly unpredictable ways. An even more reasonable possibility: a Scratch extension might choose to extend the prototype of an object it "owns" and passes into Scratch routines. It might not be clear to an extension author why our code doesn't handle that object as expected. Even if they understand what's going on, it seems kind of rude to tell them that their favorite style of programming JavaScript isn't valid when communicating with Scratch. There's also potential here for support headaches, with people reporting bugs that we can't reproduce in house.

More info about this can be found here: http://bonsaiden.github.io/JavaScript-Garden/#object.forinloop

## Won't this hurt performance?

I agree with @rschamp when he says:
> Perhaps we should just keep an eye on it, and in the most expensive cases [do something else]

Also, in my opinion, @tmickel's [benchmark above](https://github.com/LLK/scratch-vm/pull/446#issuecomment-279037496) actually indicates that this isn't very expensive.

That benchmark tests the cost of adding one line of code to a block of JavaScript:

```js
var count = 0;
for (var test in x) {
    if (!x.hasOwnProperty(test)) continue; // <---- how expensive is this line?
    count++;
}
```

The result: for Tim it runs about 90% slower, and on my system it's about 80% slower. Let's go with the 90% number since it's more dire.

When jsPerf says one piece of code is "90% slower" than another, it really means that `1 - (number of times the slower code runs per second) / (number of times the faster code runs per second)` is 0.90, approximately. Another way to put this is that the slower code takes about 10 times as long to run, compared to the faster code.

Let's algebra!

- Let `p` represent the amount of time spent to "prepare" the loop, including the `var count = 0` line, creation of the `test` variable, etc.
- Let `b` represent the amount of time spent per iteration of the "body" of the loop, not counting the `hasOwnProperty` check -- mostly just incrementing the `count` variable.
- Let `h` represent the time spent on this line of code: `if (!x.hasOwnProperty(test)) continue;`
- Note that the loop executes 10 times per test.

```
p + 10 * (h + b) = 10 * (p + 10 * b)
p + 10h + 10b = 10p + 100b
10h = 9p + 90b
h = 0.9p + 9b
```

So, roughly speaking, the line `if (!x.hasOwnProperty(test)) continue` is approximately as expensive as this code:
```js
var p;
for (var b = 0; b < 9; ++b) {}
```
While that's not the same as "free" it seems fairly cheap to me -- especially considering that most of these pieces of code don't run with high frequency.

cwillisf (approved):
👍
tested locally

PR #447, opened 2017-02-10
Switch From Return Null to Return Empty String in keyCodeToScratchKey

### LINKED ISSUE #443: toUperCase Error
### Expected Behavior

Scratch VM should run without error

### Actual Behavior

Get error:

image

### Steps to Reproduce

Press a key not recognized by scratch (like control) while scratch is focused.

Ubuntu 16.04, Chrome 56.0.2924.87

SillyInventor:
### Resolves

Resolves #443 

### Proposed Changes

Return empty string instead of null.

### Reason for Changes

You can preform checks and actions on empty strings you can't on null, which was previously throwing errors -- but will still lead to exit.

### DIFF
--- a/src/io/keyboard.js
+++ b/src/io/keyboard.js
@@ -56,7 +56,7 @@ Keyboard.prototype._keyCodeToScratchKey = function (keyCode) {
     case 39: return 'right arrow';
     case 40: return 'down arrow';
     }
-    return null;
+    return '';
 };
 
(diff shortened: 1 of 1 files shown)


rschamp (approved):
Thank you!  The empty string in this case makes sense. I think we should eventually make the `startHats` method smarter so it doesn't crash on `null`.

PR #451, opened 2017-02-10
Use Webpack to generate the whole playground

cwillisf:
### Proposed Changes

The previous configuration mixed Webpack output with static content in order to create the playground. This change moves that static content from `/playground/` into `/src/playground/` and adds a Webpack rule to copy it into the playground as part of the build.

As a side effect of this change `playground.js` is now covered by `npm run lint`. Included are a few small style changes necessary to make `playground.js` comply with our lint rules.

### Reason for Changes

On the surface this might seem unnecessary, but it provides at least two benefits:
- It's no longer possible to accidentally load stale build output   through `webpack-dev-server` in the event of a misconfiguration. This was very easy in the previous configuration, and might in fact be the only way that `webpack-dev-server` ever worked for this repository.
- It's simpler to ensure that various rules apply to the hand-authored content and not build outputs. This includes lint rules, `.gitignore`, IDE symbol search paths, etc.

### Test Coverage

This change has no effect on testing.


webpack.config.js
```diff
@@ -5,7 +5,7 @@ var webpack = require('webpack');
 
 var base = {
     devServer: {
-        contentBase: path.resolve(__dirname, 'playground'),
+        contentBase: false,
```
rschamp, line 8: Can this be removed?

rschamp (approved):
This is a good idea, thanks!

webpack.config.js
cwillisf: If I'm interpreting [the docs](https://webpack.js.org/configuration/dev-server/#devserver-contentbase) correctly the default is equivalent to `contentBase: __dirname`, which would mean that (for example) `http://localhost:8073/package.json` would be available.

Setting `contentBase` to `false` turns off static content serving altogether... I feel like that _should_ be the default, but it isn't.

rschamp: It looks like this change makes us need to navigate to /playground when running the dev server (since the content base is no longer `playground`). Was that intentional?

cwillisf: Try this:
- Check out a version of `scratch-vm` from before this change
- Remove `dist` entirely -- this means `scratch-vm.js` must be provided from in-memory build output, not from `contentBase`
- `npm run start`
- Try loading various different paths, such as `/`, `/playground`, `/scratch-vm.js`, and `/playground/scratch-vm.js`.

From my tests while working on this change, I basically became convinced that it has always been broken. I also couldn't find a way to make it provide the build output from `webpack-dev-server` as `/scratch-vm.js` instead of `/playground/scratch-vm.js` -- if you find a way to do that I would be all for it.

cwillisf: Correction: I also couldn't find a way to make it provide the build output from `webpack-dev-server` as `/scratch-vm.js` instead of `/playground/scratch-vm.js` _without_ adding the potential to pick up stale files from the `dist` directory.

PR #457, opened 2017-02-15
Match Scratch 2.0 order of execution

cwillisf:
### Resolves

Working toward LLK/scratch-render#62

### Proposed Changes

The `allStacksAndOwnersDo` function in Scratch 2.0 runtime iterates targets in reverse. The corresponding function in Scratch 3.0, `allScriptsDo`, iterates targets in forward order. This change makes Scratch 3.0 match the Scratch 2.0 execution order.

### Reason for Changes

Projects sometimes rely on that for correct initialization. If, for example, each sprite runs a "go back 999 layers" or "go to front" block as its first action, the order of execution will determine the ordering of the layers. This change causes such projects to run more correctly in Scratch 3.0.

### Test Coverage

This change adds an integration test to verify that "when green flag" hats in multiple sprites execute in the same order as when run in the Scratch 2.0 implementation.


thisandagain (approved):
This makes sense to me. Thanks for cleaning up the jsdocs within `rendered-target`. Glad to see the integration test harness being useful.

rschamp (approved):

griffpatch:
Has this change been pushed out to the develop branch?
I have the issue reproduced with this project running in dev right now:
https://llk.github.io/scratch-gui/develop/#216001494

thisandagain:
@griffpatch Yes. Could you file a new issue?

PR #459, opened 2017-02-15
Feature fencing again

griffpatch:
Fixes the reverted functionality for fencing.
This is dependent on the change in scratch-render.


cwillisf (approved):
I tested LLK/scratch-render#93 and LLK/scratch-vm#459 together locally and they worked well for me. It turns out that this was one of the missing pieces to get Cachalot elements to align with each other :)

rschamp (approved):
This looks good to me, thanks @griffpatch!

PR #460, opened 2017-02-15
Update playground paths in README.md

cwillisf:
### Proposed Changes

Update README.md to reflect the new location for playground source files and the new URL used to access the playground through the dev server.

Note that we may change the URL back, pending further discussion (see https://github.com/LLK/scratch-vm/pull/451#discussion_r100648211). For now, though, this change brings the README back in sync with the actual configuration.

### Reason for Changes

A recent change (#451) moved the playground source files and changed the URL but did not update README.md to reflect these changes.

Thanks, @picklesrus, for bringing this to my attention!

### Test Coverage

No changes to test coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ npm start
 ```
 
 ## Playground
-To run the Playground, make sure the dev server's running and go to [http://localhost:8073/](http://localhost:8073/) - you will be directed to the playground, which demonstrates various tools and internal state.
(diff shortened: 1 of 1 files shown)


picklesrus:
works for me :) Thanks @cwillisf!

rschamp (approved):
I agree there's not a way to avoid serving old builds from `/playground` while also serving the contents of `/playground` at `/`.  I think the way to do it would be to have a separate config for `npm start` and `npm run build`, but just going to `/playground/` is easy enough especially with the docs change.

I'll review the other webpack configs, they probably have this same issue since I misunderstood `contentBase` when I made them.

PR #461, opened 2017-02-16
Provide method for filtering toolbox XML

rschamp:
### Resolves

For LLK/scratch-gui#16

### Proposed Changes

Provides a method to transform Blockly toolbox XML to only contain blocks which the VM can handle. It does this by removing any blocks in the toolbox for which it does not have a registered opcode.

### Reason for Changes

We need this method so that we can strip a given toolbox to only the blocks that the VM is capable of running. This is a temporary solution until we have a way to generate the toolbox based on the editing target.

### Test Coverage

I would like to include tests, but I didn't want to require `jsdom` or a similar DOM-simulating library just for this one hacky method...  Any suggestions for a lighter DOM-mocking method would be appreciated!

I considered the possibility that the fact that I would need to include a library to test this may indicate that I'm putting this method in the wrong place. However, given the other parts of the VM that interact with Blockly, this seems like the right place for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "gh-pages": "^0.12.0",
     "highlightjs": "^9.8.0",
     "htmlparser2": "3.9.2",
+    "jsdom": "^9.11.0",
     "json": "^9.0.4",
     "lodash.defaultsdeep": "4.6.0",
     "minilog": "3.1.0",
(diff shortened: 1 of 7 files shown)


thisandagain:
This looks good in terms of functionality. For testing I actually wouldn't be opposed to adding `jsdom` to handle this kind of thing, but it does make our `devDependencies` significantly heavier.

cwillisf (approved):

PR #462, opened 2017-02-16
Optimise reducegetblockcalls

griffpatch:
This results in around a 15 to 20% speed up (not massive, but each of these types of optimisations added together does made things better).

Replaces some method parameters to pass in block objects rather than IDs when we already have the blocks.

Shuffle around some of the getBlock code to make it a little more efficient.


src/engine/execute.js
```diff
@@ -16,38 +16,42 @@ var isPromise = function (value) {
...
-    var fields = blockContainer.getFields(currentBlockId);
-    var inputs = blockContainer.getInputs(currentBlockId);
+
+    // if (eCount++ % 1000==0) {   // 603,000 times!!
+    //     console.log('Execute x' + eCount);
+    // }
+
+    //    var block = blockContainer.getBlock(currentBlockId);
+    var opcode = block.opcode; // blockContainer.getOpcode(currentBlockId);
+    var fields = block.fields; // blockContainer.getFields(currentBlockId);
```
tmickel: may be varying opinions here but if it were me I might consider removing the simple blocks getters (i.e., returning one property of one block) altogether. can't imagine these getters adding much, and it might make sense to specifically be constructing the vm code to be robust to direct access to block fields/properties like you did here (for performance as you've noticed :)) Having the getters and *also* directly accessing the properties during execution could be misleading...

tmickel: (I would wait for further review before acting on that, just interested in this PR!)

src/engine/execute.js
```diff
@@ -16,38 +16,42 @@ var isPromise = function (value) {
...
+    var currentBlockId = thread.peekStack();
+    var currentStackFrame = thread.peekStackFrame();
+
+    var blockContainer = target.blocks;
+    var block = blockContainer.getBlock(currentBlockId);
+    if (typeof block === 'undefined') {
         blockContainer = runtime.flyoutBlocks;
+        block = blockContainer.getBlock(currentBlockId);
+        // Stop if block or target no longer exists.
+        if (typeof block === 'undefined') {
```
tmickel, line 39: 👍 this structure is also much cleaner/simpler!

src/engine/execute.js
```diff
@@ -101,24 +105,20 @@ var execute = function (sequencer, thread) {
...
+        }
+        var keys = Object.keys(fields);
+        if (keys.length === 1 && Object.keys(inputs).length === 0) {
+            // One field and no inputs - treat as arg.
+            handleReport(fields[keys[0]].value);
         } else {
-            if (Object.keys(fields).length === 1 &&
-                Object.keys(inputs).length === 0) {
-                // One field and no inputs - treat as arg.
-                for (var fieldKey in fields) { // One iteration.
```
tmickel, line 112: this old version is slightly embarrassing :)

src/engine/blocks.js
```diff
@@ -91,21 +91,21 @@ Blocks.prototype.getOpcode = function (id) {
...
 
 /**
  * Get all non-branch inputs for a block.
- * @param {?string} id ID of block to query.
+ * @param {?object} block the block to query.
  * @return {!object} All non-branch inputs and their associated blocks.
  */
-Blocks.prototype.getInputs = function (id) {
-    var block = this._blocks[id];
+Blocks.prototype.getInputs = function (block) {
```
tmickel, line 105: I bet eventually it may also make sense to have this/the branch inputs separated on block-representation-construction-time, rather than determining it inline

griffpatch:
I have another branch for the caching of block inputs and procedures here:

https://github.com/griffpatch/scratch-vm/tree/optimise/cacheBlockInputs

But I didn't raise a pull request for it. Feel free to have a look, if you
want a PR then I'll raise one, but I don't want to go raising lots of PRs
for things that you guys wouldn't want right now.

src/engine/blocks.js
```diff
@@ -91,21 +91,21 @@ Blocks.prototype.getOpcode = function (id) {
...
  * Get all fields and their values for a block.
- * @param {?string} id ID of block to query.
- * @return {!object} All fields and their values.
+ * @param {?object} block The block to query.
+ * @return {?object} All fields and their values.
  */
-Blocks.prototype.getFields = function (id) {
-    var block = this._blocks[id];
+Blocks.prototype.getFields = function (block) {
+    // var block = this._blocks[id];
```
thisandagain: Remove commented code

src/engine/blocks.js
```diff
@@ -91,21 +91,21 @@ Blocks.prototype.getOpcode = function (id) {
...
 /**
  * Get all non-branch inputs for a block.
- * @param {?string} id ID of block to query.
+ * @param {?object} block the block to query.
  * @return {!object} All non-branch inputs and their associated blocks.
  */
-Blocks.prototype.getInputs = function (id) {
-    var block = this._blocks[id];
+Blocks.prototype.getInputs = function (block) {
+    // var block = this._blocks[id];
```
thisandagain: Remove commented code

src/engine/execute.js
```diff
@@ -16,38 +16,42 @@ var isPromise = function (value) {
  * @param {!Thread} thread Thread which to read and execute.
  */
 var execute = function (sequencer, thread) {
+
```
thisandagain: Extra line break

src/engine/blocks.js
```diff
@@ -120,11 +120,11 @@ Blocks.prototype.getInputs = function (id) {
...
  * Get mutation data for a block.
- * @param {?string} id ID of block to query.
- * @return {!object} Mutation for the block.
+ * @param {?object} block The block to query.
+ * @return {?object} Mutation for the block.
  */
-Blocks.prototype.getMutation = function (id) {
-    var block = this._blocks[id];
+Blocks.prototype.getMutation = function (block) {
+    // var block = this._blocks[id];
```
thisandagain: Remove commented code

src/engine/execute.js
```diff
@@ -16,38 +16,42 @@ var isPromise = function (value) {
...
+            // No block found: stop the thread; script no longer exists.
+            sequencer.retireThread(thread);
+            return;
+        }
     }
-    var opcode = blockContainer.getOpcode(currentBlockId);
-    var fields = blockContainer.getFields(currentBlockId);
-    var inputs = blockContainer.getInputs(currentBlockId);
+
+    // if (eCount++ % 1000==0) {   // 603,000 times!!
```
thisandagain: Remove commented code

src/engine/execute.js
```diff
@@ -16,38 +16,42 @@ var isPromise = function (value) {
...
+    //     console.log('Execute x' + eCount);
+    // }
+
+    //    var block = blockContainer.getBlock(currentBlockId);
```
thisandagain: Remove commented code

src/engine/execute.js
thisandagain: I agree. It seems we are loosing some performance by continually accessing individual block properties and maintaining both approaches seems unwise. I don't think this PR should expand to cover to that, but rather I'd like to discuss that during a larger conversation about refactoring the VM in a few weeks.

thisandagain (changes requested):
Couple nitpicks but this looks / works well for me. Assigning to @cwillisf for another set of eyes.

src/engine/blocks.js
```diff
@@ -75,37 +75,34 @@ Blocks.prototype.getBranch = function (id, branchNum) {
...
 
 /**
  * Get the opcode for a particular block
- * @param {?string} id ID of block to query
+ * @param {?object} block The block to query
  * @return {?string} the opcode corresponding to that block
  */
-Blocks.prototype.getOpcode = function (id) {
-    var block = this._blocks[id];
+Blocks.prototype.getOpcode = function (block) {
```
griffpatch, line 87: Since you requested to remove the direct access to the opcode from this PR, I have refactored the 'getOpcode' function to work in the same way as the other functions in the block class such that you now pass in the block to be retrieved instead of the block ID. I note though that this still introduces a redundant `=== 'undefined'` check that isn't needed when we know the object is not undefined at the point of call?

src/engine/execute.js
griffpatch: A little note here... There was a merge conflict at this point which I had to resolve... We've lost a 'hasOwnProperty' call in the resolution which I don't believe is an issue, but I just wanted to raise it.

griffpatch:
- Right, I've addressed all the comments I left in (sorry about that)
- I've reverted the direct member access on the block object, but to keep it consistent I've refactored the getOpcode call to accept a block object as the parameter instead of the block ID. See what you think.
- I've had to resolve one merge conflict - I added a github comment to the source so you will see it when you look through. I don't think it's an issue.

src/engine/blocks.js
thisandagain: That makes sense. Probably ok to leave for now. Might be a good idea to add a `@todo` note here though.

thisandagain:
Thanks @griffpatch !

This looks good to go for me. @cwillisf any thoughts?

src/engine/execute.js
cwillisf: +1, especially now these no longer even look up the block by ID. The only added "benefit" is to check if the block is actually there (`!== undefined`) but as @griffpatch points out that's redundant at least some of the time -- I suspect in the majority of cases. I do agree with @thisandagain about addressing the getters as a separate change / discussion, though.

src/engine/execute.js
```diff
@@ -19,35 +19,33 @@ var execute = function (sequencer, thread) {
...
-    var targetHasBlock = (
-        typeof target.blocks.getBlock(currentBlockId) !== 'undefined'
-    );
-    var flyoutHasBlock = (
-        typeof runtime.flyoutBlocks.getBlock(currentBlockId) !== 'undefined'
-    );
-
     // Stop if block or target no longer exists.
-    if (!target || (!targetHasBlock && !flyoutHasBlock)) {
+    if (target === null) {
```
cwillisf, line 23: I'm fine with this new code as-is but I find it interesting that the old code accessed `target` before this line, in the `targetHasBlock` test. I guess this branch isn't used very often in practice :)

src/engine/execute.js
```diff
@@ -148,19 +140,19 @@ var execute = function (sequencer, thread) {
...
+            thread.popStack();
         }
         argValues[inputName] = currentStackFrame.reported[inputName];
     }
 
     // Add any mutation to args (e.g., for procedures).
-    var mutation = blockContainer.getMutation(currentBlockId);
-    if (mutation) {
+    var mutation = blockContainer.getMutation(block);
+    if (mutation !== null) {
```
cwillisf, line 155: I'm curious about this change: why not leave it as `if (mutation)`? I don't see a way for the current implementation of `getMutation()` to return `undefined` but if it did, whether intentionally or due to a bug, the old code would ignore the return value whereas the new code will try to use it.

cwillisf (approved):
Sorry for the delay. This looks good! I do have one additional question, but I don't feel too strongly about it :)

src/engine/execute.js
griffpatch: You are right... I think there are quite a few 'unnecessary' checks in the source code, but I think the project errs on the side of being over cautions *grin*.

src/engine/execute.js
griffpatch: Yes, the old way IS better assuming you want to catch any future case where the `getMutation()` might return undefined... The reason for changing it was simple... In all my test it's quicker to compare to !== null than to allow javascript to do it's casting required to change `mutation` into a boolean. All these changes were about optimising things as best I could, but I am aware that you might prefer to err on the side of caution. The gains are all very slight, but they all add up. What we really need to a good refactor of the runtime engine to reduce the number of function calls as they are proving very expensive. This small tweak is peanuts in the grand scale of things, but it is called for 'every block' so I think is worth considering still? However, it's your call, would you prefer that line reverting?

thisandagain (approved):

src/engine/execute.js
cwillisf: I think this one's OK: I think it's more likely that `getMutation` will disappear as a function than change to something which can return `undefined`. In general, though, I'd be more comfortable avoiding micro-optimizations that trade safety for speed unless we can show (say, through profiling) that the impact is non-trivial.

That said...
> What we really need to a good refactor of the runtime engine to reduce the number of function calls as they are proving very expensive.

👍

griffpatch: I do profile everything - that's why I made the change. When I said it was peanuts in the grand scale of things I was comparing it with all the extra function calls, etc that occur continuously through the runtime. However, taken on there own, using `IF (a!==null)` is much faster than 'IF (a)'. At least under the conditions that I tested it ;).

Issue #464, opened 2017-02-17
Execution bug: can't escape from conditional branches that end with a Promise-resolution-terminating command block


tmickel:
### Expected Behavior

Some command blocks use Promises to indicate when they're finished. When the Promise resolves, execution should proceed to the next block.

I think the current set of blocks uses this in two places: playing sounds, and say/think (as a demo). But it's conceivable this could be used for extensions and other network-accessing blocks.

If one of these blocks is inside a conditional branch (the `if` or `if-else` C-block), at the end, execution should pop out of the branch (to whatever's connected to the bottom of the if/if-else).

### Actual Behavior

If you place a Promise-using command block at the end of a conditional branch (i.e., as the last block inside the C), the branch will loop forever!

I haven't looked into why this happens. It seems to be exclusive to conditionals (if and if-else) but could affect other things.

### Steps to Reproduce

2.0 demo: https://scratch.mit.edu/projects/145800579/ (the if conditional executes once instantaneously, and then the green checkmark should appear).
3.0: https://llk.github.io/scratch-vm/#145800579 (red X forever. Because "say" terminates with a Promise resolution. You can replace the "say" with any sound-playing block to achieve the same).

### Operating System and Browser

OS X 10.12.1 Beta, Chrome Version 56.0.2924.87 (64-bit)

rasmushaglund:
This is fixed, right?

rschamp:
Yes, thanks.

PR #472, opened 2017-02-20
Bug - escapeFromConditionalBranches

griffpatch:
Fixes Bug: Execution bug: can't escape from conditional branches that end with a Promise-resolution-terminating command block #464

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -241,7 +241,15 @@ var execute = function (sequencer, thread) {
                         return;
                     }
                     var nextBlockId = thread.target.blocks.getNextBlock(popped);
-                } while (nextBlockId === null);
+                    if (nextBlockId !== null) {
+                        // A next block exists so break out this loop
+                        break;
+                    }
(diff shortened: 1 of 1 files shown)


griffpatch:
Wait... hold off on that... this is not a good fix as I forgot to account for promises at the end of actual loops. This will break that scenario at present.

griffpatch:
Forgot to say, this fix is ok now and can be looked at (if it helps).
My only hesitation is that it would be nicer if the scripts to exit the current stack frame did not have to live in this bit of the codebase as this is usually done in the stepThread function rather than the execute function... but since the popThread and getNextBlockID was already being done in the later for the isPromise callback I added the other logic to look for end of execution and loop detection in there too to fix the bug.

thisandagain (approved):
Thanks @griffpatch . This looks good. Sorry for the slow review on this, but I just wanted to make sure to test it thoroughly.

griffpatch:
Very very wise :) I prefer it that way :)

PR #483, opened 2017-02-27
Show Categories that use custom code to load (variables, procedures)

TheBrokenRail:
### Resolves

_More of LLK/scratch-gui#16_

### Proposed Changes

_If a category uses the "custom" tag to generate blocks don't delete in the filtered toolbox_

### Reason for Changes

_So you can use thins like variables and custom blocks._

### DIFF
--- a/src/util/filter-toolbox.js
+++ b/src/util/filter-toolbox.js
@@ -39,6 +39,7 @@ var filterToolbox = function (toolbox, opcodes) {
             if (category.nodeName.toLowerCase() !== 'category') continue;
             var filteredCategory = filterToolboxNode(category, opcodes);
             if (filteredCategory.hasChildNodes()) filteredToolbox.appendChild(filteredCategory);
+            if (!filteredCategory.hasChildNodes() && filteredCategory.hasAttribute("custom")) filteredToolbox.appendChild(filteredCategory);
         }
     } else {
         filteredToolbox = filterToolboxNode(toolbox, opcodes);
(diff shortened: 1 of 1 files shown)


src/util/filter-toolbox.js
```diff
@@ -39,6 +39,7 @@ var filterToolbox = function (toolbox, opcodes) {
             if (category.nodeName.toLowerCase() !== 'category') continue;
             var filteredCategory = filterToolboxNode(category, opcodes);
             if (filteredCategory.hasChildNodes()) filteredToolbox.appendChild(filteredCategory);
+            if (!filteredCategory.hasChildNodes() && filteredCategory.hasAttribute("custom")) filteredToolbox.appendChild(filteredCategory);
```
rschamp: Hey @TheBrokenRail — the tests are failing because this line is too long and the string is quoted with double quotes. Would you please make this a multi-line if statement and change the quote marks to single quotes?

rschamp (changes requested):

src/util/filter-toolbox.js
rschamp: When you fix the lint errors, would you please combine the line with the one before? I.e.,
```javascript
if (filteredCategory.hasChildNodes() ||
    filteredCategory.hasAttribute('custom')) {
    ...
}
```

rschamp (changes requested):
Thanks @TheBrokenRail, nice catch. I have a suggestion to simplify this but looks like a good way to do it.

TheBrokenRail:
Fixed!

rschamp (approved):
Looks great, thanks @TheBrokenRail!

Issue #484, opened 2017-03-01
New Sprite Not Always Visible


thisandagain:
### Expected Behavior

After selecting new sprite, said sprite should appear.

### Actual Behavior

Works sometimes, but not always. Eg. apple doesn't load.

### Steps to Reproduce

Start blank project, click new sprite, click apple.

### Operating System and Browser

Ubuntu 16.04, chrome 56.0.2924.76 

---
Moved from LLK/scratch-gui#54
/cc @SillyInventor

PR #485, opened 2017-03-01
Default bitmapResolution to 1 if not provided in costume metadata

thisandagain:
### Resolves

Resolves GH-484

### Proposed Changes

Default `bitmapResolution` to `1` during `updateAllDrawableProperties` if no value is provided in the costume metadata.

### Reason for Changes

Many sprites in the sprite library (e.g. "Apple") do not specify a `bitmapResolution` in their costume metadata. This leads to a bug in the GUI when one of these sprites is added to the project and the sprite is not rendered to the screen.

Example: 
https://github.com/LLK/scratch-gui/blob/develop/src/lib/libraries/sprites.json#L569

### DIFF
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -45,7 +45,7 @@ var parseScratchObject = function (object, runtime, topLevel) {
                 skin: 'https://cdn.assets.scratch.mit.edu/internalapi/asset/' +
                     costume.baseLayerMD5 + '/get/',
                 name: costume.costumeName,
-                bitmapResolution: costume.bitmapResolution || 1,
+                bitmapResolution: costume.bitmapResolution,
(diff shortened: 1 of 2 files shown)


thisandagain:
@rschamp Tests are failing due to issues with `./test/unit/util_filter-toolbox.js` not this PR.

rschamp (changes requested):
I think it would be more robust if this default were set in `RenderedTarget`, or even in the renderer's handling of costumes. For instance, if a bitmap costume is added without a `bitmapResolution` (rather than a sprite containing a costume without `bitmapResolution`), this same problem will arise from a different place.

thisandagain:
@rschamp This is ready for re-review

rschamp (approved):
Looks great, thanks!

PR #487, opened 2017-03-01
Add draggable property to sprites

rschamp:
This property should be managed by the VM so that it can be displayed in the sprite info panel in the GUI.


thisandagain (approved):
This looks great. Thanks.

PR #491, opened 2017-03-03
Exclude more unfinished blocks from toolbox

rschamp:
Temporary for a play test, exclude procedure, variable, list, say and think blocks from the palette when filtering the toolbox.

Resolves https://github.com/LLK/scratch-gui/issues/157 and https://github.com/LLK/scratch-gui/issues/158

### DIFF
--- a/src/util/filter-toolbox.js
+++ b/src/util/filter-toolbox.js
@@ -24,9 +24,11 @@ var filterToolboxNode = function (node, opcodes) {
  * @param {HTMLElement} toolbox Blockly toolbox XML node
  * @param {Array.<string>} opcodes Valid opcodes. Blocks producing other opcodes
  * will be filtered.
+ * @param {boolean} includeCustom Include empty categories with a non-empty `custom` attribute
  * @returns {HTMLElement} filtered toolbox XML node
  */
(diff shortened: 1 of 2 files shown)


thisandagain (commented):
This looks ok. I keep wondering if it would have been easier / cleaner to just remove the opcodes for the say and think blocks (and thus simplify this change) or to just update the default toolbox in `scratch-blocks`. While this covers a lot of the issues, it doesn't cover more subtle cases like removing "loudness" and "video motion" from the "when [x] > [n]" block.

rschamp:
I thought about removing the opcodes outright but decided not to because some projects depend on the "say _ for _ secs" blocks for timing. So keeping the partial implementation present will keep those projects functioning the same in 3.0 as 2, timing-wise.  Happy to remove the block implementations until they are fully implemented if you are not worried about that.

I believe removing the unimplemented items in the blocks drop-downs needs to be resolved in a separate PR in scratch-blocks, I just haven't gotten around to that yet.

rschamp:
Just after I posted this I re-read your comment and realized what you were saying about doing it all in scratch-blocks. I think that's a good idea, so closing in preference to that.

Issue #139, opened 2016-09-02
Clear graphic effects on green flag and stop


tmickel:

tmickel:
(Actually, this should happen on green flag)

thisandagain:
To match Scratch 2.0, the stop button should also reset graphic effects.

PR #494, opened 2017-03-06
Stop all gfx reset

paulkaplan:
### Resolves

https://github.com/LLK/scratch-vm/issues/139

### Proposed Changes

Clears the graphics effects on stop all. This extends #199 which cleared graphics effects on green flag.

### Reason for Changes

@thisandagain said
> To match Scratch 2.0, the stop button should also reset graphic effects.

### Test Coverage

Added a unit test making sure that stopping resets an effect to 0. Let me know if you have any feedback about the way this test was implemented, I tried to just stick with what was already there. Better way might be to stub `clearEffects` and make sure that it is called after `stopAll`, rather than test a side effect of the `clearEffects` function, but I didn't see anything in the test harness for something like `t.wasCalled`. 

Here is what the change looks like.

what it does now.
scratch-3-no-reset

what scratch 2 does
scratch-2-reset

what this PR does
scratch-3-reset

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -706,9 +706,10 @@ RenderedTarget.prototype.onGreenFlag = function () {
 
 /**
  * Called when the project receives a "stop all"
- * Stop all sounds
+ * Stop all sounds and clear graphic effects.
  */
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
LGTM

Issue #159, opened 2016-09-08
Enforce clone limit


thisandagain:
Scratch 2.0 enforces a limit of 300 (301) clones.

mrjacobbloom:
if performance is better in Scratch 3, would there be any possibility of raising the limit?

jeandrek:
Is it 300 or 301?

tmickel:
Closing since this is now part of #150/#113. 

The current limit looks to be 300: https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/primitives/Primitives.as#L35.

Regarding raising the limit: there's a chance Scratch 3.0 will perform better with many clones. There's also a chance it won't. If it does, we may consider raising the limit. But, there may also be some non-technical reasons to have the limit be at a fathomable/not-too-large number. And, currently some projects likely depend on it. Changing the limit could affect compatibility, which we're trying to be very aware of.

PR #497, opened 2017-03-09
Change MAX_CLONES to 301 for slightly better Scratch 2.0 compatibility

GarethPW:
Scratch 2.0 allowed up to 301 clones, as opposed to the 300 currently specified in `src/engine/runtime.js`.

### Related Issues

LLK/scratch-vm#159


thisandagain:
The clone limit in Scratch 2.0 is indeed 301, but this PR doesn't match (rather we are at 302 right now due to a bug):
https://scratch.mit.edu/projects/151234984/
https://llk.github.io/scratch-gui/#151234984

PR #498, opened 2017-03-09
Load projects & costumes through scratch-storage

### LINKED ISSUE #427: Use scratch-storage
### Expected Behavior

The VM should use the [scratch-storage](/LLK/scratch-storage) module for loading files. Ideally we should do this in such a way that the storage module could be replaced with a different implementation if necessary.

### Actual Behavior

The VM (actually `playground.js` [here](https://github.com/LLK/scratch-vm/blob/fec9e5b91a1a72f78bc9b860ec6784f2641a4600/playground/playground.js#L9)) loads project files itself, relies on the renderer to load costumes, etc.

cwillisf:
NOTE: the scratch-storage module should be made public before merging this PR.

### Resolves

Resolves #427

### Proposed Changes

This change adds scratch-storage as a dependency and modifies the SB2 import process to use scratch-storage. Costumes and the project JSON are fully working, but sound loading is not yet finished. For now, sounds are still working through the old path.

Because costumes are now loaded through scratch-storage, scratch-render no longer needs to make its own HTTP requests.

There are a few other, smaller changes in here, including:
- Use the `Scratch` namespace for "global" items in the playground when possible
- Alphabetize Scratch dependencies (audio, render, storage)
- Fixed `parseScratchObject` returning `undefined` when it should return `null`

### Reason for Changes

File access should be centralized in scratch-storage.

### Test Coverage

Existing tests which interact with project or costume data effectively cover these changes. Each such test now attaches a storage module instance to its VM instance before running the test body. The `sb2_import` test has moved from `tests/unit/` into `tests/integration/` to reflect the fact that it now depends heavily on the storage module.


cwillisf:
I believe this PR is now ready for review -- the initial submission had broken tests.
It's still not ready to _merge_ until the storage module is public, of course. Coming soon!

src/import/sb2import.js
```diff
@@ -138,6 +148,79 @@ var parseScratchObject = function (object, runtime, topLevel) {
...
+    var md5 = idParts[0];
+    var ext = idParts[1].toUpperCase();
+    var assetType = (ext === 'SVG') ? AssetType.ImageVector : AssetType.ImageBitmap;
+
+    var rotationCenter = [
+        costume.rotationCenterX / costume.bitmapResolution,
+        costume.rotationCenterY / costume.bitmapResolution
+    ];
+
+    var promise = runtime.storage.load(assetType, md5);
```
rschamp, line 183: If you don't attach storage to your VM, this line throws a `TypeError`... at the top of these functions that require `runtime.storage`, maybe we should log a warning and exit before doing anything.

src/import/sb2import.js
```diff
@@ -138,6 +149,92 @@ var parseScratchObject = function (object, runtime, topLevel) {
...
+ * @property {int} skinId - the ID of the costume's render skin, once installed.
+ * @property {number} rotationCenterX - the X component of the costume's origin.
+ * @property {number} rotationCenterY - the Y component of the costume's origin.
+ * @property {number} [bitmapResolution] - the resolution scale for a bitmap costume.
+ * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
+ * @returns {?Promise} - a promise which will resolve after skinId is set, or null on error.
+ */
+var loadCostume = function (md5ext, costume, runtime) {
+    if (!runtime.storage) {
+        log('No storage module present; cannot load costume asset: ', md5ext);
```
rschamp: I think it would be more appropriate for these to be errors (with `log.error`).

rschamp (changes requested):

rschamp (approved):
🎉

PR #499, opened 2017-03-09
Switch to using number directly, instead of toNumber for costume switch

### LINKED ISSUE #489: Switching to undefined costume switches to last costume
From a Scratcher:
> Switching to a non-existent costume name now switches to the last costume, previously it didn't switch at all. This breaks some case sensing and possibly other projects.

SillyInventor:
### Resolves

Resolves #489

### Proposed Changes

Instead of using .toNumber which just calls number() and checks for NaN, just call number().

### Reason for Changes

So that you can get NaN when using unrecognized strings, so that it won't switch costume.

### Test Coverage

None added.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -101,7 +101,7 @@ Scratch3LooksBlocks.prototype._setCostumeOrBackdrop = function (target,
                    requestedCostume === 'next backdrop') {
             target.setCostume(target.currentCostume + 1);
         } else {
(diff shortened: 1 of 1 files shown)


rschamp (approved):
This looks good, thanks @SillyInventor!

griffpatch:
There is still an edge case here demonstrated by this project: https://llk.github.io/scratch-gui/develop/#188166105

Number("") still returns 0 not NaN so setting the costume to "" causes a switch to the last costume. This causes a few glitches in my projects.

PR #504, opened 2017-03-20
Don't rename sprites to empty string names

### LINKED ISSUE #488: Validate sprite renames
Empty sprite names should be rejected, and duplicate sprite names should be updated with a counter.

An invalid sprite name shouldn't throw an error, it just shouldn't rename the sprite. Maybe return `true`/`false` if successful?

paulkaplan:
~Fixes the blank string part of https://github.com/LLK/scratch-vm/issues/488 and adds some test coverage to that function which was lacking. I didn't do the name incremented, just because there is a bit more complexity there and I think it would be better to pair on that @rschamp.~

Updated: fixes https://github.com/LLK/scratch-vm/issues/488 including the name incrementing!

### DIFF
--- a/src/util/string-util.js
+++ b/src/util/string-util.js
@@ -1,0 +1,17 @@
+var StringUtil = function () {};
+
+StringUtil.withoutTrailingDigits = function (s) {
+    var i = s.length - 1;
+    while ((i >= 0) && ('0123456789'.indexOf(s.charAt(i)) > -1)) i--;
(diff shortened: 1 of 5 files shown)


rschamp (approved):
Looks great, thanks for the tests!

thisandagain (approved):
Thanks Paul. This looks good. Thanks for adding test coverage for the VM interface as well as the new string utils.

Issue #511, opened 2017-03-27
Sounds are not being loaded into the VM


paulkaplan:
Introduced (intentionally) in https://github.com/LLK/scratch-vm/pull/498/files as a `@todo`. Just wanted to file an issue for this so that we remember. To reproduce right now, just try to `play sound <meow>` on the default project and see it doesn't work.

paulkaplan:
@ericrosenbaum I know we had started working on this, make sure to tag this issue when you make a PR.

PR #512, opened 2017-03-27
Use Scratch Storage to load sounds and Audio Engine to decode and register them

ericrosenbaum:
### Resolves

Addresses https://github.com/LLK/scratch-vm/issues/511

### Proposed Changes

Remove call to loadSounds from rendered-target. Scratch Storage loads the sounds, and passes the data to the audio engine to decode and register them. 

Depends on https://github.com/LLK/scratch-audio/pull/32

### Reason for Changes

Moving all asset loading to Scratch Storage.

### Test Coverage

No tests yet.

### DIFF
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -226,11 +226,15 @@ var loadSound = function (sound, runtime) {
         log.error('No storage module present; cannot load sound asset: ', sound.md5);
         return;
     }
+    if (!runtime.audioEngine) {
+        log('No audio engine present; cannot load sound asset: ', sound.md5);
+        return;
+    }
     var idParts = sound.md5.split('.');
     var md5 = idParts[0];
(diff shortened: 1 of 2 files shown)


src/import/sb2import.js
```diff
@@ -226,11 +226,15 @@ var loadSound = function (sound, runtime) {
         log.error('No storage module present; cannot load sound asset: ', sound.md5);
         return;
     }
+    if (!runtime.audioEngine) {
+        log('No audio engine present; cannot load sound asset: ', sound.md5);
```
cwillisf: These changed to `log.error` as part of the review process on #498 so I'd recommend changing this one too

cwillisf (changes requested):

cwillisf (approved):

rschamp (approved):

PR #514, opened 2017-03-27
Use costume loading from importer in public costume and backdrop loader

paulkaplan:
Extracts the costume loader from the importer so that it can be used for both importing and adding costume/backdrop through the public APIs.

Some background:
Before scratch-storage, we used to just tack on the costume URL directly to the costume object, allowing it to be loaded by the renderer. Now that loading is happening with scratch-storage, the loading step is explicit, and wasn't available to the public API, meaning the backdrop / costume library wouldn't work. This, along with a corresponding PR to scratch-gui, fixes that.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -1,0 +1,69 @@
+var AssetType = require('scratch-storage').AssetType;
+
+/**
+ * Load a costume's asset into memory asynchronously.
+ * Do not call this unless there is a renderer attached.
+ * @param {string} md5ext - the MD5 and extension of the costume to be loaded.
+ * @param {!object} costume - the Scratch costume object.
+ * @property {int} skinId - the ID of the costume's render skin, once installed.
+ * @property {number} rotationCenterX - the X component of the costume's origin.
+ * @property {number} rotationCenterY - the Y component of the costume's origin.
+ * @property {number} [bitmapResolution] - the resolution scale for a bitmap costume.
+ * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
+ * @returns {?Promise} - a promise which will resolve after skinId is set, or null on error.
+ */
+var loadCostume = function (md5ext, costume, runtime) {
+    if (!runtime.storage) {
+        log.error('No storage module present; cannot load costume asset: ', md5ext);
+        return null;
+    }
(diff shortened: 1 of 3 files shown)


src/import/load-costume.js
```diff
@@ -0,0 +1,70 @@
...
+                removeEventListeners = function () {
+                    imageElement.removeEventListener('error', onError);
+                    imageElement.removeEventListener('load', onLoad);
+                };
+                imageElement.addEventListener('error', onError);
+                imageElement.addEventListener('load', onLoad);
+                imageElement.src = costumeAsset.encodeDataURI();
+            });
+        }).then(function (imageElement) {
+            costume.skinId = runtime.renderer.createBitmapSkin(imageElement, costume.bitmapResolution, rotationCenter);
```
rschamp, line 65: I would expect this to return the `skinId`, which the consumer could use in its own `then` rather than altering the costume object passed in.

paulkaplan: @rschamp I agree, but this code was just moved from the existing `loadCostume` function in the sb2importer. @cwillisf is there a reason it was done that way?

cwillisf: The short version: it was easier to do it this way in `sb2import` but there's no need to do it here if you'd like a different interface. If it does stay like this, I almost feel like it should be an instance method on a `Costume` object.

paulkaplan:
@rschamp do you know why this is failing on the CI?

src/import/load-costume.js
rschamp: @paulkaplan I think we should have it return the id here, then over at https://github.com/LLK/scratch-gui/pull/282/files#diff-f0dbeaecf9325f943b76a2cc14612110R21 we don't need that null `skinId`

rschamp (changes requested):

src/import/load-costume.js
paulkaplan: is there a function i can call when that promise resolves with the skinId to apply the skinId to the costume?

paulkaplan: @rschamp @cwillisf is there a function i can call when that promise resolves with the skinId to apply the skinId to the costume?

rschamp (approved):

PR #516, opened 2017-04-03
Expose VM#addSound publicly

paulkaplan:
Create a way to add a sound to a sprite through the VM. For the GUI to use, instead of reaching directly into the VM. Like the `addCostume` and `addBackdrop` interfaces, it does the required loading of the sound into the audio engine (again by extracting the required logic from `sb2import`). 

This will allow us to fix the sound tab in scratch-gui.

### DIFF
--- a/src/import/load-sound.js
+++ b/src/import/load-sound.js
@@ -1,0 +1,29 @@
+var AssetType = require('scratch-storage').AssetType;
+var log = require('../util/log');
+
+/**
+ * Load a sound's asset into memory asynchronously.
+ * @param {!object} sound - the Scratch sound object.
+ * @property {string} md5 - the MD5 and extension of the sound to be loaded.
+ * @property {Buffer} data - sound data will be written here once loaded.
+ * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
+ * @returns {?Promise} - a promise which will resolve after sound is loaded, or null on error.
+ */
+var loadSound = function (sound, runtime) {
+    if (!runtime.storage) {
+        log.error('No storage module present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(null);
+    }
+    if (!runtime.audioEngine) {
+        log.error('No audio engine present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(null);
+    }
+    var idParts = sound.md5.split('.');
+    var md5 = idParts[0];
+    return runtime.storage.load(AssetType.Sound, md5).then(function (soundAsset) {
+        sound.data = soundAsset.data;
+        return runtime.audioEngine.decodeSound(sound);
+    });
+};
+
+module.exports = loadSound;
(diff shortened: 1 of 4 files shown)


src/import/load-sound.js
```diff
@@ -0,0 +1,29 @@
+var AssetType = require('scratch-storage').AssetType;
+var log = require('../util/log');
+
+/**
+ * Load a sound's asset into memory asynchronously.
+ * @param {!object} sound - the Scratch sound object.
+ * @property {string} md5 - the MD5 and extension of the sound to be loaded.
+ * @property {Buffer} data - sound data will be written here once loaded.
+ * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
+ * @returns {?Promise} - a promise which will resolve after sound is loaded, or null on error.
```
cwillisf: This function no longer returns `null` on error

src/import/load-sound.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+var loadSound = function (sound, runtime) {
+    if (!runtime.storage) {
+        log.error('No storage module present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(null);
```
cwillisf: Perhaps this should follow #517's recent precedent and `return Promise.resolve(sound)`?

src/import/load-sound.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+    if (!runtime.audioEngine) {
+        log.error('No audio engine present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(null);
+    }
+    var idParts = sound.md5.split('.');
+    var md5 = idParts[0];
+    return runtime.storage.load(AssetType.Sound, md5).then(function (soundAsset) {
+        sound.data = soundAsset.data;
+        return runtime.audioEngine.decodeSound(sound);
```
cwillisf, line 25: Just FYI: there's currently a branch of `decodeSound` which returns the return value of `log.warn`. I'm fairly certain that's not the sound object or a promise for it (heh), so until/unless that changes this promise chain won't reliably result in a sound object even if you change the error cases here to return `Promise.resolve(sound)`.

cwillisf (changes requested):
Looks good, just a couple of details. Really the JSDoc on `loadSound` is the only reason I'm picking "Request changes", though I'm interested in your thoughts on the other comments.

src/import/load-sound.js
paulkaplan: Good point, fixed in 		33f4368

src/import/load-sound.js
paulkaplan: Interesting. I'll file that as an issue on scratch-audio, since it cannot really be addressed here.

paulkaplan:
@cwillisf thanks for the feedback, updated the PR with a couple commits and responded to your questions.

cwillisf (approved):

rschamp (approved):

Issue #515, opened 2017-03-31
Costume and sound data must be reported with sprite info


paulkaplan:
In order for the GUI to present sound and costume data in the tabs, we need to report not only the current costume but all costumes, as well as all sounds. I am labeling this a `bug` because until it is resolved we will not be able to play sounds or display costume thumbnails in GUI. 

## Questions
How can we report the costume and sound data in a way that supports

1. Displaying a costume thumbnail.
2. Playing a sound on demand.

PR #517, opened 2017-04-03
Add costume dataURI to costumes from storage

rschamp:
When costumes are fetched from storage, add the dataURI of the costume data to the costume, so that the GUI can use it in turn.

Also fix some tests that were calling addCostume and addBackdrop incorrectly (not sure how the tests passed before).

Re-arrange loadCostume, so that the returned costume contains as much data as loadCostume can provide.

Towards #515, LLK/scratch-gui#279
/cc @cwillisf 

bitmoji

### DIFF
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -53,6 +53,7 @@ var parseScratchObject = function (object, runtime, topLevel) {
                 rotationCenterY: costumeSource.rotationCenterY,
                 skinId: null
             };
+            sprite.costumes.push(costume);
             costumePromises.push(loadCostume(costumeSource.baseLayerMD5, costume, runtime));
         }
     }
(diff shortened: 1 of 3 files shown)


src/import/sb2import.js
```diff
@@ -138,7 +135,12 @@ var parseScratchObject = function (object, runtime, topLevel) {
         }
     }
     target.isStage = topLevel;
-    Promise.all(costumePromises).then(function () {
+    Promise.all(costumePromises).then(function (costumes) {
+        sprite.costumes = costumes.filter(
+            function (c) {
+                return c !== null;
+            }
+        );
```
cwillisf: It looks to me like there's no longer any way to get here with a `null`; maybe you don't need the filter?

cwillisf (approved):

paulkaplan (approved):
LG, although second @cwillisf comment

src/import/sb2import.js
rschamp: Good point!

Issue #521, opened 2017-04-05
Size property behavior in Scratch 3.0 not compatible with 2.0


rasmushaglund:
### Expected Behavior

Size measurement should be the same in Scratch 3.0 and 2.0 to ensure backward compatibility. In 2.0 size is defined as float while in 3.0 the property is rounded to integer.

### Actual Behavior

Size differs between versions making some edge cases result in Scratch 2.0 projects not being compatible with 3.0.

### Steps to Reproduce

Run and compare this identical project in Scratch 2.0 vs Scratch 3.0:
* https://scratch.mit.edu/projects/71362070/
* https://llk.github.io/scratch-gui/#71362070

Observe that the second example will stuck because Intro sprite never reaches 100%

Original issue: https://github.com/LLK/scratch-gui/issues/211

### Operating System and Browser

Linux and Google Chrome 56.0.2924.87 (64-bit)

rasmushaglund:
https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L301-L308

cwillisf:
Reopening since #522 made `set size` act like Scratch 2.0, but the `size` reporter still needs work.

Specifically, the Scratch 2.0 `size` reporter returns an integer even when the internal size value might have a fractional component. See also: https://github.com/LLK/scratch-flash/blob/develop/src/primitives/LooksPrims.as#L244

@tmickel shared this example on #522:
> But the reporter block in scripts does seem to round, e.g., in Scratch 2.0:
> <img width="204" alt="screen shot 2017-04-05 at 12 56 49 pm" src="https://cloud.githubusercontent.com/assets/120403/24717139/6d14effa-19ff-11e7-98fc-8dce61cdfbab.png">
> causes the sprite to say hello.

rasmushaglund:
Ah I see, I'll see if I can fix that right away.

rasmushaglund:
Now the above example is working at least, let me know if you find any more places that needs to change.

cwillisf:
Between #522 and #528 I believe this is now fixed.

@rschamp @paulkaplan: In the long term I'd like to consider other possible solutions to this, such as removing the rounding from the reporter in favor of rounding in the GUI combined with adding a tolerance to the `=` operator. We'd have to control that sort of change with a compatibility flag, though, since it would change project behavior in some cases.

PR #522, opened 2017-04-05
Fix so that size is set in float to ensure backward compatibility with Scratch 2.0

rasmushaglund:
### Resolves

https://github.com/LLK/scratch-vm/issues/521
https://github.com/LLK/scratch-gui/issues/211

### Proposed Changes

Represent sprite size as float instead of integer

### Reason for Changes

Scratch 2.0 represent size as float, this change will ensure backward compatibility 

### Test Coverage

None. I did try to add a integration test, but I would need some pointers on how I could write a test for this issue.

### DIFF
--- a/src/playground/index.html
+++ b/src/playground/index.html
@@ -33,6 +33,7 @@
             <canvas id="scratch-stage" style="width: 480px; height: 360px;"></canvas><br />
             x: <input id='sinfo-x' />
             y: <input id='sinfo-y' /><br />
+            size: <input id='sinfo-size' /><br />
             dir: <input id='sinfo-direction' />
             rotation style: <input id='sinfo-rotationstyle' /><br />
             visible: <input id='sinfo-visible' />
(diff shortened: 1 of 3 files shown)


thisandagain:
Thanks @rasmushaglund . We'll take a look!

tmickel:
(p.s., there is rounding apparently in 2.0's size *reporter* https://github.com/LLK/scratch-flash/blob/develop/src/primitives/LooksPrims.as#L244 which I think is the source of the original mistake)

rasmushaglund:
Yeah @tmickel it sounds reasonable. Would it make sense to make the ui to care about rounding? In my view the vm should always report a detailed state, and let whatever uses the lib handle eventual rounding.

tmickel:
@rasmushaglund
I agree that the VM should have the most detailed state as its internal representation, then various parts of the UI could reflect the real value, like your changes here do reflect :) your change is also very important because the previous version was rounding *every* time the size was set, causing cumulative error.

But the reporter block in scripts does seem to round, e.g., in Scratch 2.0:
screen shot 2017-04-05 at 12 56 49 pm
causes the sprite to say hello.

If we're going for compatibility, that (the behavior of the "size" reporter) is also probably important to maintain, not just at the UI level, if that makes sense. I think it's arguably an odd/questionable behavior, but may be important to not break projects.

anyway just making the note :)

cwillisf (approved):
Looks good!

I agree with @tmickel's concerns about the `size` reporter, but I think that can be handled with a separate change. I think we should leave #521 open until we resolve the reporter's behavior, though.

Issue #481, opened 2017-02-26
Clones don't inherit pen color


thisandagain:
Reported by a Scratcher in the forums:
> I found a bug: clones don't inherit pen colour from their parents. It can be seen in [this project](http://llk.github.io/scratch-vm/#24900838), where the drips and spray will always be the default blue regardless of the main paint colour.

### Expected Behavior

- Clones should inherit pen color from the sprite which created them

### Actual Behavior

- They inherit the default "blue" pen color

### Operating System and Browser

Mac OS 10.12.3  |  Chrome Version 56.0.2924.87 (64-bit)

SillyInventor:
This can be fixed by copying _customState and _events from renderedTarget to newClone in makeClone.

rasmushaglund:
Fixed in https://github.com/LLK/scratch-vm/pull/525. For some reason it wasn't automagically referenced in this ticket.

cwillisf:
Partially resolved by #555, tracking remained in #557

PR #525, opened 2017-04-05
Copy events and customState when creating a clone

rasmushaglund:
### Resolves

https://github.com/LLK/scratch-vm/issues/493 and https://github.com/LLK/scratch-vm/issues/481

### Proposed Changes

Copy `_events` and `_customState`

### Reason for Changes

Clones do not inherit pen color

### Test Coverage

None

### DIFF
--- a/src/import/load-sound.js
+++ b/src/import/load-sound.js
@@ -1,0 +1,29 @@
+var AssetType = require('scratch-storage').AssetType;
+var log = require('../util/log');
+
+/**
+ * Load a sound's asset into memory asynchronously.
+ * @param {!object} sound - the Scratch sound object.
+ * @property {string} md5 - the MD5 and extension of the sound to be loaded.
+ * @property {Buffer} data - sound data will be written here once loaded.
+ * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
+ * @returns {!Promise} - a promise which will resolve after sound is loaded
+ */
+var loadSound = function (sound, runtime) {
+    if (!runtime.storage) {
+        log.error('No storage module present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(sound);
+    }
+    if (!runtime.audioEngine) {
+        log.error('No audio engine present; cannot load sound asset: ', sound.md5);
+        return Promise.resolve(sound);
+    }
+    var idParts = sound.md5.split('.');
+    var md5 = idParts[0];
+    return runtime.storage.load(AssetType.Sound, md5).then(function (soundAsset) {
+        sound.data = soundAsset.data;
+        return runtime.audioEngine.decodeSound(sound);
+    });
+};
+
+module.exports = loadSound;
--- a/src/import/sb2import.js
+++ b/src/import/sb2import.js
@@ -5,9 +5,6 @@
  * scratch-vm runtime structures.
  */
 
-var ScratchStorage = require('scratch-storage');
-var AssetType = ScratchStorage.AssetType;
-
 var Blocks = require('../engine/blocks');
 var RenderedTarget = require('../sprites/rendered-target');
 var Sprite = require('../sprites/sprite');
@@ -19,6 +16,7 @@ var Variable = require('../engine/variable');
 var List = require('../engine/list');
 
 var loadCostume = require('./load-costume.js');
+var loadSound = require('./load-sound.js');
 
 /**
  * Parse a single "Scratch object" and create all its in-memory VM objects.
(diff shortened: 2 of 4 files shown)


cwillisf (changes requested):
Cloning `_customState` is a good idea, but I don't think we should clone `_events`. The two main reasons I feel that way:
1. `_events` is internal, undocumented `EventEmitter` data. In order to reduce the chances of maintenance headaches later on, I'd rather stick to public APIs when possible -- especially when dealing with code that's not part of this repository.
2. It's possible that rendered targets will eventually have events which are not related to the pen, including (possibly) some events that shouldn't be cloned when cloning the target.

This could end up being pretty tricky to fix in a clean way. The first thing that comes to mind is to let individual block packages react to a clone operation. For example, maybe cloning a target fires `EVENT_TARGET_CLONED` and "pen down" registers for that as well as `EVENT_TARGET_MOVED`. The handler for `EVENT_TARGET_CLONED` could effectively run `penDown` on the new clone. To make that easier, the contents of the `penDown` function should probably be pulled out into a new function which can be called by both `penDown` and the new clone event handler.

That's just one way to approach it, though. If you see an easier way that's fine by me :)

rasmushaglund:
Ok sounds like a better approach, I'll give it a shot :)

rasmushaglund:
@cwillisf would it make sense to only clone _customState in order to solve #493 and #481, and then create another issue/improvement for the refactoring of cloning functionality?

cwillisf:
@rasmushaglund that sounds fine to me :)

cwillisf:
On second thought... I like that idea in principle but I'm not sure it's the right way to go. Cloning the custom state will clone the `penDown` flag, but since the event isn't registered the pen won't actually behave like it's down. Even worse, running the "pen down" block won't fix it because of the `if (!penState.penDown)` in the `penDown` function. I suppose #493 and #481 would technically be fixed, but the user-observable behavior is only slightly better.

I'm starting to think that the best approach might *not* involve cloning `_customState` at all but instead letting individual block packages set custom state on a new clone however they want, potentially using the clone source as a reference. I think we should consider having an event which fires after a clone is created (like `EVENT_TARGET_CLONED` as above), and that event should get both the new clone and the original as parameters. This is an approach that @rschamp and I were discussing last week; I had anticipated we'd get to it later but maybe we shouldn't put it off...

rasmushaglund:
Thanks for the explaination @cwillisf. I agree that this probably is the correct way to solve it.

Where can I add the listeners for the cloning event? I can't find any constructors in the blocks, and the other event listeners (for example in the pen block) is created on an action like penDown and penUp. We would need to add the listeners when the block and clones are created. Any pointers?

cwillisf:
@rasmushaglund sorry for the delay in responding on this. You're right - currently there's no good place to attach event listeners like that. Most likely we'll end up adding some sort of initialization function to each "package" of blocks, but I'm not sure we're quite ready to do that yet. We've been talking within the team about how we might implement extensions in Scratch 3.0 and the decisions we make there are likely to influence other changes to the way we package blocks.

So... I guess the options at the moment are:
- Clone `_customState` as part of the cloning operation, technically fixing #481 and #493 but leaving the bug where the pen thinks it's down but the events aren't hooked up.
- Do nothing for now; wait for us to add a block package initialization function. I'm sure you could do this yourself but as it could involve more structural changes I'd prefer it to come from within the team so we can ~argue~discuss within the team how exactly we want it to happen ;)

Either option is fine by me -- you've already helped us quite a bit just by having this conversation!

SillyInventor:
I'd like to add an alternative: Rather than copying the _customState & _events directly, check _customState, and call penDown(), setpencolortonum(), setpenshadeto(), & setpensizeto() for the new clone. Rinse and repeat for all other states.

In general I'm not a fan of enumeration, but it will alleviate the problem of copying future components that shouldn't be and copying undocumented objects.

cwillisf:
@SillyInventor that seems like a reasonable idea, but it still has some of the same problems. Namely, we need a good way to run some pen-specific code at clone time. We could put the code you describe in the clone code directly, but that code shouldn't really "know about" the pen's implementation details.

Really, running some pen-specific code at clone time is the main hurdle with most of the solutions we've discussed on this PR. We'll add a way to do that at some point, but it hasn't yet sifted to the top of the priority list. We're starting to look at extensions and think about "block packages" in more detail, though, so those prototypes and discussions will likely lead to figuring out how we want to fix this.

rasmushaglund:
No worries @cwillisf, the delay coincided with me working on other things. I agree that the structural changes should be decided by the team that collaborate and discuss every day, and have most knowledge of the system. 

I'd prefer to clone _customState to close those bugs and then revisit when the initialization functions (or similar) are in place.

rasmushaglund:
As I described in https://github.com/LLK/scratch-vm/pull/555 I broke the branch, so I had to start a new pull request.

PR #526, opened 2017-04-06
Report sound costume data

### LINKED ISSUE #515: Costume and sound data must be reported with sprite info

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/515

### Proposed Changes

Includes all costume and sound data in the `toJSON` method, in order to provide that data to the GUI. 

### Reason for Changes

Previously only the current costume data was reported. With the new sound and costume tabs in gui, we also need to report all costume and sound data from the VM. 

### Test Coverage

Added unit test for covering the new methods.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -441,6 +441,14 @@ RenderedTarget.prototype.getCostumes = function () {
 };
 
 /**
+ * Get full sound list
+ * @return {object[]} list of sounds
+ */
+RenderedTarget.prototype.getSounds = function () {
+    return this.sprite.sounds;
+};
+
+/**
  * Update all drawable properties for this rendered target.
  * Use when a batch has changed, e.g., when the drawable is first created.
  */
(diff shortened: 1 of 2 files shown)


src/sprites/rendered-target.js
```diff
@@ -767,6 +775,8 @@ RenderedTarget.prototype.toJSON = function () {
         direction: this.direction,
         draggable: this.draggable,
         costume: this.getCurrentCostume(),
```
rschamp, line 777: Now that we have the full costume list, perhaps this should update to the current costume id rather than the costume data?  (In a subsequent PR)

rschamp (approved):

PR #528, opened 2017-04-08
Fix rounded get size reporting

rasmushaglund:
### Resolves

https://github.com/LLK/scratch-vm/issues/521

### Proposed Changes

Round the result from getSize

### Reason for Changes

To ensure backward compatibility with Scratch 2.0 where size is reported in integers

### Test Coverage

None


cwillisf (approved):
Thanks!

PR #535, opened 2017-04-18
Separate downloading from installation of sprite info

### LINKED ISSUE #278: Sprite selector doesn't render in target pane
### Expected Behavior

Should see the sprite selectors render
### Actual Behavior

image

---

This is because of a change to what the VM reports when we introduced `scratch-storage`. It used to report `skin` on the costume, which was a full URL, but now it doesn't.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/278

### Proposed Changes

Changes the order of operations when loading a project so that the downloading of all sprite assets completes before installing the new targets into the runtime.

### Reason for Changes

The introduction of scratch-storage made the loading happen in an indeterminate order, which caused some issues when loading projects / rendering sprites. This forces all downloading to complete before telling the GUI anything. 

### Test Coverage

Updated the tests to work correctly.

### DIFF
--- a/src/import/load-sound.js
+++ b/src/import/load-sound.js
@@ -7,7 +7,7 @@ var log = require('../util/log');
  * @property {string} md5 - the MD5 and extension of the sound to be loaded.
  * @property {Buffer} data - sound data will be written here once loaded.
  * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
- * @returns {!Promise} - a promise which will resolves to the sound when ready.
+ * @returns {!Promise} - a promise which will resolve to the sound when ready.
  */
 var loadSound = function (sound, runtime) {
     if (!runtime.storage) {
--- a/src/import/load-sound.js
+++ b/src/import/load-sound.js
@@ -7,7 +7,7 @@ var log = require('../util/log');
  * @property {string} md5 - the MD5 and extension of the sound to be loaded.
  * @property {Buffer} data - sound data will be written here once loaded.
  * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
- * @returns {!Promise} - a promise which will resolve after sound is loaded
+ * @returns {!Promise} - a promise which will resolves to the sound when ready.
  */
 var loadSound = function (sound, runtime) {
     if (!runtime.storage) {
@@ -22,7 +22,9 @@ var loadSound = function (sound, runtime) {
     var md5 = idParts[0];
     return runtime.storage.load(AssetType.Sound, md5).then(function (soundAsset) {
         sound.data = soundAsset.data;
-        return runtime.audioEngine.decodeSound(sound);
+        return runtime.audioEngine.decodeSound(sound).then(function () {
+            return sound;
+        });
     });
 };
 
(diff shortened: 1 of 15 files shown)


src/import/load-sound.js
```diff
@@ -7,7 +7,7 @@ var log = require('../util/log');
  * @property {string} md5 - the MD5 and extension of the sound to be loaded.
  * @property {Buffer} data - sound data will be written here once loaded.
  * @param {!Runtime} runtime - Scratch runtime, used to access the storage module.
- * @returns {!Promise} - a promise which will resolve after sound is loaded
+ * @returns {!Promise} - a promise which will resolves to the sound when ready.
```
rschamp: Nit: object/verb agreement grammar

src/virtual-machine.js
```diff
@@ -150,17 +150,27 @@ VirtualMachine.prototype.postIOData = function (device, data) {
 /**
  * Load a project from a Scratch 2.0 JSON representation.
  * @param {?string} json JSON string representing the project.
+ * @return {!Promise} Promise that resolves after targets are installed.
  */
 VirtualMachine.prototype.loadProject = function (json) {
-    this.clear();
+    var that = this;
```
rschamp: Please use `.bind(this)` on the callback function (this will make the merge easier with the ES6 PR).

src/virtual-machine.js
```diff
@@ -185,11 +195,15 @@ VirtualMachine.prototype.downloadProjectId = function (id) {
  */
 VirtualMachine.prototype.addSprite2 = function (json) {
     // Select new sprite.
-    this.editingTarget = sb2import(json, this.runtime, true);
-    // Update the VM user's knowledge of targets and blocks on the workspace.
-    this.emitTargetsUpdate();
-    this.emitWorkspaceUpdate();
-    this.runtime.setEditingTarget(this.editingTarget);
+    var that = this;
```
rschamp: Same here, using `bind(this)` on the function will let eslint detect and convert it to an arrow function.

rschamp (changes requested):
Thanks for taking this on! Looks great overall, just a few nits.

paulkaplan:
@rschamp addressed your comments in 4b5e7a5

src/virtual-machine.js
```diff
@@ -153,24 +153,23 @@ VirtualMachine.prototype.postIOData = function (device, data) {
...
 
         // Update the VM user's knowledge of targets and blocks on the workspace.
-        that.emitTargetsUpdate();
-        that.emitWorkspaceUpdate();
-        that.runtime.setEditingTarget(that.editingTarget);
-    });
+        this.emitTargetsUpdate();
+        this.emitWorkspaceUpdate();
+        this.runtime.setEditingTarget(this.editingTarget);
+    }.bind(this));
```
towerofnix: Is this the best method of scoping `this`, currently? In ES6 arrow functions can be used to keep the value of `this`, ex.:

```
setTimeout(() => {
  this.spin()
}, 500)

// ..instead of
setTimeout(function() {
  this.spin()
}.bind(this), 500)
```

I suppose the ES6 PR isn't merged yet, but when it is, it might be better to use arrow functions rather than `fn.bind(this)`. (Sorry if this was already discussed somewhere else!)

rschamp: Right — eslint can recognize `function () {}.bind(this)` callbacks as equivalent to an arrow function and automatically convert them. http://eslint.org/docs/rules/prefer-arrow-callback

rschamp (approved):

PR #534, opened 2017-04-17
Use ES6, convert to ES6-style classes

rschamp:
### Resolves

_What Github issue does this resolve (please include link)?_
Resolves #3 

### Proposed Changes

_Describe what this Pull Request does_
Updates the build system to use Babel, and updates class syntax to ES6 classes.

### Reason for Changes

_Explain why these changes should be made_
In our discussion last week we decided to settle on ES6 everywhere, including here.

### Test Coverage

_Please show how you have added tests to cover your changes_
It works!

PR #539, opened 2017-04-20
Use ES6 linting rules in the project root

### LINKED ISSUE #537: Use ES6 lint rules for the root
Since we're moving our base Node version to 6, we can use ES6 syntax everywhere in the project. Less context-switching = nicer experience = convert our tests to ES6.

rschamp:
Follow up from #534. Update all tests for `no-var` and `prefer-arrow-callback` (using `--fix`).

Resolves #537

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,3 +1,3 @@
 module.exports = {
-    extends: ['scratch', 'scratch/node']
+    extends: ['scratch', 'scratch/node', 'scratch/es6']
 };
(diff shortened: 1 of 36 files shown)


cwillisf (approved):
LGTM

rschamp:
@cwillisf @thisandagain Rebased onto the new tests, and re-`--fix`ed.

cwillisf (approved):
Still LGTM ;)

PR #541, opened 2017-04-21
Get `Asset` and `AssetType` from storage instance

cwillisf:
### Proposed Changes

Access `Asset` and `AssetType` through an instance of `ScratchStorage`. Remove all instances of `require('scratch-storage')` from this repository.

This requires LLK/scratch-storage#9

### Reason for Changes

This change means `scratch-storage` is no longer built into the `scratch-vm` build output.

### Test Coverage

The existing storage-related test code has been updated.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -1,4 +1,3 @@
-const AssetType = require('scratch-storage').AssetType;
 const log = require('../util/log');
 
 /**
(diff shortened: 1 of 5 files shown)


rschamp (approved):
I discovered this when I was converting the VM to ES6, but it's existed forever apparently. If we wanted to maintain `Asset` and `AssetType` as class properties, we could avoid importing scratch-storage and access them from the instance with `runtime.storage.constructor.AssetType`.

It kind of depends on your opinion of what type of data these are. But if we want configurable asset types eventually I think the instance getters are correct.

cwillisf:
Note: the Travis CI build will fail until LLK/scratch-storage#9 is merged. I'll trigger a rebuild after that merge and before this merge, just to make sure we're all ✅

Issue #550, opened 2017-05-01
Discussion - Many larger projects do not load / unusable


thisandagain:
Filing an issue to track this as it's something we should discuss. Many mid-sized projects (such as the one below) currently cannot be loaded into the GUI. I haven't spent too much time in the profiler yet, but a cursory glance shows that we are burning through memory and CPU even when the editor is idle.

#### Mid-level Complexity Project
- `https://llk.github.io/scratch-gui/#155128646`
- Press green flag
- Profile / observe performance

#### Simple Project
- `https://llk.github.io/scratch-gui/#158487479`
- Press green flag
- Add a new sprite
- Profile / observe performance

rschamp:
The serialization fix only addressed the issue with adding a new sprite in the simple project. There is still work to be done for the mid-level complexity project performance. Though I wonder if we need a general purpose issue like this for it, or should we investigate and file more specific performance issues?

cwillisf:
IMO the performance problems we were having before were bad enough to count as "broken" rather than just "not done yet" so this issue was justified at the time. If we've improved to just "not done yet" levels of performance problems (have we?) then I would be OK with closing this. If not, I'd prefer to leave this open until we identify the main things that are pushing into "broken" territory.

PR #552, opened 2017-05-04
Add project serialization

rschamp:
### Resolves
Requires https://github.com/LLK/scratch-storage/pull/11. Tests will fail until this is merged and published.

Resolves https://github.com/LLK/scratch-vm/issues/550 and https://github.com/LLK/scratch-gui/issues/329
Also is a start on https://github.com/LLK/scratch-vm/issues/194 (not final until we have a parser).

### Proposed Changes

This is a merge of the work done by @TheBrokenRail, @morantsur, and @thisandagain into the current VM.
- Add JSON serialization of project data, to serve as a first iteration of SB3 JSON
- Add sb3 json de/serializer
- Refactor sb2 loading to use a sb2 deserializer
- Stop storing costume and sound data in the VM state. 

### Reason for Changes

- Gives us a way to save/load sb3 files.
- Improve GUI performance by not sending costume and sound data in target updates. The GUI wants to compare the targets, and walking ArrayBuffers is not very fun

/cc @thisandagain @paulkaplan


src/virtual-machine.js
```diff
@@ -180,11 +166,78 @@ class VirtualMachine extends EventEmitter {
...
+     * @returns {Promise} Promise that resolves after the project has loaded
+     */
+    fromJSON (json) {
+        // Clear the current runtime
+        this.clear();
+
+        // Validate & parse
+        if (typeof json !== 'string') return;
+        json = JSON.parse(json);
+        if (typeof json !== 'object') return;
```
cwillisf: These early-outs should probably log an error or something (same in `addSprite2`)

src/virtual-machine.js
```diff
@@ -180,11 +166,78 @@ class VirtualMachine extends EventEmitter {
...
+        return deserializer.deserialize(json, this.runtime).then(targets => {
+            this.clear();
+            for (let n = 0; n < targets.length; n++) {
+                if (targets[n] !== null) {
+                    this.runtime.targets.push(targets[n]);
+                    targets[n].updateAllDrawableProperties();
+                }
+            }
+            // Select the first target for editing, e.g., the first sprite.
+            this.editingTarget = this.runtime.targets[1];
```
cwillisf: There's no requirement for a project to include a sprite. I'd recommend something equivalent to:
`this.editingTarget = this.runtime.targets[targets.length > 1 ? 1 : 0];`

That way the stage is selected if there are no sprites. If there isn't even a stage then the editing target will be `undefined`, which I guess is the best we can hope for in that situation...

cwillisf (changes requested):

cwillisf:
BTW: what happened with the history on this branch?

rschamp:
That's the actual history :)  I will squash when I merge.

rschamp:
@cwillisf Updated

cwillisf (approved):

PR #556, opened 2017-05-05
Add WeDo 2.0 blocks

cwillisf:
### Proposed Changes

Add block implementations, and supporting code, for the LEGO WeDo 2.0. Note that these blocks are not yet exposed through scratch-blocks; that's intentional: I'd like to experiment with exposing these blocks to the GUI and to scratch-blocks dynamically, similar to the way we might in the case of a user-supplied extension.

Note that these blocks are not yet exposed at the scratch-blocks level; the current extensions prototype implements that in a way that we might not want to merge into the main development line.

**For this reason I think it's worth discussing whether or not we currently want to merge** these blocks into `develop`, or whether it would be better to hold off and use these blocks as a way to experiment with extensions as described here: https://github.com/LLK/scratch-vm/wiki/Extensions-Proposal

### Reason for Changes

This is part of the current extension "track" of Scratch 3.0 development, working toward an extension playtest.

### Test Coverage

No coverage as this involves a separate physical hardware device. We could consider mocking a WeDo 2.0 device for testing purposes.

/cc @ericrosenbaum

### DIFF
--- a/src/blocks/scratch3_wedo2.js
+++ b/src/blocks/scratch3_wedo2.js
@@ -1,3 +1,4 @@
+const color = require('../util/color');
 const log = require('../util/log');
 
 /**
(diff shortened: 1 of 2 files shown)


rschamp (approved):
This is a pretty small and debatable thing since these classes are all within the same file, but the only thing I'm wondering about is the relationship between `WeDo2Motor`, `WeDo2`, and `Scratch3WeDo2Blocks`. For instance, it seems a little un-OOP to require `WeDo2#motor` only so that you can call into `WeDo2Motor` from `Scratch3WeDo2Blocks`. That seems like something `WeDo2` could handle. Similarly, `WeDo2Motor` calls `_send`, when the sending seems like the task of `WeDo2`. For `Scratch3WeDoBlocks` I wonder about `_forEachMotor` — that also seems like something that `WeDo2` could handle.

cwillisf:
Yeah, I think I mostly agree with your concerns, @rschamp. I put `_forEachMotor` on `Scratch3WeDoBlocks` because I was thinking of the motor ID enum as a "blocks" thing, but that caused me to need `WeDo2#motor` which isn't awesome. I'm a little bit less concerned about `WeDo2Motor` calling `_send` -- I sort of think of it as "behind the curtain" -- but I agree that ideally it wouldn't. I've filed a low-prio issue (#596) to clean this stuff up a bit: I think it's worth doing since this code is likely to be used as an example.

PR #558, opened 2017-05-08
Run blocks with boxes checked next to them

fsih:
When a checkbox is checked in the flyout, calculate the value of that block every frame in the VM. Does not yet deal with single sprite scoped variables.

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -85,7 +85,7 @@ const execute = function (sequencer, thread) {
         } else {
             // In a non-hat, report the value visually if necessary if
             // at the top of the thread stack.
-            if (typeof resolvedValue !== 'undefined' && thread.atStackTop()) {
+            if (typeof resolvedValue !== 'undefined' && thread.showVisualReport && thread.atStackTop()) {
                 runtime.visualReport(currentBlockId, resolvedValue);
             }
             // Finished any yields.
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -350,6 +350,8 @@ class VirtualMachine extends EventEmitter {
      * @param {!Blockly.Event} e Any Blockly event.
      */
     monitorBlockListener (e) {
+        // Filter events by type, since monitor blocks only need to listen to these events.
+        // Monitor blocks shouldn't be destroyed when flyout blocks are deleted.
         if (['create', 'change'].indexOf(e.type) !== -1) {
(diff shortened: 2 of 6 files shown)


fsih:
Does this need tests?

src/virtual-machine.js
```diff
@@ -345,6 +346,16 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Handle a Blockly event for the flyout to be passed to the monitor container.
+     * @param {!Blockly.Event} e Any Blockly event.
+     */
+    monitorBlockListener (e) {
+        if (['create', 'change'].indexOf(e.type) !== -1) {
```
rschamp: We should explain this whitelist in the docstring

rschamp (approved):

rschamp:
It would be great to have tests for this. But since this is still incomplete, we can write them as we finish the feature... we might be making more work for ourselves writing them before we finish figuring out exactly how it should work.

mrjacobbloom:
something to watch out for: right now no reporters should have any side effects, but if custom reporters are implemented that might become an issue

rschamp:
@mrjacobbloom That's a good point! For that reason we will probably not allow monitoring of custom reporters if they are added. (And probably the reason you can't monitor custom reporters in Snap!)

src/engine/execute.js
```diff
@@ -85,8 +87,22 @@ const execute = function (sequencer, thread) {
...
-                runtime.visualReport(currentBlockId, resolvedValue);
+                if (thread.showVisualReport) {
+                    runtime.visualReport(currentBlockId, resolvedValue);
+                }
+
+                if (thread.updateMonitor) {
+                    runtime.updateMonitors([{
+                        id: currentBlockId, // @todo(dd) this will collide if multiple sprites use same block
+                        category: 'data',
+                        label: blockContainer.getOpcode(blockContainer.getBlock(currentBlockId)), // @todo(dd) how to handle translation here?
```
CosmicWebServices: 100:1   error    Line 100 exceeds the maximum line length of 120       max-len

Travis error

CosmicWebServices (changes requested):
Travis does not pass

src/virtual-machine.js
fsih: Done

fsih:
This code won't do anything until the blockly change to fire checkbox events is in. https://github.com/paulkaplan/scratch-blocks/tree/emit-checkbox-event

PR #560, opened 2017-05-10
Update monitors action

fsih:
### Resolves
LLK/scratch-gui#356

### Proposed Changes
Sends the monitor update event from the VM to the GUI. We have to keep track of which threads are started by click (which shows value pop up) vs monitor (which sends update monitor action). We also have to prevent the green flag from turning on whenever a monitor script is run.


src/engine/runtime.js
```diff
@@ -769,16 +787,16 @@ class Runtime extends EventEmitter {
      * Emit run start/stop after each tick. Emits when `this.threads.length` goes
      * between non-zero and zero
      *
-     * @param {number} threadCount The new threadCount
+     * @param {number} nonMonitorThreadCount The new nonMonitorThreadCount
      */
-    _setThreadCount (threadCount) {
-        if (this._threadCount === 0 && threadCount > 0) {
+    _maybeEmitProjectRunStartOrStop (nonMonitorThreadCount) {
```
rschamp: How about `emitProjectRunStatus`? And perhaps we should do the setThreadCount separately if we're giving it a distinct name? This name doesn't indicate that the thread count is being set anymore.

src/engine/blocks.js
```diff
@@ -288,6 +290,24 @@ class Blocks {
...
+                        category: 'data',
+                        // @todo(dd) how to handle translation here?
+                        label: block.opcode,
+                        value: '',
+                        x: 0,
+                        // @todo(dd) place below the last monitor instead
+                        y: 0
+                    }]
+                );
+            }
```
rschamp: Firing the add/remove events here makes the GUI need to do more management than necessary. Let's try to only emit the current state of the monitor list each tick.

src/engine/execute.js
```diff
@@ -85,8 +87,17 @@ const execute = function (sequencer, thread) {
...
+            if (typeof resolvedValue !== 'undefined' && thread.atStackTop()) {
+                if (thread.showVisualReport) {
+                    runtime.visualReport(currentBlockId, resolvedValue);
+                }
+
+                if (thread.updateMonitor) {
+                    runtime.updateMonitors([{
+                        id: currentBlockId,
+                        value: String(resolvedValue)
+                    }]);
```
rschamp: One way to aggregate these changes would be to have this method just update the monitor state, rather than firing the event.

src/engine/runtime.js
```diff
@@ -650,14 +686,32 @@ class Runtime extends EventEmitter {
         this._pushMonitors();
         const doneThreads = this.sequencer.stepThreads();
         this._updateGlows(doneThreads);
-        this._setThreadCount(this.threads.length + doneThreads.length);
+        // Add done threads so that even if a thread finishes within 1 frame, the green
+        // flag will still indicate that a script ran.
+        this._maybeEmitProjectRunStartOrStop(
+            this.threads.length - this._getMonitorThreadCount(this.threads) +
+                doneThreads.length - this._getMonitorThreadCount(doneThreads));
```
rschamp: Not a big deal but this could just do one function call with `this.threads.length + doneThreads.length - this._getMonitorThreadCount([...this.threads, ...doneThreads])`.

rschamp (changes requested):

src/engine/runtime.js
fsih: Renamed. Nothing else was using _threadCount so I replaced it with _nonMonitorThreadCount which is specific to this function. Do you think we should still keep track of _threadCount?

src/engine/runtime.js
fsih: Done

src/engine/runtime.js
rschamp: No, I don't think we need to.

src/engine/blocks.js
```diff
@@ -270,11 +270,13 @@ class Blocks {
...
      * Block management: change block field values
      * @param {!object} args Blockly change event to be processed
+     * @param {?Runtime} optRuntime Optional runtime to allow changeBlock to emit actions.
      */
-    changeBlock (args) {
+    changeBlock (args, optRuntime) {
         // Validate
         if (['field', 'mutation', 'checkbox'].indexOf(args.element) === -1) return;
         const block = this._blocks[args.id];
+        let wasMonitored = block.isMonitored;
```
CosmicWebServices: not ok Cannot read property 'isMonitored' of undefined - Travis

rschamp: Thanks @CosmicWebServices for noting the lines that fail in Travis. But you don't need to comment because we get emails whenever a build fails. GitHub also doesn't let us merge the PR if the Travis build failed.

src/engine/runtime.js
```diff
@@ -234,6 +239,30 @@ class Runtime extends EventEmitter {
...
+        return 'MONITORS_REMOVED';
+    }
+
+    /**
+     * Event name for monitors added.
+     * @const {string}
+     */
+    static get MONITORS_ADDED () {
+        return 'MONITORS_ADDED';
+    }
```
rschamp: We no longer need `MONITORS_ADDED` or `MONITORS_REMOVED`, please remove those.

src/virtual-machine.js
```diff
@@ -55,6 +55,15 @@ class VirtualMachine extends EventEmitter {
...
         });
+        instance.runtime.on(Runtime.MONITORS_UPDATE, monitorList => {
+            instance.emit(Runtime.MONITORS_UPDATE, monitorList);
+        });
+        instance.runtime.on(Runtime.MONITORS_REMOVED, monitorList => {
+            instance.emit(Runtime.MONITORS_REMOVED, monitorList);
+        });
+        instance.runtime.on(Runtime.MONITORS_ADDED, monitorList => {
+            instance.emit(Runtime.MONITORS_ADDED, monitorList);
+        });
```
rschamp: Same comment here re: `MONITORS_ADDED/REMOVED`.

rschamp (changes requested):
Looks good other than the leftover code!

src/engine/runtime.js
```diff
@@ -105,11 +105,10 @@ class Runtime extends EventEmitter {
         this._cloneCounter = 0;
 
         /**
-         * Flag to emit a targets update at the end of a step. When target data
-         * changes, this flag is set to true.
-         * @type {boolean}
+         * List of all monitors.
          */
-        this._refreshTargets = false;
```
rschamp: Looks like a bad merge here — this removes a part of https://github.com/LLK/scratch-vm/pull/563. Please leave this in (along with the docstring).

src/engine/runtime.js
```diff
@@ -658,12 +679,33 @@ class Runtime extends EventEmitter {
...
+        // Add done threads so that even if a thread finishes within 1 frame, the green
+        // flag will still indicate that a script ran.
+        this._emitProjectRunStatus(
+            this.threads.length + doneThreads.length -
+                this._getMonitorThreadCount([...this.threads, ...doneThreads]));
         if (this.renderer) {
             // @todo: Only render when this.redrawRequested or clones rendered.
             this.renderer.draw();
         }
-        if (this._refreshTargets) this.emit(Runtime.TARGETS_UPDATE);
```
rschamp: This is also needed for https://github.com/LLK/scratch-vm/pull/563, please keep this in.

rschamp (changes requested):

src/engine/runtime.js
```diff
@@ -658,12 +685,35 @@ class Runtime extends EventEmitter {
...
+        this._emitProjectRunStatus(
+            this.threads.length + doneThreads.length -
+                this._getMonitorThreadCount([...this.threads, ...doneThreads]));
         if (this.renderer) {
             // @todo: Only render when this.redrawRequested or clones rendered.
             this.renderer.draw();
         }
         if (this._refreshTargets) this.emit(Runtime.TARGETS_UPDATE);
+
+        // @todo(vm#570) only emit if monitors has changed since last time.
```
rschamp: Maybe we should think about moving to immutable objects for state, so we can get this for free everywhere we need it? https://facebook.github.io/immutable-js/

rschamp (approved):
Looks good!

PR #562, opened 2017-05-12
Fix the next thread is skipped after clone deleted

bigeyex:
This resolved scratch-gui [issue 206](https://github.com/LLK/scratch-gui/issues/206).

Problem:
When a clone is deleted, its thread is removed from the thread
list; but the index "i" in the execution for-loop will still +1,
making the next thread skipped.

Changes:
Added a flag "isKilled" on a thread;
Set the flag when a thread is removed from the thread list;
reduce the "i" counter if the thread of the current loop is removed

### Test Coverage

No test affected.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -590,6 +590,7 @@ class Runtime extends EventEmitter {
                 continue;
             }
             if (this.threads[i].target === target) {
+                this.threads[i].isKilled = true;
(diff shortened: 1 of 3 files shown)


thisandagain (approved):

thisandagain:
Thanks for the PR @bigeyex! Also thanks for being so patient on the review.

PR #563, opened 2017-05-12
Stop emitting SPRITE_INFO_REPORT

rschamp:
Instead, set the `_refreshTargets` flag, and emit a full target list at the end of the step.

This reduces the amount of work the GUI has to do to update the UI, since it can stop merging sprite data into its list of targets.

/cc @cwillisf @thisandagain @fsih

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -105,6 +105,13 @@ class Runtime extends EventEmitter {
         this._cloneCounter = 0;
 
         /**
+         * Flag to emit a targets update at the end of a step. When target data
+         * changes, this flag is set to true.
+         * @type {boolean}
+         */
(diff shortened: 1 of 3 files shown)


paulkaplan:
Does this have any performance impact for the vm ticks, since it has to emit a full json-ified targets list on every tick? Is that (on the whole) balanced by emitting only a single event per tick instead of the possibility of many events per tick?

rschamp:
@paulkaplan Yes it's outweighed by not doing the many events per tick.  From what I can see, there is not much weight to the targets update event itself since it's not stringifying that "JSON" object, it's just creating a new object with references it already has.

paulkaplan:
LG to me!

PR #567, opened 2017-05-14
Use stages rather than travis-after-all to deploy

rschamp:
_I tried this out just because the feature was available, but if we decide to merge it, we shouldn't til after Youth I/O._

Travis has released beta support for build stages, which eliminates the need for the `travis-after-all` work-around module.  With stages, we can choose to deploy only if all test jobs pass.

https://blog.travis-ci.com/2017-05-11-introducing-build-stages

Example build: https://travis-ci.org/rschamp/scratch-vm/builds/231994024

This change does a few things:
* Remove `travis-after-all` config
* Create separate jobs for linting, unit, and integration tests in Node 6 and latest
* Create a separate stage for the Github Pages and NPM releases, to be run only if all test jobs pass (and on a release branch)
* Speed up build times because running 2 sequential stages of jobs is slower than 2 alone in parallel:
  * Split up test jobs from each other and linting allows them to run in parallel
  * Run tap tests in parallel, which speeds up integration and unit tests. Integration tests now run in ~15s vs ~45s on my local machine.
  * Npm < 4 runs `prepublish` after a local `npm install`. This adds time to build jobs which will not even use the output of the build. Avoid this by using `in-publish` to only run `prepublish` if `npm publish` is running.
  * All together, the build time for building with these changes is down to ~4:30 on a branch without any cache, vs the current build time of ~4:15 on a branch with a cache. With a cache, the build takes ~2.30-2:45 depending on the job queue.

Why: This allows us to use Travis's native configuration for deploys, which makes the travis config more understandable.

Q: Should we start using this even though it's in beta?
_IMO Travis is pretty reliable even with their beta features, and since it's pretty easy to revert this wouldn't put us at much risk. We could pick out the performance improvements even if we don't want to use stages._


cwillisf (approved):
Sounds good to me. I don't mind using a beta feature from Travis. Worst case, they remove it or make a breaking change to the config API and our tests fail for probably <24 hours while we fix or config or revert this PR.

PR #568, opened 2017-05-15
Fix update emitting for new costumes

paulkaplan:
### Resolves

Fixes an issue where adding a costume from the library does not update the costume tab list. 

### Proposed Changes

~~This PR emits a targets update event directly. Previously it relied on the fact that `setCostume` eventually called `emitTargetsUpdate`, but now those operations are requested instead of emitted directly. However, this is an operation that we do want to emit directly.~~

Update: see @rschamp comment below. The real cause of this was a race condition with refreshTargets code in runtime#_step. Gifs still depict the new behavior. 

Here is the old behavior. You can see that the costume doesn't show up until you do something else.
add-costume-no-refresh

With this change, the costume shows up on its own. 
add-costume-force-refresh

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -645,7 +645,6 @@ class Runtime extends EventEmitter {
      * inactive threads after each iteration.
      */
     _step () {
-        this._refreshTargets = false;
         // Find all edge-activated hats, and add them to threads to be evaluated.
         for (const hatType in this._hats) {
             if (!this._hats.hasOwnProperty(hatType)) continue;
@@ -663,7 +662,10 @@ class Runtime extends EventEmitter {
             // @todo: Only render when this.redrawRequested or clones rendered.
             this.renderer.draw();
         }
-        if (this._refreshTargets) this.emit(Runtime.TARGETS_UPDATE);
+        if (this._refreshTargets) {
+            this.emit(Runtime.TARGETS_UPDATE);
+            this._refreshTargets = false;
+        }
     }
 
     /**


rschamp:
> Previously it relied on the fact that `setCostume` eventually called `emitTargetsUpdate`, but now those operations are requested instead of emitted directly. However, this is an operation that we do want to emit directly.

It should do the `emitTargetsUpdate` at the end of the current step, is it not doing so? Why?

On the one hand, I think it makes sense to do this directly since we probably would want the data  even if the vm isn't running (edge case). On the other, this may lead to more `targetsUpdate` events than necessary, since we're emitting it directly at almost the same time as it should be emitted by way of `setCostume`.

I think I would also argue that the VM is doing more than it should here, managing the target's costume list, rather than the target doing that itself (what if the target doesn't want to add the costume?). I think this should probably live within `RenderedTarget#addCostume`, and do the `requestTargetsUpdate` there.  If there's a case where a `requestTargetsUpdate` can be ignored, then maybe we should update that method to do the update directly if we're in one of those situations? Though I'm kind of baffled what that situation would be.

paulkaplan:
> It should do the emitTargetsUpdate at the end of the current step, is it not doing so? Why?

@rschamp I believe that doing a `requestTargetsUpdate` would not have any impact on the GUI if there are no threads waiting for the VM to step. It isn't really an edge case, it is whenever nothing is running.

rschamp:
`Runtime#_step` is responsible for initiating the threads. It runs once per interval regardless of whether there are any threads or not. This is why I'm curious why it isn't working.

I think there's a race condition between the `_refreshTargets = false` at the start of `_step` and the `_refreshTargets = true` happening at any time during the step. To make it more transactional we can move `_refreshTargets = false` to after its use and I believe this fixes the issue.
```
        if (this._refreshTargets) {
            this.emit(Runtime.TARGETS_UPDATE);
            this._refreshTargets = false;
        }
```

paulkaplan:
@rschamp Oh I see thanks for pointing that out. I didn't realize that was how it worked. I did what you suggested and confirmed it worked. Pushed a fix to this branch and updated the PR.

rschamp (approved):
Thanks @paulkaplan!

PR #569, opened 2017-05-15
Add public API for deleting costumes and sounds

paulkaplan:
This provides methods on the public VM object to delete sounds and costumes on the editing target. These will be used by the GUI instead of reaching into the internal vm data.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -401,10 +401,10 @@ class RenderedTarget extends Target {
             .slice(0, index)
             .concat(this.sprite.costumes.slice(index + 1));
 
-        if (index === originalCostumeCount - 1) {
+        if (index === this.currentCostume && index === originalCostumeCount - 1) {
             this.setCostume(index - 1);
-        } else {
-            this.setCostume(index);
+        } else if (index < this.currentCostume) {
+            this.setCostume(this.currentCostume - 1);
         }
 
         this.runtime.requestTargetsUpdate(this);
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -415,7 +415,6 @@ class RenderedTarget extends Target {
      * @param {number} index Sound index to be deleted
      */
     deleteSound (index) {
-        if (this.sprite.sounds.length === 1) return;
         this.sprite.sounds = this.sprite.sounds
             .slice(0, index)
             .concat(this.sprite.sounds.slice(index + 1));
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -390,6 +390,39 @@ class RenderedTarget extends Target {
     }
 
     /**
+     * Delete a costume by index.
+     * @param {number} index Costume index to be deleted
+     */
+    deleteCostume (index) {
+        const originalCostumeCount = this.sprite.costumes.length;
+        if (originalCostumeCount === 1) return;
+
+        this.sprite.costumes = this.sprite.costumes
+            .slice(0, index)
+            .concat(this.sprite.costumes.slice(index + 1));
+
+        if (index === originalCostumeCount - 1) {
+            this.setCostume(index - 1);
+        } else {
+            this.setCostume(index);
+        }
+
+        this.runtime.requestTargetsUpdate(this);
+    }
+
+    /**
+     * Delete a sound by index.
+     * @param {number} index Sound index to be deleted
+     */
+    deleteSound (index) {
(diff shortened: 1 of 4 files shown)


src/virtual-machine.js
```diff
@@ -284,6 +284,20 @@ class VirtualMachine extends EventEmitter {
...
+     * Delete a costume from the current editing target.
+     * @param {int} costumeIndex - the index of the costume to be removed.
+     */
+    deleteCostume (costumeIndex) {
+        this.editingTarget.sprite.costumes = this.editingTarget.sprite.costumes
+            .slice(0, costumeIndex)
+            .concat(this.editingTarget.sprite.costumes.slice(costumeIndex + 1));
+        if (costumeIndex === this.editingTarget.currentCostume) {
+            this.editingTarget.setCostume(costumeIndex - 1);
+        }
```
rschamp: Would you please add some tests? Curious about these cases:
* `costumeIndex === currentCostume === 0`: Would expect currentCostume to remain 0, but I believe it will wrap to the last costume
* `costumeIndex === 0`, `currentCostume === 2`, `costumes.length === 3`: Would expect currentCostume to decrease to maintain the same costume, but I believe it will wrap to 0
* `costumes.length === 1`: We should probably reject attempts to delete the last costume?

rschamp (changes requested):

src/virtual-machine.js
rschamp: And jumping off of my comment on https://github.com/LLK/scratch-vm/pull/568, can we move this logic to the `RenderedTarget`? The target should maintain its own asset lists.

paulkaplan: Will do. I'll just lean on scratch2 behavior for those test cases.

paulkaplan:
@rschamp I paired with @fsih on e0e8235, added tests for deleting sounds and costumes, which should address your questions. We also moved the responsibility of deleting into rendered target.

src/sprites/rendered-target.js
```diff
@@ -390,6 +390,39 @@ class RenderedTarget extends Target {
...
+
+        this.runtime.requestTargetsUpdate(this);
+    }
+
+    /**
+     * Delete a sound by index.
+     * @param {number} index Sound index to be deleted
+     */
+    deleteSound (index) {
+        if (this.sprite.sounds.length === 1) return;
```
rschamp: I think deleting the last sound would be ok — sprites don't require a sound to work.

src/sprites/rendered-target.js
```diff
@@ -390,6 +390,39 @@ class RenderedTarget extends Target {
...
+
+        this.sprite.costumes = this.sprite.costumes
+            .slice(0, index)
+            .concat(this.sprite.costumes.slice(index + 1));
+
+        if (index === originalCostumeCount - 1) {
+            this.setCostume(index - 1);
+        } else {
+            this.setCostume(index);
+        }
```
rschamp: This still seems to do an odd thing - say you have 5 costumes:
```
Costume 1
Costume 2
Costume 3
Costume 4   <-- current costume
Costume 5

Costume 1
-Costume 2- <-- deleted costume
Costume 3
Costume 4   <-- current costume
Costume 5

Behavior                           Expected
Costume 1                          Costume 1                      
Costume 3   <-- current costume    Costume 3                      
Costume 4                          Costume 4   <-- current costume
Costume 5                          Costume 5                      
```
If we allow deleting costumes other than the currently selected one, this is weird behavior — we should maintain the current costume if we delete an index < than the current costume index.

rschamp (changes requested):
Thank you for refactoring that into `RenderedTarget` and adding tests. I think we need to consider when costumes.length > 3 though.

paulkaplan:
@rschamp Added a few more test cases, used that ascii art syntax to describe them because I ran out of words. Good catch.

rschamp (changes requested):
When I delete the current costume, the change is not reflected on the renderer? It's not immediately apparent how that would have regressed with this code though (this behavior isn't present on the current build). To repro: load the default project and delete the first costume. The renderer also seems to be painting — you can use move blocks and the sprite will move still using the old deleted costume.

paulkaplan:
@rschamp good catch, the problem was that there was a branch of the `if` that didn't call `setCostume` because the current costume index doesn't change. however, the _costume_ that index was referring to changed, so I needed to add an extra `else` to the bottom to make sure `setCostume` is always called.

rschamp (approved):
Looks good!

PR #573, opened 2017-05-18
Fix sprite rename incrementing when renamed to itself

### LINKED ISSUE #370: Sprite names handle numbers incorrectly
### Expected Behavior

Entering a sprite name of "sprite_1" changes sprite name to "sprite_1".
Names can be arbitrary strings of letters, numbers, and allowed special characters. 

### Actual Behavior

Entering a sprite name of "sprite_1" changes sprite name to "sprite_2".
Entering the name "sprite_2" changes sprite name to "sprite_3".
Entering "sprite_4" changes sprite name to "sprite_2".
Entering "sprite_999" changes sprite name to "sprite_2".

The number 2 is automatically appended to any letters entered; "hello" becomes "hello2".

Every number but 2 is changed to 2, except for 2 itself, which becomes 3 when a user enters a string ending in a single 2 as a new sprite name. Entering "sprite_2" when a sprite name is already "sprite_2" has no effect. 

### Steps to Reproduce

Enter any string into the name field of a sprite, and it follows this pattern. 

Ubuntu 16.04.1, Firefox 53.0 (64-bit)
Windows 10, Firefox 53.0.2 (32-bit)

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/370

### Proposed Changes

Filter out your own sprite name when creating the list of names to consider "used" when renaming. 

### Reason for Changes

See comment on https://github.com/LLK/scratch-gui/issues/370
### Test Coverage

Added a test for this case.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -329,9 +329,8 @@ class VirtualMachine extends EventEmitter {
             }
             if (newName && RESERVED_NAMES.indexOf(newName) === -1) {
                 const names = this.runtime.targets
-                    .filter(runtimeTarget => runtimeTarget.isSprite())
+                    .filter(runtimeTarget => runtimeTarget.isSprite() && runtimeTarget.id !== target.id)
                     .map(runtimeTarget => runtimeTarget.sprite.name);
-
                 sprite.name = StringUtil.unusedName(newName, names);
             }
             this.emitTargetsUpdate();
--- a/test/unit/virtual-machine.js
+++ b/test/unit/virtual-machine.js
@@ -94,3 +94,18 @@ test('renameSprite increments from existing sprite names', t => {
     t.equal(vm.runtime.targets[0].sprite.name, 'that name2');
     t.end();
 });
+
+test('renameSprite does not increment when renaming to the same name', t => {
+    const vm = new VirtualMachine();
+    vm.emitTargetsUpdate = () => {};
+    vm.runtime.targets = [{
+        id: 'id1',
+        isSprite: () => true,
+        sprite: {
+            name: 'this name'
+        }
+    }];
+    vm.renameSprite('id1', 'this name');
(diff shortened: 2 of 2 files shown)


src/virtual-machine.js
```diff
@@ -329,9 +329,8 @@ class VirtualMachine extends EventEmitter {
             }
             if (newName && RESERVED_NAMES.indexOf(newName) === -1) {
                 const names = this.runtime.targets
-                    .filter(runtimeTarget => runtimeTarget.isSprite())
+                    .filter(runtimeTarget => runtimeTarget.isSprite() && runtimeTarget.id !== target.id)
                     .map(runtimeTarget => runtimeTarget.sprite.name);
-
                 sprite.name = StringUtil.unusedName(newName, names);
```
fsih, line 334: Another weird thing about thing about this bug was

"Every number but 2 is changed to 2, except for 2 itself, which becomes 3 when a user enters a string ending in a single 2 as a new sprite name."

Not needed for this fix, but is it weird that when you try to name two sprites TurboBlaster9000, the second one gets renamed to TurboBlaster2? Maybe it would make more sense to go to TurboBlaster9001 or TurboBlaster9000_2?

src/virtual-machine.js
```diff
@@ -329,9 +329,8 @@ class VirtualMachine extends EventEmitter {
             }
             if (newName && RESERVED_NAMES.indexOf(newName) === -1) {
```
fsih, line 330: Should we find where this is firing twice and remove the second time?

paulkaplan: It is coming from the GUI, so wouldn't be part of this fix.

src/virtual-machine.js
paulkaplan: Yeah @rschamp and I discussed that. Since it is currently the way scratch2 works, we decided to leave it that way for now.

fsih (approved):

Issue #570, opened 2017-05-15
Only emit MONITORS_UPDATE action if the state has changed since last step


fsih:
This shouldn't have visible side effects, but will be more efficient.

Relevant code: https://github.com/LLK/scratch-vm/pull/560/files#diff-38d875e1d50b3158dc96404b65275a54R691

PR #574, opened 2017-05-19
Only send update monitor events when changes happen

fsih:
### Resolves
https://github.com/LLK/scratch-vm/issues/570

### Proposed Changes
Changes the monitor state object to an Immutable Map object. Each time we update it, check if anything changed, and only fire the update monitors action if it has.

### Reason for Changes
Sending update monitors every frame is unnecessary and inefficient.
Switching to the immutable libraries makes some code cleaner and is supposed to be less error-prone.

### Test Coverage
Added tests for applying monitor state changes confirming that the state gets overridden correctly and isEquals returns correctly, since isEquals is used to determine if we need to emit an action.

/cc @paulkaplan


fsih:
<strike>My performance logging says it saves .7ms/tick when the monitors are on but not updating</strike>

I was comparing it to the monitor update PR before it was hooked up to the gui. The gui update actually added 3 or 4 ms and this PR cuts that back off.

rschamp:
> Added tests for the monitorWouldChange function which checks if an update changes the object. We could avoid needing this function though if we had nested immutables

I think we should use nested immutables, since without them we don't get the benefit of using immutables. The `_monitorWouldChange` method is not very much different than a flag to track whether a change has occurred or not.  Other than the complexity of the PR, is there a reason you decided not to move the monitor state to a Map as well?

fsih:
That sounds good. I wasn't sure how far to go with switching things over, if we should expect to start replacing every object/array we pass around with immutables or if that would be considered too heavy weight

rschamp:
I think this is a nice test case to see how it works out.  I think it would help with performance on the React side to use it in all state, but that would be a lot all at once.

src/engine/runtime.js
```diff
@@ -699,10 +705,11 @@ class Runtime extends EventEmitter {
             this._refreshTargets = false;
         }
 
-        // @todo(vm#570) only emit if monitors has changed since last time.
-        this.emit(Runtime.MONITORS_UPDATE,
-            Object.keys(this._monitorState).map(key => this._monitorState[key])
-        );
+        if (!this._prevMonitorState.equals(this._monitorState)) {
+            const monitorStateObj = this._monitorState.toJS();
+            this.emit(Runtime.MONITORS_UPDATE, Object.keys(monitorStateObj).map(key => monitorStateObj[key]));
```
rschamp: I think you could simplify this with `this._monitorState.map(value => value).toJS()`. But I think we should think about passing the immutable object over to the GUI, since `toJS` incurs a performance penalty, so we should avoid it if possible.  If we need to cast to JS on the GUI side then we can do it there, but if we don't need to, all the better.

fsih: I've changed it and added https://github.com/LLK/scratch-gui/pull/381

It works, but it gives me a weird warning in the console:
Warning: Failed prop type: Invalid prop `monitors` of type `Map` supplied to `MonitorList`, expected instance of `Map`.
    in MonitorList (created by MonitorList)

Also I had to convert to JS in the GUI since it gave me a warning that having Maps as UI elements was a temporary feature that might go away.

rschamp (approved):
Thanks for investigating using immutable data! LGTM.

paulkaplan:
@rschamp @fsih do you want to hold off on merging this until we can get together and do a dev-talk type of thing to explore the pros/cons of immutable?

fsih:
Per discussion with @rschamp and @paulkaplan, switched to using Record instead of map for the state of each monitor, and OrderedMap instead of Map for the list of all monitors (to prevent flickering as the monitors fight each other for z position)

My concern with using Record is that default values have to be present, and most of the values that are required for monitors are null by default.

fsih:
@paulkaplan can you take another look, even though it's marked good to merge?

paulkaplan (approved):
LG!

fsih:
@paulkaplan please check the merge

paulkaplan:
Looks like travis is failing, can you check into that?

fsih:
@paulkaplan should be fixed

PR #575, opened 2017-05-21
Fix SB2 Imports

TheBrokenRail:
### Resolves

_Fixes #501_

### Proposed Changes

_Changes the SB2 Spec Map so that it is up to date on what things are fields and what things are shadows_

### Reason for Changes

_To Fix SB2 Imports_


thisandagain:
Thanks @TheBrokenRail ! This looks good.

thisandagain (approved):

PR #579, opened 2017-05-24
Specify dataFormat when loading asset for import

cwillisf:
### Resolves

This is follow-up work on LLK/scratch-storage#7 and will not work correctly until that PR is merged.

### Proposed Changes

When importing a project we know the file extension for each asset to be loaded. This change provides that information to the storage system so that we can load assets which don't use the default.

In support of this change, there's a new function on `StringUtil` called `splitFirst`, which splits a string on the first instance of a separator character.

### Reason for Changes

Without these changes, assets with non-default file types fail to load. This includes JPEG format backdrops, for example.

### Test Coverage

This change includes unit tests for the new `StringUtil.splitFirst` function.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -1,3 +1,4 @@
+const StringUtil = require('../util/string-util');
 const log = require('../util/log');
 
 /**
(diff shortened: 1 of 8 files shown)


src/import/load-costume.js
```diff
@@ -19,17 +20,17 @@ const loadCostume = function (md5ext, costume, runtime) {
     }
 
     const AssetType = runtime.storage.AssetType;
-    const idParts = md5ext.split('.');
+    const idParts = StringUtil.splitFirst(md5ext, '.');
     const md5 = idParts[0];
-    const ext = idParts[1].toUpperCase();
-    const assetType = (ext === 'SVG') ? AssetType.ImageVector : AssetType.ImageBitmap;
+    const ext = idParts[1].toLowerCase();
+    const assetType = (ext === 'svg') ? AssetType.ImageVector : AssetType.ImageBitmap;
```
rschamp, line 26: We could simplify a little,
```javascript
    const [md5, ext] = StringUtil.splitFirst(md5ext.toLowerCase(), '.');
```

rschamp (approved):
Looks good!

PR #581, opened 2017-05-25
Add variable creation and variable serialization

### LINKED ISSUE #350: Variables don't persist
### Expected Behavior

After creating a global variable, it should be available for all sprites.

### Actual Behavior

The variable disappears after switching to another sprite.

### Steps to Reproduce

1. Go to https://llk.github.io/scratch-gui/
2. Add a new sprite, for the "A"
3. Switch to the cat, and add a new variable.
4. Switch to the "A" - the variable is gone.
5. Switch back to the cat - the variable is still gone.

**Note that this is not broken on http://llk.github.io/scratch-vm/**

paulkaplan:
Explicitly create global variables (on the stage) and serialize variables when updating blockly. Depends on variable serialization from https://github.com/LLK/scratch-blocks/pull/903.

Will fix https://github.com/LLK/scratch-gui/issues/350

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -427,11 +427,7 @@ class Blocks {
      * @return {string} String of XML representing this object's blocks.
      */
     toXML () {
-        let xmlString = '<xml xmlns="http://www.w3.org/1999/xhtml">';
-        for (let i = 0; i < this._scripts.length; i++) {
-            xmlString += this.blockToXML(this._scripts[i]);
-        }
-        return `${xmlString}</xml>`;
+        return this._scripts.map(script => this.blockToXML(script)).join();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


rschamp (commented):
Does this interact with the existing behavior where variables are created as the blocks that use them are run?

paulkaplan:
It doesn't change that behavior, it just creates a way for them to be created explicitly. My thinking was that we could fully separate the lookup vs. create behavior at a later time when we implement scoped variables. 

One reason not we can't separate the lookup/create behavior at this point is that the create action is invoked directly by the GUI, not wired to scratch-blocks in a way that would allow it to work independently. We might want to reconsider that, but again I was thinking we have some breathing room on that.

paulkaplan:
(woops didn't mean to close/comment, just meant to comment!)

rschamp (approved):
Sounds good!

Issue #582, opened 2017-05-26
SB3 serialization exports clones


rschamp:
### Expected Behavior

_Please describe what should happen_
SB3s should not contain clones.

### Actual Behavior

_Describe what actually happens_
They do.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Save an SB3 of the default project.
Make a clone and save an SB3 of the default project.
Observe the clone saved in the second SB3 that isn't in the first SB3.

PR #583, opened 2017-05-26
Don't serialize clones when saving

rschamp:
Fixes https://github.com/LLK/scratch-vm/issues/582


paulkaplan:
LG. Travis failed, not sure why, restarted the job.

PR #584, opened 2017-05-26
Fix no-useless-call todo

rschamp:
The linter is correct that `devObject[func].call(devObject, args)` is unnecessary because it's equivalent to `devObject[func](args)`.

@tmickel probably either meant to allow passing an array of arguments to be applied, or to call the function with the provided argument. Since we probably want to be able to use multi-argument functions, use `apply` and fix the one place that uses `ioQuery` with an argument.


tmickel:
your guess is as good as mine... who knows what was happening there ;)

cwillisf (approved):
LGTM

PR #587, opened 2017-06-01
Pin to `latest` for 3.0 modules

rschamp:
Reverts #551

Pinning to `^0.1.0.prerelease.0` is not equivalent to `latest`, since has the effect of not installing the latest version on `npm install`, you have to also do `npm update`. This hurts us ergonomically on `npm install`, and doesn't help us since we already are releasing breaking changes on `prerelease`.


cwillisf:
Does `^0.1.0-prerelease.0` not even install the latest `0.1.0-prerelease.*` on `npm install`? That's disappointing...

My hope was that we'd soon transition to real version numbers for everything so we could stay stable-ish without worrying about the order of merges on multi-repository changes -- something that `latest` makes difficult. Do you think it's still too early to do that?

rschamp:
> Does `^0.1.0-prerelease.0` not even install the latest `0.1.0-prerelease.*` on npm install?

Only if you have not installed the package at all yet. Otherwise any installed `0.1.0-prerelease.*` version satisfies the dependency, so it's not updated.

I think we should transition to real version numbers given the number of people who are both working on and using these repos now. But until we have semantic release set up on all of the packages, I think `latest` is the most ergonomic way to handle these dependencies.  We should talk over how to use semantic release on all of these, since I would hope outside contributors wouldn't have to worry about that.

cwillisf (approved):

Issue #591, opened 2017-06-05
"Pick random" value as input to "wait" is ignored 


ericrosenbaum:
### Expected Behavior

The "pick random" block should output a number that the "wait" block can use as a number of seconds. 

### Actual Behavior

The value from the "pick random" block seems to be ignored, and the "wait" block always waits one second.

### Steps to Reproduce

Create and run this stack:

<img width="286" alt="screen shot 2017-06-05 at 2 34 19 pm" src="https://cloud.githubusercontent.com/assets/567844/26797914/9f46824a-49fc-11e7-9a5f-f46c3d689117.png">

It should play the "pop" sound at a random interval of 1 to 5 seconds, but it plays every second.

### Operating System and Browser

Mac OS 10.11.6 Chrome 58.0

thisandagain:
This also feels like a blocker for the upcoming play test.

ericrosenbaum:
I have a clue about this. Logging the value of the input (args.DURATION) in the wait block implementation [here](https://github.com/LLK/scratch-vm/blob/ecf535f4c3d4f5820ea969bb23a24ff4f6f4770b/src/blocks/scratch3_control.js#L76-L88), shows it running dozens of times, once per frame, with a different random value each time. 

So, it seems to be re-evaluating the inputs every frame as it waits. So I suppose the input needs to be evaluated just the first time, and stored.

The wait block uses util.stackFrame.timer, but for other blocks that wait we've used a Promise resolved by a setTimeout, which is simpler and wouldn't have this problem. But maybe there's something different about the way the stackFrame timer works which we need here?

ericrosenbaum:
The "rest" block (in the sounds category) is implemented using a Promise and setTimeout, which you can see [here](https://github.com/LLK/scratch-audio/blob/19e1136e3b3a44a041f0f89eea2d2a19ec238175/src/index.js#L289-L301). This is equivalent to the example shown above, but behaves correctly:

<img width="296" alt="screen shot 2017-06-05 at 3 09 02 pm" src="https://cloud.githubusercontent.com/assets/567844/26799117/3d6dd248-4a02-11e7-8914-2a21a3e6eb12.png">

ericrosenbaum:
[...]

PR #593, opened 2017-06-05
Change wait to use a promise instead of being called every frame

fsih:
### Resolves
https://github.com/LLK/scratch-vm/issues/591
### Proposed Changes

Change wait block to work like wait beats block, use a promise which makes it not be called every frame

### Reason for Changes

Calling the wait function every frame with different random numbers generally causes it to finish after the smallest random number


tmickel (approved):
seems reasonable to me! :p

fsih:
Thanks for helping us look at this bug @tmickel!

PR #594, opened 2017-06-07
Select the last target instead of the first after deleting

### LINKED ISSUE #294: Deleting a sprite selects the stage
### Expected Behavior

Deleting a sprite should select the next sprite in the sprite list. The stage should only become selected after deleting the last sprite in the sprite list.

### Actual Behavior

Deleting any sprite with in the sprite list results in the stage becoming selected.
deleting-sprite-stage-selected2

### Steps to Reproduce

1. Add some sprites
2. Delete any sprite
3. Stage becomes selected

Browser: Chrome 57.0.2987.133
OS: macOS 10.11.6

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/294
### Proposed Changes

Prefer selecting the last target instead of the first (the stage). This way it will always try to stay selected on a target. 

### Reason for Changes

https://github.com/LLK/scratch-gui/issues/294

--- 

@rschamp sorry I know this was assigned to you, but I figured it was just a one liner. Is this what you had in mind?

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -379,7 +379,8 @@ class VirtualMachine extends EventEmitter {
                 this.runtime.disposeTarget(sprite.clones[i]);
                 // Ensure editing target is switched if we are deleting it.
                 if (clone === currentEditingTarget) {
-                    this.setEditingTarget(this.runtime.targets[0].id);
+                    const lastTargetIndex = this.runtime.targets.length - 1;
+                    this.setEditingTarget(this.runtime.targets[lastTargetIndex].id);
                 }
             }
             // Sprite object should be deleted by GC.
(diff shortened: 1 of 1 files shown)


fsih (approved):

rschamp (changes requested):
I would hope to match Scratch 2.0 behavior, which I think will end up being a bit more than a one-liner.  That behavior is to select the sprite before the deleted sprite, or the first sprite if you deleted the first sprite.

Happy to work on this unless you want to revise what you have to behave that way.

paulkaplan:
@rschamp addressed the comments in 9c5d43e

src/virtual-machine.js
```diff
@@ -364,6 +364,8 @@ class VirtualMachine extends EventEmitter {
      */
     deleteSprite (targetId) {
         const target = this.runtime.getTargetById(targetId);
+        const targetIndexBeforeDelete = this.runtime.targets.map(t => t.id).indexOf(target.id);
```
rschamp, line 367: Is this different from `this.runtime.targets.indexOf(target)`?

PR #598, opened 2017-06-08
Improve cleanup on clone disposal

### LINKED ISSUE #78: TypeError - Cannot read property 'getFastBounds' of undefined

cwillisf:
### Resolves

Resolves LLK/scratch-render#78

### Proposed Changes

The `dispose()` method on `RenderedTarget` now:
- informs the runtime that it should end any threads corresponding the target being disposed, and
- removes the clone from its sprite.

### Reason for Changes

The problem described by LLK/scratch-render#78 is due to threads continuing to run on a target (clone) which has been disposed. This manifests as an exception within the renderer when a block attempts to adjust properties of the `Drawable` object responsible for the visual representation of the clone: that `Drawable` no longer exists because it was cleaned up at `dispose()` time.

Removing the clone from its sprite's clone list is just a matter of fixing a leak that I noticed while I was in the area.

### Test Coverage

I added a new integration test called "clone-cleanup" -- this new test verifies that clones and their associated threads are cleaned up properly by the `delete this clone` block. Several items in the test fail prior to the changes in this PR.

The clones run two stacks each: one which waits and then deletes the clone, and another which includes a `forever` loop: this is to verify that the thread running the `forever` loop is ended when the clone itself is deleted. The project does this with two batches of clones to ensure there are no problems with reusing array indices previously occupied by now-removed threads or clones.

Here's the code for `clone-cleanup.sb2`:
image


rschamp (approved):
I think this makes sense for how we currently have things set up.

I have a couple concerns, though I don't think they're worth addressing in this PR:

The test for this is clever! But it's hard to understand if you don't know what the contents of the project are. Existing integration tests we have take more of a "black box" approach where they just run the project, and test that nothing went wrong, rather than testing specific things in the project. I wonder if there is some other way to write tests like the one here which wouldn't depend on an `sb2` file?  Something like a functional test with more explicit set-up in the test?

The logic for clone disposal seems inverted. The `RenderedTarget` is responsible for calling methods on its owners, `sprite` and `runtime`. I think ideally the `runtime` or `sprite` would be responsible for as much of this logic as possible.

cwillisf:
> I wonder if there is some other way to write tests like the one here which wouldn't depend on an sb2 file

I think that's a good idea in principle, but wouldn't it be tricky to inject all the necessary blocks? I guess we could have JSON snippets in the test itself, and that could be a step in the right direction.

> The logic for clone disposal seems inverted.

Yeah, I felt the same way. A large part of the problem is that `dispose()` is called on a particular `Target` (which might or might not be a `RenderedTarget`) by the `Runtime` -- and the `Runtime` doesn't know about `Sprite`. Even so, it smells like this particular corner is ripe for a small refactor... it's just not trivial enough for me to feel comfortable rolling it into this PR.

PR #602, opened 2017-06-12
Remember data format after image or sound import

cwillisf:
### Resolves

Resolves LLK/scratch-gui#407

### Proposed Changes

When importing an asset, record its actual data format (file type / extension) instead of the `AssetType` (runtime data type).

### Reason for Changes

We need to know if an asset is (for example) a PNG or JPG when attempting to load it, but the `AssetType` only expresses the default.

### Test Coverage

This PR includes a new test which verifies that asset details survive a round trip through SB3 serialization & deserialization. In support of this new test, and to correct an existing misuse, the `attachTestStorage` function has been replaced with `makeTestStorage`, which now simply returns a prepared storage instance instead of automatically attaching it to a VM instance.

### DIFF
--- a/test/integration/import_sb2.js
+++ b/test/integration/import_sb2.js
@@ -1,6 +1,6 @@
 const path = require('path');
 const test = require('tap').test;
-const attachTestStorage = require('../fixtures/attach-test-storage');
+const makeTestStorage = require('../fixtures/make-test-storage');
 const extract = require('../fixtures/extract');
 
 const renderedTarget = require('../../src/sprites/rendered-target');
@@ -20,7 +20,7 @@ test('default', t => {
 
     // Create runtime instance & load SB2 into it
     const rt = new runtime();
-    attachTestStorage(rt);
+    rt.attachStorage(makeTestStorage());
     sb2.deserialize(json, rt).then(targets => {
         // Test
         t.type(file, 'string');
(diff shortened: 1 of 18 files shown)


test/integration/import_sb2.js
```diff
@@ -20,7 +20,7 @@ test('default', t => {
 
     // Create runtime instance & load SB2 into it
     const rt = new runtime();
-    attachTestStorage(rt);
```
cwillisf, line 23: Strictly speaking, this call is incorrect: `attachTestStorage` expects a VM, not a runtime. It was working anyway because both `VirtualMachine` and `Runtime` have a method called `attachStorage` and that's all `attachTestStorage` was calling, but IMO moving the attachment out is safer and simpler than making `attachTestStorage` officially accept both types.

test/integration/sb3-roundtrip.js
```diff
@@ -0,0 +1,110 @@
...
+        runtime1.targets = [stageClone, spriteClone];
+
+        testRuntimeState('original', runtime1);
+    });
+
+    const serializeAndDeserialize = installThings.then(() => {
+        // `Clone.simple` here helps in two ways:
+        // 1. it ensures that any non-serializable data is thrown away, and
+        // 2. `sb3.deserialize` does some `hasOwnProperty` checks which fail on the object returned by `sb3.serialize`
+        //    due to its inheritance structure.
```
cwillisf: It took me far longer than it should have to figure out that this was a problem...

rschamp: What inheritance structure specifically caused this issue?

rschamp (approved):
The backwards compatibility works! Looks great, thanks!

PR #611, opened 2017-06-15
Add ability to refresh workspace for current editing target

fsih:
### Resolves

Workaround for https://github.com/LLK/scratch-blocks/issues/929

### Proposed Changes

Refresh the workspace similar to setEditingTarget, but without changing the target

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -470,6 +470,17 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Repopulate the workspace with the blocks of the current editingTarget. This
+     * allows us to get around bugs like gui#413.
+     */
+    refreshWorkspace () {
+        if (this.editingTarget) {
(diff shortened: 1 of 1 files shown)


src/virtual-machine.js
```diff
@@ -470,6 +470,17 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Repopulate the workspace with the blocks of the current editingTarget. This
+     * allows us to get around bugs like gui#413.
+     */
+    refreshWorkspace () {
+        if (this.editingTarget) {
+            this.emitWorkspaceUpdate();
+            this.runtime.setEditingTarget(this.editingTarget);
```
rschamp, line 479: Why is setting the editing target necessary?

fsih: runtime.setEditingTarget is responsible for checking which threads are running and putting the glows back on the right blocks

rschamp (approved):

PR #612, opened 2017-06-15
Fix setting Infinity to direction

TheBrokenRail:
### Resolves

_Fixes LLK/scratch-gui#455_

### Proposed Changes

_Makes sure a number is finite when setting direction_

### Reason for Changes

_So you can't set a infinite direction_

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -228,9 +228,6 @@ class RenderedTarget extends Target {
         if (this.isStage) {
             return;
         }
-        if (!isFinite(direction)) {
-            return;
-        }
         // Keep direction between -179 and +180.
         this.direction = MathUtil.wrapClamp(direction, -179, 180);
         if (this.renderer) {
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -228,6 +228,9 @@ class RenderedTarget extends Target {
         if (this.isStage) {
             return;
         }
+        if (!isFinite(direction)) {
+            return;
+        }
         // Keep direction between -179 and +180.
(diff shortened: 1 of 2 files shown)


src/sprites/rendered-target.js
```diff
@@ -228,6 +228,9 @@ class RenderedTarget extends Target {
         if (this.isStage) {
             return;
         }
+        if (!isFinite(direction)) {
+            return;
+        }
```
rschamp, line 233: I think this might be better handled by the `wrapClamp` utility, since we will probably have other cases which will need to handle `Infinity`.  Would you mind moving this into `wrapClamp`?

rschamp (changes requested):
Thanks for the pull request @TheBrokenRail!

src/sprites/rendered-target.js
TheBrokenRail: Fixed

src/util/math-util.js
```diff
@@ -41,6 +41,9 @@ class MathUtil {
      */
     static wrapClamp (n, min, max) {
         const range = (max - min) + 1;
+        if (!isFinite(n)) {
+            return min;
+        }
```
rschamp: I'm sorry @TheBrokenRail, I didn't really think through what I was asking you to "just move to `wrapClamp`". I think the behavior you had before actually was more appropriate (leave the direction unchanged).  Would you please change it back to what you had? I think that makes the most sense for now.  Thanks for the quick update!

TheBrokenRail: Done!

rschamp (approved):

rschamp:
Thanks @TheBrokenRail!

Issue #586, opened 2017-05-31
Variable rename and delete does not work


paulkaplan:
Scratch-blocks has it's own rename/update management but it isn't getting sent to the vm. so creating and renaming variables does not work correctly. This will depend on scratch-blocks variable create/update/destroy events.

PR #614, opened 2017-06-16
Updated variables and var_fire event listener.

marisaleung:
### Resolves

https://github.com/LLK/scratch-vm/issues/586

### Proposed Changes

Adds a listener for variable events.
Variable models now include ID's.

### Reason for Changes

We want to be able to delete variables and rename them.

### Test Coverage

---
This change is [Reviewable](https://reviewable.io/reviews/llk/scratch-vm/614)

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -181,16 +181,19 @@ class Blocks {
     // ---------------------------------------------------------------------
 
     /**
-     * Create event listener for blocks. Handles validation and serves as a generic
-     * adapter between the blocks and the runtime interface.
-     * @param {Object} e Blockly "block" event
+     * Create event listener for blocks and variables. Handles validation and
+     * serves as a generic adapter between the blocks, variables, and the
+     * runtime interface.
+     * @param {object} e Blockly "block" or "variable" event
      * @param {?Runtime} optRuntime Optional runtime to forward click events to.
      */
-
     blocklyListen (e, optRuntime) {
         // Validate event
         if (typeof e !== 'object') return;
-        if (typeof e.blockId !== 'string') return;
+        if (typeof e.blockId !== 'string' && typeof e.varId !== 'string') {
+            return;
+        }
+        const stage = optRuntime.getTargetForStage();
 
         // UI event: clicked scripts toggle in the runtime.
         if (e.element === 'stackclick') {
@@ -242,6 +245,15 @@ class Blocks {
             this.deleteBlock({
                 id: e.blockId
             });
+            break;
+        case 'var_create':
+            stage.createVariable(e.varId, e.varName);
+            break;
+        case 'var_rename':
+            stage.renameVariable(e.varId, e.newName);
+            break;
+        case 'var_delete':
+            stage.deleteVariable(e.varId);
             break;
         }
     }
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -72,24 +72,25 @@ class Target extends EventEmitter {
     /**
      * Look up a variable object, and create it if one doesn't exist.
      * Search begins for local variables; then look for globals.
-     * @param {!string} name Name of the variable.
+     * @param {string} id Id of the variable.
+     * @param {string} name Name of the variable.
      * @return {!Variable} Variable object.
      */
-    lookupOrCreateVariable (name) {
+    lookupOrCreateVariable (id, name) {
         // If we have a local copy, return it.
-        if (this.variables.hasOwnProperty(name)) {
-            return this.variables[name];
+        if (this.variables.hasOwnProperty(id)) {
+            return this.variables[id];
         }
         // If the stage has a global copy, return it.
         if (this.runtime && !this.isStage) {
             const stage = this.runtime.getTargetForStage();
-            if (stage.variables.hasOwnProperty(name)) {
-                return stage.variables[name];
+            if (stage.variables.hasOwnProperty(id)) {
+                return stage.variables[id];
             }
         }
         // No variable with this name exists - create it locally.
-        const newVariable = new Variable(name, 0, false);
-        this.variables[name] = newVariable;
+        const newVariable = new Variable(id, name, 0, false);
+        this.variables[id] = newVariable;
         return newVariable;
     }
 
@@ -115,6 +116,44 @@ class Target extends EventEmitter {
         const newList = new List(name, []);
         this.lists[name] = newList;
         return newList;
+    }
+
+    /**
+     * Creates a variable with the given id and name and adds it to the
+     * dictionary of variables.
+     * @param {string} id Id of variable
+     * @param {string} name Name of variable.
+     */
+    createVariable (id, name) {
+        if (!this.variables.hasOwnProperty(id)) {
+            const newVariable = new Variable(id, name, 0,
+                false);
+            this.variables[id] = newVariable;
+        }
+    }
+
+    /**
+     * Renames the variable with the given id to newName.
+     * @param {string} id Id of renamed variable.
+     * @param {string} newName New name for the variable.
+     */
+    renameVariable (id, newName) {
+        if (this.variables.hasOwnProperty(id)) {
+            const variable = this.variables[id];
+            if (variable.id === id) {
+                variable.name = newName;
+            }
+        }
+    }
+
+    /**
+     * Removes the variable with the given id from the dictionary of variables.
+     * @param {string} id Id of renamed variable.
+     */
+    deleteVariable (id) {
+        if (this.variables.hasOwnProperty(id)) {
+            delete this.variables[id];
+        }
     }
 
     /**
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -191,11 +191,13 @@ const parseScratchObject = function (object, runtime, topLevel) {
     if (object.hasOwnProperty('variables')) {
         for (let j = 0; j < object.variables.length; j++) {
             const variable = object.variables[j];
-            target.variables[variable.name] = new Variable(
+            const newVariable = new Variable(
+                null,
                 variable.name,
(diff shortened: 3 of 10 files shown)


marisaleung:
@rschamp @paulkaplan @rachel-fenichel @RoboErikG

marisaleung:
There is a bug with the checkbox listener. It keeps checking for the variable by name, not finding it (because the variables are mapped by id now), and then creating an extra variable by the name of 'undefined.'

We can fix this by making the checkbox listener call "getVariable" with id and not name. @rschamp can you help me trace how the listener works?

src/engine/runtime.js
```diff
@@ -1012,6 +1013,51 @@ class Runtime extends EventEmitter {
             this._step();
         }, interval);
     }
+
+    // ---------------------------------------------------------------------
+
+    /**
+     * Create event listener for variables.
+     * @param {Object} e Blockly "var" event
```
RoboErikG: via travis test, use 'object' instead of 'Object'

src/engine/runtime.js
```diff
@@ -1012,6 +1013,51 @@ class Runtime extends EventEmitter {
             this._step();
         }, interval);
     }
+
+    // ---------------------------------------------------------------------
```
RoboErikG: What's this line and the one below for?

src/virtual-machine.js
```diff
@@ -447,6 +448,14 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Handle a Blockly event for the flyout.
+     * @param {!Blockly.Event} e Any Blockly event.
+     */
+    variableListener (e) {
+        this.runtime.blocklyListen(e);
```
RoboErikG: You may want to filter for only variable events similar to monitorBlockListener() above.

RoboErikG:
And here's the checkbox event. https://github.com/LLK/scratch-blocks/blob/35c48ab72e9ce0902ab93f7d1c022e1f7b71addf/core/flyout_vertical.js#L578

And the listener in the VM https://github.com/LLK/scratch-vm/blob/d3d82bae69aa8ea6bc80560d953180f58f853e47/src/engine/blocks.js#L298

It looks like it's doing it by the block, not the variable, so keep digging down on the vm side until you find where it's looking up the variable.

src/engine/runtime.js
marisaleung: copy pasta from:
L81
https://github.com/LLK/scratch-vm/blob/develop/src/engine/blocks.js

Should I not follow this style?

src/engine/runtime.js
marisaleung: Done

src/virtual-machine.js
marisaleung: Done

src/engine/runtime.js
```diff
@@ -1012,6 +1013,51 @@ class Runtime extends EventEmitter {
...
+            {
+                const id = e.varId;
+                if (stage.variables.hasOwnProperty(id)) {
+                    delete stage.variables[id];
+                }
+                break;
+            }
+        }
+    }
+    // ---------------------------------------------------------------------
```
rschamp: I am wondering about this listener being in Runtime. I think it might be better placed in the main [`Blocks#blocklyListen`](https://github.com/LLK/scratch-vm/blob/b9aefa780f40f2bdc1c44dca34683aadf044cc6e/src/engine/blocks.js#L190), for a few reasons:
* The VM may not be running in a context with scratch-blocks, so it would be nice to keep the scratch-blocks-specific code contained. I think it would help us later on if we avoided spreading it into multiple parts of the VM. Eventually we may refactor all of it out of the VM and into an adapter, so having it self-contained would make that easier to do.
* I believe it is possible to keep this contained in the main blocks listener: Although `optRuntime` is technically optional, in practice it's always available for the events coming from scratch-blocks. So we should have access to get the stage target (or access other methods on the runtime if we reorganize variables later).
* Currently all variables are global, so we are keeping the list of variables on the stage only. When we have scoped variables, we will want to store them by target. The current setup of piping events from scratch-blocks to the current editing target's block container would help with that, since if the variable is scoped, then we would attach it to whatever the listener's target is, otherwise the stage.

/cc @paulkaplan @fsih @cwillisf @thisandagain

rschamp: For testability and organization, I also think we should push the logic for manipulating the variable list into methods on `Target`, and then call them from the listener rather than having it inline in the listener.

marisaleung:
The variable's bug has been fixed by making the monitor search for the variable by id. It is added to this CL now. Thanks for the starting points @RoboErikG.

src/engine/runtime.js
paulkaplan: @rschamp those are good points. @marisaleung reached out to me about this and I told her the opposite! (sorry!) simply based on the naming and the way we use block listeners for the monitor blocks also, but i think your points are all very good. @marisaleung sorry for leading you down the wrong path on that one!

paulkaplan: @rschamp so to be explicit for @marisaleung, you are saying we should implement a `Target#addVariable`, `Target#deleteVariable` and `Target#renameVariable` right?

src/serialization/sb2.js
```diff
@@ -191,11 +191,13 @@ const parseScratchObject = function (object, runtime, topLevel) {
     if (object.hasOwnProperty('variables')) {
         for (let j = 0; j < object.variables.length; j++) {
             const variable = object.variables[j];
-            target.variables[variable.name] = new Variable(
+            const newVariable = new Variable(
+                variable.id,
```
paulkaplan: I do not think a variable `id` is available from the sb2 file format.

src/virtual-machine.js
```diff
@@ -554,10 +567,11 @@ class VirtualMachine extends EventEmitter {
     /**
      * Create a variable by name.
      * @todo this only creates global variables by putting them on the stage
-     * @param {string} name The name of the variable
+     * @param {string} id The id of the variable.
+     * @param {string} name The name of the variable.
      */
-    createVariable (name) {
```
paulkaplan: Just wanted to point out that we should file an issue for any external API changes before we make them so we do not lose track of this, as it'll break the GUI. 

Come to think of it, this function was actually only being used as a kind of crutch to wire together the blocks -> gui -> VM. Now that we have the events coming directly to the VM through the listener we can probably get rid of this and the accompanying call from the GUI. Feel free to remove this function entirely!

src/serialization/sb2.js
RoboErikG: That should just cause 'undefined' to be passed in which will cause a uuid to be generated for it.

paulkaplan: Ah I see, that is why it works. I think we should explicitly use `undefined`, using `variable.id` when we know it is always `undefined` is bound to confuse someone eventually.

src/virtual-machine.js
marisaleung: Done

src/serialization/sb2.js
marisaleung: Done

src/engine/runtime.js
marisaleung: Done! I used "createVariable" instead of "addVariable" to be consistent with the event names.

marisaleung: No worries, this all makes sense. And I like the idea of the refactoring all of it out into adaptor eventually.

Also, done.

src/engine/target.js
```diff
@@ -118,6 +119,44 @@ class Target extends EventEmitter {
     }
 
     /**
+     * Creates a variable with information given in the event and adds it to
+     * the dictionary of variables.
+     * @param {object} e Blockly "variable" event.
+     */
+    createVariable (e) {
```
paulkaplan: Could you change the arguments to these so that they do not contain the blockly event, but instead the stuff that will be used from the event. E.g.

```
createVariable(id, name) { ...
renameVariable(id, newName) { ...
deleteVariable(id) { ...
```

paulkaplan (changes requested):

src/engine/target.js
marisaleung: Done!

RoboErikG:
As discussed offline, add tests since we're not trying to get this in for tomorrow.

---

Review status: 0 of 9 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/llk/scratch-vm/614#-:-Kn6_jevI2bCWea3OIgV:bxving1)*

marisaleung:
Done

---

Review status: 0 of 10 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/llk/scratch-vm/614#-:-KnAtMVRox4qefIsrA_t:bb6un1j)*

paulkaplan:
@marisaleung looks great, thanks for the tests! I'm going to be doing all the merging today, I'll probably connect with you at some point today to make sure I've got all the changes in the right order. I also have to make a corresponding PR on GUI to remove the `createVariable` usage, which I'll also have to test against everything.

src/engine/blocks.js
```diff
@@ -511,7 +523,11 @@ class Blocks {
     _getBlockParams (block) {
         const params = {};
         for (const key in block.fields) {
-            params[key] = block.fields[key].value;
```
paulkaplan: Noticed a problem here when connecting this with the GUI. The variable ID is being reported in the monitor instead of the variable name. This actually should use `.value` because it is used by the monitor reporting so it should include the name of the variable, not the ID. Sorry there is no useful indication of that.

paulkaplan (changes requested):

paulkaplan (approved):

PR #619, opened 2017-06-16
Clone custom state

### LINKED ISSUE #557: Cloning _customState: clones don't inherit correct pen state
### Expected Behavior

When a rendered target is cloned, the new clone should inherit the pen state of the original. That includes:
- the size and color of the pen,
- whether the pen is down or not,
- and the events which implement drawing on move.

We should implement this in such a way that the cloning code doesn't have to "know about" pen implementation details. Ideally we should allow a block package to decide how to clone its custom state to a new target.

### Actual Behavior

As of #555 we are now cloning the contents of a rendered target's `_customState` when that target is cloned. This means we're cloning the size and color of the pen, as well as the flag indicating whether or not the pen is down, but we're not cloning the events. This means that pen projects involving clones often behave incorrectly.

There may also be audio-related consequences, since the audio effects use `_customState` as well.

### Steps to Reproduce

image

### See Also

Previous issues: #481, #493
Previous discussion: #525
Current partial solution: #555

cwillisf:
### Resolves

Resolves #557

### Proposed Changes

The biggest change here is the introduction of a `targetWasCreated` emitted by `Runtime`. This event is to be emitted any time a target is created -- currently, only by `createClone()` in `Sprite` -- and contains the new target as well as the target it was cloned from, if any.

The pen blocks use this new event to clone pen state information, including event registration when necessary.

### Reason for Changes

I debated a variety of different ways to handle clones, including:
- An event emitted by `Runtime` only when a target is cloned, and
- An event emitted by any `Target` used as a clone source.

I settled on this model because I suspect there will also be cases where we want to initialize custom state on a target which wasn't created by cloning.

### Test Coverage

The existing test in `test/ingration/pen.js` has been extended to include coverage for pen state cloning. Specifically, it verifies that a clone inherits its source's pen size but does not share the same state object.


rschamp (approved):
I like it! If we did this with the other blocks, we could greatly simplify `makeClone`.  I think that could happen in a different PR though.

rschamp:
@paulkaplan @fsih @thisandagain Worth checking out this PR if you haven't seen it.

Issue #628, opened 2017-06-27
Dropdown uses old variable.


marisaleung:
### Expected Behavior
When a variable dropdown is changed to a different variable, the new variable is used when the block is run.

### Actual Behavior
When a variable dropdown is changed to a different variable, the first one is still referenced when the block is run.
![variable-dropdowns-broken](https://user-images.githubusercontent.com/27254516/27614350-e2f9b614-5b55-11e7-9e80-ad1715dd5f64.gif)

### Steps to Reproduce
1. Create variable 'a.'
2. Create variable 'b.'
3. Create 'Change a to 10' block on workspace.
4. Create 'go to x: a y:b' block on the workspace.
5. Change Dropdown in change variable block to 'b'.
6. Click change variable block to run it and notice how the monitor value for 'a' changes isteadd of 'b.'

### Operating System and Browser

Chrome Ubuntu

PR #629, opened 2017-06-27
ChangeBlock uses id's of variable instead of name.

marisaleung:
### Resolves
https://github.com/LLK/scratch-vm/issues/628

### Proposed Changes

On a block change, checks if what we are changing is a variable, if it is we use the value (which is the variable id) to find the variable name and set the block's field VARIABLE id and value correctly.

### Reason for Changes

Without this, the value of the block's field VARIABLE would be set to the passed in id, instead of the name.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -296,7 +296,16 @@ class Blocks {
         case 'field':
             // Update block value
             if (!block.fields[args.name]) return;
-            block.fields[args.name].value = args.value;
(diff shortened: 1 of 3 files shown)


paulkaplan (approved):
LGTM. 

Merging this is blocked on https://github.com/google/blockly/pull/1196 and merging that into scratch-blocks

marisaleung:
google/blockly#1196 is merged into blockly now.

PR #633, opened 2017-07-06
Fix sound indexing and add tests

### LINKED ISSUE #626: Drums and instruments should wrap-clamp using 1-indexing
### Expected Behavior

When selecting a drum or an instrument using a reporter block dropped onto the menu, the value of the reporter should always result in a drum or instrument being selected, wrapping the number if it is out of range. For example, since there are 18 drums, selecting drum 19 using a variable should play drum 1, drum 0 should give drum 18, etc.   

### Actual Behavior

The code assumes that drums and instruments are 0-indexed instead of 1-indexed, so certain values (such as 0 and 19 for the drums) cause an error.

### Steps to Reproduce

Running this stack should play a drum sound, but it causes an error:

screen shot 2017-06-27 at 11 11 04 am

Mac OS 10.11.6 Chrome 59.0

### LINKED ISSUE #627: Sounds selected by reporter should be 1-indexed
### Expected Behavior

When you select a sound in the "play sound" or "play sound until done" block by dropping a reporter onto the menu, it should treat the sprite's sounds a 1-indexed list, and wrap-clamp appropriately.

This bug is similar to https://github.com/LLK/scratch-vm/issues/626.

### Actual Behavior

The sounds are selected using 0-indexing.

### Steps to Reproduce

Create a sprite with two or more sounds. This code should play the first sound:

screen shot 2017-06-27 at 11 28 09 am

Instead it plays the second sound in the list.

Mac OS 10.11.6 Chrome 59.0

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/627
Fixes https://github.com/LLK/scratch-vm/issues/626

### Proposed Changes

Fixes the way one indexing is handled in the sound blocks. Also includes a new set of tests!

### Reason for Changes

Bugs linked above.

### Test Coverage

new unit tests on the scratch3_sound file

/cc @cwillisf @rschamp about how i made tests for this. It would be nice to have a stubbing library, but not really needed. I'm not sure I like how much these tests rely on the internal workings :( I'm totally open to refactoring ideas that would make this easier to test.

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -183,7 +183,7 @@ class Scratch3SoundBlocks {
         let drum = Cast.toNumber(args.DRUM);
         drum -= 1; // drums are one-indexed
         if (typeof this.runtime.audioEngine === 'undefined') return;
-        drum = MathUtil.wrapClamp(drum, 0, this.runtime.audioEngine.numDrums);
+        drum = MathUtil.wrapClamp(drum, 0, this.runtime.audioEngine.numDrums - 1);
         let beats = Cast.toNumber(args.BEATS);
         beats = this._clampBeats(beats);
         if (util.target.audioPlayer === null) return;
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
I think this way of testing is OK for the moment, though I do agree with you that we're not in a great place for this sort of test. Another approach is to set up a proper VM/runtime, manipulate them, then check the state. I'm not sure how well that work in your situation, but it worked for me in #619, for example.

PR #634, opened 2017-07-07
Add public API for costume and sound renaming 

### LINKED ISSUE #630: Need public API for changing a sound/costumes name
For the costume/sound editors, we need a way for the GUI to trigger a name change.

It should use the same duplicate validation that the sprites use (i.e. add number on duplicate). That will address some of the recording numbering issues from the playtest with the sound editor.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/630

### Proposed Changes

Adds public `renameCostume` and `renameSound` methods to the `virtual-machine` object. This will support the sound and costume editors in the GUI. 

### Test Coverage

Added tests for the usage of these methods, including tests to make sure that duplicate naming is handled correctly.

---

Questions for the reviewer (particularly @rschamp):

1. Can you think of a way to DRY up the `unusedName` code? It is now repeated in three places (rename sprites, costumes and sounds) and the way it works is always the same (gather up the other names, then use `unusedName(newName, otherNames)`
2. Should this renaming be done by the target instead of directly by the VM? @rschamp I know you previously suggested this for some other piece of code, but I'm forgetting exactly when.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -288,6 +288,19 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Rename a costume on the current editing target.
+     * @param {int} costumeIndex - the index of the costume to be renamed.
+     * @param {string} newName - the desired new name of the costume (will be modified if already in use).
+     */
+    renameCostume (costumeIndex, newName) {
+        const usedNames = this.editingTarget.sprite.costumes
+            .filter((costume, index) => costumeIndex !== index)
+            .map(costume => costume.name);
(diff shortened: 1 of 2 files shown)


fsih (approved):
Might want to test naming something hello2, or your expected behavior when someone already has variables ending in numbers but they are discontinuous

paulkaplan:
@fsih good point. Those behaviors are already covered in the tests for `StringUtil#unusedName`, since they get pretty complicated.

Issue #623, opened 2017-06-22
Monitor not removed when variable is deleted


paulkaplan:
Because https://github.com/LLK/scratch-vm/pull/614 will fix variable deleting, we need to support it with monitors!

Monitors should go away when the variable they are attached to is deleted.

Issue #622, opened 2017-06-22
Monitor label is not updated when a variable is renamed


paulkaplan:
Since https://github.com/LLK/scratch-vm/pull/614 will make variable renaming work, we need to properly support it!

Monitor labels should change when the name of the variable changes

PR #636, opened 2017-07-10
Update monitors when deleting or renaming variables

fsih:
### Resolves

Part of solution for LLK/scratch-vm#623
and https://github.com/LLK/scratch-vm/issues/622

### Proposed Changes

Update monitors when variables are renamed or deleted

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -298,7 +298,7 @@ class Blocks {
             if (!block.fields[args.name]) return;
             if (args.name === 'VARIABLE') {
                 // Get variable name using the id in args.value.
-                const variable = optRuntime.getEditingTarget().lookupVariableById(args.id);
+                const variable = optRuntime.getEditingTarget().lookupVariableById(args.value);
                 if (variable) {
                     block.fields[args.name].value = variable.name;
                     block.fields[args.name].id = args.value;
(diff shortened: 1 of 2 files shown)


src/engine/blocks.js
```diff
@@ -298,7 +298,7 @@ class Blocks {
             if (!block.fields[args.name]) return;
             if (args.name === 'VARIABLE') {
                 // Get variable name using the id in args.value.
-                const variable = optRuntime.getEditingTarget().lookupVariableById(args.value);
```
fsih: Was this a bug originally? I don't see why you would pass the name of the variable to fetch the variable, since the ID never matched the name

paulkaplan: Do you mean that `args.value` is the variable name? I thought it was the variable id?

paulkaplan: Hmmm looking at line 304 below, it looks like the `args.value` is the variable id?

paulkaplan:
CI is failing

src/engine/blocks.js
fsih: The only usage of changeBlock is line 217, and args.value passes in the name of the variable

fsih: Changed this back

paulkaplan (approved):
WOOOOOOÔOŒ

PR #637, opened 2017-07-12
Fix monitor not updating every frame

fsih:
1. Evaluate monitors every frame (they were only being evaluated every other frame) and add test
2. Instead of activating monitors by using toggleScript, make another function that's almost the same as toggleScript but never toggles off (always toggles on if the thread isn't running). Monitors try to evaluate every frame, and if they don't finish evaluating during one frame, then toggleScript on the next frame would toggle the monitor script off. Then if you had a monitor that took too long to run (which we shouldn't), it would never finish.


src/engine/runtime.js
```diff
@@ -587,7 +597,8 @@ class Runtime extends EventEmitter {
                 // give up if any threads with the top block are running.
                 for (let j = 0; j < instance.threads.length; j++) {
                     if (instance.threads[j].topBlock === topBlockId &&
-                        instance.threads[j].target === target) {
+                        instance.threads[j].target === target &&
+                        !instance.threads[j].status === Thread.STATUS_DONE) {
```
paulkaplan: What does this line address?

paulkaplan:
The behavior here is quite subtle, and seems easy to break. I think we should try to find a way to write a test that would fail if we broke some part of the behavior, specifically:

1. Hats being evaluated on every tick instead of every other
2. Making sure hat stacks run on stack click
3. Allowing stack click threads to coexist with regular threads

src/engine/runtime.js
fsih: 3. Start hats aren't added every other frame because we keep them around as "done" threads

fsih:
At Long Last, an update!

@paulkaplan I did my best to add tests for the weird interactions between threads, but in order to do so I had to expose a lot deep pieces of the VM, in particular the _step() function. Please take a look and see if you think it's acceptable.

If you look at the latest commit, along with the tests there's also a bit of refactoring and some bug fixes.

Some of the problems I encountered (which are admittedly getting pretty tangential to this bug) I felt like were because we are using toggle script in a way that doesn't make sense. We're having problems with monitors and hat blocks toggling themselves off when they're trying to turn on, but we never actually want things that run every frame to turn anything off. So I wrote a separate function for monitors which only adds the thread and never accidentally removes it.

fsih:
Poke

paulkaplan:
[from offline: pls break up the monitor fixes and the stack fixes into separate PRs with corresponding tests]

fsih:
Paul and I decided that 5 things in one pull request is too confusing and I should at least pull out the changes unrelated to the 2 bugs mentioned, so you can hold off on review for a bit

fsih:
https://github.com/LLK/scratch-vm/pull/655 where I removed the monitor changes is ready

fsih:
This PR is now transformed into the just-monitor-bugs PR

rschamp (approved):
This makes sense to me, but we should keep an eye on performance, since this is adding a new loop through the threads where there wasn't one before.

fsih:
It shouldn't be looping through anything more than it was before, because add monitor script replaces toggle script which has the same loop.

PR #640, opened 2017-07-13
Implement message dispatch system

cwillisf:
### Proposed Changes

This message dispatch system allows a "service" to register itself under a particular name, and then anyone can call any method on that service from any context (the main "window" thread or any worker). Return values are passed back through promise resolution.

### Reason for Changes

This will make it much easier to put extensions inside workers: we can expose any services needed by extensions through this message dispatch system, but otherwise extensions will be unable to interact with the editor. If we find that it's necessary, we could even add some form of permissions to the message dispatch system with relatively little work.

### Test Coverage

New tests cover the basics of the message dispatch system. Because the system is built around web workers, which aren't present in our Node.js-based testing process, I added the `tiny-worker` library. This library emulates web workers by running multiple Node processes.

Unfortunately this exposes the difference between webpack's build-time handling and Node's run-time handling of `require` statements. The `dispatch-test-worker-shim.js` file monkey-patches `require` **in the worker only** to deal with these differences.

### DIFF
--- a/src/dispatch/central-dispatch.js
+++ b/src/dispatch/central-dispatch.js
@@ -106,15 +106,19 @@ class CentralDispatch {
 
     /**
      * Set a local object as the global provider of the specified service.
+     * WARNING: Any method on the provider can be called from any worker within the dispatch system.
      * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui', 'extension9'.
      * @param {object} provider - a local object which provides this service.
-     * WARNING: Any method on the provider can be called from any worker within the dispatch system.
+     * @returns {Promise} - a promise which will resolve once the service is registered.
      */
     setService (service, provider) {
-        if (this.services.hasOwnProperty(service)) {
-            log.warn(`Replacing existing service provider for ${service}`);
-        }
-        this.services[service] = provider;
+        return new Promise(resolve => {
+            if (this.services.hasOwnProperty(service)) {
+                log.warn(`Replacing existing service provider for ${service}`);
+            }
+            this.services[service] = provider;
+            resolve();
+        });
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/dispatch/central-dispatch.js
```diff
@@ -0,0 +1,173 @@
...
+        } else {
+            resolve(result);
+        }
+    }
+}
+
+const dispatch = new CentralDispatch();
+module.exports = dispatch;
+self.Scratch = self.Scratch || {};
+self.Scratch.dispatch = dispatch;
```
rschamp: Is this left in for debugging? Seems like it should be something the app does.

cwillisf:
Rebased on `develop` @ e249c5c2525316a369a2f6a5220467eb6c1d60b5 to fix `package.json` conflicts

src/dispatch/central-dispatch.js
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Set a local object as the global provider of the specified service.
+     * WARNING: Any method on the provider can be called from any worker within the dispatch system.
+     * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui', 'extension9'.
+     * @param {object} provider - a local object which provides this service.
+     * @returns {Promise} - a promise which will resolve once the service is registered.
+     */
+    setService (service, provider) {
+        return new Promise(resolve => {
```
rschamp: We should add an explanatory comment about why this returns a Promise (for consistency).

src/dispatch/central-dispatch.js
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Handle a message event received from a connected worker.
+     * @param {Worker} worker - the worker which sent the message.
+     * @param {MessageEvent} event - the message event to be handled.
+     * @private
+     */
+    _onMessage (worker, event) {
+        const [service, method, callbackId, ...args] = /** @type {[string, string, *]} */ event.data;
```
rschamp: Maybe `event.data` should come as an object with a `type` key. Either way we should `try...except` here in case the message isn't well formed.

src/dispatch/central-dispatch.js
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * Handle a callback from a worker. This should only be called as the result of a message from a worker.
+     * @param {int} callbackId - the ID of the callback to call.
+     * @param {*} result - if `error` is not truthy, resolve the callback promise with this value.
+     * @param {*} [error] - if this is truthy, reject the callback promise with this value.
+     * @private
+     */
+    _callback (callbackId, result, error) {
```
rschamp: `result, error` looks reversed to me.

rschamp (changes requested):
@cwillisf Looks great overall! Just a few notes from our walkthrough.

rschamp (approved):
Looks good @cwillisf! Can't wait to see this in action.

rschamp:
@paulkaplan @fsih @ericrosenbaum should also take a look at this, at least just to be familiar with what's going in for the extensions work.

PR #641, opened 2017-07-17
Fix variable import scoping

### LINKED ISSUE #638: Loaded Project's have broken variable assignments
### Expected Behavior

Loading an existing project in Scratch 3 should have working variable assignment blocks.

### Actual Behavior

Existing variable assignments [set (var) = "value"] do not assign their values, whereas newly created ones, and duplicated versions (of the bad ones) do work.

### Steps to Reproduce

> Load a project: https://llk.github.io/scratch-gui/#168754184
> Make the two variables visible (if they are not)
> Run the project
Both variables stay at 0 rather than being assigned any values.
> Duplicate the two assignment blocks
> Run the new blocks
Both values are assigned correctly.

Chrome on windows 10.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/638

### Proposed Changes

1. Adds variables private to the editing target to the published XML.
2. Creates deterministic variable IDs for imported variables, and correctly scopes them.

### Reason for Changes

As shown in https://github.com/LLK/scratch-vm/issues/638, the changes to the variable field to include ID broke the way importing worked. 

### Test Coverage

Added tests for both the XML creation and the correct variable scope importing. In order to do this, I added a local variable to the `data.sb` file.

---

I don't like the use of closures in the sb2 importing. @rschamp thoughts on a way to clean that up?


paulkaplan:
@cwillisf I'm adding you to this since it is near your wheelhouse. Let me know if you can think of a better way to do it.

rschamp:
I guess the alternative would be to pass a map around and then have a non-closured function you could pass the map to with the targetId and name whenever you want a variable. That might be a little more understandable, but much more spaghetti-like. I think this method is good enough for now.  Maybe if we turned the parser into a stateful object, we could do this kind of thing more easily?

rschamp (approved):

PR #653, opened 2017-07-25
Use sound ID instead of md5 for playing sounds

paulkaplan:
### Proposed Changes

Uses the sound ID as it comes back from the audio engine (requires https://github.com/LLK/scratch-audio/pull/57) to play sounds. This is to set up for being able to edit sounds that have the same content without conflicts. 

### Test Coverage

Updated the relevant sound block tests.


ericrosenbaum (approved):
LG

Issue #153, opened 2016-09-07
Two issues with "toggle script": any hat block should activate its stack when clicked


tmickel:
- [ ] Scripts that start with edge-activated hats can't currently be started by clicking. This is because the hat is evaluated, and the predicate fails. I think we should just skip the hat block when a user explicitly clicks on a stack starting with an edge-activated hat.
- [ ] "Untoggle" is based on the top-block of the thread, instead of the "script-block" of the thread. If the thread moves onto a different script, this can lead to weird behavior.

rschamp:
For July: clicking hats should run their scripts without evaluating the hat first

fsih:
@tmickel I don't understand what the second check box means, can you give repro steps?

tmickel:
Sure! I don't remember precisely all the weird behavior that happens because of this. But one such repro... Start with this:
<img width="330" alt="screen shot 2017-07-11 at 5 16 43 pm" src="https://user-images.githubusercontent.com/120403/28091096-b92ccc10-665c-11e7-9685-83ae2e0da689.png">
The left forever is running (i.e. cat is spinning).

Drag the left blocks into the right forever. The glow snaps to the outer loop:
<img width="214" alt="screen shot 2017-07-11 at 5 17 37 pm" src="https://user-images.githubusercontent.com/120403/28091121-d5e5ef12-665c-11e7-9280-d46fe8e710e1.png">

Now, click on any block in the stack to untoggle it. Notice that the cat spins twice as fast - i.e., that you spawned a second thread - instead of stopping the stack.

This is basically because the toggle/untoggle for a stack is currently related to the first block that ran in the thread, rather than the first block of the *current* stack of the thread.


I think Scratch 2.0 also has some weird weird behavior with this sort of thing, but I think it should be fundamentally solvable...

fsih:
For the first issue, this seems to be working already. I tried every hat block and the only one that didn't work was "when timer > 10", which must be a different bug, because timer is much greater than 10.

<img width="477" alt="screen shot 2017-07-11 at 17 51 57" src="https://user-images.githubusercontent.com/2855464/28092534-c0c5604a-6661-11e7-8bd0-5d2e7f66cfd0.png">

(So I guess https://github.com/LLK/scratch-blocks/issues/845 was not a dupe)

fsih:
Thanks Tim!

In Scratch 2.0 when you add the outer forever, the inner one stays glowing, and you can toggle the glow on the outer one, so that's cool

[...]

PR #655, opened 2017-08-03
Activate the greater than hat block by clicking

### LINKED ISSUE #845: "When timer > x" block does not activate script when it is clicked
### Expected Behavior

In 2.0, clicking a stack with "When timer greater than 0" as the hat block activates the stack.

### Actual Behavior

In 3.0, clicking a stack with "When timer greater than 0" as the hat block does not activate the stack.
The stack will be activated on green flag (when timer conditional is true).

### Steps to Reproduce

- Compare behaviour in 2.0 & 3.0
- Put "when timer greater than x" block in the scripting area
- Change x to 0
- Add the stamp block to the stack
- Put  "when this sprite clicked" block in the scripting area
- Add the stamp block to that stack
- Click on each stack, then drag the sprite to see if the stamp occurred
- Note the stamp occurred for the "when this sprite clicked" stack but not for the "when timer > 0" stack
- Delete the "when this sprite clicked" stack
- Press green flag
- Note that the stamp does occur for the "when timer > 0" block

_Mac El Capitan, Chrome 56_

fsih:
https://github.com/LLK/scratch-vm/pull/637 without the monitor changes

- Rename show visual report option to stack click, since it's used to decide more than just show visual report now
- Circumvent evaluating the predicate on the hat block if it's activated via stack click
- Fix startHats, which was only evaluating the hat blocks every other frame instead of every frame
- Allow stack clicked thread to coexist with edge activated hat block thread, which (now) runs every frame, so that stack clicking works

These things together fix LLK/scratch-blocks#845 (clicking the timer > 10 hat block stack should run the stack)
Also fixes first bullet of #153


paulkaplan (approved):
LG, thanks for splitting this up!

PR #657, opened 2017-08-08
Fix variable deserializing

neurosie:
### Resolves

Fixes LLK/scratch-gui#461.

### Proposed Changes

Fixed a small error in the sb3 deserializer preventing variables from loading to the proper targets.

### Reason for Changes

Fixes above bug.

### Test Coverage

None, the sb3 save/load tests aren't written yet, so there was nothing to change.

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -100,7 +100,7 @@ const parseScratchObject = function (object, runtime) {
     const target = sprite.createClone();
     // Load target properties from JSON.
     if (object.hasOwnProperty('variables')) {
-        for (let j = 0; j < object.variables.length; j++) {
+        for (let j in object.variables) {
             const variable = object.variables[j];
             const newVariable = new Variable(
                 variable.id,


src/serialization/sb3.js
```diff
@@ -100,7 +100,7 @@ const parseScratchObject = function (object, runtime) {
     const target = sprite.createClone();
     // Load target properties from JSON.
     if (object.hasOwnProperty('variables')) {
-        for (let j = 0; j < object.variables.length; j++) {
+        for (let j in object.variables) {
             const variable = object.variables[j];
```
towerofnix, line 104: This could be changed to...

```js
for (let variable of Object.values(object.variables)) {
  const newVariable ...
}
```

..which would get rid of the less-obviously named (JS-)variable `j`, and function the same. (Though I don't know if [`for..of`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is part of the set of ES6 features scratch-* repositories use, considering `for (let k = 0; ...)` is used frequently.)

PR #660, opened 2017-08-10
Add [string] contains [string] block

Kenny2github:
### Resolves

LLK/scratch-gui#601

### Proposed Changes

Adds `contains` as a block to `scratch-vm/src/blocks/scratch3_operators.js` - it returns `string1.includes(string2)`

This block is also added to the blocks list in LLK/scratch-blocks#1045

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -109,8 +109,10 @@ class Scratch3OperatorsBlocks {
     }
     
     contains (args) {
-        return Cast.toString(args.STRING1).toLowerCase()
-            .includes(Cast.toString(args.STRING2).toLowerCase());
+        const format = function (string) {
+            return Cast.toString(string).toLowerCase()
+        }
+        return format(args.STRING1).includes(format(args.STRING2))
     }
 
     mod (args) {
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -30,6 +30,7 @@ class Scratch3OperatorsBlocks {
             operator_join: this.join,
             operator_letter_of: this.letterOf,
             operator_length: this.length,
+            operator_contains: this.contains,
             operator_mod: this.mod,
             operator_round: this.round,
             operator_mathop: this.mathop
@@ -105,6 +106,10 @@ class Scratch3OperatorsBlocks {
 
     length (args) {
         return Cast.toString(args.STRING).length;
+    }
+    
+    contains (args) {
+        return Cast.toString(args.STRING1).includes(Cast.toString(args.STRING2));
     }
 
     mod (args) {
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -110,9 +110,9 @@ class Scratch3OperatorsBlocks {
     
     contains (args) {
         const format = function (string) {
-            return Cast.toString(string).toLowerCase()
-        }
-        return format(args.STRING1).includes(format(args.STRING2))
+            return Cast.toString(string).toLowerCase();
+        };
+        return format(args.STRING1).includes(format(args.STRING2));
     }
 
     mod (args) {
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -109,7 +109,7 @@ class Scratch3OperatorsBlocks {
     }
     
     contains (args) {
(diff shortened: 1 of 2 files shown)


src/blocks/scratch3_operators.js
```diff
@@ -106,6 +107,10 @@ class Scratch3OperatorsBlocks {
     length (args) {
         return Cast.toString(args.STRING).length;
     }
+    
+    contains (args) {
+        return Cast.toString(args.STRING1).includes(Cast.toString(args.STRING2));
```
griffpatch: Isn't includes case sensitive?

Kenny2github: Uhh, true 🙄 lemme change that...

thisandagain:
Thanks @Kenny2github. In addition to the case sensitivity fix that @griffpatch mentioned it would be helpful for this to include unit test coverage (see: https://github.com/LLK/scratch-vm/blob/develop/test/unit/blocks_operators.js).

Kenny2github:
@thisandagain I have now added tests :) boy doing this on an iPad is hard 😓

thisandagain:
@Kenny2github Getting close! Looks like you are just failing the linter:
```bash
 112:58  error  Expected line break before `.includes`  newline-per-chained-call
```

I would recommend breaking up the line a bit to make it more readable / pass lint.

src/blocks/scratch3_operators.js
```diff
@@ -106,6 +107,11 @@ class Scratch3OperatorsBlocks {
     length (args) {
         return Cast.toString(args.STRING).length;
     }
+    
+    contains (args) {
+        return Cast.toString(args.STRING1).toLowerCase()
```
thisandagain: This could be written in various ways to reduce duplication and improve readability:
```js
const format = function (string) {
    return Cast.toString(string).toLowerCase();
};

return format(args.STRING1).includes(format(args.STRING2);
```

or even ...
```js
const format = function (string) {
    return Cast.toString(string).toLowerCase();
};

const string1 = format(args.STRING1);
const string2 = format(args.STRING2);

return string1.includes(string2);
```

thisandagain (changes requested):

src/blocks/scratch3_operators.js
Kenny2github: Sure! I’ll do that now - I was hesitant about using extra variables

Kenny2github:
Just quickly wait for Travis to finish and this should be done and dusted. I would guess that this would be merged with the scratch-blocks pr at once?
EDIT: uhh, what now?

Kenny2github:
@thisandagain how’s this? It seems like lint failed - the logs say “use const instead of var” but I did use const... what happened?

src/blocks/scratch3_operators.js
```diff
@@ -106,6 +107,13 @@ class Scratch3OperatorsBlocks {
     length (args) {
         return Cast.toString(args.STRING).length;
     }
+    
+    contains (args) {
+        const format = function (string) {
+            return Cast.toString(string).toLowerCase()
```
thisandagain: Missing semicolon.

src/blocks/scratch3_operators.js
```diff
@@ -106,6 +107,13 @@ class Scratch3OperatorsBlocks {
...
+        }
```
thisandagain: Missing semicolon.

src/blocks/scratch3_operators.js
```diff
@@ -106,6 +107,13 @@ class Scratch3OperatorsBlocks {
...
+        return format(args.STRING1).includes(format(args.STRING2))
```
thisandagain: Missing semicolon.

thisandagain (changes requested):
Getting super close! Once resolved we'll land this first and then look at the blocks PR.

thisandagain:
Note to self: this PR should be squashed before merge.

Kenny2github:
Okay, this is where my Python history kicks in 🤦‍♂️
Hopefully that’s fixed now :)
Blocks PR should be fairly straightforward - I’ll just check that I didn’t miss any *semicolons* on that while waiting for travis

Kenny2github:
@thisandagain Seems like it’s done - checks passed and code looks good. Any final comments?

thisandagain (approved):

thisandagain:
Looks good! Thanks @Kenny2github

Kenny2github:
No problem! Glad I could help 😁

PR #662, opened 2017-08-10
Add glide (number) secs to [dropdown] block

Kenny2github:
### Resolves

LLK/scratch-gui#598

### Proposed Changes

Adds the block to `motion.js`, basically copying target selection from go to and copying the glide from glide.

Comes with LLK/scratch-blocks#1046 and LLK/scratch-gui#634

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -73,7 +73,7 @@ class Scratch3MotionBlocks {
     }
 
     goTo (args, util) {
-        const targetXY = getTarget(args.TO, util);
+        const targetXY = this.getTarget(args.TO, util);
         if (targetXY) {
             util.target.setXY(targetXY[0], targetXY[1]);
         }
@@ -149,9 +149,9 @@ class Scratch3MotionBlocks {
     }
     
     glideTo (args, util) {
-        const targetXY = getTarget(args.TO, util);
+        const targetXY = this.getTarget(args.TO, util);
         if (targetXY) {
-            glide({SECS:args.SECS, X:targetXY[0], Y:targetXY[1]}, util);
+            this.glide({SECS:args.SECS, X:targetXY[0], Y:targetXY[1]}, util);
         }
     }
 
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -52,24 +52,29 @@ class Scratch3MotionBlocks {
         util.target.setXY(x, y);
     }
 
-    goTo (args, util) {
+    getTarget (TO, util) {
         let targetX = 0;
         let targetY = 0;
-        if (args.TO === '_mouse_') {
+        if (TO === '_mouse_') {
             targetX = util.ioQuery('mouse', 'getX');
             targetY = util.ioQuery('mouse', 'getY');
-        } else if (args.TO === '_random_') {
+        } else if (TO === '_random_') {
             const stageWidth = this.runtime.constructor.STAGE_WIDTH;
             const stageHeight = this.runtime.constructor.STAGE_HEIGHT;
             targetX = Math.round(stageWidth * (Math.random() - 0.5));
             targetY = Math.round(stageHeight * (Math.random() - 0.5));
         } else {
-            const goToTarget = this.runtime.getSpriteTargetByName(args.TO);
+            const goToTarget = this.runtime.getSpriteTargetByName(TO);
             if (!goToTarget) return;
             targetX = goToTarget.x;
             targetY = goToTarget.y;
         }
-        util.target.setXY(targetX, targetY);
+        return (targetX, targetY);
+    }
+
+    goTo (args, util) {
+        let targetXY = getTarget(args.TO, util);
+        util.target.setXY(targetXY[0], targetXY[1]);
     }
 
     turnRight (args, util) {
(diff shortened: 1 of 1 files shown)


Kenny2github:
@ericrosenbaum seeing as you were requested for review, any comments? /cc @thisandagain

ericrosenbaum:
We should consider refactoring to reduce duplication of the code (which could be a source of bugs later). So instead of copying and pasting the code from goto and glide, we should make two new helper functions, one for target selection and one for gliding, that can be called by the other blocks. What do you think?

ericrosenbaum:
Also note that to update the menu with the names of the sprites, we'll need to add a function in scratch-gui like [this](https://github.com/LLK/scratch-gui/blob/67885bca3280979a5f460271f9e455945583aceb/src/lib/blocks.js#L95-L101).

Kenny2github:
Ah, I see. Will change this tomorrow :D

Kenny2github:
@ericrosenbaum Note: I made `glideTo` just use the original `glide` for relatively obvious reasons.

Kenny2github:
Hmm, I knew something was wrong but for some reason I can’t see Travis logs. Will look into it when I can.

src/blocks/scratch3_motion.js
```diff
@@ -52,24 +52,29 @@ class Scratch3MotionBlocks {
...
             targetY = Math.round(stageHeight * (Math.random() - 0.5));
         } else {
-            const goToTarget = this.runtime.getSpriteTargetByName(args.TO);
+            const goToTarget = this.runtime.getSpriteTargetByName(TO);
             if (!goToTarget) return;
             targetX = goToTarget.x;
             targetY = goToTarget.y;
         }
-        util.target.setXY(targetX, targetY);
+        return (targetX, targetY);
```
towerofnix: Arrays use the syntax `[targetX, targetY]`, not `(targetX, targetY)`.

Kenny2github: Ah, yes, I was getting confused with Python’s `tuple` syntax 😒

Kenny2github:
Hmm, I was wondering if that would be a problem. I don’t really know how JavaScript global/locals work though so a little help would be appreciated (assuming I don’t research it first).

src/blocks/scratch3_motion.js
```diff
@@ -51,24 +52,29 @@ class Scratch3MotionBlocks {
...
             targetX = goToTarget.x;
             targetY = goToTarget.y;
         }
-        util.target.setXY(targetX, targetY);
+        return [targetX, targetY];
+    }
+
+    goTo (args, util) {
+        let targetXY = getTarget(args.TO, util);
+        util.target.setXY(targetXY[0], targetXY[1]);
```
towerofnix: Hmm.. compare this with line 68, `if (!goToTarget) return;`. In that case, `targetXY` will be the value `undefined`, and so doing `targetXY[0]` (or `targetXY[1]`) will throw an error (since doing `undefined[0]` doesn't make sense).

This would practically be a problem if `goTo` is given some nonexistant sprite's name; for example "go to (join (nonexistant) (sprite))".

Probably the simple solution would just be to check if targetXY isn't undefined: `if (targetXY) { util.target.setXY(...) }`.

This is also probably a problem in the `glideTo` definition.

Kenny2github: Ah, okay. Doing that now!

Kenny2github:
Ah, “getTarget” is not defined - I’ll really need some global/local help here.

src/blocks/scratch3_motion.js
```diff
@@ -51,24 +52,31 @@ class Scratch3MotionBlocks {
...
             if (!goToTarget) return;
             targetX = goToTarget.x;
             targetY = goToTarget.y;
         }
-        util.target.setXY(targetX, targetY);
+        return [targetX, targetY];
+    }
+
+    goTo (args, util) {
+        const targetXY = getTarget(args.TO, util);
```
towerofnix: `getTarget` isn't defined here (you noticed that!). See, the issue is that JavaScript expects there to be a variable named `getTarget`, and there isn't; it can't find any `let`, `var`, or `const getTarget`.

What *you* expected is that by declaring a *method* named `getTarget`, JavaScript would automatically know that "`getTarget`" means "the method named `getTarget`". But JavaScript doesn't know that.

There's a specific syntax for getting a method, and it's actually the same as the syntax to get any other property: `this.methodName`. (You see, methods are (basically) just properties of the class instance, just like `this.runtime`.)

(In fact, you can already see `this.methodName` being used a bunch in this very file -- check the `getPrimitives` function, and you'll see that, for example, the line `motion_glideto: this.glideTo` is getting the method `glideTo` of `this` (so, the class instance) and storing it in the `motion_glideto` property of the return object!)

So, what you want here is `this.getTarget(args.TO, util)`, not `getTarget(args.TO, util)`.

Kenny2github: Ah, okay, thanks! I see that I really don’t know much about js classes either lol.
I’ll get to it right away!

Kenny2github:
Finally, Travis is happy :)

So I guess the last question is:
Status on this PR?

Kenny2github:
@towerofnix Finally, Travis is happy :)

Final question is, I guess:
Status on this PR?

src/blocks/scratch3_motion.js
```diff
@@ -51,24 +52,31 @@ class Scratch3MotionBlocks {
         util.target.setXY(x, y);
     }
 
-    goTo (args, util) {
+    getTarget (TO, util) {
```
rschamp: Would you please rename this function and its argument to `getTargetXY (targetName, util)`? I think that would be clearer.

rschamp (changes requested):

src/blocks/scratch3_motion.js
Kenny2github: *shrug* as you will!

Kenny2github:
@rschamp Done! Travis is still happy so that's a plus :)

Merge when you will.

rschamp (approved):

PR #668, opened 2017-08-22
Switch from parseInt to Math.floor

### LINKED ISSUE #601: Variable stops increasing eventually, but should go up to "infinity"
### Expected Behavior

The value should go up until it displays "infinity"; it shouldn't get stuck at a specific numerical value and stop increasing.

### Actual Behavior

The value gets fairly large quickly in the stack as shown, then slows down in its rate of increase. If you edit the value in the forever loop for how much the value is increasing by, you can lower the value to get the increase of the variable to speed up again. Eventually the value of the variable stops increasing at all, even if you lower the rate of change back to 1. However, you can get to a much larger value if you change what values are in the loop.

### Steps to Reproduce

- Add a new variable
- Drag out the blocks as shown in the screenshots for Stack 1
- Click to start the stack
- While the stack is active, edit the values in the pick random block to change how quickly the variable value increases
- Eventually the value will get "stuck"
- Then even if you lower the values in the pick random block, it will remain stuck
- If you change to Stack 2, note you can get to a much larger value
  - We should probably be consistent

**Stack 1**
screen shot 2017-06-12 at 1 20 56 pm
screen shot 2017-06-12 at 12 46 03 pm

**Stack 2**
Increasing the number of 0's in the divisor:
screen shot 2017-06-12 at 2 04 30 pm
screen shot 2017-06-12 at 2 29 32 pm

I may have hit some kind of limit eventually - after I dropped the block stack in the toolbox and it was deleted, I noticed the variable value was "infinity":
screen shot 2017-06-12 at 2 33 15 pm
screen shot 2017-06-12 at 2 31 43 pm

_Mac OS El Capitan, Chrome 59_

SillyInventor:
### Resolves

Resolves #601

### Proposed Changes

Switches from parseInt to Math.floor

### Reason for Changes

ParseInt cannot handle integers greater than 1e21 because such numbers are expressed in scientific notation. Consequently, it retrieves only the first digit. Math.floor performs truncation correctly.

### Test Coverage

None added.

### Special Note

Both parseInt and floor return NaN when asked to parse very large (or low) numbers because they are cast to infinity (or negative infinity). It's unclear if this is the desired behavior as it's consistent with scratch 2 but not technically correct. Perhaps a check should be added to provide a helpful error.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -85,7 +85,7 @@ class Scratch3OperatorsBlocks {
         if (low === high) return low;
         // If both arguments are ints, truncate the result to an int.
         if (Cast.isInt(args.FROM) && Cast.isInt(args.TO)) {
-            return low + parseInt(Math.random() * ((high + 1) - low), 10);
+            return low + Math.floor(Math.random() * ((high + 1) - low));
         }
         return (Math.random() * (high - low)) + low;
     }
@@ -107,7 +107,7 @@ class Scratch3OperatorsBlocks {
     length (args) {
         return Cast.toString(args.STRING).length;
     }
-    
+
     contains (args) {
         const format = function (string) {
             return Cast.toString(string).toLowerCase();


rschamp (approved):
Looks and works great, thanks @SillyInventor!

PR #669, opened 2017-08-25
Add pen transparency blocks

Kenny2github:
### Resolves

Resolves LLK/scratch-gui#602

### Proposed Changes

Adds the two new pen transparency blocks.

### Note

I didn't include any sort of wrapper around the possible value for the transparency (like change hue by does) since the only way I found to wrap the maximum was for HSV and this is RGB. Some advice on how to do this would be appreciated.

Comes with LLK/scratch-blocks#1055

### DIFF
--- a/src/blocks/scratch3_pen.js
+++ b/src/blocks/scratch3_pen.js
@@ -375,7 +375,7 @@ class Scratch3PenBlocks {
     }
     
     /**
-     * The pen "change pen transparency by {number}" block changes the "transparency" of the pen, related to the RGB value.
+     * The pen "change pen transparency by {number}" block changes the RGBA "transparency" of the pen.
      * @param {object} args - the block arguments.
      *  @property {number} TRANSPARENCY - the amount of desired transparency change.
      * @param {object} util - utility object provided by the runtime.
@@ -383,12 +383,12 @@ class Scratch3PenBlocks {
     changePenTransparencyBy (args, util) {
         const penState = this._getPenState(util.target);
         
-        penState.penAttributes.color4f[3] = penState.penAttributes.color4f[3] + (args.TRANSPARENCY / 255);
+        penState.penAttributes.color4f[3] = penState.penAttributes.color4f[3] + (args.TRANSPARENCY / 100);
         this._updatePenColor(penState);
     }
     
     /**
-     * The pen "set pen transparency to {number}" block sets the "transparency" of the pen, related to the RGB value.
+     * The pen "set pen transparency to {number}" block sets the RGBA "transparency" of the pen.
      * @param {object} args - the block arguments.
      *  @property {number} TRANSPARENCY - the amount of desired transparency change.
      * @param {object} util - utility object provided by the runtime.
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
If I understand your question about the wrapper for possible transparency values- we should just constrain the value to 0-100. Here's an [example](https://github.com/LLK/scratch-vm/blob/fc4f0735269d93a570edadcbca74ecab5d0ebbc8/src/blocks/scratch3_sound.js#L265) using our utility function MathUtil.clamp().

Kenny2github:
@ericrosenbaum I clamped the value for "set" but I'm not sure how to do it for "change" - how can I clamp the value? Is it possible to do `MathUtil.clamp(TRANSPARENCY, 0, 1);` and preserve the float (aka double) value?

ericrosenbaum:
The clamping should actually be in _updatePenColor, so it applies for both set and change.

I'm not sure about your question about numeric precision, when clamping 0 to 1- it seems like it should work fine but we should verify using the debugger.

@Kenny2github how are you testing your code? I noticed that this [line](https://github.com/Kenny2github/scratch-vm/blob/f61366643d7e36eb80ecd2d946a4a7c8bfbf9714/src/blocks/scratch3_pen.js#L186) will prevent any of these transparency changes from having an effect.

Kenny2github:
Clamping: Ah, I see.

The question was, in essence: will clamping 0.5 to 0 and 1 preserve 0.5 or convert it into an integer?

I'm not, really, I'm moving blind ._.
I'm not sure the best way I can go about testing (I don't have the access rights to install the necessary dependencies on what I'm using).

I'll try to fix those problems though!

Kenny2github:
I went with a different overall approach to the block:
1. Wrapping of the transparency value is now handled by a new function, `_wrapTransparency`, which wraps the value within the range 0, 100;
2. `_updatePenColor` now handles the transparency value instead of setting it to 1 (i.e. full opacity);
3. `DEFAULT_PEN_STATE` now includes a `transparency` value (100).
@ericrosenbaum you may want to take a look at this change with that keen bug-spotting eye of yours, if Travis fails but I can't find anything.

Kenny2github:
I went with a different overall approach to the block:

1. Wrapping of the transparency value is now handled by a new function, `_wrapTransparency`, which wraps the value within the range 0, 100;
2. `_updatePenColor` now handles the transparency value instead of setting it to 1 (i.e. full opacity);
3. `DEFAULT_PEN_STATE` now includes a transparency value (100).

@ericrosenbaum you may want to take a look at this change when you have time.

ericrosenbaum:
Thanks for your help on this @Kenny2github. There are still some minor issues, so I am going to finish this up by making a new PR that includes your commits plus my fixes. I hope that's cool. 

In the future, if there's a way you can set up a development environment to run your code, that would make your contributions most useful to us. Thanks!

ericrosenbaum:
I'm moving these changes to https://github.com/LLK/scratch-vm/pull/674

PR #671, opened 2017-08-31
Add methods to retrieve and update SVGs

fsih:
### Resolves
This change needs to go in after https://github.com/LLK/scratch-render/pull/160
It allows updating costume SVGs, which is a step toward having editable costumes

### Proposed Changes
Retrieve costumes from storage and update them in the renderer. Eventually we will need to update them in storage but that's a bigger task.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -341,6 +341,20 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Get an SVG string from the renderer.
+     * @param {int} costumeIndex - the index of the sound to be got.
+     * @return {string} the costume's SVG string, or null if it's not an SVG costume.
+     */
+    getCostumeSVG (costumeIndex) {
+        const id = this.editingTarget.sprite.costumes[costumeIndex].assetId;
+        if (id && this.runtime && this.runtime.storage &&
+                this.runtime.storage.get(id).dataFormat === 'svg') {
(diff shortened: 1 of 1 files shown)


src/virtual-machine.js
```diff
@@ -362,6 +362,40 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Get an SVG string from storage.
+     * @param {int} costumeIndex - the index of the sound to be got.
```
rschamp: Nit: `s/sound/costume/`

src/virtual-machine.js
```diff
@@ -362,6 +362,40 @@ class VirtualMachine extends EventEmitter {
...
+            const rotationCenter = [
+                costume.rotationCenterX / costume.bitmapResolution,
+                costume.rotationCenterY / costume.bitmapResolution
+            ];
+
+            this.runtime.renderer.updateSVGSkin(costume.skinId, svg, rotationCenter);
+        }
+        // TODO: Also update storage in addition to renderer. Without storage, if you switch
+        // costumes and switch back, you will lose your changes in the paint editor.
+        // TODO: emitTargetsUpdate if we need to update the storage ID on the updated costume.
```
rschamp: I think elsewhere we are using `@todo` which is picked up by jsdoc.

rschamp (approved):
Only nits. Looks good!

PR #675, opened 2017-09-06
Feature/extensions

### LINKED ISSUE #643: Register extension blocks with scratch-blocks & the VM
When an extension loads and registers with the extension system, its blocks should be registered with the VM and scratch-blocks so that:
- the user can find the new extension blocks in the toolbox / block palette and
- the blocks can run correctly when placed in a block stack in the project's code.

This should be implemented in such a way that we avoid / fix #580

See also: https://github.com/LLK/scratch-vm/wiki/Extensions-Proposal

- [x] Extension manager collects info from extension
- [x] Convert Scratch 3.0 extension info into scratch-blocks JSON & XML
- [x] scratch-{gui,blocks} collect block info from scratch-vm

cwillisf:
### Resolves

Along with LLK/scratch-gui#655, this resolves #643

### Proposed Changes

This change introduces an extension-worker framework to the VM, including:
- A way to request that a particular extension should be loaded
- A web worker-based extension "harness" to facilitate communication between the extension and the rest of the VM
- Code to convert Scratch 3.0 extension information into JSON and XML to be consumed by scratch-blocks
- A way for the scratch-gui to collect that JSON and XML for registration with scratch-blocks

Some features are not yet complete, including:
- There is not yet any way to translate text within extensions
- Boolean inputs are not yet supported
- Conditionals / branching blocks are implemented but may not be entirely correct

### Reason for Changes

See https://github.com/orgs/LLK/projects/6

### Test Coverage

No test coverage yet.

/cc @paulkaplan @ericrosenbaum


package.json
```diff
@@ -54,6 +54,7 @@
     "tap": "^10.2.0",
     "tiny-worker": "^2.1.1",
     "webpack": "^2.4.1",
-    "webpack-dev-server": "^2.4.1"
+    "webpack-dev-server": "^2.4.1",
+    "worker-loader": "0.8.1"
```
thisandagain, line 58: This should probably use a fuzzy (`^`) version number.

src/engine/runtime.js
```diff
@@ -321,6 +386,160 @@ class Runtime extends EventEmitter {
...
     /**
+     * Register the primitives provided by an extension.
+     * @param {ExtensionInfo} extensionInfo - information about the extension (id, blocks, etc.)
+     * @private
+     */
+    _registerExtensionPrimitives (extensionInfo) {
+        const categoryInfo = {
+            id: extensionInfo.id,
+            name: extensionInfo.name,
+            color1: '#FF6680',
```
thisandagain, line 397: If these color definitions are now going to come in through the VM should we find a good place to define / manage them globally?

package.json
cwillisf: I thought we were pinning exact versions for packages which have direct impact on the build output, which `worker-loader` does. Has that advice changed?

src/engine/runtime.js
cwillisf: Yep - ideally I'd like these to somehow come from the GUI but I haven't yet come up with a good way to do that -- actually I haven't put a whole lot of thought into it yet. Input welcome!

PR #676, opened 2017-09-07
Add sprite duplication method

paulkaplan:
Adds a method to the VM to duplicate a sprite

There are several parts to duplicating a sprite
- [ ] Copying over all the properties (I may have missed some?)
- [ ] Copying over the blocks
- [ ] Copying the sounds
- [ ] Copying the costumes

For sounds and costumes, I made the duplicates load their costumes ands sounds from storage based on the asset ID. That means (at least for now) that to work properly, the asset id needs to correspond to the up-to-date version of the asset in storage. 

I plan to accomplish this by having the GUI update storage with the encoded asset whenever it makes a change, creating a new asset ID. The GUI already has the encoders required, and is the single point where assets are modified. My one concern with this plan is that a new asset id will be created, and the old asset will not be removed from memory.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -32,12 +32,12 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
     return new Promise((resolve, reject) => {
         const imageElement = new Image();
         const onError = function () {
-                    // eslint-disable-next-line no-use-before-define
+            // eslint-disable-next-line no-use-before-define
             removeEventListeners();
             reject();
         };
         const onLoad = function () {
-                    // eslint-disable-next-line no-use-before-define
+            // eslint-disable-next-line no-use-before-define
             removeEventListeners();
             resolve(imageElement);
         };
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -2,47 +2,28 @@ const StringUtil = require('../util/string-util');
 const log = require('../util/log');
 
 /**
- * Load a costume's asset into memory asynchronously.
+ * Initialize a costume from an asset asynchronously.
  * Do not call this unless there is a renderer attached.
- * @param {string} md5ext - the MD5 and extension of the costume to be loaded.
  * @param {!object} costume - the Scratch costume object.
  * @property {int} skinId - the ID of the costume's render skin, once installed.
  * @property {number} rotationCenterX - the X component of the costume's origin.
  * @property {number} rotationCenterY - the Y component of the costume's origin.
  * @property {number} [bitmapResolution] - the resolution scale for a bitmap costume.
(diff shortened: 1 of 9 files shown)


src/virtual-machine.js
```diff
@@ -440,6 +440,31 @@ class VirtualMachine extends EventEmitter {
...
+            if (!sprite) {
+                throw new Error('No sprite associated with this target.');
+            }
+
+            target.duplicate().then(newTarget => {
+                this.runtime.targets.push(newTarget);
+                this.setEditingTarget(newTarget.id);
+            });
+        } else {
+            throw new Error('No target with the provided id.');
```
cwillisf: `if (target)` at the top makes sense if you're trying to avoid early-outs, but then there are two early-out `throw`s inside... I suggest sticking to one pattern or the other. Personally, I'd move this `throw` into an `if (!target)` near the top and un-indent the current body of `if (target)` but I'm OK with either pattern.

src/import/load-costume.js
```diff
@@ -25,55 +25,54 @@ const loadCostume = function (md5ext, costume, runtime) {
...
+        imageElement.addEventListener('error', onError);
+        imageElement.addEventListener('load', onLoad);
+        imageElement.src = costumeAsset.encodeDataURI();
+    }).then(imageElement => {
+        costume.skinId = runtime.renderer.createBitmapSkin(imageElement, costume.bitmapResolution, rotationCenter);
+        return costume;
+    });
 };
 
+loadCostume.loadCostumeFromAsset = loadCostumeFromAsset;
```
cwillisf: Hmm... putting a function on a non-constructor function feels a bit out-of-convention to me. Maybe we should make a `LoadCostume` object whose job is to contain the various (possibly static) methods used in loading costumes? Alternatively, maybe `module.exports = {loadCostume, loadCostumeFromAsset};`? (Kinda the same thing, but maybe more lightweight...)

src/import/load-sound.js
```diff
@@ -23,18 +23,22 @@ const loadSound = function (sound, runtime) {
...
+        {},
+        sound,
+        {data: soundAsset.data}
+    )).then(soundId => {
+        sound.soundId = soundId;
+        return sound;
+    });
+};
+
+loadSound.loadSoundFromAsset = loadSoundFromAsset;
```
cwillisf: I have similar comments here, compared to `loadCostume` above.

cwillisf (changes requested):
The functionality looks good to me... just a few style comments/questions.

src/virtual-machine.js
paulkaplan: Good call, I was kind of blindly working from the other methods in this file. Refactored in a885e7d

paulkaplan:
@cwillisf responded to all your comments

cwillisf (approved):

PR #679, opened 2017-09-22
Use unqualified paths in package.json scripts

aeons:
This allows stuff like yarn workspaces to function with node_modules hoisted.


rschamp (approved):
Thanks @aeons!

PR #681, opened 2017-09-27
Reduce clock precision

### LINKED ISSUE #659: Timer goes from 0 to 0.01 and back again if reset timer is in forever loop
### Expected Behavior
In Scratch 2 if you put a reset timer in a forever loop the timer stays in 0. See https://scratch.mit.edu/projects/172704222/

### Actual Behavior

Timer goes from 0 to 0.01 and back again multiple times.

### Steps to Reproduce

1. Put reset timer in a forever loop.
2. Start project.
3. Observe

Firefox and Mac OS

yongabyte:
### Resolves
fixed https://github.com/LLK/scratch-gui/issues/659 

### Proposed Changes
Limited clock accuracy to 1 decimal place by adding .tofix(1)

### Reason for Changes
in scratch 3, time was displayed to 1ms while in scratch 2, time is only displayed to 0.1 second. 

### Test Coverage
None

### DIFF
--- a/src/io/clock.js
+++ b/src/io/clock.js
@@ -17,7 +17,7 @@ class Clock {
         if (this._paused) {
             return this._pausedTime / 1000;
         }
-        return (this._projectTimer.timeElapsed() / 1000).toFixed(1);
+        return parseFloat((this._projectTimer.timeElapsed() / 1000).toFixed(1));
     }
 
     pause () {
--- a/src/io/clock.js
+++ b/src/io/clock.js
@@ -17,7 +17,7 @@ class Clock {
         if (this._paused) {
             return this._pausedTime / 1000;
         }
-        return this._projectTimer.timeElapsed() / 1000;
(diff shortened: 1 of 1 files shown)


thisandagain:
Thanks @yongabyte! I don't think we want to modify this in the actual project timer itself as that will likely impact performance characteristics across the entire VM, but rather in the opcode definition here:
https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_sensing.js#L150

thisandagain (changes requested):
Thanks @yongabyte! I don't think we want to modify this in the actual project timer itself as that will likely impact performance characteristics across the entire VM, but rather in the opcode definition here:
https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_sensing.js#L150

thisandagain:
Please feel free to open a new PR if you would like to take another shot at this! 😄

Issue #644, opened 2017-07-17
Expose Device Manager interface to extensions


cwillisf:
Hardware extensions will need access to the [Device Manager interface](https://github.com/LLK/scratch-vm/blob/develop/src/io/deviceManager.js). The API may need to change slightly in order to make it comfortable to use through the message dispatch system.

cwillisf:
This is no longer a priority since we introduced a way to load extensions in the main thread (without a worker sandbox).

We'll still need to do this before we can support third-party hardware extensions, but I'm putting it in the backlog for now.

PR #684, opened 2017-10-02
Allow workerless extensions

cwillisf:
### Resolves

This change lowers the priority of #644, such that we could wait until post-alpha to do that work. IMO we should do it sooner than that, but the advantage of this change is that internal work is no longer blocked on #644 and other similar items.

### Proposed Changes

This change adds a `_registerInternalExtension` method to the extension manager. This method is similar to the existing `loadExtensionURL` method, with two important differences:
- the new method takes an extension object instance (an object with a `getInfo()` method) instead of a URL, and
- the extension is registered as-is, likely in the same context as the extension manager, rather than sandboxing it in a web worker.

### Reason for Changes

This will allow us to convert the implementation of our existing blocks to use the extension system without having to build service interfaces first. We can later convert such extensions into "worker extensions" by implementing the necessary services, but that work is no longer prerequisite.

See 7fdaaaa25987eafdf9c841fcd79e05925aab6e5a to see a quick-and-dirty conversion of our pen blocks into an "internal" extension. After this conversion, the extension must be registered before use: `extensionManager._registerInternalExtension(new Scratch3PenBlocks(runtime));` would work.

### Test Coverage

A new integration test registers a small internal extension and calls a block implementation on that extension through the runtime. The test then verifies that the extension state is as expected -- and it does so in a way that wouldn't work in a worker context.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -459,10 +459,17 @@ class Runtime extends EventEmitter {
             const argInfo = blockInfo.arguments[placeholder] || {};
             const argTypeInfo = ArgumentTypeMap[argInfo.type] || {};
             const defaultValue = (typeof argInfo.defaultValue === 'undefined' ? '' : argInfo.defaultValue.toString());
+
+            // <value> is the ScratchBlocks name for a block input.
+            // The <shadow> is a placeholder for a reporter and is visible when there's no reporter in this input.
             inputList.push(`<value name="${placeholder}"><shadow type="${argTypeInfo.shadowType}">`);
+
+            // <field> is a text field that the user can type into. Some shadows, like the color picker, don't allow
+            // text input and therefore don't need a field element.
             if (argTypeInfo.fieldType) {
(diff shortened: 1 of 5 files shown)


src/engine/runtime.js
```diff
@@ -456,13 +459,11 @@ class Runtime extends EventEmitter {
...
-                  `<shadow type="${argTypeInfo.shadowType}">` +
-                    `<field name="${argTypeInfo.fieldType}">${defaultValue}</field>` +
-                  '</shadow>' +
-                '</value>'
-            );
+            inputList.push(`<value name="${placeholder}"><shadow type="${argTypeInfo.shadowType}">`);
+            if (argTypeInfo.fieldType) {
+                inputList.push(`<field name="${argTypeInfo.fieldType}">${defaultValue}</field>`);
+            }
+            inputList.push('</shadow></value>');
```
rschamp: Would you please add a comment explaining the conditional here?

rschamp (approved):
This is way simpler than I was imagining! Looks good other than some explanation.

PR #686, opened 2017-10-04
Remove label generation for extensions

paulkaplan:
### Resolves

Fixes the problem where there are duplicate labels when you add an extension in the GUI. 

### Proposed Changes

Don't manually generate labels when generating extension block xml. The blocks already handle this for all categories because of the all-in-one scrolling list.


rschamp (approved):

cwillisf (commented):
Ha! I was just about to do this, too :)

PR #691, opened 2017-10-05
Add say/think functionality to looks blocks

paulkaplan:
### Proposed Changes

Adds say/think bubble functionality. Depends on https://github.com/LLK/scratch-render/pull/177 and https://github.com/LLK/scratch-render/pull/176.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -280,7 +280,8 @@ class Runtime extends EventEmitter {
     }
 
     /**
-     * Event name for glowing the green flag
+     * Event name when threads start running.
+     * Used by the UI to indicate running status.
      * @const {string}
      */
     static get PROJECT_RUN_START () {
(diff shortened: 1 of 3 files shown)


src/engine/runtime.js
```diff
@@ -280,22 +280,33 @@ class Runtime extends EventEmitter {
     }
 
     /**
-     * Event name for glowing the green flag
+     * Event name when threads start running.
+     * Used by the UI to indicate running status.
```
cwillisf, line 284: 🥇

src/engine/runtime.js
```diff
@@ -1228,6 +1242,16 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Report that a new target has been created, possibly by cloning an existing target.
+     * @param {Target} newTarget - the newly created target.
+     * @param {Target} [sourceTarget] - the target used as a source for the new clone, if any.
+     * @fires Runtime#targetWasCreated
+     */
+    fireTargetWasRemoved (newTarget, sourceTarget) {
+        this.emit('targetWasRemoved', newTarget, sourceTarget);
```
cwillisf: Looks like the comment & arg list need to be updated here

cwillisf (changes requested):
Looks great! Just one C&P error :)

src/engine/runtime.js
paulkaplan: Good catch. Fixed in 
fa2db05

cwillisf (approved):
💬
:shipit:

PR #698, opened 2017-10-06
Changing parsing of custom procedures for new style

### LINKED ISSUE #697: Create two blocks when importing a procedure from SB2
Current XML for a custom block definition and call:
```
<xml xmlns="http://www.w3.org/1999/xhtml">
<variables>
</variables>
<block id="]){{Y!7N9ezN+j@Vr`8p" type="procedures_defnoreturn" x="47.175" y="91.96000000000001">
    <mutation proccode="say %s %n times if %b" argumentnames="[&quot;something&quot;,&quot;this many&quot;,&quot;this is true&quot;]" argumentdefaults="[&quot;&quot;,1,false]" warp="false"/>
    <next>
        <block id="tIVT2u+!X}IS8z4nu1x9" type="procedures_callnoreturn">
            <mutation proccode="say %s %n times if %b"/>
            <value name="input0">
                <shadow id="V0~M:TxRk0Ua%osjGzh," type="text">
                    <field name="TEXT"/>
                </shadow>
            </value>
            <value name="input1">
                <shadow id="uPx3si(KkbL1)-XpbXoQ" type="math_number">
                    <field name="NUM">1</field>
                </shadow>
            </value>
        </block>
    </next>
</block>
</xml>
```

New XML for the same block definition and call:

```
<xml xmlns="http://www.w3.org/1999/xhtml">
<variables>
</variables>
<block id="]){{Y!7N9ezN+j@Vr`8p" type="procedures_defnoreturn" x="47.175" y="91.96000000000001">
    <statement name="custom_block">
        <shadow type="procedures_callnoreturn_internal">
            <mutation proccode="say %s %n times if %b" argumentnames="[&quot;something&quot;,&quot;this many&quot;,&quot;this is true&quot;]" argumentdefaults="[&quot;&quot;,1,false]" warp="false"/>
        </shadow>
    </statement>
    <next>
        <block id="tIVT2u+!X}IS8z4nu1x9" type="procedures_callnoreturn">
            <mutation proccode="say %s %n times if %b"/>
            <value name="input0">
                <shadow id="V0~M:TxRk0Ua%osjGzh," type="text">
                    <field name="TEXT"/>
                </shadow>
            </value>
            <value name="input1">
                <shadow id="uPx3si(KkbL1)-XpbXoQ" type="math_number">
                    <field name="NUM">1</field>
                </shadow>
            </value>
        </block>
    </next>
</block>
</xml>
```

Looks like this now:
image

In the new definition the inner "say something" block is a shadow block of type `procedures_callnoreturn_internal`.  It is connected to the input `custom_block` on the `procedures_defnoreturn` block.  The inner block has the mutations that define the proccode, argument names, etc.  The first block in the definition is connected to the next connection of the `procedures_defnoreturn` block.

In the old definition the mutations are on the `procedures_defnoreturn` block, and there is no inner block.  The first block of the definition is connected to the next connection of the `procedures_defnoreturn` block.

Importing a procedure block should create a new block and attach it to the `custom_block` input [here](https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb2.js#L484-L490).

```
else if (oldOpcode === 'procDef') {
        // Mutation for procedure definition:
        // store all 2.0 proc data.
        const procData = sb2block.slice(1);
        activeBlock.mutation = {
            tagName: 'mutation',
            proccode: procData[0], // e.g., "abc %n %b %s"
            argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
            argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
            warp: procData[3], // Warp mode, e.g., true/false.
            children: []
        };
    } 
```

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/697 which allows us to display and run custom blocks correctly!

image

(the color change is actually a separate scratch-blocks change)

### Proposed Changes

Parse sb2 files according to the issue @rachel-fenichel filed. 

### Testing

@rschamp fixed tests!

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -148,6 +148,8 @@ const execute = function (sequencer, thread) {
     // Recursively evaluate input blocks.
     for (const inputName in inputs) {
         if (!inputs.hasOwnProperty(inputName)) continue;
+        // Do not evaluate the internal custom command block within definition
+        if (inputName === 'custom_block') continue;
         const input = inputs[inputName];
         const inputBlockId = input.block;
         // Is there no value for this input waiting in the stack frame?
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -225,7 +225,9 @@ class Sequencer {
             // Look for warp-mode flag on definition, and set the thread
             // to warp-mode if needed.
             const definitionBlock = thread.target.blocks.getBlock(definition);
-            const doWarp = definitionBlock.mutation.warp;
+            const innerBlock = thread.target.blocks.getBlock(
+                definitionBlock.inputs.custom_block.block);
+            const doWarp = innerBlock.mutation.warp;
             if (doWarp) {
                 thread.peekStackFrame().warpMode = true;
             } else if (isRecursive) {
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -481,14 +481,31 @@ const parseBlock = function (sb2block, getVariableId) {
         // Mutation for procedure definition:
         // store all 2.0 proc data.
         const procData = sb2block.slice(1);
-        activeBlock.mutation = {
-            tagName: 'mutation',
-            proccode: procData[0], // e.g., "abc %n %b %s"
-            argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
-            argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
-            warp: procData[3], // Warp mode, e.g., true/false.
-            children: []
+        // Create a new block and input metadata.
+        const inputUid = uid();
+        const inputName = 'custom_block';
+        activeBlock.inputs[inputName] = {
+            name: inputName,
+            block: inputUid,
+            shadow: inputUid
         };
+        activeBlock.children = [{
+            id: inputUid,
+            opcode: 'procedures_callnoreturn_internal',
+            inputs: {},
+            fields: {},
+            next: null,
+            shadow: true,
+            children: [],
+            mutation: {
+                tagName: 'mutation',
+                proccode: procData[0], // e.g., "abc %n %b %s"
+                argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
+                argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
+                warp: procData[3], // Warp mode, e.g., true/false.
+                children: []
+            }
+        }];
     } else if (oldOpcode === 'call') {
         // Mutation for procedure call:
         // string for proc code (e.g., "abc %n %b %s").
--- a/test/unit/engine_sequencer.js
+++ b/test/unit/engine_sequencer.js
@@ -156,7 +156,7 @@ test('stepToProcedure', t => {
 
     th.target.blocks.createBlock({
         id: 'internalId',
-        opcode: 'procedures_defnoreturn_internal',
+        opcode: 'procedures_callnoreturn_internal',
         mutation: {
             proccode: 'othercode'
         }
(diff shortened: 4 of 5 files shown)


src/serialization/sb2.js
```diff
@@ -481,14 +481,31 @@ const parseBlock = function (sb2block, getVariableId) {
...
-            argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
-            argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
-            warp: procData[3], // Warp mode, e.g., true/false.
-            children: []
+        // Create a new block and input metadata.
+        const inputUid = uid();
+        const inputName = 'custom_block';
+        activeBlock.inputs[inputName] = {
+            name: inputName,
+            block: inputUid,
```
rachel-fenichel, line 489: What's with the double use of inputUid?

src/engine/sequencer.js
```diff
@@ -225,7 +225,9 @@ class Sequencer {
             // Look for warp-mode flag on definition, and set the thread
             // to warp-mode if needed.
             const definitionBlock = thread.target.blocks.getBlock(definition);
-            const doWarp = definitionBlock.mutation.warp;
+            const innerBlock = thread.target.blocks.getBlock(
+                definitionBlock.inputs.custom_block.block);
```
rachel-fenichel, line 229: This is assuming that a whole chain of things exists (include a block connected to the input on the definition block).  What happens if it's missing?

rachel-fenichel (commented):
A few things here look fishy to me, but I may just be misunderstanding the scratch-vm code.  Let me know.

test/unit/engine_sequencer.js
```diff
@@ -153,9 +153,19 @@ test('stepToProcedure', t => {
     t.strictEquals(th.peekStack(), expectedBlock);
     s.stepToProcedure(th, 'faceCode');
     t.strictEquals(th.peekStack(), expectedBlock);
-    s.stepToProcedure(th, 'faceCode');
-    th.target.blocks.getBlock(th.stack[th.stack.length - 4]).mutation.proccode = 'othercode';
+
+    th.target.blocks.createBlock({
+        id: 'internalId',
+        opcode: 'procedures_defnoreturn_internal',
```
rachel-fenichel: Not a block type.  Did you mean `procedures_callnoreturn_internal`?

paulkaplan: Ah good catch. It isn't actually used in the code (only the outer block opcode is used) but good to have it there to make it clear what is going on.

src/engine/sequencer.js
paulkaplan: Yeah, fixed in follow up commit e8c5bbb

src/serialization/sb2.js
paulkaplan: I was following the style of the above sb2 import code, where it sets `block` to the same as `shadow` if there is no "real" block. I believe it is because the VM is expecting the `block` property to always exist, and only consults the `shadow` property when removing a block?

paulkaplan:
@rschamp can you review this?

rschamp (approved):
This all makes sense to me, exciting to see this working!

src/engine/execute.js
```diff
@@ -148,6 +148,8 @@ const execute = function (sequencer, thread) {
     // Recursively evaluate input blocks.
     for (const inputName in inputs) {
         if (!inputs.hasOwnProperty(inputName)) continue;
+        // Do not evaluate the internal custom command block within definition
+        if (inputName === 'custom_block') continue;
```
cwillisf, line 152: What do you think about adding an opcode check here? As-is, this would suppress an input named `custom_block` on an extension block, for example. Not that I expect that to be a common case -- it may never happen -- but this check does feel a bit imprecise to me.

cwillisf (approved):
Looks good, apart from one nitpick.

PR #700, opened 2017-10-07
Extensions: implement drop-down menus

### LINKED ISSUE #689: Extensions: drop-down menus
Implement drop-down menus for extensions.

For example, the `getInfo()` snippet below should display a drop-down menu. See [the extension spec](https://github.com/LLK/scratch-vm/wiki/Extensions-Proposal) for more detail, including more advanced examples.
```js
menus: {
    lessMore: ['<', '>']
},
blocks: [
    {
        opcode: 'whenDistance',
        text: 'when distance [OP] [REFERENCE]',
        blockType: BlockType.HAT,
        arguments: {
            OP: {
                type: ArgumentType.STRING,
                menu: 'lessMore',
                defaultValue: '<'
            },
            REFERENCE: {
                type: ArgumentType.NUMBER,
                defaultValue: 50
            }
        }
    }
]
```

cwillisf:
### Resolves

Resolves #689 

### Proposed Changes

This change implements basic drop-down menu support for extension blocks. Dynamic menus -- ones which collect menu items fresh each time the menu opens -- are not yet supported.

Since I was in the area, I also fixed Boolean inputs and added angle picker support.

### Reason for Changes

This is another step toward implementing the whole extension spec.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -33,6 +33,10 @@ const defaultBlockPackages = {
  */
 const ArgumentTypeMap = (() => {
     const map = {};
+    map[ArgumentType.ANGLE] = {
+        shadowType: 'math_angle',
+        fieldType: 'NUM'
+    };
     map[ArgumentType.COLOR] = {
         shadowType: 'colour_picker'
     };
@@ -389,6 +393,18 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Generate an extension-specific menu ID.
+     * @param {string} menuName - the name of the menu.
+     * @param {string} extensionId - the ID of the extension hosting the menu.
+     * @returns {string} - the constructed ID.
+     * @private
+     */
+    _makeExtensionMenuId (menuName, extensionId) {
+        /** @TODO: protect against XML characters in menu name */
+        return `${extensionId}.menu.${menuName}`;
+    }
+
+    /**
      * Register the primitives provided by an extension.
      * @param {ExtensionInfo} extensionInfo - information about the extension (id, blocks, etc.)
      * @private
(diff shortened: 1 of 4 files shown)


ericrosenbaum (approved):
Looks good!

The menu field's color looks like it needs adjusting, but we can fix that later.

src/engine/runtime.js
```diff
@@ -400,19 +415,71 @@ class Runtime extends EventEmitter {
...
+     * @param {array|string} menuItems - the list of menu items, or the name of an extension method to collect them.
+     * @param {CategoryInfo} categoryInfo - the category for this block
+     * @returns {object} - a JSON-esque object ready for scratch-blocks' consumption
+     * @private
+     */
+    _buildMenuForScratchBlocks (menuName, menuItems, categoryInfo) {
+        const menuId = this._makeExtensionMenuId(menuName, categoryInfo.id);
+
+        /** @TODO: support dynamic menus when 'menuItems' is a method name string (see extension spec) */
+        const options = menuItems.map(item => {
```
rschamp: Seeing as this will be encountered frequently by developers, it might be nice to put some validation on the `menuItems` type and throw a warning if it's not an array/function that returns an array?

src/engine/runtime.js
```diff
@@ -400,19 +415,71 @@ class Runtime extends EventEmitter {
...
+        return {
+            json: {
+                message0: '%1',
+                type: menuId,
+                inputsInline: true,
+                output: 'String',
+                colour: categoryInfo.color1,
+                colourSecondary: categoryInfo.color2,
+                colourTertiary: categoryInfo.color3,
+                outputShape: ScratchBlocksConstants.OUTPUT_SHAPE_ROUND,
```
rschamp, line 476: Should all menus be round, or are we going to support droppable vs non-droppable (square, don't accept blocks) dropdowns?

rschamp (approved):

src/engine/runtime.js
cwillisf: Good question. My initial instinct is to make all menus accept blocks until someone complains :)
We could certainly add a flag controlling that. I'm not sure if that flag would make the most sense on the block or menu...

towerofnix: Can I complain? :)

I mean, my honest opinion is that non-droppable menus should be done away with entirely, so that every menu is programmable - but I doubt that's going to happen (and there's reasons for it not to), so..

But practically I think it makes sense for extensions to be able to do what normal Scratch blocks can do. It seems like it should be possible for every block in the entire Scratch programming language to be converted to extensions (or at least "builtin extensions").

PR #703, opened 2017-10-11
⭐️ Fix the infinite loop in say/think ⭐️

### LINKED ISSUE #702: Long string in "say" block causes an error
### Expected Behavior

You should be able to enter a long string into the say block and get a reasonable behavior, such as the say bubble rendering truncated text. 

### Actual Behavior

The GUI locks up if you enter text that is more than a few words long. 

### Steps to Reproduce

In the "say" block, type something like: "word word word word word word".

Definitely do not paste in the lyrics, in their entirety, to the classic song "All Star" by Smash Mouth. 

Mac OS 10.12.6 Chrome 61

paulkaplan:
### Resolves

Fixes the infinite loop that results if.... (deep breath)... you try to say/think the chorus to "all star" by smash mouth. Strangely specific bug report? Yup, thank @ericrosenbaum 

Only flip the bubble if it would fit the other way. 

Fixes https://github.com/LLK/scratch-vm/issues/702

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -125,7 +125,11 @@ class Scratch3LooksBlocks {
         } else {
             this.runtime.renderer.updateDrawableProperties(bubbleState.drawableId, {
                 position: [
-                    bubbleState.onSpriteRight ? targetBounds.right : targetBounds.left - bubbleWidth,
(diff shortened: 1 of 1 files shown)


bates64:
🤔

cwillisf (approved):
🌀

Issue #696, opened 2017-10-06
Add property in extension spec for hats: edgeActivated


ericrosenbaum:
### Expected Behavior

You should be able to specify that an extension hat block is activated on an edge (i.e. it activates when its function, which is called every frame, returns true after returning false) or on an event (i.e. it activates when its function is called).

### Actual Behavior

It's not available yet.

thisandagain:
@cwillisf @ericrosenbaum Is this blocking? Putting in the October milestone but feel free to push back to November or December if it is not blocking progress on the extensions playtest.

ericrosenbaum:
If I understand correctly, the default behavior is edgeActivated, which we use for most things, so this is not blocking anything as far as I know. I'll push it back to November for now.

PR #713, opened 2017-10-13
Fix extension hats

### LINKED ISSUE #705: Extension hat blocks not working
### Expected Behavior

Extension blocks that are hats should activate on an edge as expected. 

### Actual Behavior

The Lego WeDo 2.0 extension hats "When distance < 50" and "When tilted any" do not activate.

### Steps to Reproduce

- Load the current extension prototype: https://ericrosenbaum.github.io/scratch-gui/extensions-oct2017/
- Add the Lego WeDo extension
- Connect a Lego WeDo (note that this requires an edit to the Scratch Device Manager config file)
- Make a stack such as "when distance < 50, play sound meow" 
- Note that it does not trigger when the distance drops below 50

cwillisf:
### Resolves

Resolves #705 

### Proposed Changes

These changes make extension hat blocks function. For now they are assumed to be edge-activated, but see #696.

This PR also includes a fix for what seems to be a long-standing issue affecting hats with more than one input. In this situation, the code to collect hat inputs would permanently assign new values to the block's `fields` list, causing problems in the runtime later on. This was a tricky bug to find since the symptom is so distant from the cause; this might be an argument for using `Object.freeze` and similar features.

As a side effect of fixing extension hat blocks I noticed that the WeDo 2.0 `whenDistance` block wasn't scaled correctly, so I fixed that too. I also put in a workaround for #580.

### Reason for Changes

Further progress on extensions & preparation for Monday's playtest.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -409,6 +409,9 @@ class Runtime extends EventEmitter {
             const opcode = convertedBlock.json.type;
             categoryInfo.blocks.push(convertedBlock);
             this._primitives[opcode] = convertedBlock.info.func;
+            if (blockInfo.blockType === BlockType.HAT) {
+                this._hats[opcode] = {edgeActivated: true}; /** @TODO let extension specify this */
+            }
         }
 
         this.emit(Runtime.EXTENSION_ADDED, categoryInfo.blocks);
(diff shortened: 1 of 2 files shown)


src/engine/runtime.js
```diff
@@ -802,13 +805,14 @@ class Runtime extends EventEmitter {
...
             if (Object.keys(hatFields).length === 0) {
+                hatFields = {}; // don't overwrite the block's actual fields list
                 const hatInputs = blocks.getInputs(block);
                 for (const input in hatInputs) {
                     if (!hatInputs.hasOwnProperty(input)) continue;
                     const id = hatInputs[input].block;
                     const inpBlock = blocks.getBlock(id);
                     const fields = blocks.getFields(inpBlock);
-                    hatFields = Object.assign(fields, hatFields);
+                    Object.assign(hatFields, fields);
```
cwillisf, line 815: My guess is the original line here was written thinking that it would act like `hatFields = Object.assign({}, fields, hatFields)`. This new code should be equivalent except that it throws away fewer intermediate objects in cases where there are multiple inputs. In my opinion it's also easier to understand: in particular, it's now obvious that `Object.assign` is expected to modify one of the objects.

ericrosenbaum (approved):
I tip my hat to you sir!

PR #718, opened 2017-10-13
Update storage when the SVG updates

fsih:
### Resolves

This will allow drawing in the paint editor to update storage and thus save new assets.


rschamp (approved):
🎉

PR #722, opened 2017-10-17
Playground: use UMD to load storage as global

### LINKED ISSUE #719: Missing Scratch.Storage in Playground
### Expected Behavior

It should be possible to run Playground

### Actual Behavior

It's not possible to run Playground because of the javascript error 'Uncaught TypeError: Scratch.Storage is not a constructor'

### Steps to Reproduce

Open Playground in a browser

61.0.3163.100 (Official Build) (64-bit) on Linux

cwillisf:
### Resolves

Resolves #719 

### Proposed Changes

This change causes the playground to load `scratch-storage` as an independent script and relies on its UMD loader to expose it as a global. Bonus points: we get better source mapping this way.

### Reason for Changes

It appears that `expose-loader` is not compatible with WebPack's UMD exporter, so the VM playground broke when I converted `scratch-storage`'s web build to UMD.

### Test Coverage

The playground is not covered by our tests.

### DIFF
--- a/src/playground/index.html
+++ b/src/playground/index.html
@@ -66,7 +66,9 @@
 
     <!-- FPS counter, Syntax highlighter, Blocks, Renderer -->
     <script src="./vendor.js"></script>
-    <!-- VM Worker -->
+    <!-- Storage module -->
+    <script src="./scratch-storage.js"></script>
+    <!-- VM -->
     <script src="./scratch-vm.js"></script>
     <!-- Playground -->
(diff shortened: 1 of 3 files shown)


cwillisf:
Hmm... it appears that the "complex" integration test failed because of a TLS error. Trying again...

rschamp (commented):
Thanks a lot @cwillisf!

rschamp (approved):
Oops, approved!

PR #724, opened 2017-10-18
Prevent duplicate extensions

### LINKED ISSUE #694: Prevent loading two copies of the same extension
### Expected Behavior

The user should be prevented from loading a second copy of a particular extension.

### Actual Behavior

Loading a second copy of an extension causes duplicate blocks and other internal runtime/VM issues.

### Steps to Reproduce

1. Load an extension using the extension library
2. Load the same extension again
3. Note the duplicate blocks

cwillisf:
### Resolves

Resolves #694 

### Proposed Changes

This change prevents two extensions / extension instances from being registered under the same extension ID (from `getInfo`). For internal extensions this is relatively easy: we just unregister the service when we find out it's a duplicate. For a worker-based extension, though, we don't find out about the duplicate until the extension worker has started up and the extension object has been instantiated in that worker.

Since we can't be sure that this all happens with predictable timing -- especially if an extension doesn't register before the extension framework starts up -- I chose to handle the situation with a watchdog. Any duplicate extension ID registration coming from a worker is ignored, and if no extension has been successfully registered by a worker for a specific amount of time (currently 5 seconds) then the worker is terminated. This also guards against some probably-would-never-happen-but-maybe edge cases, such as two extensions being registered from one worker where one of the extensions duplicates the ID of a third extension, provided by another worker.

If the GUI knows an extension ID, it can call `isExtensionLoaded` to short-circuit all of the above at the UI layer.

### Reason for Changes

Without these changes it's possible to load multiple copies of the same extension, leading to duplicate blocks and mass hysteria.

### DIFF
--- a/src/dispatch/central-dispatch.js
+++ b/src/dispatch/central-dispatch.js
@@ -38,7 +38,7 @@ class CentralDispatch extends SharedDispatch {
     /**
      * Set a local object as the global provider of the specified service.
      * WARNING: Any method on the provider can be called from any worker within the dispatch system.
-     * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui'.
+     * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui', 'extension9'.
      * @param {object} provider - a local object which provides this service.
      * @returns {Promise} - a promise which will resolve once the service is registered.
      */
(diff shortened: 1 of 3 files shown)


rschamp (changes requested):
Off thread we discussed splitting this into a separate PR for the "watchdog" functionality, and simplifying the dedupe functionality to prevent registration altogether by checking against which extension URLs have been loaded.

cwillisf:
I've updated this branch and the PR description with a simpler VM-side duplication prevention mechanism.

@rschamp, since we probably don't need the watchdog system for now do you think it would still make sense to make a separate PR for that? I'm happy to have it hang out in the Git history like this -- I added a comment with the relevant commit hash so if/when we want to bring it back, we can.

ericrosenbaum (approved):
It works!

PR #725, opened 2017-10-18
Set/change pen color blocks with color, brightness, saturation, transparency

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-blocks/issues/1091
Resolves https://github.com/LLK/scratch-gui/issues/783

And progress on https://github.com/LLK/scratch-gui/issues/725

Note that testing these changes will require pulling a PR on Scratch-gui that updates the extension library modal: https://github.com/LLK/scratch-gui/pull/793

### Proposed Changes

Change the way the pen color is set to use a menu of four color parameters: color, brightness, saturation and transparency. All are scaled 0-100. Color is wrapped, the rest are clamped.

### Reason for Changes

The earlier version of pen color blocks used different blocks with different ranges for the color parameters (and the "shade" block combined saturation and brightness). This change simplifies the color parameters by putting them all into the same range. It also reduces the total number of pen blocks by putting the four parameters into a menu.

### DIFF
--- a/src/blocks/scratch3_pen.js
+++ b/src/blocks/scratch3_pen.js
@@ -13,7 +13,7 @@ const RenderedTarget = require('../sprites/rendered-target');
  * @enum {string}
  */
 const ColorParam = {
-    HUE: 'hue',
+    COLOR: 'color',
     SATURATION: 'saturation',
     BRIGHTNESS: 'brightness',
     TRANSPARENCY: 'transparency'
@@ -22,7 +22,7 @@ const ColorParam = {
 /**
  * @typedef {object} PenState - the pen state associated with a particular target.
  * @property {Boolean} penDown - tracks whether the pen should draw for this target.
- * @property {number} hue - the current hue of the pen.
+ * @property {number} color - the current color (hue) of the pen.
  * @property {number} shade - the current shade of the pen.
  * @property {PenAttributes} penAttributes - cached pen attributes for the renderer. This is the authoritative value for
  *   diameter but not for pen color.
@@ -68,7 +68,7 @@ class Scratch3PenBlocks {
     static get DEFAULT_PEN_STATE () {
         return {
             penDown: false,
-            hue: 33,
+            color: 33,
             saturation: 100,
             brightness: 100,
             shade: 50,
@@ -185,7 +185,7 @@ class Scratch3PenBlocks {
         }
     }
 
-    _wrapHue (value) {
+    _wrapColor (value) {
         return MathUtil.wrapClamp(value, 0, 100);
     }
 
@@ -267,7 +267,7 @@ class Scratch3PenBlocks {
                         COLOR_PARAM: {
                             type: ArgumentType.STRING,
                             menu: 'colorParam',
-                            defaultValue: ColorParam.HUE
+                            defaultValue: ColorParam.COLOR
                         },
                         VALUE: {
                             type: ArgumentType.NUMBER,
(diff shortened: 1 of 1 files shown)


paulkaplan:
Lint is failing
```
/home/travis/build/LLK/scratch-vm/src/blocks/scratch3_pen.js
  324:1   error  Expected indentation of 24 spaces but found 20  indent
  415:13  error  'rgb' is never reassigned. Use 'const' instead  prefer-const
  436:1   error  Expected indentation of 8 spaces but found 12   indent
  437:1   error  Expected indentation of 12 spaces but found 16  indent
  438:1   error  Expected indentation of 12 spaces but found 16  indent
  439:1   error  Expected indentation of 8 spaces but found 12   indent
  440:1   error  Expected indentation of 12 spaces but found 16  indent
  441:1   error  Expected indentation of 12 spaces but found 16  indent
  442:1   error  Expected indentation of 8 spaces but found 12   indent
  443:1   error  Expected indentation of 12 spaces but found 16  indent
  444:1   error  Expected indentation of 12 spaces but found 16  indent
  445:1   error  Expected indentation of 8 spaces but found 12   indent
  446:1   error  Expected indentation of 12 spaces but found 16  indent
  447:1   error  Expected indentation of 12 spaces but found 16  indent
  448:1   error  Expected indentation of 8 spaces but found 12   indent
  449:1   error  Expected indentation of 12 spaces but found 16  indent
```

ericrosenbaum:
Note that we are hoping to do a workshop on Wednesday Oct 25th using this version on develop, so lets try to merge and deploy by then if possible

paulkaplan (approved):
Code looks good to me, and confirmed it plays nice with GUI (that PR mentioned in the description is already merged). 

@cwillisf I think it is probably important for you to take a look at this also though.

chrisgarrity:
@cwillisf please /cc me when this is ready so I know if we can use it on Wednesday for the workshop. Thanks!

thisandagain (approved):
This looks good to me and is working well locally. Thanks @ericrosenbaum

src/blocks/scratch3_pen.js
```diff
@@ -171,44 +184,22 @@ class Scratch3PenBlocks {
...
     /**
-     * Clamp a pen transparency value to the range (0,100).
-     * @param {number} value - the pen transparency value to be clamped.
+     * Clamp a pen color parameter to the range (0,100).
+     * @param {number} value - the value to be clamped.
      * @returns {number} the clamped value.
      * @private
      */
-    _clampTransparency (value) {
+    _clampColorParam (value) {
```
cwillisf, line 202: prio-low: the names `_wrapColorInput` and `_clampColorParam` are very similar and don't immediately communicate the difference between the two, other than wrap vs. clamp. If that's the only intended difference then I would prefer consistency between "input" and "param" in these names. I don't think this needs to be fixed immediately.

cwillisf (commented):
The code itself looks good (with one minor comment on naming or documentation) but I'm concerned about compatibility. It seems like this will break compatibility with pen projects from Scratch 2.0, especially with the change to the wrapping behavior. Is there a plan for that?

src/blocks/scratch3_pen.js
ericrosenbaum: Sorry for merging this before getting your comments!

re: compatibility, I have on my list (but not in an issue yet) that we will need to update the sb2 importer and add back in the scratch 2.0 implementations (including different wrapping and clamping behavior, the shade block, and accepting hex values). Then, once we have auto-loading of then pen extension for projects containing pen blocks, they should work.

I'm not totally sure what to do about the wrap and clamp function names. We will need at least two more variants of wrapping and clamping to different ranges when we implement the 2.0 style pen color blocks, so maybe we can think about the naming scheme again then. The use of "color" as one of the color parameters complicates things- I was thinking about switching to using "hue" in the code, but I tried to just keep it simple and match what the user   sees.

PR #729, opened 2017-10-23
Fix several say/think bugs

### LINKED ISSUE #728: Say/think bubbles stop moving with sprite
### Expected Behavior

Say / Think bubbles always stay attached to the sprite they originated from (See: #691)

### Actual Behavior

Say / Think bubble can become fixed in their position and not move with their sprite / clone

### Steps to Reproduce

* Create a stack: When Green Flag Clicked -> Move (10) steps -> Say "Hello!" for (2) seconds.
* Click the green flag a couple of times
* Move the sprite and watch the say block stay where it was.

image

If it helps, make the delay larger from 2 seconds :). I kind of feel the issue is to do with the promise on a timer and a new say block being instanced which the promise is still active? Perhaps the promise needs cancelling when a new say or think block is executed? However, playing around with this, it is really rather unpredictable :)

Also getting this exception in the console:

image

This test setup shows how the say until needs cancelling after another say:

image

And it also has the same issue with the "say for 2" not moving with the sprite, but the second "say for 2" actually moving ok weirdly...

Chrome latest on windows 10

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/728

@griffpatch thanks very much for reporting those issues. This PR contains commits to fix that memory leak with event listeners, fixes up the show/hide timeout through green flag clicks and fixes the moving bug. Tiny fixes individually, but together led to some very strange behavior.

Thank you for the thorough repro instructions, that really helps with these bugs.  

For the reviewer, this is one you want to review one commit at a time. The commit description contains more info for each fix.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -6,7 +6,7 @@ const RenderedTarget = require('../sprites/rendered-target');
  * @typedef {object} BubbleState - the bubble state associated with a particular target.
  * @property {Boolean} onSpriteRight - tracks whether the bubble is right or left of the sprite.
  * @property {?int} drawableId - the ID of the associated bubble Drawable, null if none.
- * @property {Boolean} drawableVisible - if drawable has been hidden by blank text.
+ * @property {Boolean} drawableVisible - false if drawable has been hidden by blank text.
  *      See _renderBubble for explanation of this optimization.
  * @property {string} text - the text of the bubble.
  * @property {string} type - the type of the bubble, "say" or "think"
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -8,7 +8,6 @@ const RenderedTarget = require('../sprites/rendered-target');
  * @property {?int} drawableId - the ID of the associated bubble Drawable, null if none.
  * @property {string} text - the text of the bubble.
  * @property {string} type - the type of the bubble, "say" or "think"
- * @property {Boolean} visible - whether the bubble is hidden along with its sprite.
  */
 
 class Scratch3LooksBlocks {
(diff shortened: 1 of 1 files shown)


rschamp (changes requested):
Discussed offline, there's still some "unique" visibility behavior, so @paulkaplan is going to look into refactoring the handling of the `visible` state.

paulkaplan:
@rschamp thanks for talking through this. I removed the `visible` state, but then realized what it was originally for: it is an optimization specifically targeted at making "typewriter" style projects work without "flashing" of the bubble by removing it and recreating it. 

I added that back in in the last commit, but what I think is a more stable and sane way. The optimization is only used if the skin has already been initialized, so the relevant code is confined to the single "update" branch of the `renderBubble` function.

rschamp (approved):

src/blocks/scratch3_looks.js
```diff
@@ -6,9 +6,10 @@ const RenderedTarget = require('../sprites/rendered-target');
  * @typedef {object} BubbleState - the bubble state associated with a particular target.
  * @property {Boolean} onSpriteRight - tracks whether the bubble is right or left of the sprite.
  * @property {?int} drawableId - the ID of the associated bubble Drawable, null if none.
+ * @property {Boolean} drawableVisible - if drawable has been hidden by blank text.
+ *      See _renderBubble for explanation of this optimization.
```
cwillisf: Based on the description here it sounds like `drawableVisible` being true means that the Drawable is hidden. Based on the name of the property and the logic in `_renderBubble` it looks like `drawableVisible` being true means that the Drawable is visible. Unless I'm misunderstanding something, could you adjust for consistency? (Maybe just `false if drawable has been hidden by blank text.`?)

cwillisf (changes requested):
One nitpick / misunderstanding, but overall LGTM!

src/blocks/scratch3_looks.js
paulkaplan: Oh good catch, sorry I went back and forth adjusting these property names about a dozen times, must have got messed up in the shuffle. Fixed in c14155c

PR #730, opened 2017-10-25
Fix undefined procedure execution

### LINKED ISSUE #809: Uncaught TypeError: Cannot read property 'length' of null
### Expected Behavior

- Project should start on green flag press

### Actual Behavior

- Throws exception

```
lib.min.js:17 Uncaught TypeError: Cannot read property 'length' of null
    at e.value (lib.min.js:17)
    at e (lib.min.js:17)
    at e.value (lib.min.js:17)
    at e.value (lib.min.js:17)
    at t.value (lib.min.js:17)
    at lib.min.js:17
```

Needs investigation, but this appears to be related to the execution of custom procedures within the VM. This may be a useful test case for the next iteration.

### Steps to Reproduce

- Load either
    - https://llk.github.io/scratch-gui/#178966496
    - https://llk.github.io/scratch-gui/develop#178966496
- Wait for project to load
- Press "green flag" button

Mac OS 10.12.6 Chrome Version 61.0.3163.100 (Official Build) (64-bit)

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/809, where a procedure is trying to be run without an accompanying define block. Generally you can't get into this state (deleting a procedure requires no other instances of the call block), but you can do it by dragging the custom block from one sprite to another, as happened in that project. 

### Proposed Changes

Like scratch 2, just noop if that happens.

### Reason for Changes

So that these projects can run without crashing!

### Test Coverage

Added a new test to the procedure block documenting this behavior

### DIFF
--- a/src/blocks/scratch3_procedures.js
+++ b/src/blocks/scratch3_procedures.js
@@ -27,11 +27,20 @@ class Scratch3ProcedureBlocks {
         if (!util.stackFrame.executed) {
             const procedureCode = args.mutation.proccode;
             const paramNames = util.getProcedureParamNames(procedureCode);
+
+            // If null, procedure could not be found, which can happen if custom
+            // block is dragged between sprites without the definition.
+            // Match Scratch 2.0 behavior and noop.
+            if (paramNames === null) {
+                return;
+            }
+
             for (let i = 0; i < paramNames.length; i++) {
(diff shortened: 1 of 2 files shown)


thisandagain (approved):
🎉 🌮 🐱

PR #737, opened 2017-10-30
Fix say/think with non-string inputs

### LINKED ISSUE #817: “say” block can't accept number input
screen shot 2017-10-29 at 10 09 01 pm

### Expected Behavior
show “4” in stage

### Actual Behavior

nothing happened

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/817

### Proposed Changes

`String( <> )` things before say/thinking them.

### Reason for Changes

These kinds of blocks didn't work before, now they do!

image

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -240,7 +240,7 @@ class Scratch3LooksBlocks {
 
     say (args, util) {
         // @TODO in 2.0 calling say/think resets the right/left bias of the bubble
-        this._updateBubble(util.target, 'say', args.MESSAGE);
+        this._updateBubble(util.target, 'say', String(args.MESSAGE));
     }
 
     sayforsecs (args, util) {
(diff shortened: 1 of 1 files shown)


src/engine/blocks.js
```diff
@@ -316,6 +317,14 @@ class Blocks {
                 if (variable) {
                     block.fields[args.name].value = variable.name;
                     block.fields[args.name].id = args.value;
+                } else {
```
fsih: This whole variable logic is super weird, and I feel like it would be easy to miss bugs. Is it possible to change the source of this information to provide information in a cleaner way, where it always provides an ID and the value field isn't different things?

fsih (changes requested):

src/engine/blocks.js
paulkaplan: Oh jeez sorry this is totally unrelated, didn't mean for it to be part of this PR.

paulkaplan:
@fsih sorry about that, resubmitted with the right code.

fsih (approved):

PR #739, opened 2017-10-31
Register extension with URL and category name

ericrosenbaum:
### Resolves

This is support for https://github.com/LLK/scratch-gui/issues/821

### Proposed Changes

When emitting the EXTENSION_ADDED event, in addition to info about the blocks themselves, include the category name and URL.

### Reason for Changes

When the GUI receives this event, it can store the extension URL, in order to prevent loading duplicate copies of the extension. The GUI can also use the category name to ask the blocks flyout to scroll to that category.   

### Test Coverage

No tests added.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -447,7 +447,11 @@ class Runtime extends EventEmitter {
             }
         }
 
-        this.emit(Runtime.EXTENSION_ADDED, categoryInfo.blocks.concat(categoryInfo.menus));
+        const blocksInfo = categoryInfo.blocks.concat(categoryInfo.menus);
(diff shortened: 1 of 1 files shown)


cwillisf (changes requested):
I think I'd prefer making these three different properties on a simple object, rather than two extra properties tacked onto an `Array` object. Something like this:
```js
this.emit(Runtime.EXTENSION_ADDED, {
  blocks: ...,
  extensionURL: ...,
  categoryName: ...
});
```
...or however you'd like to assemble it.

That'll require changes to places which receive this event but I think it'll be easier to understand & deal with, for example when debugging.

PR #745, opened 2017-11-01
Support legacy hue and shade pen blocks

### LINKED ISSUE #736: Pen extension should support 2.0 pen blocks
### Expected Behavior

Scratch 3.0 should be able to load projects that use 2.0 pen blocks.

### Actual Behavior

Loading a 2.0 project with pen blocks does not work.

There's a set of Scratch 2.0 pen blocks that the 3.0 pen extension needs to provide compatibility support for (note that the first two look similar to blocks in 3.0, but have a different range, so we need to provide distinct legacy opcodes): 

- change pen color by
- set pen color to 
- change shade by
- set shade to

Using these legacy opcodes will require auto-loading the extension (https://github.com/LLK/scratch-vm/issues/734) and including hidden legacy blocks in the extension (https://github.com/LLK/scratch-vm/issues/735). 

The fix will require changes to the sb2 specmap, and adding the legacy opcode implementations to the pen extension.

paulkaplan:
_Paired with @ericrosenbaum on this._

### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/736

### Proposed Changes

Adds 4 legacy pen block definitions to the pen extension.

The `set/change hue` blocks can use the new HSV color model directly as `hue` is identically twice what the new `color` property is. 

The shade blocks are somewhat trickier. For the `set shade to <number>` block, we can use the old scratch 2 shade model code to create an RGB color based on the current hue and the given shade, and then decompose that to HSV.

For the `change shade by <number>`, I unfortunately needed to store the `shade` information in the color state because I couldn't reliably extract shade from HSV, and because the "ping-pong" wrapping of shade makes it impossible to correctly implement without some state tracking. So I stored an extra value in `penState._shade`, and update it in two places. First is when you use `set pen color to color`. This stores a rough estimate of shade (`brightness / 2`), which is the same as the scratch 2 behavior. The other is when shade is explicitly updated with `set/change shade`. 

Scratch 2 pen projects should work the same way they used to. One limitation is that using the new `set pen <saturation/brightness>` blocks will not update the stored shade value, as they are incompatible. I think that this is an acceptable limitation. 

---

The second commit hides the new legacy blocks from the palette. I introduced a new flag because I was unsure about the future of the `filter` attribute on extension blocks. @cwillisf let me know your thoughts on this.

### DIFF
--- a/src/blocks/scratch3_pen.js
+++ b/src/blocks/scratch3_pen.js
@@ -71,6 +71,7 @@ class Scratch3PenBlocks {
             saturation: 100,
             brightness: 100,
             transparency: 0,
+            _shade: 50, // Used only for legacy `change shade by` blocks
             penAttributes: {
                 color4f: [0, 0, 1, 1],
                 diameter: 1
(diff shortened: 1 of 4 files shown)


ericrosenbaum (approved):
LGTM!

PR #753, opened 2017-11-02
Provide argument id list with callers and procdef blocks

### LINKED ISSUE #747: Scratch-vm should provide argument IDs when serializing procedure mutators
Context from scratch-blocks: https://github.com/LLK/scratch-blocks/issues/1195

Scratch Blocks is going to keep blocks connected to the correct inputs when procedure calls are mutated.  To do so, arguments need IDs as well as names and default values.  These are provided by scratch-vm.  The next time I land procedure code it's going to be looking for an `argumentid` attribute on the mutator.  Here's an example mutator: 
```
<mutation proccode="say %s %n times if %b" argumentnames="[&quot;something&quot;,&quot;this many&quot;,&quot;this is true&quot;]" argumentdefaults="[&quot;&quot;,1,false]" warp="false" argumentids="[&quot;a42&quot;, &quot;b23&quot;, &quot;c99&quot;]"/>
```

Scratch-vm needs to provide IDs by then, to be able to load procedure blocks without crashing.  

For now it's fine to use the name as the ID.  Longer term that doesn't work, because renaming an argument would make it lose its connected block.

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/747, unblocking scratch-blocks custom procedure work.

### Proposed Changes

Use the input "names" as argument ids for both the caller and the procdef mutations. This is required to allow moving around procedure parameters. Since args are purely positional in scratch 2, we can use anything we want to create these ids. Use the existing procedure parsing code because it is easy.


rachel-fenichel (approved):
LG

PR #754, opened 2017-11-03
Extension block icons (including pen icon)

ericrosenbaum:
### Proposed Changes

Add an extension-specific icon to the left side of each extension block. 

This PR includes adding the icon for the pen extension.

Note that the blocks rendering for extensions needs to be changed to bring the extension blocks up to spec (see https://github.com/LLK/scratch-blocks/issues/1207)

### Reason for Changes

These icons help to visually distinguish extensions from each other, and communicate what they do.

### DIFF
--- a/src/blocks/scratch3_pen.js
+++ b/src/blocks/scratch3_pen.js
@@ -8,6 +8,13 @@ const RenderedTarget = require('../sprites/rendered-target');
 const log = require('../util/log');
 
 /**
+ * Icon svg to be displayed at the left edge of each extension block, encoded as a data URI.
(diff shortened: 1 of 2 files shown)


paulkaplan:
Does this need to wait on blocks rendering work or is it good to go out without that?

ericrosenbaum:
@paulkaplan without the blocks rendering work, it doesn't lay out quite right, but it works, so I'd say it's good to go without it.

cwillisf (approved):
⚽️🐐!!!
no, wait...
⚽️🥅!!!
...or maybe both!

Issue #744, opened 2017-11-01
Investigate performance issues in "Sprite Caching Benchmark" project


thisandagain:
### Expected Behavior

- Project should run at 30fps on testing Chromebook

### Actual Behavior

- Project runs slowly

### Steps to Reproduce

https://scratch.mit.edu/projects/14844969/

mzgoddard:
This project spends a lot of time in overhead with `execute` like https://github.com/LLK/scratch-vm/issues/740#issuecomment-341288496and #741 and `getProcedure*` calls like https://github.com/LLK/scratch-vm/issues/741#issuecomment-341532400. After those functions `needsConvexHullPoints` in `Drawable` is another contributor. `scratch-vm`'s isPromise spends a lot of time likely due to it being deoptimized like `execute`, especially with how simple a function it is.

## `needsConvexHullPoints`

`needsConvexHullPoints` is showing up a lot with how many calls to position changing block functions are made. It might not be `needsConvexHullPoints` itself but that the performance inspector sampled there frequently instead of other points spread out in functions `setXY` calls.

Improving `needsConvexHullPoints` and functions used by `setXY` more generally as it may need lots of small changes in logic order or variable assignment.

## `isPromise`

There are enough code blocks being executed to demonstrate the time isPromise uses. It'll likely become more obvious for other projects as the larger problems are solved. isPromise is getting enough different input types that javascript can't optimize it.

Looking at bluebird or other promise polyfills/implementations may show a good way to implement this function so it doesn't deoptimize.

## Some Numbers (MBP 13" 2013)

Here are the top operations and the part of a sample of 5 seconds of the project running.

operations | % of 5 seconds
--------|--------
indexOf | 14.0
execute | 13.7
getProcedureParamNames | 10.5
getProcedureDefinition | 5.8
stepThreads | 5.5
isPromise | 2.8

thisandagain:
This project is now running at > 30 fps in the VM Playground on our testing hardware.

Issue #740, opened 2017-10-31
Investigate performance issues in "Stacky Build" project


thisandagain:
### Expected Behavior

- Project should run at 30 fps on testing Chromebook (2016 Samsung Chromebook 3)

### Actual Behavior

- Project runs slowly (5 – 10 fps)

### Steps to Reproduce

https://scratch.mit.edu/projects/155128646/

mzgoddard:
This project is heavily bound by the virtual machine's internal performance. The virtual machine's performance is dependent on how it has been written in javascript and some of the data structures used. `execute` in `scratch-vm/src/engine/execute` and `Sequencer#stepThreads` in `scratch-vm/src/engine/sequencer` are responsible.

## `execute`

The `execute` function is being deoptimized by at least Chrome. Chrome in trying to optimize the function is make assumptions like a variable is always a string and then seeing that assumption broken when its set to a number. Happening enough times Chrome is giving up and running the function in a safe but very slow way. This can be improved by either figuring what assumptions Chrome is making and ensures those don't break or break up execute into smaller functions where most of them can be optimized.

Each call of `execute` that calls a blockFunction (instead of returning early) creates at least 16 objects, both costing time to allocate the objects and later contributing to some regular `Minor GC` operations the javascript VM is performing. Many of these objects would be best turned into static reused instances with internal values set by each `execute` call. The first one to tackle is the util argument to blockFunction since it should be safe as a single instance. That would save 12 allocations, one for the util object, and one for each javascript closure letting the functions refer to the higher scopes they're created in.

## `stepThreads`

`stepThreads` in this project is spending a lot of time in `indexOf` calls and its loops and tests on threads. This project is running a lot of threads in other words. And a lot of that is likely churning threads, starting and ending threads frequently, growing doneThreads every step to the point indexOf spends a fair amount of time checking for `stepThreads` if a thread is not in doneThreads so it can be added or if it is in doneThreads so it can be dropped from runtime threads.

While still using an array for doneThreads adding another strategy to test if a thread is in doneThreads or not would offer a recognizable performance boost.

## Some numbers

Over a quick 5 second sample of the game running here are the amount of time spent in some of these expensive operations:


Operation   | % of 5 seconds
------------|---------
stepThreads | 10.4%
execute | 24.8%
indexOf | 9.7%
Minor GC | 6.4%

griffpatch:
[...]

PR #758, opened 2017-11-03
Constant time check if a thread is in doneThreads

mzgoddard:
### Resolves

Related to #740 and #744.

Speeds up `Sequencer.stepThreads` in projects like https://scratch.mit.edu/projects/155128646/ and https://scratch.mit.edu/projects/14844969/ which use a lot threads.

### Proposed Changes

Maintain doneThreads as a list of null and done threads as runtime threads are iterated. This way any thread was done when it was interpreted will appear in the same index as it appear in the list of runtime threads to be able to do a constant time check if the thread was done at the right time.

After filtering the runtime threads, the done threads can be filtered done to just thread objects, filtering out all the null values.

### Reason for Changes

`indexOf` may need to step over all of the already `doneThreads` to know if it has been added previously. This adds up in projects that start and finish a lot of threads regularly.

### Test Coverage

This replaces existing behaviour to help the scratch-vm perform faster. It passes the existing engine_sequencer tests.

### DIFF
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -38,7 +38,7 @@ class Sequencer {
         let numActiveThreads = Infinity;
         // Whether `stepThreads` has run through a full single tick.
         let ranFirstTick = false;
-        const doneThreads = [];
+        const doneThreads = this.runtime.threads.map(() => null);
         // Conditions for continuing to stepping threads:
         // 1. We must have threads in the list, and some must be active.
         // 2. Time elapsed must be less than WORK_TIME.
@@ -54,10 +54,12 @@ class Sequencer {
                 if (activeThread.stack.length === 0 ||
                     activeThread.status === Thread.STATUS_DONE) {
                     // Finished with this thread.
-                    if (doneThreads.indexOf(activeThread) < 0) {
-                        doneThreads.push(activeThread);
-                    }
+                    doneThreads[i] = activeThread;
                     continue;
+                }
+                // A thread was removed, added or this thread was restarted.
+                if (doneThreads[i] !== null) {
+                    doneThreads[i] = null;
                 }
                 if (activeThread.status === Thread.STATUS_YIELD_TICK &&
                     !ranFirstTick) {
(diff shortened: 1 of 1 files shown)


mzgoddard:
Happy to add any tests y'all think are needed.

I talk about a side-effect in https://github.com/LLK/scratch-vm/pull/758/commits/4e24a3f3807bd30dca6785172408b66ba4bf2c61 relating to threads that are restarted in `_restartThread` or removed with `_removeThread`. Since those can edit `runtime.threads` while under `stepThreads` they may appear in `doneThreads` before this PR. Those won't appear after this PR. Best I can tell this will not have a real effect just a virtual one. Restarted threads still exist in `runtime.threads` and removed threads can be removed before stepThreads can track them as done anyways.

I made this as two commits. The first commit adds the constant time check. The second commit unravels the `.filter` calls into in place filtering. Happy to drop the second commit, the real performance gain is in the first.

src/engine/sequencer.js
```diff
@@ -84,13 +84,28 @@ class Sequencer {
...
+        // Filter undefined and null values from `doneThreads`.
+        let numDoneThreads = 0;
+        for (let i = 0; i < doneThreads.length; i++) {
+            const maybeThread = doneThreads[i];
+            if (maybeThread !== null) {
+                doneThreads[numDoneThreads] = maybeThread;
+                numDoneThreads++;
+            }
+        }
+        doneThreads.length = numDoneThreads;
```
rschamp, line 106: TIL!

rschamp (approved):
Looks and works great. Thanks for pointing out the minor functional difference — since it doesn't affect glows or monitors, it's not a big deal.

Incidentally I beat my PR in Stacky Build while testing this PR so 👍 👍  from me!

src/engine/sequencer.js
mzgoddard: A lot of JS data libraries use this technique to quickly change the size of the internal structure of collection-like objects.

Like this code from Backbone that quickly dumps the contents of its internal array and replaces it with a new set of values.

https://github.com/jashkenas/backbone/blob/0052c2fc5c8b49dd74d5fc8af62a6ce5d0d6ec10/backbone.js#L875-L877

PR #759, opened 2017-11-04
Add back pen transparency from color number

paulkaplan:
### Resolves

Fix the dropped scratch2 pen transparency model to support imported projects. Uses the same code as used to be in the pen extension. 

image

image


ericrosenbaum (approved):
Looks good!

PR #764, opened 2017-11-06
Ask/answer opcode implementation with runtime events.

paulkaplan:
### Resolves

Implements the first part of the VM implementation of ask/answer: emit QUESTION events with the asked question and bind ANSWER events to set the global answer reporter value. 

### Test Coverage

Added a test for the sensing block code to test asking and answering questions. 

---

Future work will involve triggering a bubble instead of sending the question with the event if the sprite is visible.


rschamp (approved):
❓ 💬 🎉

PR #765, opened 2017-11-06
Write src/engine/execute so it can be optimized

mzgoddard:
### Resolves

Related to #744, performance in execute and isPromise.

### Proposed Changes

Use explicit tests to detect if a value is a promise.

### Reason for Changes

Testing with implicitly casted `value` and `value.then` is deoptimizing
isPromise, which deoptimizes execute. In this case deoptimizing means
that the JavaScript VM cannot compile the functions into a form that
can be run faster.

### Test Coverage

Existing part of execute behaviour.


paulkaplan (approved):
LGTM

PR #768, opened 2017-11-07
Broadcast and wait chained

### LINKED ISSUE #760: Broadcast and wait should not yield when dirty until return?
Here is a simplified project demonstrating a feature that I use in most of my scratch projects that acts differently in Scratch 2 & 3.

https://llk.github.io/scratch-gui/develop/#184390187

In Scratch 2 you can execute blocks that update the screen and then issue a broadcast and wait without the Scratch runtime yielding to do a screen update until the broadcast and wait completes.

In Scratch 3 the scratch vm apparently stops to update the screen before the broadcast and wait is issued? This results in projects running at half speed in Scratch 3 even when they are doing very little.

Another interesting issue... Try the project on turbo mode. It then runs at 30fps exactly... But remove the broadcast and wait and it jumps to run at 2730 fps. Now in Scratch 2 it runs super fast with the broadcast and wait as well as without. So something about broadcast and waits appear to be holding up turbo mode too?

_More Info_

I have identified that it is not to do with loops. See: http://llk.github.io/scratch-vm/#184430433
In this project, if it runs at 30fps, then both cats should 'just' pass the line.

Turbo Off, 60fps mode  -> Runs at 30fps
Turbo Off, 30fps mode -> Runs at 15fps
Turbo On, 60fps mode -> Runs at 60fps
Turbo On, 30fps mode -> Runs at 30fps

So the questions are, why does a broadcast and wait cause the turbo mode to yield? And why do we get two yields instead of one when running without turbo.

### Steps to Reproduce

Try this project in Scratch 2 and Scratch 3.

https://llk.github.io/scratch-gui/develop/#184390187

Windows 10 Chrome

mzgoddard:
### Resolves

#760

### Proposed Changes

- Add restarted threads to `newThreads` list returned by `startHats`.
- Return false from `isActiveThread` if thread's status is done or stack.length is 0 or if the thread is not contained in the `runtime.threads` list.

### Reason for Changes

`broadcastAndWait` and `switchBackdropAndWait` store a `startedThreads` member on their stack frames to track their `Wait` part. The waiting depends on `startHats` returning an array with a length greater than 1 and `isActiveThread` returning false for all threads once they are all done. The proposed changes fix two edge cases for these `Wait` blocks.

### Test Coverage

Should this have a regression test?

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -825,7 +825,11 @@ class Runtime extends EventEmitter {
      * @return {boolean} True if the thread is active/running.
      */
     isActiveThread (thread) {
-        return this.threads.indexOf(thread) > -1;
+        return (
+            (
+                thread.stack.length > 0 &&
+                thread.status !== Thread.STATUS_DONE) &&
+            this.threads.indexOf(thread) > -1);
     }
 
     /**
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -804,6 +804,7 @@ class Runtime extends EventEmitter {
      * This is used by `startHats` to and is necessary to ensure 2.0-like execution order.
      * Test project: https://scratch.mit.edu/projects/130183108/
      * @param {!Thread} thread Thread object to restart.
+     * @return {Thread} The restarted thread.
      */
     _restartThread (thread) {
         const newThread = new Thread(thread.topBlock);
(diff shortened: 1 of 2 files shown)


rschamp (approved):
I think a regression test would be nice, but not a big deal if it's difficult to construct.  Thanks @mzgoddard!

mzgoddard:
@rschamp will do.

mzgoddard:
@rschamp Added a regression test https://github.com/LLK/scratch-vm/pull/768/commits/d173326290e592c79f006a97f36db5a62c441a26

rschamp:
Great! Thanks!

PR #769, opened 2017-11-07
Move music blocks into an extension

ericrosenbaum:
### Proposed Changes

Move the music blocks (play note, play drum, set instrument, rest, and the tempo blocks) into their own extension. 

A PR in the GUI will add this extension to the extension library: https://github.com/LLK/scratch-gui/pull/870

We also plan to re-implement most of the functionality of these blocks in the extension, but for now I am moving them over to the extension in their current form.

### Reason for Changes

Some aspects of the music blocks can make them confusing for beginners, such as the use of MIDI note numbers for the play note block, and setting duration and tempo in beats per minute. These blocks are also used relatively infrequently. Moving them into an extension (as with the pen blocks) allows us to reduce the size of the blocks menu, allowing beginners to focus on things that are easier to tinker with. It also will allow us to provide support specific to the music blocks when you add them from the extension library. Note that the extension will automatically load for Scratch 2.0 projects that include any of these blocks (as with the pen extension).

### Test Coverage

I moved the existing unit tests for these blocks into their own file

### DIFF
--- a/src/blocks/scratch3_music.js
+++ b/src/blocks/scratch3_music.js
@@ -4,6 +4,10 @@ const Clone = require('../util/clone');
 const Cast = require('../util/cast');
 const MathUtil = require('../util/math-util');
 
+/**
+ * An array of drum names, used in the play drum block.
+ * @type {string[]}
+ */
 const drumNames = [
     'Snare Drum',
     'Bass Drum',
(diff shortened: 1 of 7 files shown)


paulkaplan (approved):
Looks good!

PR #770, opened 2017-11-08
Caching.2017.11.07

griffpatch:
### Resolves

Improves execution speed of scratch runtime by preventing constant reevaluation / mutation of structures that do not change at runtime:
* *blocks.getInputs()*
* *blocks.getProcedureDefinition()*
* *blocks.getProcedureParamNames()*

### Proposed Changes

* Add caching to blocks.getInputs() - Store the cached result within each block (this._inputs). On said block being updated invalidate the cache and bubble invalidation back to parent block, etc).
* Add caching to blocks.getProcedureDefinition() - Cache the result in a global cache (blocks._procedureDefinitions) which is simply invalidated verbatim when any change is made to any block... I'm sure this could be improved, but I don't think it would add any visible effect to the end user at present.
* Add caching to blocks.getProcedureParamNames() - Also cached in a global cache (blocks._procedureParamNames)

### Reason for Changes

Test results:

http://llk.github.io/scratch-vm/#141514080

Test runs for 12 seconds executing various scratch blocks and a custom block and results in a benchmark score based on the number of iterations it achieves in that timeframe.

Scratch 3 Pre PR - Score: 179
Scratch 3 After PR - Score: 518
Scratch 2 Comparison - Score 899

The results are quite unstable, but in this case it's quite obvious that it makes a substantial difference. In this case almost a x3 boost in speed on this project. What's more exciting is that it's really bringing the speed up closer to Scratch 2.

### Test Coverage

Now there's a thing! Is there any way we can add tests to ensure the changes will always be invalidated when required... We don't want to cache something and ignore a change to a block stack or input, etc...


src/engine/blocks.js
```diff
@@ -25,6 +25,20 @@ class Blocks {
          * @type {Array.<String>}
          */
         this._scripts = [];
+
+        /**
+         * Cache procedure Param Names by block id
+         * @type {object.<string, ?Array.<string>>}
+         * @private
+         */
+        this._procedureParamNames = {};
```
mzgoddard: This PR has two caching patterns. Storing cached values on blocks themselves and storing cached values on the Blocks instance.

I think you can reduce this to one pattern and simplify cache invalidation.

Stored cached values in category members of a cache member on Blocks.

Like maybe:

```js
this._cached = {
  inputs: {},
  procedureParamNames: {},
  procedureDefinitions: {},
};
```

Then have a method `_resetCache` that resets `_cached` to the same object expression.

You can call `_resetCache` at the end of `addBlock`, `changeBlock`, `moveBlock`, and `deleteBlock` to reset the caches.

src/engine/blocks.js
```diff
@@ -149,17 +168,22 @@ class Blocks {
      * @return {?string} ID of procedure definition.
      */
     getProcedureDefinition (name) {
+        if (this._procedureDefinitions.hasOwnProperty(name)) {
```
mzgoddard: Do you mean `typeof this._procedureDefinitions[name] !== 'undefined'`?

src/engine/blocks.js
```diff
@@ -149,17 +168,22 @@ class Blocks {
...
                 block.opcode === 'procedures_defreturn') {
                 const internal = this._getCustomBlockInternal(block);
                 if (internal && internal.mutation.proccode === name) {
-                    return id; // The outer define block id
+                    this._procedureDefinitions[name] = id; // The outer define block id
+                    return id;
                 }
             }
         }
+        this._procedureDefinitions[name] = null; // This is not the same as undefined
```
mzgoddard: `hasOwnProperty` doesn't test if `[name]` is `undefined`. Its equivalent logically to `Object.keys(this._procedureDefinitions).indexOf(name) > -1`. You could set this to `undefined` and `hasOwnProperty` will return `true`.

src/engine/blocks.js
```diff
@@ -105,10 +119,14 @@ class Blocks {
     /**
      * Get all non-branch inputs for a block.
      * @param {?object} block the block to query.
-     * @return {!object} All non-branch inputs and their associated blocks.
+     * @return {?object} All non-branch inputs and their associated blocks.
      */
     getInputs (block) {
         if (typeof block === 'undefined') return null;
+        const cachedInputs = block._inputs;
```
mzgoddard: It would be best to not store cached values on blocks themselves.

mzgoddard (changes requested):
I think you can take this a little further and provide a caching pattern in `Blocks` that can be easily followed for anything else that may benefit from caching results.

Nice work.

griffpatch:
@mzgoddard I have made the changes you suggested, thanks

mzgoddard (approved):

mzgoddard:
@griffpatch @rschamp LGTM. @rschamp how does this look to you?

thisandagain:
Attempting to resolve conflicts. Will merge if tests pass.

thisandagain:
Thanks a ton @griffpatch and @mzgoddard ! 😄

PR #772, opened 2017-11-08
Improving test coverage

kchadha:
### Resolves

Improved test coverage for src/virtual-machine.js.

### Proposed Changes

Adding unit tests for deleteSprite and duplicateSprite functions.
Modifies these functions in the source file to address minor bugs found during testing:
- [In `deleteSprite`] accessing target.id before checking that it's null
- [In `deleteSprite`] resetting current editing target to null if last target gets deleted
- [In `duplicateSprite`] returning a promise for use in testing

### Reason for Changes

Testing is good. ~10% more coverage of this file.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -474,9 +474,12 @@ class VirtualMachine extends EventEmitter {
      */
     deleteSprite (targetId) {
         const target = this.runtime.getTargetById(targetId);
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):

rschamp (approved):
Thank you @kchadha! 🎉

PR #777, opened 2017-11-09
Preliminary localization

chrisgarrity:
Very preliminary - @ericrosenbaum @cwillisf thoughts?

* localize the block and menu strings in the pen extension
* adds .tx/config to be able to push translations to transifex
* includes format-message to localize strings and extracting them.
* For now just extracting messages from the pen extension into their own file. We’ll need to decide if each category gets its own file or group them. It will probably depend on where these files end up.


src/blocks/scratch3_pen.js
```diff
@@ -3,6 +3,8 @@ const BlockType = require('../extension-support/block-type');
 const Cast = require('../util/cast');
 const Clone = require('../util/clone');
 const Color = require('../util/color');
+const formatMessage = require('format-message');
+const translations = require('../../translations/pen.json');
```
chrisgarrity: This is a temporary way to include translations for testing (line 7), not the permanent solution.

src/blocks/scratch3_pen.js
```diff
@@ -26,6 +28,11 @@ const ColorParam = {
     TRANSPARENCY: 'transparency'
 };
 
+formatMessage.setup({
```
chrisgarrity: This is also a temporary solution to see different translations in the editor.

cwillisf:
This looks like a great start to me! I do have a couple of questions:
1. Do the translations need to be in their own top-level directory? If not I think I'd prefer that they go under `dist` -- maybe `/dist/translations` for example.
2. Would this setup allow for changing the language without reloading the extension? It looks like maybe it would just be a matter of calling `formatMessage` with new `opts`? If that's the case, would it make sense to move the `formatMessage` call out into the GUI, and the GUI will know to call it after changing language or loading an extension?

chrisgarrity:
@cwillisf good questions

1. The translations directory is only temporary in this version to let me force a language change. The plan is for the gui to pass in the locale and messages. In the next version the translations directory will be gone.
2. `formatMessage.setup` definitely needs to move out of the extension manager. I think there needs to be a `VM.updateLocale` function that GUI can call with the new locale and built-in translations that calls `setup`.

src/extension-support/extension-manager.js
```diff
@@ -133,6 +133,25 @@ class ExtensionManager {
...
 
+    /**
+    * regenerate blockinfo for any loaded extensions
+    */
+    refreshBlocks () {
+        const extensionsInfo = {};
+        this._loadedExtensions.forEach(extensionURL => {
+            if (builtinExtensions.hasOwnProperty(extensionURL)) {
+                const extension = builtinExtensions[extensionURL];
+                const extensionInstance = new extension(this.runtime);
```
chrisgarrity: There's no record of the service name assigned for the built-in extension, so a new instance was needed.  Perhaps `_loadedExtensions` should be a `new Map()` instead of Set, and then extensions get registered with the serviceName as value: `_loadedExtensions.set(extensionURL, serviceName)`assuming that the serviceName can be available.

chrisgarrity:
@cwillisf @rschamp ping - I know this got moved into the December iteration, but it would be good to get a final review sooner rather than later as it's blocking the gui PR.

src/extension-support/extension-manager.js
rschamp: @cwillisf Any thoughts on this? It seems like we should have a way to get the existing instance?

cwillisf: I'm in favor of the `Map` option that @chrisgarrity mentioned. I think the easiest way to accomplish that would be for `_registerInternalExtension` to return a Promise for the service name instead of an empty promise. That way the `then` in `loadExtensionURL` could become `...then(serviceName => {...` and we'd have both the service name and the extension URL/ID in the place where we add a `_loadedExtensions` entry.

Getting this working will be more tricky with non-built-in extensions, but right now those aren't being added to `_loadedExtensions` at all so this change wouldn't make it any worse.

cwillisf (changes requested):
I'd really prefer to avoid creating a new instance of the extension object in `refreshBlocks`, and I think it's probably feasible to avoid it (see [comment above](https://github.com/LLK/scratch-vm/pull/777#discussion_r155724292)). Other than that (and the merge conflict) I think this looks great as a starting point!

src/engine/runtime.js
```diff
@@ -497,6 +505,41 @@ class Runtime extends EventEmitter {
...
+    /**
+     * Reregister the primitives for an extension
+     * @param  {ExtensionInfo} extensionInfo - new info (results of running getInfo)
+     *                                         for an extension
+     * @private
+     */
+    _refreshExtensionPrimitives (extensionInfo) {
+        let extensionBlocks = [];
+        for (const categoryInfo of this._blockInfo) {
+            if (extensionInfo.id === categoryInfo.id) {
```
chrisgarrity, line 517: Now this gets called for each extension instead of for the whole list. Does the VM know which index in _blockInfo corresponds to a particular extension as that would save having to look through the whole array?

cwillisf: The VM doesn't currently index blocks by category or extension ID. It could in the future if there's enough benefit, though.

I think this is fine for now, but if you, say, don't like the extra indent you could use `if (extensionInfo.id !== categoryInfo.id) continue;`, or if you want to be fancy you could do `for (const categoryInfo of this._blockInfo.filter(x => extensionInfo.id === x.id)) {`. Like I said, though, I think this is fine as it is :)

src/extension-support/extension-manager.js
```diff
@@ -133,6 +133,21 @@ class ExtensionManager {
         });
     }
 
+    /**
+    * regenerate blockinfo for any loaded extensions
+    */
+    refreshBlocks () {
+        this._loadedExtensions.forEach(serviceName => {
```
cwillisf, line 140: This function is so tiny now! 🐭

cwillisf (approved):
👍 Thanks for making those changes!

src/extension-support/extension-manager.js
chrisgarrity: 💯  yes I like how simple this one ends up!

PR #779, opened 2017-11-13
Music extension handles timing

ericrosenbaum:
### Proposed Changes

This is continued progress on moving music functionality into the music extension. Large chunks will need to be removed from the audio engine once this is done (i.e. after additional PRs here). 

- Move tempo state here
- Move control of timing here
- Change the note, drum and rest blocks to use the stackframe timer (instead of using promises with setTimeouts). 

### Reason for Changes

Generally, we are moving all functionality related to the note, drum, rest and tempo blocks into the music extension. 

The change to using the stackframe timer results in a noticeable improvement in the accuracy of musical timing.

### Test Coverage

Unit tests required a small fix.


paulkaplan (approved):
LGTM

Issue #749, opened 2017-11-02
Runtime Inspector


mzgoddard:
Need a way to discover where the VM is spending time on opcodes and threads during a step. Many operations are quick and hard to get a handle on time spent in them and overhead for them in tools like Chrome's inspector performance tab.

Goal: Build an API of hooks to record opcodes and thread usage. This could be a single or set of simple hooks that are called during a frame's operation. The hooked functions can record that information as they wish and later display the information.

Useful data to record:
- opcodes per step
  - `{opcode, threadId, targetId, duration, prepDuration, postDuration}`
- threads per step
  - `{threadId, targetId, duration}`
- drawRequests `{opcode, threadId, targetId}`
- drawing `{drawableId, 
- threads started per step
- threads finished started per step

Visual Output Goal:
stats.js view displaying bar graph of opcode category colors proportioned to time spent per category. A stats.js view in the scratch-vm playground can offer quick insights into what operations contribute to each frames duration. #738 would show a lot of time in sensing operations, and #740 would show a lot of time in overhead (time spent outside of operations) and procedure operations.

Issue #750, opened 2017-11-02
Consider creating a scratch-vm performance benchmark


boazsender:
To help track performance improvement or regressions, a tool that can run a project in a browser and record information from #749 will give us regular recorded information that can be stored and used to compare performance between different versions of the vm, renderer, and gui.

It could be useful to run a project with a renderer but without rendering to test how many steps a second it can perform as a metric.

mzgoddard:
Here is a spreadsheet with some devices I'm testing with and there respective scores in the (deprecated) Google Octane benchmark and (2015) Unity 3D's WebGL benchmark. I'm using these to have a relative handle on the performance of the devices in regards to JavaScript performance and WebGL (and JavaScript) performance.

https://docs.google.com/spreadsheets/d/1Jo3wyZw6l3ey9-FKX87FGgpqPAkz8NCVhmA6zl980xs/edit#gid=0

PR #780, opened 2017-11-13
Runtime profiler

mzgoddard:
### Resolves

https://github.com/LLK/scratch-vm/issues/749

Add an internal Profiler/Inspector API to record information on the performance of different parts of the vm and renderer.

### Proposed Changes

Add a Profiler object and runtime reference to one. When the runtime has a profiler instance, any instrumented areas like the call to a `blockFunction` in `src/engine/execute` can add `start` and `stop` events to the Profiler instance. Later `reportFrames` can be called to run through the events, collapse them into frames and call the Profiler's callback which can reduce the information down to the desired set.

Parts of Runtime, Sequencer, and execute are instrumented to provide a view of how much time is spent in block functions and different parts of the VM's overhead to track thread status and built up arguments for block functions.

An example use case is included in this PR that adds a Stats tab to the playground that displays the % of time used to call the block functions versus that mentioned overhead iterating threads and generating the arguments for a block function call.

Here is an gif example of the Stats tab:

### Reason for Changes

We need better data to track where time is spent in the VM and RenderWebGL. Builtin performance inspectors like Chrome's is one useful perspective for performance insights, but it doesn't provide a complete view. This internal Profiler provides the benefit through instrumenting specific calls the ability to count and get time per call measurements.

### Test Coverage

The instrumentation of parts of Runtime, Sequencer, and execute, pass existing tests when Profiler is disabled.

### Example of Further Work

I built a benchmark prototype as a further example of the Profiler API in use, and towards the work for https://github.com/LLK/scratch-vm/issues/750. Here is a screenshot of that output. Breaking out the collected frame and opcode information for a 2 second run of the project id in the location bar.

screen shot 2017-11-13 at 12 20 17 pm

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -10,6 +10,18 @@ const {Map} = require('immutable');
 const blockUtility = new BlockUtility();
 
 /**
+ * Profiler frame name for block functions.
+ * @const {string}
+ */
+const blockFunctionProfilerFrame = 'blockFunction';
+
+/**
+ * Profiler frame ID for 'blockFunction'.
+ * @type {number}
+ */
+let blockFunctionProfilerId = -1;
+
+/**
  * Utility function to determine if a value is a Promise.
  * @param {*} value Value to check for a Promise.
  * @return {boolean} True if the value appears to be a Promise.
@@ -207,7 +219,23 @@ const execute = function (sequencer, thread) {
     let primitiveReportedValue = null;
     blockUtility.sequencer = sequencer;
     blockUtility.thread = thread;
+    if (runtime.profiler !== null) {
+        if (blockFunctionProfilerId === -1) {
+            blockFunctionProfilerId = runtime.profiler.idByName(blockFunctionProfilerFrame);
+        }
+        // The method commented below has its code inlined underneath to reduce
+        // the bias recorded for the profiler's calls in this time sensitive
+        // execute function.
+        //
+        // runtime.profiler.start(blockFunctionProfilerId, opcode);
+        runtime.profiler.records.push(
+            runtime.profiler.START, blockFunctionProfilerId, opcode, performance.now());
+    }
     primitiveReportedValue = blockFunction(argValues, blockUtility);
+    if (runtime.profiler !== null) {
+        // runtime.profiler.stop(blockFunctionProfilerId);
+        runtime.profiler.records.push(runtime.profiler.STOP, performance.now());
+    }
 
     if (typeof primitiveReportedValue === 'undefined') {
         // No value reported - potentially a command block.
--- a/src/engine/profiler.js
+++ b/src/engine/profiler.js
@@ -1,0 +1,311 @@
+/**
+ * @fileoverview
+ * A way to profile Scratch internal performance. Like what blocks run during a
+ * step? How much time do they take? How much time is spent inbetween blocks?
+ *
+ * Profiler aims for to spend as little time inside its functions while
+ * recording. For this it has a simple internal record structure that records a
+ * series of values for each START and STOP event in a single array. This lets
+ * all the values be pushed in one call for the array. This simplicity allows
+ * the contents of the start() and stop() calls to be inlined in areas that are
+ * called frequently enough to want even greater performance from Profiler so
+ * what is recorded better reflects on the profiled code and not Profiler
+ * itself.
+ */
+
+/**
+ * The next id returned for a new profile'd function.
+ * @type {number}
+ */
+let nextId = 0;
+
+/**
+ * The mapping of names to ids.
+ * @const {Object.<string, number>}
+ */
+const profilerNames = {};
+
+/**
+ * The START event identifier in Profiler records.
+ * @const {number}
+ */
+const START = 0;
+
+/**
+ * The STOP event identifier in Profiler records.
+ * @const {number}
+ */
+const STOP = 1;
+
+/**
+ * The number of cells used in the records array by a START event.
+ * @const {number}
+ */
+const START_SIZE = 4;
+
+/**
+ * The number of cells used in the records array by a STOP event.
+ * @const {number}
+ */
+const STOP_SIZE = 2;
+
+/**
+ * Stored reference to Performance instance provided by the Browser.
+ * @const {Performance}
+ */
+const performance = typeof window === 'object' && window.performance;
+
+
+/**
+ * Callback handle called by Profiler for each frame it decodes from its
+ * records.
+ * @callback FrameCallback
+ * @param {ProfilerFrame} frame
+ */
+
+/**
+ * A set of information about a frame of execution that was recorded.
+ */
+class ProfilerFrame {
+    /**
+     * @param {number} depth Depth of the frame in the recorded stack.
+     */
+    constructor (depth) {
+        /**
+         * The numeric id of a record symbol like Runtime._step or
+         * blockFunction.
+         * @type {number}
+         */
+        this.id = -1;
+
+        /**
+         * The amount of time spent inside the recorded frame and any deeper
+         * frames.
+         * @type {number}
+         */
+        this.totalTime = 0;
+
+        /**
+         * The amount of time spent only inside this record frame. Not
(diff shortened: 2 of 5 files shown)


src/engine/profiler.js
```diff
@@ -0,0 +1,289 @@
...
+const START_SIZE = 4;
+
+/**
+ * The number of cells used in the records array by a STOP event.
+ * @const {number}
+ */
+const STOP_SIZE = 2;
+
+/**
+ * Stored referenced to Browser provided Performance instance.
```
kchadha: minor typo

src/engine/profiler.js
```diff
@@ -0,0 +1,289 @@
...
+ * @const {Performance}
+ */
+const performance = typeof window === 'object' && window.performance;
```
kchadha: I'm slowly re-learning the ways of JS so I appreciate the comment above this line. If it didn't exist, I definitely would have thought that performance would always be a boolean.

src/engine/profiler.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+class Profiler {
+    /**
+     * @param {function} onFrame a handle called for each recorded frame. The
+     * passed frame value may not be stored as it'll be updated with later
+     * frame information. Any information that is further stored by the handler
+     * should make copies or reduce the information.
+     */
```
kchadha: It would definitely be helpful to me as a reader of this code to understand the expected type signature for the onFrame function (e.g. what type of argument does it expect and what does it return)? I'm also coming from Haskell which is a very strongly typed language, so if this isn't known or not really expected JS convention, please let me know!

src/engine/execute.js
```diff
@@ -207,7 +219,19 @@ const execute = function (sequencer, thread) {
     let primitiveReportedValue = null;
     blockUtility.sequencer = sequencer;
     blockUtility.thread = thread;
+    if (runtime.profiler !== null) {
+        if (blockFunctionProfilerId === -1) {
+            blockFunctionProfilerId = runtime.profiler.idByName(blockFunctionProfilerFrame);
+        }
+        // runtime.profiler.start(blockFunctionProfilerId, opcode);
+        runtime.profiler.records.push(
+            runtime.profiler.START, blockFunctionProfilerId, opcode, performance.now());
```
kchadha: why not the commented out line above (same question for stop below)?

src/engine/profiler.js
```diff
@@ -0,0 +1,289 @@
...
+                    stack.push(new ProfilerFrame(depth));
+                }
+
+                // Store id, arg, totalTime, and initialize selfTime.
+                const frame = stack[depth++];
+                frame.id = this.records[i + 1];
+                frame.arg = this.records[i + 2];
+                // totalTime starts as the start time for this frame and when
+                // the we reach the stop event will be used to calculate and
+                // store the totalTime in this member.
```
kchadha: typo/slightly confusing...

src/engine/profiler.js
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    /**
+     * Decode records and report all frames to `this.onFrame`.
+     */
+    reportFrames () {
+        const stack = this._stack;
+        let depth = 1;
+
+        for (let i = 0; i < this.records.length;) {
```
kchadha: It would be good to have a high level 1-2 sentence description of what's happening in this loop

src/engine/runtime.js
```diff
@@ -1413,6 +1468,24 @@ class Runtime extends EventEmitter {
             this._step();
         }, interval);
     }
+
+    /**
+     * Turn on profiling.
+     * @param {function} onFrame A callback handle based a profiling frame when
+     * the profiler reports back.;
+     */
```
kchadha: I'm not sure I understand this comment...

src/engine/sequencer.js
```diff
@@ -129,7 +185,19 @@ class Sequencer {
             // Execute the current block.
             // Save the current block ID to notice if we did control flow.
             currentBlockId = thread.peekStack();
+            if (this.runtime.profiler !== null) {
+                if (executeProfilerId === -1) {
+                    executeProfilerId = this.runtime.profiler.idByName(executeProfilerFrame);
+                }
+                // this.runtime.profiler.start(executeProfilerId, null);
```
kchadha: same question here about why this is commented out/replaced w/the code below...

src/playground/playground.js
```diff
@@ -40,6 +40,57 @@ const getAssetUrl = function (asset) {
...
+    vm.runtime.profiler = null;
+
+    const stepThreadsProfilerId = profiler.idByName('Sequencer.stepThreads');
+    const blockFunctionProfilerId = profiler.idByName('blockFunction');
+
+    let executeTime = 0;
+    const stepValues = [];
+    let lastUpdate = Date.now();
+
+    profiler.onFrame = function ({id, totalTime}) {
```
kchadha: I'd put a comment here about what's happening in this function

src/playground/playground.js
```diff
@@ -40,6 +40,57 @@ const getAssetUrl = function (asset) {
...
+    vm.runtime.enableProfiling();
+    if (vm.runtime.profiler === null) {
+        // Profiler isn't available on the local system.
+        return;
+    }
+
+    const blockPercentPanel = stats.addPanel(new window.Stats.Panel(panelName, panelFg, panelBg));
+
+    const profiler = vm.runtime.profiler;
+    vm.runtime.profiler = null;
```
kchadha: Why are you setting this to null here?

src/playground/playground.js
```diff
@@ -40,6 +40,57 @@ const getAssetUrl = function (asset) {
...
+        }
+    };
+
+    blockPercentPanel.dom.style.display = 'none';
+
+    stats.dom.addEventListener('click', () => {
+        if (blockPercentPanel.dom.style.display === 'block') {
+            vm.runtime.profiler = profiler;
+        } else {
+            vm.runtime.profiler = null;
```
kchadha, line 102: I also don't understand why the profiler is being set to null here...

kchadha (commented):
Most of my comments are just questions about what's happening in certain places or asking for a bit more clarity in the comments. Overall, everything looks good and is very thoroughly documented! =]

src/playground/playground.js
mzgoddard: This line sets profiler to null to disable profiling until we want to record since active profiling slows down the scratch-vm a little.

The runtime's profiler member is null by default. To ensure errors don't occur the instrumented areas check if profiler is not null before calling `profiler.start` as that would throw a TypeError because you cannot read any properties of null.

So toggling the profiler on, is setting profiler to a Profiler instance. And toggling off, is setting it to null.

mzgoddard: I guess another to mention is this line is assuming that profiling is initially off.

src/playground/playground.js
mzgoddard: This if/else block is checking if the Block % stats panel is visible or not. If its visible `=== 'block'`, enable profiling, which will call onFrame, which will update the stats panel every second onFrame is called. If `!== 'block'` it sets the profiler to null, disabling the profiling.

That way the profiler records when that stats panel is visible, and doesn't record when it is not visible.

src/engine/execute.js
mzgoddard: This is the code of `Profiler.start` manually inlined. For small functions in jS, the virtual machines can end up doing more to make the call than perform the logic in the call. Its one of those techniques you want to balance for sections of JavaScript that are critical to performance.

Its true for start and stop, but only need it here and the calls around the call to execute in Sequencer to have the recorded information better reflect normal performance instead of having added bias over time from `Profiler.start`.

This is one of the things I'd like to do through a babel plugin like I briefly mentioned yesterday, to separate the performance concern into a different part of the scratch-vm code through code generation.

mzgoddard:
Thanks @kchadha for the feedback! I made changes to fix the typos and add or expand on the comments in the code.

kchadha:
LGTM!

PR #782, opened 2017-11-13
List typed variable

kchadha:
### Proposed Changes

Lists now behave like variables with variable type string (`''` and `'list'`) differentiating between the two.

src/engine/list.js has been removed. src/engine/variable.js now handles (and properly serializes to XML) lists as well as scalar variables.

### Reason for Changes

Lists were previously not working/being serialized to XML (especially when importing SB2 projects).

### Test Coverage

Existing tests in test/unit/engine_target.js have been updated to reflect new implementation of lists, and new tests have been added to test target.lookupOrCreateList as well as the fact that the Variable constructor now throws an error if the type field has an unexpected value.

The demo fixture has been updated for compatibility with the new implementation in the integration tests.


paulkaplan:
@fsih @rschamp since I was pairing on some of this, would one/both of you mind reviewing this?

src/engine/variable.js
```diff
@@ -9,19 +9,45 @@ class Variable {
...
-        return `<variable type="" id="${this.id}">${this.name}</variable>`;
+        return `<variable type="${this.type}" id="${this.id}">${this.name}</variable>`;
+    }
+
+    /**
+     * Type representation for scalar variables.
+     * @const {string}
+     */
+    static get SCALAR_TYPE () {
+        return '';
```
paulkaplan: I think this specific value requires some explanation. "Empty string" is how scratch-blocks denotes the scalar type, basically for backwards compatibility with untyped variables. @rschamp I was wondering if we should have an explicit adapter layer between scratch-blocks variable type strings and the vm variable types, particularly because there are some dangers in using the empty string as a meaningful value (e.g. it is falsey).

src/serialization/sb2.js
```diff
@@ -251,10 +251,14 @@ const parseScratchObject = function (object, runtime, extensions, topLevel) {
         for (let k = 0; k < object.lists.length; k++) {
             const list = object.lists[k];
             // @todo: monitor properties.
-            target.lists[list.listName] = new List(
+            const newVariable = new Variable(
+                getVariableId(list.listName),
                 list.listName,
-                list.contents
+                Variable.LIST_TYPE,
+                list.isPersistent
```
paulkaplan: @rschamp can you confirm that lists are never persistent (cloud) coming from SB2? If so, this should just be changed to `false`, or just left out and forced to `false` for all list types?

src/serialization/sb2_specmap.js
```diff
@@ -1201,7 +1201,8 @@ const specMap = {
         argMap: [
             {
                 type: 'field',
-                fieldName: 'VARIABLE'
+                fieldName: 'VARIABLE',
+                variableType: ''
```
paulkaplan: @kchadha do we want to use the `Variable.SCALAR_TYPE` constant here instead of the raw string?

kchadha: I was actually going to pose this question in my actual PR description. I can make the change so that variableType always refers to the static constants we defined.

src/serialization/sb2.js
rschamp: Correct, cloud variables are only scalar.

src/engine/variable.js
rschamp: I agree with your impulse to make our internal representation of the type explicit and distinct from Blockly's. However I think that would add some unnecessary complexity right now. But I think we should keep the adapter idea in mind if it presents itself as a problem down the line.

kchadha:
@rschamp @fsih, I just pushed the changes @paulkaplan requested. Please let me know if there's anything else that needs to be changed or explained. Thank you!

src/engine/variable.js
```diff
@@ -9,19 +9,47 @@ class Variable {
     /**
      * @param {string} id Id of the variable.
      * @param {string} name Name of the variable.
-     * @param {(string|number)} value Value of the variable.
+     * @param {string} type Type of the variable, one of '' or 'list'
```
fsih, line 12: Since it's only '' or 'list', would it be cleaner to take an isList boolean instead of a string?

fsih (commented):
I'm unfamiliar with this code, so it might be better to wait for Ray's review

src/engine/variable.js
rschamp: We may be including broadcast messages here soon, so I think we want to leave it as a choice rather than a boolean.

src/serialization/sb2_specmap.js
```diff
@@ -1252,7 +1252,7 @@ const specMap = {
         ]
     },
     'contentsOfList:': {
-        opcode: 'data_list',
+        opcode: 'data_listcontents',
```
rschamp, line 1262: Curious about this change, was this broken before?

rschamp (approved):
Looks great, thanks @kchadha!

src/serialization/sb2_specmap.js
paulkaplan: Yup, it was.

Issue #564, opened 2017-05-12
Make monitors work for per-sprite variables like x position


fsih:
### Expected Behavior

Different sprites' x position monitors have different names (e.g. apple x position) and record to separate monitors

### Actual Behavior

The 2 sprites' x positions share one monitor and overwrite each other

### Steps to Reproduce

Check the box next to x position for 2 different sprites and move them around
Relevant part of code: https://github.com/LLK/scratch-vm/pull/560/files#diff-c2e52c759bccc44bfc3789bbfe5d9b2eR299

fsih:
Same as https://github.com/LLK/scratch-gui/issues/749

PR #784, opened 2017-11-14
Execute monitors on a given target ID when block is sprite-specific

fsih:
Working towards https://github.com/LLK/scratch-vm/issues/564

- Move data about whether a block is a monitor/sprite specific into src/blocks files
- Use that information to pass targetId and spriteName into monitors that are sprite-specific
- Also stop adding non-monitors to runtime.monitorBlocks, which would accumulate infinite blocks before

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -327,7 +327,11 @@ class Blocks {
             break;
         case 'checkbox':
             block.isMonitored = args.value;
-            block.targetId = args.isSpriteSpecific ? this._getTargetIdFromBlockId(block.id) : null;
+            if (optRuntime) {
+                const isSpriteSpecific = optRuntime.monitorBlockInfo.hasOwnProperty(block.opcode) &&
+                    optRuntime.monitorBlockInfo[block.opcode].isSpriteSpecific;
+                block.targetId = isSpriteSpecific ? optRuntime.getEditingTarget().id : null;
+            }
             if (optRuntime && wasMonitored && !block.isMonitored) {
                 optRuntime.requestRemoveMonitor(block.id);
             } else if (optRuntime && !wasMonitored && block.isMonitored) {
@@ -414,11 +418,6 @@ class Blocks {
                 runtime.addMonitorScript(blockId, targetId ? runtime.getTargetById(targetId) : null);
             }
         });
-    }
-
-    _getTargetIdFromBlockId (blockId) {
-        // First word of block ID. See makeToolboxXML in scratch-gui
-        return blockId.split('_')[0];
     }
 
     /**
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -334,6 +334,7 @@ class Blocks {
                 optRuntime.requestAddMonitor(MonitorRecord({
                     // @todo(vm#564) this will collide if multiple sprites use same block
                     id: block.id,
+                    targetId: block.targetId,
                     spriteName: block.targetId ? optRuntime.getTargetById(block.targetId).getName() : null,
                     opcode: block.opcode,
                     params: this._getBlockParams(block),
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -226,8 +226,7 @@ class Blocks {
                 id: e.blockId,
                 element: e.element,
                 name: e.name,
-                value: e.newValue,
-                isSpriteSpecific: e.isSpriteSpecific
+                value: e.newValue
             }, optRuntime);
             break;
         case 'move':
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -226,7 +226,8 @@ class Blocks {
                 id: e.blockId,
                 element: e.element,
                 name: e.name,
-                value: e.newValue
+                value: e.newValue,
+                isSpriteSpecific: e.isSpriteSpecific
             }, optRuntime);
             break;
         case 'move':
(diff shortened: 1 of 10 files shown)


fsih:
I'm worried about how reliant this code is on the block ID being formatted exactly as <targetId_opcode>

src/engine/blocks.js
```diff
@@ -326,12 +327,15 @@ class Blocks {
             break;
         case 'checkbox':
             block.isMonitored = args.value;
+            block.targetId = args.isSpriteSpecific ? this._getTargetIdFromBlockId(block.id) : null;
```
paulkaplan: since checkbox events can only be triggered when on the current editing target, maybe we can simplify this and avoid the brittle "parse the target id out of the block id" thing by just using the current editing target?

src/virtual-machine.js
```diff
@@ -577,6 +578,15 @@ class VirtualMachine extends EventEmitter {
         // Filter events by type, since monitor blocks only need to listen to these events.
         // Monitor blocks shouldn't be destroyed when flyout blocks are deleted.
         if (['create', 'change'].indexOf(e.type) !== -1) {
+            let blockType = e.blockId.split('_');
+            blockType = blockType[blockType.length - 1];
```
paulkaplan: Can you use the opcode that is attached to the block instead of parsing out the ID? Block containers (like `runtime.monitorBlocks`) have `getBlock(id)` and `getOpcode(block)` functions that could help with this?

paulkaplan (changes requested):

src/engine/blocks.js
fsih: Great idea, thanks Paul!

src/virtual-machine.js
fsih: You're right, the data was there all along.

paulkaplan:
@cwillisf @rschamp could you both take a look at this? In particular, is this going to work with extensions in an ok way in the future?

src/util/uid.js
```diff
@@ -6,9 +6,10 @@
  * Legal characters for the unique ID.
  * Should be all on a US keyboard.  No XML special characters or control codes.
  * Removed $ due to issue 251.
+ * Removed _ which denotes word separation in XML.
  * @private
  */
-const soup_ = '!#%()*+,-./:;=?@[]^_`{|}~' +
```
paulkaplan: I don't think this change is required anymore.

src/engine/blocks.js
```diff
@@ -226,7 +226,8 @@ class Blocks {
                 id: e.blockId,
                 element: e.element,
                 name: e.name,
-                value: e.newValue
+                value: e.newValue,
+                isSpriteSpecific: e.isSpriteSpecific
```
paulkaplan: I don't think this line is needed anymore

paulkaplan (changes requested):

src/util/uid.js
fsih: Done, thanks!

fsih:
@cwillisf  do you mind taking a look?

cwillisf:
@fsih sorry for the delay. Looking at it now...

src/engine/runtime.js
```diff
@@ -451,6 +461,22 @@ class Runtime extends EventEmitter {
...
     }
 
+    /**
+     * Populate this.monitorBlockInfo
+     */
+    _registerMonitorInfo () {
+        for (const packageName in defaultBlockPackages) {
+            if (defaultBlockPackages.hasOwnProperty(packageName)) {
+                // @todo pass a different runtime depending on package privilege?
+                const packageObject = new (defaultBlockPackages[packageName])(this);
```
cwillisf: I think it would be best to only have one instance of each blocks package. It looks like this instance is short-lived, but still I'd rather this function be blended into or called from `_registerBlockPackages` such that the instance used for monitors is the same instance used for other block registration. For built-in blocks this isn't THAT important, but it's pretty important for extensions.

src/engine/runtime.js
```diff
@@ -838,7 +864,7 @@ class Runtime extends EventEmitter {
     /**
      * Enqueue a script that when finished will update the monitor for the block.
      * @param {!string} topBlockId ID of block that starts the script.
-     * @param {?string} optTarget target ID for target to run script on. If not supplied, uses editing target.
+     * @param {?string} optTarget target Target to run script on. If not supplied, uses editing target.
```
cwillisf: If this really is taking a `Target` now, please update `{?string}` => `{?Target}`

cwillisf (changes requested):
There are a couple small things that I think should be addressed before merging, but I think your overall approach is sound and I don't see any major obstacles in the way of making this work for extensions. Nice work!

src/engine/runtime.js
fsih: Good catches, thanks!

paulkaplan (approved):

paulkaplan:
DD and I discussed a related issue offline, no impact on this PR (it LG) but for posterity:

For variable monitors, the opcode does not specify whether the monitor should be sprite specific. It is determined by whether the variable is local or global. We haven't touched local variables in any real way yet, so I'm cool leaving this out for now, but wanted to bring up this caveat. DD pointed out we could have separate opcodes for global vs. local variables, but that'll take some more unrelated work to decouple variable palette blocks from scratch-blocks.

PR #787, opened 2017-11-15
Delete monitor block when variable is deleted

fsih:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/775

### Proposed Changes

- Delete the associated monitor block when deleting a variable
- Clean up delete block to take just ID


paulkaplan (approved):
LGTM

PR #792, opened 2017-11-17
Only uglify if building for production.

### LINKED ISSUE #774: In range update for webpack-dev-server does not work
### Expected Behavior

Clean install right now gives webpack-dev-server@2.9.4, which does not compile correctly when using npm start.

### Actual Behavior

I'm on webpack-dev-server@2.4.1, which works. Had to downgrade from 2.9.4 (which is in-range for the package.json) to 2.4.1 to get npm start to work. 

### Steps to Reproduce

Try updating webpack-dev-server to most recent version.

Mac OS Sierra

paulkaplan:
Uglify was trying to run always, including the dev-server output, which
included es6 that broke uglifier.

### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/774

### Proposed Changes

Only uglify things when run in production environment. 

### Reason for Changes

It was not possible to use the playground from a fresh install.

### Test Coverage

Ran the npm run start from clean install, no errors. 

---

The same problem exists on render, i'll update that repo with the same fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ node_js:
 env:
   - NPM_SCRIPT="tap:unit -- --jobs=4"
   - NPM_SCRIPT="tap:integration -- --jobs=4"
+  - NODE_ENV=production
 sudo: false
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


rschamp (approved):
Thanks @paulkaplan!

PR #793, opened 2017-11-17
Make _removeThread safe during thread iteration

cwillisf:
### Resolves

Resolves the latest cause of LLK/scratch-render#123

### Proposed Changes

Before: `_removeThread` stops a thread and immediately removes it from the runtime's thread array. If this happens while iterating over the thread array, such as in the case of clone deletion, some elements of the array can be missed.

After: `_removeThread` has been renamed to `_stopThread`. It still stops the thread immediately but it no longer removes the thread from the runtime's thread array. Instead, `_stopThread` marks the thread for later removal. Such threads are removed at the top of the next `_step`.

### Reason for Changes

In certain very specific circumstances, deleting a clone could cause a thread to run on a target which no longer existed. Depending on the blocks running in that thread the results could be harmless or could cause crashes. In the case of project 173499153, for example, getting hit by an alien would cause an endless flood of `Uncaught TypeError: Cannot read property 'skin' of undefined` in the JS console as the non-existent sprite kept asking the renderer to check if its disposed Drawable was touching any others.

### Test Coverage

So far I've been unable to create an automated test which sets up the right conditions. If anyone is able to do so, please feel free to add a test case for this!

Here's the type of thing I was trying:
image

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -786,17 +786,14 @@ class Runtime extends EventEmitter {
     }
 
     /**
-     * Remove a thread from the list of threads.
-     * @param {?Thread} thread Thread object to remove from actives
+     * Stop a thread: stop running it immediately, and remove it from the thread list later.
+     * @param {!Thread} thread Thread object to remove from actives
      */
-    _removeThread (thread) {
+    _stopThread (thread) {
+        // Mark the thread for later removal
+        thread.isKilled = true;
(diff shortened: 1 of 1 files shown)


kchadha:
This looks good to me, but I think someone else who's also familiar w/this code should take a look. @rschamp?

rschamp (approved):
Thanks @cwillisf! Looks and works great.

PR #805, opened 2017-11-20
Replace 'got' module with 'nets'

### LINKED ISSUE #802: Remove / replace "got" from DeviceManager IO module
We have been removing the `got` module from our dependencies throughout 3.0. There is still one instance hanging around here:
https://github.com/LLK/scratch-vm/blob/3970883e45810fe6163ffd8213077df321fb817c/src/io/deviceManager.js#L1

/cc @cwillisf @rschamp

thisandagain:
### Resolves

Resolves GH-802
Resolves GH-585 

### Proposed Changes

- Replace 'got' module with 'nets'
- Adds test coverage for the Device Manager IO module's `list` method

### Reason for Changes

- Normalize package dependencies across repos

### Test Coverage

https://github.com/LLK/scratch-vm/compare/develop...thisandagain:bugfix/802?expand=1#diff-e65e552a369fef0421fe08422bb31fd0R23


.gitignore
```diff
@@ -2,6 +2,7 @@
 .DS_Store
 
 # NPM
+package-lock.json
```
thisandagain, line 5: This may be a little controversial. My thinking is to omit the `package-lock.json` file until we normalize on Node 8 / NPM 5 across the team.

radioblahaj: Yeah, The Package-lock.json file will be Good For This.
Hopeefully, The NodeNpm Can Be Installted.

radioblahaj:
Looking Good!

paulkaplan:
@cwillisf can you review this? I don't have any context on these changes.

test/unit/io_deviceManager.js
```diff
@@ -19,3 +19,16 @@ test('default connected', t => {
...
+    deviceManager
+        .list('test', 'test', null)
+        .catch(err => {
+            t.true((typeof err === 'undefined') || (typeof err === 'object'));
+        })
+        .then(body => {
+            t.true((typeof body === 'undefined') || (typeof body === 'object'));
+            t.end();
+        });
+});
```
cwillisf: Is the idea with this test that the `list(...)` will fail (either because the SDM isn't running or because it doesn't offer any `test` plugin) so you'll end up in the `catch` clause and just verify the "shape" of the error?

If so: I think it's a little deceptive that the expected path involves running through both the `catch`, which "returns" `undefined`, and the `then`, which receives that `undefined`. I'd recommend using the two-clause form of `then`... maybe something like this:
```js
deviceManager
    .list('test', 'test', null);
    .then(
        body => {
            // body should always be an object -- never undefined or null
            // or maybe we just always fail in this case since success is unexpected?
            t.end();
        },
        err => {
            t.true(...); // failure is expected and should pass the test
            t.end();
        }
    );
```

src/io/deviceManager.js
```diff
@@ -312,7 +312,17 @@ class DeviceManager {
...
         const url = `${this._serverURL}/${encodeURIComponent(deviceType)}/list?${querystring.stringify(queryObject)}`;
-        return got(url).then(response => JSON.parse(response.body));
+        return new Promise((resolve, reject) => {
+            nets({
+                method: 'GET',
+                url: url,
+                json: {}
+            }, (err, res, body) => {
+                if (err) return reject(err);
+                if (res.statusCode !== 200) reject(body);
```
cwillisf: I recommend `return reject(body);` here, or switching it all to an `if`/`else` chain.

cwillisf (changes requested):
Thanks for doing this! I have a couple of nits to pick but overall it looks pretty good. 🥅🥅

src/io/deviceManager.js
thisandagain: Good catch. That was simply an oversight on my part.

thisandagain:
@cwillisf This should be good for re-review.

cwillisf (approved):
👍

PR #806, opened 2017-11-20
Note player

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-audio/issues/20
Resolves https://github.com/LLK/scratch-audio/issues/61

### Proposed Changes

This PR moves the remaining music functionality into the music extension, with the addition of a note player. Previously, the audio engine provided a prototype "soundfont player" for the "play note" blocks using an existing module that caused several issues: it loaded instrument samples "on demand", causing a delay before playing notes; it used one sample per note, requiring dozens of files per instrument; and as a result it needed to download several megabytes of audio samples per instrument. We need a system that downloads the instrument samples on load, with only 2MB or so total for all 21 instruments.   

I've implemented a custom player for the play note blocks that:

- Loads instrument samples from a static directory (under 2MB in total),
- Uses multiple samples at different pitches for some instruments, which otherwise sound too distorted in high or low ranges,
- Use playbackRate to change the pitch of the sample to play each note,
- Includes a release time, optionally set per instrument, to allow the note to decay after its duration has ended. This creates a more natural sound for string and percussive instruments than in Scratch 2.0, and reduces or eliminates the gaps in between notes.

Some known problems that I will put into issues for future work:
  
- Generally the loudness needs adjusting across the entire set (e.g. the flute is too quiet, and the synth lead is too loud)
- Some instruments need more samples to improve the timbre across the range (e.g. the choir sound is chipmunk-y below note 60)
- All instruments have a limited maximum duration, because most of the samples are only a few seconds long. This max duration also gets shorter as the samples get shifted up in pitch. Scratch 2.0 uses loop points to sustain instrument sounds indefinitely, but I think this strategy compromises sound quality. We might want to extend the duration of some of our samples, and add more samples per instrument, to improve the maximum duration.


src/extensions/scratch3_music/index.js
```diff
@@ -503,17 +563,112 @@ class Scratch3MusicBlocks {
                 Scratch3MusicBlocks.MIDI_NOTE_RANGE.min, Scratch3MusicBlocks.MIDI_NOTE_RANGE.max);
             let beats = Cast.toNumber(args.BEATS);
             beats = this._clampBeats(beats);
-            const musicState = this._getMusicState(util.target);
-            const inst = musicState.currentInstrument;
-            if (typeof this.runtime.audioEngine !== 'undefined') {
-                this.runtime.audioEngine.playNoteForBeatsWithInstAndVol(note, beats, inst, 100);
+            // If the duration is 0, do not play the note.
+            // In Scratch 2.0, "play drum for 0 beats" plays the drum, but "play note for 0 beats" is silent.
+            if (beats === 0) {
```
thisandagain: Nit: this could probably be condensed to a single line.

src/extensions/scratch3_music/index.js
```diff
@@ -26,11 +26,11 @@ class Scratch3MusicBlocks {
         this.tempo = 60;
 
         /**
-         * The number of drum sounds currently being played simultaneously.
+         * The number of drum or instrument sounds currently being played simultaneously.
```
thisandagain: Nit: should this be *and* rather than *or*?

thisandagain (approved):
This looks great. Few really minor nits, otherwise I think this is good to merge once all tests are passing.

thisandagain:
Note for @ericrosenbaum : Can you squash this when you merge?

PR #807, opened 2017-11-21
Benchmark prototype

mzgoddard:
### Resolves

Related to #750 

### Proposed Changes

Add a benchmark page that starts by sharing the setup and event integration from playground. Load a project and automatically execute it, record for a period of time after a warm up time, then stop and report the performance info that was collected.

Highlight rows in red if they take a significant portion of the overall time. This doesn't strictly mean the opcode is slow but may be heavily used in a project. Highlighted frames are any row that other than `blockFunction` that take a significant portion of overall time.

### Reason for Changes

Tools like Chrome and Firefox's Inspector Performance tabs are an important tool for determining the performance of parts of Scratch 3. To help increase the amount of information, this tool collects information with some knowledge of the scratch-vm, collecting time for large frames of reference, and opcodes. The `execute` frame and `opcodes` provide the most regularly useful information. `execute` presents the amount of time spent executing block opcode functions and the overhead time building the argument information to run those block functions with. The `opcodes` let us view the time spent in individual opcodes and how many times they were executed.

### Test Coverage

This is currently a type of manually executed test.

### Further Work

I'm not sure if we need to merge this PR per say, but it would be a great avenue for feedback either way at this point. I'm of the mind that perhaps the benchmark should be its own repo for so that we may store "frozen" copies of projects to reduce time spent loading assets over the web to run the benchmark. The "frozen" element of these copies also provides assurance that a project doesn't suddenly run faster because a new version of the project was saved.

Further work after this PR includes:

- "Test harness" like operation, `npm run benchmark` would run the benchmark in a default browser over a set of project fixtures and report the desired metrics from those projects for comparison to another version of the vm or renderer.
- Annotation: What do all the strings on this page even mean? What does a red row mean?


src/benchmark/index.html
```diff
@@ -0,0 +1,81 @@
...
+    <title>Scratch VM Benchmark</title>
+    <link rel="stylesheet" href="./benchmark.css" type="text/css" media="screen">
+</head>
+<body>
+  <h2>Scratch VM Benchmark</h2>
+  <p>
+    Welcome to the scratch-vm benchmark. This tool helps you profile a scratch
+    project. When you load the page, it:
+    <ol>
+      <li>loads the default project
```
rschamp: An important note to add here is that it runs the project in turbo mode.

rschamp (changes requested):
Thanks @mzgoddard this is awesome! I think we would like to replace the current playground with this tool, since at this point, it gives us the information we need when working on the VM — the rest has been moved into the GUI.  Just one note other than that change.

rschamp (approved):
@mzgoddard I will take care of these things before pulling in. Thanks again!

PR #810, opened 2017-11-21
Select target after drag end

paulkaplan:
### Proposed Changes

Change the way drags are handled by the VM in order to delay selecting the editing target until after the drag has finished. This is the behavior in scratch 2, and it also improves perceived performance by delaying the blocks reload (and other editing target updates) until after dragging. 

### Reason for Changes

It is really annoying to have the whole UI feel like it freezes when I drag a sprite. This was a rage-driven feature.

Scratch 2 behavior
select-after-drag--s2

Scratch 3 behavior on develop
select-after-drag--s3-develop

Scratch 3 after this PR 
select-after-drag--s3-fixed

### Test Coverage

Added tests for start/stop drag behavior.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -36,6 +36,13 @@ class VirtualMachine extends EventEmitter {
          * @type {Target}
          */
         this.editingTarget = null;
+
+        /**
+         * The currently dragging target, for redirecting IO data.
+         * @type {Target}
+         */
+        this._dragTarget = null;
+
         // Runtime emits are passed along as VM emits.
         this.runtime.on(Runtime.SCRIPT_GLOW_ON, glowData => {
             this.emit(Runtime.SCRIPT_GLOW_ON, glowData);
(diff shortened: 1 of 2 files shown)


fsih (commented):
drag jump

I found this issue where if you move the sprite quickly, it jumps after the sprite change. But it's definitely better than before!

I think it's a bit strange to take over postSpriteInfo, but you probably have a better idea of this code than me. It looks like it handles target updates, but only via the target pane and drag right now? We would just have to be careful about the drag case whenever using postSpriteInfo.

paulkaplan:
Yeah i don't really like the way it diverts postSpriteInfo either... Maybe we should the part that selects the sprite from the VM and just do that manually in the GUI? seems like a better place to deal with questions of interactivity.

fsih (approved):

paulkaplan:
Revisiting this, and talking it over with @fsih, I think this is the correct behavior. If the GUI (or any consumer of the VM wants to fine-tune the stop drag behavior, they can do that. It is only a quirk of timing specific to the GUI that makes that last movement seem to hitch.

Issue #819, opened 2017-11-27
Music extension drum and instrument menus cause errors


ericrosenbaum:
### Expected Behavior

I just want to choose cowbell or saxophone, without pesky errors.

### Actual Behavior

The menus in the music extension blocks that let you select a drum or instrument cause an error in several situations: changing the selection on the menu and then: deleting the block, duplicating the block, or switching sprites. The error in each case is `Uncaught TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'`

It looks like this is caused the "value" field in the menu being a number, but it should be a string.

### Steps to Reproduce

Add the music extension
Drag a "play drum" block onto the workspace
Change the menu selection to "bass drum"
Try to delete the block

PR #820, opened 2017-11-27
Convert music extension menu values to string

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/819

### Proposed Changes

Convert the value field of the menu items in the music extension to strings.

### Reason for Changes

Menus on blocks optionally consist of an array of objects with text and value fields. It turns out that for scratch-blocks to behave correctly, the values cannot be numbers- they must be strings.

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -103,7 +103,7 @@ class Scratch3MusicBlocks {
         return info.map((entry, index) => {
             const obj = {};
             obj.text = entry.name;
-            obj.value = index + 1;
+            obj.value = String(index + 1);
             return obj;
         });
     }


mrjacobbloom:
Numbered menu items seems like a common enough use-case that it'll confuse a lot of extension developers, maybe this is worth fixing on the scratch-blocks side instead of in the extension?

rachel-fenichel:
I'm fine with doing this as a temporary fix but let's keep a bug open about it--scratch-blocks should be capable of handling this.

@ericrosenbaum let me know when you have a minimal test case that I can poke at in the playground (block JSON + XML to load).

cwillisf (approved):
I agree with @rachel-fenichel: ideally this would get fixed on the scratch-blocks side but I'm fine with it as a temporary workaround.

PR #823, opened 2017-11-28
Add browser field

cwillisf:
### Resolves

This change resolves build errors exposed by #805.

Thanks, @ericrosenbaum, for bringing the issue to my attention!

### Proposed Changes

This PR contains two functional changes:
- Add a `"browser"` field to `package.json`, pointing to the web build output, and
- Use UMD for the web and playground build outputs.

This is similar to LLK/scratch-storage#20 which has been working well in that repository.

### Reason for Changes

The `"browser"` field fixes the build issues affecting both Travis and local builds since #805 was merged. The UMD change is just a matter of flexibility for other developers and a step toward converging on a standard build config for our webpack-built repositories.

### Test Coverage

On my local machine I tested the playground and building `scratch-gui` against this. We don't currently have a way to automatically test a `scratch-vm` change in the context of a `scratch-gui` build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "url": "git+ssh://git@github.com/LLK/scratch-vm.git"
   },
   "main": "./dist/node/scratch-vm.js",
+  "browser": "./dist/web/scratch-vm.js",
   "scripts": {
     "build": "webpack --progress --colors --bail",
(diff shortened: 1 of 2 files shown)


thisandagain:
Geeze. Thanks for catching that.

PR #825, opened 2017-11-28
Clones get a reference to parent’s audioPlayer

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-audio/issues/51

### Proposed Changes

Clones no longer get their own audioPlayer instance. Instead, each clone gets a reference to its parent sprite's audioPlayer.

Note that I'm not certain that I'm accessing the parent in a 'good' way, though it works (`this.sprite.clones[0]`).

### Reason for Changes

In projects that created lots of clones, the audio was dropping out completely. This change prevents that problem. 

For example, this project plays an audio loop continuously while creating clones. On develop currently, the audio crackles and then drops out after 15 or 20 seconds:

https://llk.github.io/scratch-gui/develop/#189096081

After this change, the audio loop can play indefinitely.   

### Test Coverage

We don't currently have a way to test audio performance.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -137,7 +137,11 @@ class RenderedTarget extends Target {
         */
         this.audioPlayer = null;
         if (this.runtime && this.runtime.audioEngine) {
-            this.audioPlayer = this.runtime.audioEngine.createPlayer();
+            if (this.isOriginal) {
+                this.audioPlayer = this.runtime.audioEngine.createPlayer();
+            } else {
+                this.audioPlayer = this.sprite.clones[0].audioPlayer;
+            }
         }
     }
 


mrjacobbloom:
Will this affect projects like [this one](https://llk.github.io/scratch-gui/develop/#189348981) where a sprite and/or multiple clones are trying to play different sounds/notes?

ericrosenbaum:
Thanks for asking @mrjacobbloom. This change makes the behavior of clones playing sounds the same as in Scratch 2.0. 

A single audioPlayer, shared between a sprite and its clones, can play many different sounds simultaneously. It does not _play more than one copy of the same sound at the same time_. Instead, if you try to play the same sound again before it ends, it restarts. So, with this change, if a sprite plays meow, and then its clone tries to play meow before the sprite's meow has finished, it will stop the sprite's meow and start a new one. That's how Scratch 2.0 behaves too.

There's more detail here: https://github.com/LLK/scratch-audio/issues/51

cwillisf (approved):
Glad a simple fix like this could help so much!

AFAIK `this.sprite.clones[0]` is currently the best way to access the "original" clone from another. Everything in `Sprite` is based around the idea that `clones[0]` will always be the original, so until & unless we change that I think the way you're doing it is safe.

rschamp (approved):
+1 to @cwillisf's comment. We should probably add a `getOriginal` method to abstract the clone implementation.

PR #838, opened 2017-12-01
Move pen and wedo into extensions folder

ericrosenbaum:
### Proposed Changes

We created an extensions folder when we set up the music extension- this is just moving the other internal extensions (currently pen and wedo) into that folder.


paulkaplan (approved):
LG

PR #840, opened 2017-12-01
Add icon for music extension blocks

ericrosenbaum:
### Resolves

https://github.com/LLK/scratch-gui/issues/881

### Proposed Changes

Add the svg for the icon to appear on each music extension block. Thanks @carljbowman!

Note that the layout of the icon on the block is not correct because I still need to get the rendering up to the spec in https://github.com/LLK/scratch-blocks/issues/1207


paulkaplan (approved):
LG

PR #845, opened 2017-12-04
Broadcast Msg Bugfix

kchadha:
### Resolves

[scratch-gui #994](https://github.com/LLK/scratch-gui/issues/994)

### Proposed Changes

Fixes a bug where executing a broadcast message from the flyout before having dragged it out to the workspace for the first time leads to a fatal error.

### Reason for Changes

Executing a broadcast message no longer creates a new variable if one doesn't exist. There was no reason to actually create a variable for the broadcast message, as the variable creation should be taken care of elsewhere.

### Test Coverage

Changed tests for the function that was changed (`lookupOrCreateBroadcastMsg` -> `lookupBroadcastMsg`).

Fixed a failing test for broadcast/broadcast and wait blocks. Variable must now be explicitly created. Also I'm not sure why this test was passing before (given that I had previously forgotten to propogate this change _everywhere_: `{BROADCAST_OPTION: 'message'}` -> `{BROADCAST_OPTION: {id: 'testBroadcastID', name: 'message'}`

### DIFF
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -56,36 +56,40 @@ class Scratch3EventBlocks {
     }
 
     broadcast (args, util) {
-        const broadcastVar = util.runtime.getTargetForStage().lookupOrCreateBroadcastMsg(
+        const broadcastVar = util.runtime.getTargetForStage().lookupBroadcastMsg(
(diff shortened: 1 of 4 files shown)


paulkaplan (approved):

paulkaplan:
LG

PR #850, opened 2017-12-06
Do not return "null" when cannot find argument param.

### LINKED ISSUE #699: Implement argument reporters in the VM
See https://github.com/LLK/scratch-blocks/issues/1102 and https://github.com/LLK/scratch-blocks/issues/1110

On the VM side:
- [x] Track duplication of argument reporters as they are dragged out of the define block
- [x] If used in a custom block with that argument name, they act like local variables
- [ ] If used in a custom block without the argument name, evaluate to zero/the empty string/false
- [ ] If used outside of custom blocks, evaluate to zero/the empty string/false
- [x] They may have the same names as global variables

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/699

### Proposed Changes

Returns default falsey values for argument reporters that are either unused (like empty boolean slot) or used outside of the define context. 

### Reason for Changes

I changed my mind on whether these should all return 0, because i realized it was very simple, makes the behavior more clear and seems to have an extremely low chance of impacting scratch 2 projects (it would require a project that depends on a boolean/string argument reporter reporting exactly "0" outside of a define stack...)

### Test Coverage

Added a test to execute_thread

### DIFF
--- a/src/blocks/scratch3_procedures.js
+++ b/src/blocks/scratch3_procedures.js
@@ -15,8 +15,8 @@ class Scratch3ProcedureBlocks {
         return {
             procedures_definition: this.definition,
             procedures_call: this.call,
-            argument_reporter_string_number: this.param,
-            argument_reporter_boolean: this.param
+            argument_reporter_string_number: this.argumentReporterStringNumber,
+            argument_reporter_boolean: this.argumentReporterBoolean
         };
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LGTM

PR #854, opened 2017-12-11
Fix tertiary color of extension blocks

ericrosenbaum:
### Proposed Changes

Fix a typo (colorTertiary should be colourTertiary) that was preventing the correct loading of the tertiary color for extension blocks.

### Reason for Changes

Matching the color spec from https://github.com/LLK/scratch-blocks/issues/523


cwillisf (approved):
Ha! Sorry about that... hopefully it wasn't too much of a pain to find!
I guess my fingers just refused to keep spelling it that way... :p

PR #859, opened 2017-12-13
Parse the `warp` property from the mutation instead of using directly.

### LINKED ISSUE #858: Warp needs to be parsed from string on definition mutation
Since the `warp` property is on the definition block mutation, and all mutation properties are just read as strings from XML, we need to `JSON.parse` the warp property in order to use it correctly. We are using the string 'false' (which is truthy) resulting in all custom procedures being warped.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/858

### Proposed Changes

Parse the `warp` property from the mutation instead of using directly.

### Reason for Changes

Note the way we are always parsing from json the things on the mutator in other places. This fixes the way it seemed like custom procedures were always created with "run without screen refresh". They were not actually just string/boolean truthiness problem.

### DIFF
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -316,7 +316,7 @@ class Sequencer {
             const definitionBlock = thread.target.blocks.getBlock(definition);
             const innerBlock = thread.target.blocks.getBlock(
                 definitionBlock.inputs.custom_block.block);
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM!

PR #869, opened 2017-12-18
Add separator to extension block definition

ericrosenbaum:
### Proposed Changes

When generating extension block definitions, include a vertical separator field after the icon. 

### Reason for Changes

Implementing that remaining parts of the design spec in https://github.com/LLK/scratch-blocks/issues/1207


cwillisf (approved):
👍

PR #870, opened 2017-12-19
Return promise from backdrop loading

paulkaplan:
@ericrosenbaum forgot to also fix the `addBackdrop` function when I fixed `addCostume`. Adding this will make switching to the backdrop after adding it work.


ericrosenbaum (approved):
🤙

PR #871, opened 2017-12-19
Category menu icons for extensions

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/855

Depends on https://github.com/LLK/scratch-blocks/pull/1309

### Proposed Changes

Extensions can now specify a blockIconURI and a menuIconURI. These are both optional. If a menu icon is specified, the toolbox category menu will display it. If there's no menu icon but there's a block icon, the menu will display that instead. If neither is present, the menu displays its default colored circle. 

In this PR we've added a menu icon for the music extension (thanks @carljbowman!)- it's got a dark fill, since the block icon is white and doesn't show up well in the menu. For the pen, we just use the block icon.

### Reason for Changes

This change makes extension categories pop out more clearly in the menu, especially since extensions will probably all have the same category color.


cwillisf (approved):
LGTM! Would you mind updating the wiki page with this change?
https://github.com/LLK/scratch-vm/wiki/Extensions-Proposal
Thanks!

PR #873, opened 2017-12-20
Reset sound effects

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-audio/issues/34

### Proposed Changes

Bind the green flag and stop events to a new function that clears the sound effects for all targets. This includes clearing the effects state for each target, and calling the reset function on the audio player for each target. I refactored the clearing code a bit here to make this simpler.

### Reason for Changes

This change makes sure that the state of the effects in the target and the audio player don't get out of sync.

### Test Coverage

No new tests


paulkaplan (approved):
LGTM

Issue #733, opened 2017-10-27
API `loadProject` should accept objects


rschamp:
### Expected Behavior

_Please describe what should happen_
Passing an object (a parsed JSON representation of a project) to `loadProject` should work.

### Actual Behavior

_Describe what actually happens_
It fails with a warning `Failed to parse project. Non-string supplied to fromJSON`.

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_
Parse the JSON for a project and pass it to `loadProject`

### Suggestion for Resolution
Check the `typeof` the passed in argument, and only try to parse it if it's a string. If it's an object just use the object as if it's already parsed. If it's not either of these then emit a warning.

PR #874, opened 2017-12-20
Accept both objects and strings for 'loadProject'

thisandagain:
### Resolves

GH-733

### Proposed Changes

- Adds some handling to the `loadProject` method (in the underlying `fromJson` method) of the VM to accept both stringified JSON as well as an object when attempting to load a project.

### Reason for Changes

- Improves developer ergonomics (particularly when working in the browser console)

### Test Coverage

- See new integration test case (`test/integration/import-sb2-from-object.js`)

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -217,15 +217,13 @@ class VirtualMachine extends EventEmitter {
         this.clear();
 
         // Validate & parse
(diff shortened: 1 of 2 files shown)


ericrosenbaum (approved):
LGTM

Issue #862, opened 2017-12-14
Stop all sounds doesn't work?


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Stop all sounds should stop all playing sounds **on all sprites**

### Actual Behavior

_Describe what actually happens_

Only stops sounds playing on this sprite

### Steps to Reproduce

_Explain what someone needs to do in order to see what's described in *Actual behavior* above_

1. Add a long sound to the cat sprite (e.g. "birthday bells")
2. Start it playing using a block
3. Switch to the stage.
4. Click "stop all sounds" using a block from the stage.
5. It doesn't stop the sound playing from the sprite in 3.0. In 2.0, it does. 

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

Mac, Chrome latest.

---

@ericrosenbaum Can you look into this? I couldn't find any old issues referring to this, but I might have been on the wrong repo. Is this a known issue or new?

ericrosenbaum:
Good catch- this is a long-standing uncaught issue. I'll have to figure out how to implement- should I make a new runtime event (STOP_ALL_SOUNDS)? Or something else?

ericrosenbaum:
Okay, the plan is to iterate over all targets call stopAll

PR #875, opened 2017-12-20
Stop all sounds for all targets

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/862

### Proposed Changes

The "stop all sounds" block now iterates over all targets and stops all sounds for each.

### Reason for Changes

This makes the behavior consistent with Scratch 2.0 (good catch @paulkaplan!)


paulkaplan (approved):
LG

PR #876, opened 2017-12-20
Allow size to be set through postSpriteInfo, and emit updates when it is

paulkaplan:
Required to allow size to be added to sprite info panel in the GUI.

### Resolves

Part of implementing https://github.com/LLK/scratch-gui/issues/1064

### Proposed Changes

Allow `size` to be set through the `postSpriteInfo` method, and emit target update events when it is.

Both of these changes just follow the pattern of the other params that can be set through `postSpriteInfo`.

### Reason for Changes

So that the GUI can set the target size and receive up-to-date information about it when it is changed by the blocks.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -329,6 +329,7 @@ class RenderedTarget extends Target {
                 this.runtime.requestRedraw();
             }
         }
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
LGTM

ericrosenbaum (approved):
LGTM

PR #877, opened 2017-12-22
Fix legacy `setPenHueToNumber` and `changePenHueBy`

### LINKED ISSUE #855: Pen project incompatibility
### Expected Behavior

https://scratch.mit.edu/projects/113451797/#editor

### Actual Behavior

https://llk.github.io/scratch-gui/develop/#113451797

image

---

@ericrosenbaum I believe this is an issue with the translation of pen blocks into scratch 3. I assume that the brightness/saturation is set all the way down, and it isn't being reset correctly?

cwillisf:
### Resolves

Resolves #855 

### Proposed Changes

This change pulls the hue-shade color update logic into a separate function, now called by all the legacy pen blocks.

### Reason for Changes

These two blocks were not doing the same shade-oriented math as `setPenShadeToNumber`. Effectively these blocks were using the new color model instead of Scratch 2.0's hue-shade model.

### Test Coverage

We don't have any automated tests for visual output, but I tested the example projects in #855 and they work correctly for me now.

### DIFF
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -446,7 +446,7 @@ class Scratch3PenBlocks {
                     blockType: BlockType.COMMAND,
                     text: formatMessage({
                         id: 'pen.setHue',
-                        default: 'set pen hue to [HUE]',
+                        default: 'set pen color to [HUE]',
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
Code looks good, testing on the projects linked in the issue.

PR #878, opened 2017-12-28
Optimize stepThread by reducing the number of calls to peekStack

davidaylaian:
### Resolves

Make stepThread faster by reducing the number of times it calls peekStack.

### Proposed Changes

Instead of calling `thread.peekStack()` two times, once for the condition in the while loop and once to save the current block ID, call it one time to do both.

### Reason for Changes

Assuming that it didn't break anything (see below), performance is improved.

### Test Coverage

Unfortunately, whenever I run the tests, I get `No storage/audio/rendering module present, cannot load costume/asset`, so I wasn't able to even run them.

### DIFF
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -174,7 +174,8 @@ class Sequencer {
             // A "null block" - empty branch.
             thread.popStack();
         }
-        while (thread.peekStack()) {
+        // Save the current block ID to notice if we did control flow.
+        while ((currentBlockId = thread.peekStack()) !== 0) {
             let isWarpMode = thread.peekStackFrame().warpMode;
             if (isWarpMode && !thread.warpTimer) {
                 // Initialize warp-mode timer if it hasn't been already.
@@ -183,8 +184,6 @@ class Sequencer {
                 thread.warpTimer.start();
             }
             // Execute the current block.
-            // Save the current block ID to notice if we did control flow.
-            currentBlockId = thread.peekStack();
             if (this.runtime.profiler !== null) {
                 if (executeProfilerId === -1) {
                     executeProfilerId = this.runtime.profiler.idByName(executeProfilerFrame);


paulkaplan:
Hi @davidaylaian, just wanted to give you a heads up that most of the team is away for the holidays, so it might take a bit for this get reviewed. Thanks for the submission though, we'll be able to take a look after the new year.

paulkaplan:
@rschamp or @cwillisf this came in over the holidays. Would either of you be able to take a look at it?

rschamp (commented):
The `peekStack` method doesn't actually do very much work, so I'm not sure we're saving anything on performance with this change. However it makes the `while` loop harder to read. 

@davidaylaian Would you please provide a comparison using either the metrics provided by the VM benchmark tool (http://llk.github.io/scratch-vm/) or some other method to quantify how this improves performance?

davidaylaian:
@rschamp sure thing! Here's my findings:

On average, the optimized program can...
- loop 9.5% more steps per second
- execute 9.1% more blocks per second

Tests were done using the default project ([119615668](http://llk.github.io/scratch-vm/#119615668)) on Firefox 57.0.4, Linux.

cwillisf:
I ran this (specifically, this merged onto `develop`) through the whole VM benchmark suite on both Chrome and Firefox, then compared the results to doing so on the current `develop` branch. On my computer I seem to get a 1-5% speedup in Firefox, but the change in Chrome performance seems inconsistent. I think it's effectively no change in Chrome and that I'm just seeing sampling noise.

I do think the new form of the code is slightly more "correct" in that it eliminates the possibility that the return value of `peekStack()` could change between the two calls, but that's only a theoretical concern right now. I also agree that it's a little less readable.

Thoughts, @rschamp?

rschamp:
If there is an improvement somewhere, even if it's only in certain browsers, then that's good enough to me. Also I agree that this is more correct, albeit less readable.

Thanks @davidaylaian!

rschamp (approved):

PR #881, opened 2017-12-29
No pen trails when dragging

sjhuang26:
### Resolves

[Render issue #144](https://github.com/LLK/scratch-render/issues/144)

### Proposed Changes

When a sprite is dragged while its pen is down, no pen trails are made.
In addition, only one setXY() call is made when dragging changes the X and Y positions at the same time, so even if pen trails were made while dragging, the pen trail would be smooth, not jagged and right-angle-like.

### Reason for Changes

Consistency with Scratch 2.

### Test Coverage

None.


paulkaplan (approved):
LG thanks!

PR #882, opened 2017-12-29
Reformat string keys for costume changes

sjhuang26:
### Resolves

This pull request is a dependency for [GUI PR #1123](https://github.com/LLK/scratch-gui/pull/1123).

### Proposed Changes

String keys that could be passed into setCostumeOrBackdrop changed from "next costume" to "\_nextCostume\_", "next backdrop" to "\_nextBackdrop\_", etc.

### Reason for Changes

The clone blocks use the same style string keys (underscores around them), so costume blocks should be like this too.

### Test Coverage

None.


sjhuang26:
Closing because the GUI PR doesn't need this anymore.

PR #890, opened 2018-01-09
Update monitor on dropdown change

fsih:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/785

### Proposed Changes

Update the monitor label when the choices in the dropdown are changed

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -402,7 +402,7 @@ class Blocks {
         case 'mutation':
             block.mutation = mutationAdapter(args.value);
             break;
-        case 'checkbox':
+        case 'checkbox': {
             block.isMonitored = args.value;
             if (!optRuntime) {
                 break;
@@ -427,6 +427,7 @@ class Blocks {
                 }));
             }
             break;
+        }
         }
 
         this.resetCache();
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -3,6 +3,7 @@ const mutationAdapter = require('./mutation-adapter');
 const xmlEscape = require('../util/xml-escape');
 const MonitorRecord = require('./monitor-record');
 const Clone = require('../util/clone');
+const {Map} = require('immutable');
 
 /**
  * @fileoverview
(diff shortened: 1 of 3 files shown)


paulkaplan:
Linting is failing

paulkaplan:
It looks like this only works for blocks that have inputs rather than fields (that is, you can drop something into the input). You can see this because the "costume name/number" reporter doesn't change labels, but the "current year/month/day" block does. 

monitor labels

As an aside, it looks like the "current" block should not be droppable (according to scratch 2). That is tracked https://github.com/LLK/scratch-blocks/issues/1199

src/engine/blocks.js
```diff
@@ -382,22 +383,38 @@ class Blocks {
...
+                        params: this._getBlockParams(parent)
+                    }));
+                }
             }
             break;
         case 'mutation':
             block.mutation = mutationAdapter(args.value);
             break;
-        case 'checkbox':
+        case 'checkbox': {
```
paulkaplan, line 405: What is the curly brace here for?

paulkaplan (changes requested):
see my other comments, tl;dr doesn't seem to work on the "costume name/number" block.

fsih:
Hmm, tangentially, costume number should definitely be sprite specific

fsih:
Why are the dropdowns droppable? Are there inputs you could drop in that would make sense?
image
image

src/engine/blocks.js
fsih: This was required by the lint you mentioned earlier. It ensures that the variable isSpriteSpecific is scoped to just the case, instead of the whole switch statement.

fsih:
Talked to @ericrosenbaum and "current ()" is going to be changed to not droppable. "() of ()" is going to be changed so that the first is not droppable, but the second is.

fsih:
Costume number wasn't sprite specific because the getMonitored function still had the old block names. I've updated the getMonitored functions along with this PR (but I can move that to a separate PR if you want)

paulkaplan (approved):

paulkaplan:
LG!

PR #891, opened 2018-01-09
Audio effects state for clones

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-audio/issues/50

Depends on https://github.com/LLK/scratch-audio/pull/71

### Proposed Changes

- Each clone has its own AudioPlayer (reverting the change [here](https://github.com/LLK/scratch-vm/pull/825/)), so that they can apply audio effects (volume, pitch and pan) independently. 
- When a sprite or clone is deleted, its AudioPlayer is disposed, disconnecting its webaudio nodes. This prevents the bug where audio drops out after too many AudioPlayers have been created (https://github.com/LLK/scratch-audio/issues/27).
- Clones get a copy of the audio effects state from their parent sprite, and sync their own AudioPlayer to it.

### Reason for Changes

If clones didn't have their own AudioPlayers, they couldn't have independent audio effects settings. 

### Test Coverage

Let me know if you have suggestions for tests.

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -18,7 +18,9 @@ class Scratch3SoundBlocks {
         }
 
         this._onTargetCreated = this._onTargetCreated.bind(this);
-        runtime.on('targetWasCreated', this._onTargetCreated);
+        if (this.runtime) {
+            runtime.on('targetWasCreated', this._onTargetCreated);
+        }
     }
 
     /**
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -102,7 +102,7 @@ class Scratch3SoundBlocks {
      */
     _onTargetCreated (newTarget, sourceTarget) {
         if (sourceTarget) {
-            const soundState = this._getSoundState(sourceTarget);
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_sound.js
```diff
@@ -88,6 +93,23 @@ class Scratch3SoundBlocks {
...
+    /**
+     * When a Target is cloned, clone the sound state.
+     * @param {Target} newTarget - the newly created target.
+     * @param {Target} [sourceTarget] - the target used as a source for the new clone, if any.
+     * @listens Runtime#event:targetWasCreated
+     * @private
+     */
+    _onTargetCreated (newTarget, sourceTarget) {
+        if (sourceTarget) {
+            const soundState = this._getSoundState(sourceTarget);
```
cwillisf: This `_getSoundState` call will create the state on `sourceTarget` if it doesn't exist. That's not a huge problem but I'd consider avoiding that just so that this code doesn't modify targets which don't use these blocks.

cwillisf (changes requested):
One nitpick / discussion point, but otherwise this looks good.

src/blocks/scratch3_sound.js
ericrosenbaum: Good catch!

PR #893, opened 2018-01-11
Unique sprite and backdrop names

fsih:
### Resolves
https://github.com/LLK/scratch-gui/issues/1209

### Proposed Changes
Do duplicate name checking when adding sprites or backdrops from the library

### Reason for Changes
Before, you could do this to make sprites/backdrops with identical names, which is confusing and doesn't happen in Scratch 2.0.


paulkaplan:
Looks like a couple of the tests are failing?

paulkaplan (approved):
LG

PR #894, opened 2018-01-12
Fix dragging duplicated sprites

### LINKED ISSUE #224: Duplicate Sprite breaks sprite dragging
### Expected Behavior

After duplicating a sprite you should still be able to drag sprites around on the stage.

### Actual Behavior

After duplicating a sprite, dragging any sprite on the stage no longer works and the following exception is written to the console:

```
Uncaught TypeError: Cannot read property 'rotationCenter' of null
    at e.value (lib.min.js:132)
    at e.value (lib.min.js:132)
    at lib.min.js:132
    at Array.filter (<anonymous>)
    at t.value (lib.min.js:132)
    at t.value (lib.min.js:35)
    at t.o [as onStartDrag] (lib.min.js:1)
    at t.value (lib.min.js:35)
    at HTMLDocument.o (lib.min.js:1)
```

It is also worth noting that after a duplicate, the second sprite has the same name as the original sprite? But this is unrelated as far as I can tell. Nothing I could do would restore the draggability of the sprites except reloading the page.

### Steps to Reproduce

* Open a new scratch 3 page.
* Right click on Sprite1 and select Duplicate
* Try to drag either of the two sprites on the stage - exception is thrown and no dragging takes place.

Windows PC chrome.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-render/issues/224

### Proposed Changes

Although the problem occurs in the renderer, it is because we are loading the duplicate in a way that erases the linked skins. Specifically, `initDrawable` is called in `createClone` previously, where the new costumes
are installed. Calling this again here erases those loaded skins and
causes an error when trying to calculate the uniforms when a drag
starts.

This was a bug with the original implementation, but it was hidden until a change to the renderer caused a uniform access (and convex hull calculation) to occur on new drawables on drag start.

Test manually by trying to duplicate and then drag a sprite in the stage

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -831,7 +831,6 @@ class RenderedTarget extends Target {
             newTarget.effects = JSON.parse(JSON.stringify(this.effects));
             newTarget.variables = JSON.parse(JSON.stringify(this.variables));
             newTarget.lists = JSON.parse(JSON.stringify(this.lists));
-            newTarget.initDrawable();
             newTarget.updateAllDrawableProperties();
             newTarget.goBehindOther(this);
             return newTarget;


ericrosenbaum (approved):
LGTM!

fsih:
Should we also fix the name duplication?

(Never mind, I see you added a comment on the other bug)

PR #896, opened 2018-01-13
Update sb2 specmap for sensing blocks that are now fixed menus.

paulkaplan:
The updates in https://github.com/LLK/scratch-gui/pull/1223 forgot to include an update to the specmap to make sure that sb2 blocks get imported with fields instead of inputs on `sensing_of` and `sensing_keypressed` blocks. 

You can test this by creating a project like this in scratch 2 
image

And importing it. On develop it gives warnings about "ignoring non existent input", a sure-fire sign that something is wrong. It then just uses the default menu item instead of the one in the file. This change corrects that behavior.

### DIFF
--- a/src/serialization/sb2_specmap.js
+++ b/src/serialization/sb2_specmap.js
@@ -861,9 +861,8 @@ const specMap = {
         opcode: 'sensing_keypressed',
         argMap: [
             {
-                type: 'input',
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
Sorry I missed this. LGTM!

PR #897, opened 2018-01-16
Use text-encoding library for text encoder

fsih:
### Resolves

Fixes https://github.com/LLK/scratch-paint/issues/255#issuecomment-358088363

### Proposed Changes

Add polyfill for TextEncoder. Microsoft Edge doesn't support TextEncoder.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -1,3 +1,4 @@
+import {TextEncoder} from 'text-encoding';
 const EventEmitter = require('events');
 
 const centralDispatch = require('./dispatch/central-dispatch');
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
LGTM

src/virtual-machine.js
```diff
@@ -1,3 +1,4 @@
+import {TextEncoder} from 'text-encoding';
```
rschamp: We don't use `import` in this project yet, so this is causing test failures.

rschamp (changes requested):
I will update the import syntax after testing locally

Issue #554, opened 2017-05-05
Add performance benchmarking suite


thisandagain:
It can currently be a little difficult to work through refactors on the VM because we don't have a way to track performance regressions. Building a simple benchmark suite based on how we are currently doing integration testing shouldn't be too difficult.

/cc @rschamp @cwillisf @fsih

thisandagain:
From conversation with @fsih today – a few items to focus on first:

- General thread stepping times
- Variable getter / setter times
- Custom procedure times
- Custom procedure (run without screen refresh) times
- Broadcast times
- Block getters / setters (CRUD cycle)

thisandagain:
Resolved in #901 🎉

PR #901, opened 2018-01-17
Benchmark prototype

mzgoddard:
### Resolves

Related https://github.com/LLK/scratch-vm/issues/554

### Proposed Changes

- Add benchmark suite
- The individual benchmark at `playground/index.html` displays a link to the suite page after running
- Control individual benchmark warm up and recording time as url components in the hash
- Serialize individual benchmark data into its URL
- A share link is displayed after the run with that encoded URL
- Suite reports can be opened in the running frame for easy viewing
- `Save Reports` as a JSON download and `Compare Reports` to load that JSON in a future run to compare changes made to the VM

### Reason for Changes

The individual benchmark page was a stepping stone to this to provide a more representative view of performance across varying Scratch projects. The suite executes multiple projects in order with varying warm up and recording times.

Two high level performance statistics are immediately presented, steps per second and blocks per second. The steps per second are the number of times any thread is run until it yields or completes during one second. The blocks per second are the number of VM script blocks executed in a second. VM bound projects regularly show a similar number of blocks per second with steps per second being a representation of the complexity of the threads. Lower steps per second with a VM bound number of blocks per seconds relates to a higher number of blocks per step. Blocks per second lower than the VM bound value likely indicates a currently expensive block function being executed like `pen_stamp`. Lower blocks per second may also indicate a simple project or that the project is render bound.

More details can be viewed by click the steps per second and blocks per second values. Either opening in the benchmark frame or in a new tab.

All of the benchmark suite's reports can be saved together with the `Save Reports` link. This creates a JSON download to the local system that can be loaded with `Compare Reports` in a future run, during or after the run, to see side by side steps and blocks per second values. Or the second values can be clicked to see the other recorded performance values. This is great for recording a run and comparing later against the VM with an optimization implemented. There will be some margin for error between any two runs depending on other work the testing system may be doing.

Example benchmark run:
1-alt-suite

Example saving reports and re-running against the loaded report:
2-alt-suite

Example navigating detailed reports between two executions:
3-alt-suite

### DIFF
--- a/src/playground/suite.js
+++ b/src/playground/suite.js
@@ -1,0 +1,385 @@
+const soon = (() => {
+    let _soon;
+    return () => {
+        if (!_soon) {
+            _soon = Promise.resolve()
+                .then(() => {
+                    _soon = null;
+                });
+        }
+        return _soon;
+    };
+})();
+
+class Emitter {
+    constructor () {
+        Object.defineProperty(this, '_listeners', {
+            value: {},
+            enumerable: false
+        });
+    }
+    on (name, listener, context) {
+        if (!this._listeners[name]) {
+            this._listeners[name] = [];
+        }
+
+        this._listeners[name].push(listener, context);
+    }
+    off (name, listener, context) {
+        if (this._listeners[name]) {
+            if (listener) {
+                for (let i = 0; i < this._listeners[name].length; i += 2) {
+                    if (
+                        this._listeners[name][i] === listener &&
+                        this._listeners[name][i + 1] === context) {
+                        this._listeners[name].splice(i, 2);
+                        i -= 2;
+                    }
+                }
+            } else {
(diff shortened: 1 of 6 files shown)


boazsender:
Pulled and tested, LGTM.

Also, love the bench and perma link links. Nice work!

thisandagain:
This looks great at first glance running locally @mzgoddard. 👍

src/playground/suite.js
```diff
@@ -0,0 +1,600 @@
...
+    return () => {
+        if (!_soon) {
+            _soon = Promise.resolve()
+                .then(() => {
+                    _soon = null;
+                });
+        }
+        return _soon;
+    };
+})();
```
rschamp, line 12: I don't have any suggestions, I just like this function.

rschamp (approved):
This is an amazing amount of work that looks and runs great! Thank you @mzgoddard!

Issue #902, opened 2018-01-17
Music extension does not play sounds on Safari


ericrosenbaum:
### Expected Behavior

The music extension blocks "play note" and "play drum" should play sounds.

### Actual Behavior

On Safari, they don't work.

### Operating System and Browser

Mac OS 10.12.6 Safari 11.0.2

PR #908, opened 2018-01-22
Fix music extension asset loading on safari

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/902

### Proposed Changes

When loading music extension sound assets, include a fallback from the promise-based API for audio decoding to the callback API. 

Note that this reveals another safari bug where the bufferSource onended function doesn't fire... I'll file another issue for that.

### Reason for Changes

This change is required to support both Safari (callback only) and Firefox (promise only).

This code is unfortunately partially redundant with code in the audio engine [here](https://github.com/LLK/scratch-audio/blob/84fdf5c76a41a4960e34282ea96687d9ef19d6aa/src/index.js#L204), but that doesn't seem important enough to refactor it.


paulkaplan (approved):
LG!

Issue #910, opened 2018-01-22
Music extension stops playing notes on Safari


ericrosenbaum:
### Expected Behavior

The music extension should be able to play an unlimited number of drum and instrument sounds.

### Actual Behavior

On Safari, after you've played about 30 sounds, it stops working.

This is due to the concurrency counter never getting decremented, because the audioBufferSourceNode's onended event (usually) does not fire. 

### Steps to Reproduce

- Add the music extension
- This stack works for a a few seconds, then stops playing:

![screen shot 2018-01-22 at 11 31 47 am](https://user-images.githubusercontent.com/567844/35231877-f1804cfa-ff67-11e7-816a-158da4ac681e.png)

### Operating System and Browser

Mac OS 10.13.2 Safari 11.0.2

kyleplo:
I can't recreate this in Chrome on Windows 10.

PR #911, opened 2018-01-22
Keep playing notes and drums forever on Safari

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/910

### Proposed Changes

In the music extension, store a reference to the audio bufferSourceNode each time you play a note or drum sound, then delete it when the sound ends. 

### Reason for Changes

On Safari, the onended event (usually) does not fire unless we store a reference to the bufferSourceNode.


paulkaplan (approved):
Looks good. Interesting catch with the buffer source not calling onended if there are no references to it.

PR #914, opened 2018-01-22
Keep block xml up-to-date after renaming a variable.

### LINKED ISSUE #723: After renaming a variable on the workspace, exported JSON doesn't reflect blocks on workspace
### Expected Behavior

After renaming a variable on the workspace then exporting to JSON, importing that JSON on a new workspace should show the new variable name.

### Actual Behavior

The old variable name shows after importing the JSON.

### Steps to Reproduce

1. Load playground in browser
2. Create new variable "TestVariable"
3. Drag new "TextVariable" oval block as well as "set x to 0" data block to workspace
4. Change variable dropdown in set block to "TestVariable"
5. In set block, select "Rename variable..."
6. In dialog, enter "NewTestVariable" and tap OK
7. Export to JSON
8. Refresh page and import the JSON

Note: Notice that in blocks.js, _getBlockParams receives an undefined block parameter and then a TypeError is encountered in that method: 
TypeError: Cannot read property 'fields' of undefined
    at Blocks._getBlockParams (http://localhost:8073/playground/scratch-vm.js:6376:35)
    at RenderedTarget.renameVariable (http://localhost:8073/playground/scratch-vm.js:31890:44)
    at Blocks.blocklyListen (http://localhost:8073/playground/scratch-vm.js:6074:27)
    at VirtualMachine.blockListener (http://localhost:8073/playground/scratch-vm.js:35526:43)
    at Blockly.WorkspaceSvg.Blockly.Workspace.fireChangeListener (http://localhost:8073/playground/vendor.js:18037:432)
    at Blockly.Events.fireNow_ (http://localhost:8073/playground/vendor.js:18128:467)

Mac OS 10.11.6 Chrome 61

### LINKED ISSUE #909: Go through blocks list and update xml on var_rename
### Expected Behavior

Renaming a variable should update the block xml (for each block using that variable) to reflect the new name of the variable.

### Actual Behavior

The variable model gets updated so that the variables list stays up-to-date, but the xml for each block that uses the variable keeps the old name.

This issue is mentioned in [scratch-blocks#1350](https://github.com/LLK/scratch-blocks/pull/1350)
Related to #723

kchadha:
### Resolves

Resolves #909 
Resolves #723

### Proposed Changes

Keep scratch-vm up-to-date after being notified that a variable has been renamed. Go through all of the blocks on all of the runtime targets and change the value of the field of the blocks that refer to the renamed variable. Update the value of these fields with the name of the new variable.

### Reason for Changes

As of [the variables_by_id work in scratch-blocks](https://github.com/LLK/scratch-blocks/pull/1350), scratch-blocks no longer emits individual block change events when a variable is renamed. This is because the block field values now refer to the id of the variable, and the id of the variable does not change if a variable is renamed. Thus, the vm emits outdated xml and JSON for the blocks that refer to any variable that has been renamed.

### Test Coverage

Existing tests pass.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -319,6 +319,10 @@ class Blocks {
             break;
         case 'var_rename':
             stage.renameVariable(e.varId, e.newName);
+            // Update all the blocks that use the renamed variable.
+            if (optRuntime) {
+                optRuntime.updateBlocksAfterVarRename(e.varId, e.newName);
+            }
             break;
         case 'var_delete':
             stage.deleteVariable(e.varId);
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -321,7 +321,11 @@ class Blocks {
             stage.renameVariable(e.varId, e.newName);
             // Update all the blocks that use the renamed variable.
             if (optRuntime) {
-                optRuntime.updateBlocksAfterVarRename(e.varId, e.newName);
+                const targets = optRuntime.targets;
+                for (let i = 0; i < targets.length; i++) {
+                    const currTarget = targets[i];
+                    currTarget.blocks.updateBlocksAfterVarRename(e.varId, e.newName);
+                }
             }
             break;
         case 'var_delete':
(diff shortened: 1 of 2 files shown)


paulkaplan:
Cool, I'll take a look at this soon. One thing I wanted to mention is that we have a possibly related issue when costumes, sounds and sprites are renamed. Do you think we should generalize this to handle those situations also?

kchadha:
@paulkaplan, it seems like those situations should be handled by the emitTargetsUpdate in the virtual-machine...? I'll look into this.

paulkaplan:
@kchadha the code looks good. But i noticed this is accessing the `blocks._blocks`, do you think we should move this into a method on the blocks container so the runtime can just call `target.blocks.updateAfterVarRename(...)` on each target? It would keep the private stuff private :)

kchadha:
@paulkaplan, good call! Done.

paulkaplan (approved):
LGTM

Issue #128, opened 2016-09-02
Project loaders: SB/1.4


tmickel:
We need to support the 1.4 format. Possibly even 1.4-formatted data could come from the website? But especially for any offline scenario.

tmickel:
Reference:

https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/util/ObjReader.as

tmickel:
This, too:
https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/util/OldProjectReader.as

TheBrokenRail:
I found a javascript GitHub project that might help https://github.com/trumank/sb.js

TheBrokenRail:
And from what I see with this library the names are pretty much the same as 2.0 so I think if the licence is compatible if it is detected as a 1.4 file you could use this library and give parseScratchObject the stage object of the return value

towerofnix:
This is definitely an important feature - why was this issue closed? I don't see any other issue (more recent or otherwise) for the general implementation of parsing 1.4 `sb` files.

rschamp:
Good catch @towerofnix — we had closed it because it was being tracked here https://github.com/LLK/scratch-gui/issues/972, but this is the original issue.  We're implementing 1.4 imports imminently.

PR #915, opened 2018-01-23
Support Scratch 1.4 costumes with text (that are imported from 2.0)

### LINKED ISSUE #672: Text in costumes from 1.4 SBs converted to SB2s is not loaded
(This is related to #128, but is partially specific to 1.4 SBs which were imported to 2.0, then loaded in 3.0.)

### Expected Behavior

The costume's text displays; for example:

2.0 screenshot

### Actual Behavior

The costume's text does not display:

3.0 screenshot

### Steps to Reproduce

1. Create and upload a 1.4 project that has a costume containing text, and have this project be saved as an SB2 (on the Scratch servers). I've already created [a project](https://scratch.mit.edu/projects/173088631) you can use to test (so you don't need to go through all these steps).
    1. Open Scratch 1.4. Edit (or add) a costume so that [it has a text in it](http://i.cubeupload.com/1URnWF.png).
    2. Save the project (it'll be a `.sb` file).
    3. Import the project into the online Scratch 2.0 editor.
    4. Make a note of the URL, and load the project into Scratch 3.0 (`#<project ID>`).
2. Look at the costume with the text in it. Note that the text doesn't appear.

Linux / Debian (testing), Firefox 52.2.0.

---

#### 2.0 Implementation

If a costume has a 1.4 text in it, Scratch 2.0 actually gives the costume *two* images: the `baseLayer` and the `textLayer`. That means there's four respective JSON fields:

```
{
  "baseLayerID": <id>,
  "baseLayerMD5": <md5>,
  "textLayerID": <id>,
  "textLayerMD5": <md5>
}
```

In the Scratch 2.0 editor, this makes it act as a vector image. For example, given this 1.4 costume:

Screenshot of 1.4 paint editor

..the costume displays as this vector image, in 2.0's costume editor:

Screenshot of 2.0 vector editor

Note that the text is displayed in Helvetica in the paint editor, but in the actual player it's displayed as it looks in Scratch 1.4:

Screenshot of entire 2.0 UI, with vector editor open

However, as soon as it's changed, the player shows it in Helvetica (that is, exactly as it appears in the vector editor):

Screenshot after changing the content of the text

#### Flattening

However, *flattening* the vector image -- that is, clicking on "bitmap editor" -- *without editing it*, converts it to how it looks in 1.4:

Screenshot after converting the unmodified image to a bitmap

Flattening is really just concatenating the base layer and text layer images (as gotten from the project JSON) together.

#### Intended behavior?

Perhaps the image should be flattened as soon as 3.0 sees it, rather than imported as a "pseudo-vector" costume. It's confusing for the image to display as a vector in 2.0, and for the font to change as soon as you edit the text. It's also complex to implement; there'd need to be a separate way of rendering these pseudo-vector costumes, since they're actually two bitmap images. (Rendering two images could also be slower performance-wise?)

I'd argue that flattening the images should be the behavior of 2.0 as well, but scratch-flash is now deprecated/frozen.

This is, of course, only a problem for 1.4 projects which were uploaded to and edited in the 2.0 editor, but those do exist; epninja's projects come to mind. For example, in [Extreme Ninja Parkour](https://scratch.mit.edu/projects/11858411/), the text at the bottom of the first screen is from 1.4 and [does not display in 3.0](https://llk.github.io/scratch-gui/develop/#11858411).

towerofnix:
**This pull request is a work-in-progress.** Most of the code was written several months ago and may be more complicated than is now necessary.

### Resolves

Fixes #672.

### Proposed Changes

Adds support for importing Scratch 1.4 costumes that contain text from Scratch 2.0 projects. This is done by fetching both the "base" and "text" layers of the costume, which are both bitmap images, then flattening the two into a single bitmap image.

This implementation differs from Scratch 2.0, which creates a vector image including a "pseudo"-text element and a bitmap image embed. That seemed rather confusing, and certainly difficult to work with.

### Reason for Changes

To implement features which Scratch 2.0 has.

### Test Coverage

No test coverage.


towerofnix:
The current Big Challenge: These costumes cannot be duplicated (or rather, sprites containing these costumes cannot deleted), potentially because their assetIds are not set (correctly). However, the costumes of modified sprites can be deleted.. so looking at how assetIds of modified (or newly-created) costumes works should help.

Edit: this is because of these lines, which make `costumeAsset` = null, by the way:

https://github.com/LLK/scratch-vm/blob/115184a2a39b990dee9218aa15081b334516e1a8/src/sprites/sprite.js#L92-L97

towerofnix:
The relevant code for this is in VM.updateSvg:

https://github.com/LLK/scratch-vm/blob/115184a2a39b990dee9218aa15081b334516e1a8/src/virtual-machine.js#L432-L437

So the updated SVG is stored in a cache of sorts. Now to figure out how to store a bitmap(?)...

src/import/load-costume.js
```diff
@@ -183,9 +183,36 @@ const loadOldTextCostume = function (baseMD5ext, textMD5ext, costume, runtime) {
...
+                        removeEventListeners();
+                        reject();
+                    };
+                    const onLoad = function () {
+                        // eslint-disable-next-line no-use-before-define
+                        removeEventListeners();
+                        costume.assetId = runtime.storage.builtinHelper.cache(
+                            assetType,
+                            runtime.storage.DataFormat.PNG,
+                            new Buffer(reader.result)
```
towerofnix, line 200: ESLint is failing because `Buffer` is not defined, but when I test this, Buffer *is* defined (the code works). Removing `Buffer` (to just pass cache `reader.result`) stops the code from working again.

Several documentation files mention taking `Buffer` objects, but (seemingly) nowhere are Buffers actually `new`ly-created. I'm not sure how I should actually construct one from my own scratch-gui code?

thisandagain:
@cwillisf Can you please weigh in here?

src/import/load-costume.js
cwillisf: Hmm... this is a bit tricky because of the differences between Node.js and the web. We've avoided the problem so far by letting `nets` and `binary-loader` create all of our buffer objects for us, but that doesn't help in your case. There's a workaround for this in `handleNewBlankCostume` but I'd rather not propagate that code elsewhere. I'm not certain it would work for this case, anyway.

The basic problem is that `Buffer` is a type in Node.js but it doesn't exist on the web. Webpack transforms `Buffer` into `Uint8Array` or similar, with a bit of wrapping code. However, `scratch-gui`'s lint settings are set to allow only browser-style code, which means there's currently no way to do this in `scratch-gui`.

I'll have to think about it a bit more, but my initial thought is to do one of these things in `scratch-storage`:
- Alter `cache` (and probably `setData` too, for consistency) to handle more types of data, including `Blob`
- Add a way to create a new `Asset` with an empty `data` which you can then fill and provide to `cache` afterward
- Expose the `Buffer` constructor, ensuring that any code which can access `scratch-storage` can make a `Buffer` (or augmented `Uint8Array`, depending on environment and settings) of the same type that `scratch-storage` is already using.

I should be able to do one of these things next week. @rschamp (or anyone else), any thoughts on which one(s) of the above would be useful in the largest number of situations?

cwillisf:
I'm marking this as blocked for now, and I filed LLK/scratch-storage#30 to track the issue.

cwillisf:
Sorry for letting this hang for so long. I think the best move is to close this, since there have been significant changes to `scratch-storage` as well as the costume loading pipeline, including the creation of `scratch-svg-renderer`. Parts of `scratch-svg-renderer`, such as the font replacement step, may be relevant to this work. We also plan on supporting the import of Scratch 1.4 projects (see #128) so it might make sense to hold onto this until after that work is completed.

If you'd like to update these changes to work with those new features then I'm happy to take another look; if not I certainly understand. Either way, thanks for your contributions!

towerofnix:
Thanks for checking in on this!

> We also plan on supporting the import of Scratch 1.4 projects (see #128) so it might make sense to hold onto this until after that work is completed.

That's fine with me! It should definitely be noted that this is exclusively for 1.4 projects which were then opened and saved in 2.0, although the "merge the two images together" step will probably be relevant in 1.4 import too.

Definitely agreed that most of this code is unusable, or at least not directly usable, presently. Best to leave this PR closed and as needing further discussion (then to open a new PR with fresh code).

---

Just some quick notes, for picking up work again on this whenever:

The core of what we need to do here is:

1. If there's no "text" layer, just stop and behave as normal.
2. Fetch both the base *and* text layers.
3. Merge the two together.

That second step is totally exclusive to 1.4 projects saved in 2.0, since in Scratch 1.4 projects, while there might indeed be a separate bitmap layer for text (I don't know for sure since I don't know the `.sb` project format), that would be included inside the Scratch 1.4 project binary. Assets in 2.0 projects are (typically) stored separately from the project JSON, i.e. on Scratch's servers, so we need to make sure to fetch the text layer as well as the base layer, when present.

Assuming there *is* a separate text bitmap layer in Scratch 1.4 projects, the third step, merging the base layer and text layer, will be relevant when importing 1.4 projects, since obviously they should be flattened the same way.

>Parts of `scratch-svg-renderer`, such as the font replacement step, may be relevant to this work.

Maybe, but I don't really think so. The text layer in 1.4 projects (or at least ones which were later edited in 2.0) is its own bitmap image. Assuming that we would want to convert the font used in 1.4 projects to a 3.0-valid font, the same way we do with converting the fonts used in 2.0 projects, I'm not sure that would even be possible. *If* the font family is even stored in the SB2 (it must be in 1.4, but I'm not sure it is in 2.0), there would be no way to realistically translate it to a 3.0-valid font, because *any* font which is installed on the author's system may be used in a 1.4 project. The magic is that, since the text is saved as a bitmap image, that font will "work" on any computer even if it's not installed, since really Scratch (in version 1.4 and 2.0) is loading a screenshot of the text. (If you, if you do not have the font family installed on your computer, go to edit the text in 1.4, the font family will be lost; it will reset to the default Helvetica font. In 2.0 editing the text will just scrap the font altogether regardless of whether you previously had it installed, I think, because 2.0 only supports the basic sans-serif/Marker/Scratch/etc fonts.)

Anyways, importing a 1.4 costume, which is composed of two bitmap images, as a *vector* image would probably be a mess anyways. The resulting image would be a vector image composed of a bitmap image (the base layer) and a text object (generated according to the text). A user would probably end up converting the image to a bitmap image (the "convert to bitmap" button), in which case you'd have been better off just merging the base and text layers in the first place - which would have been the most accurate anyways.

Maybe I'm totally misinterpreting what you were thinking with using `scratch-svg-renderer` here..?

My proposed solution, and what I started work on here, is to take the two bitmap layers, base and text, and merge/flatten them, creating a new bitmap asset in the process (which is used for the resulting costume).

PR #917, opened 2018-01-24
Push reported

### LINKED ISSUE #781: Rework pushReportedValue for performance
### Expected Behavior

Run faster than it does now.

### Actual Behavior

While not itself slow, `Thread.pushReportedValue` is called a lot in Scratch projects so they can return a value to a higher level of the Thread stack. This makes it a prime opportunity to have a large performance gain from a fairly small amount of code distributed in `thread.js` and `execute.js`.

`Thread.pushReportedValue` and related codes and structures can be rewritten to decrease assignments to and reading from computed member expressions (e.g. `currentStackFrame.reported[inputName]`. These aren't slow operations, but it is used frequently enough that using faster operations will provide a measurable bump in performance.

One way this might be replaced would be using a dedicated member on currentStackFrame and assigning to it, and having a argValues object cached here that execute can assign to. If execute can build all the args synchronously it'll be able to build argValues and continue normally. If asynchronous, it can store argValues (kind of like how reported is on the currentStackFrame) and read from the dedicated reporting member next time. The saved time here is from assigning to a known currentStackFrame member that JS vms can speed up compared to computed member assignments. Another simple change is to remove the if statement in `pushReportedValue` by assigning to a fake stack frame if there isn't a parent. Assuming there is always a stack frame and assigning to a dedicated member instead of needing to test first has an opportunity to also provide a boost.

It will still be important to measure and make sure these possible changes and other changes around pushReportedValue actually do so.

### Steps to Reproduce

Measure performance with a performance inspector like Chrome's Performance DevTools tab. pushReportedValue will have a larger presentation as scratch features are used like reading variables, math operators, and other blocks that return values to a higher stack member of the Thread.

mzgoddard:
### Resolves

#781

### Proposed Changes

- Report at `parentStackFrame.justReported`
- Fill `currentStackFrame.reported` if the thread pauses.
- Test that `thread`.target is not null before calling `execute` from a Sequencer.
- Assign a thread's Blocks to `blockContainer`.
- Add a third optional argument argument that lets execute calls let executes they make know they are being recursed and let some logic shortcut faster.
- Add a Blocks cache only available to `execute`.

#### `justReported`

Speed up block reporting by normally building one object, `argValues`, dynamically. Reporting block results to `justReported` lets `execute` read that value from a member statically. If the thread pauses values already stored on argValues can be copied to `currentStackFrame.reported`. For execute overall this makes passing a resulting value from one execute call to the parent execute call faster with less dynamic stores on and lookups in `reported`.

#### `target.thread` / `blockContainer`

Move `target.thread` null check from `execute` to Sequencer. Move `blockContainer` lookup to a new Thread member. This moves some branching out of `execute` so they happen less (`target.thread`) or not at all (`blockContainer`).

#### `BlocksExecuteCache`

A lot of the test results and lookups in `execute` relate to the executed block. `BlocksExecuteCache` presents an access point to `execute` from `Blocks` to a private cache that is reset with the other `Blocks` caches. `execute` then also initializes the per-block-cache with branch test results and lookups it would make up for every execution of that individual block, saving time reconsidering a test whose results will not change for that block or on lookups that will not change.

- Cache a reference to blockFunction and the isHat value (saves dynamic string lookups)
- Cache the shadow value if the block is a shadow value (saves calling Object.keys on fields and inputs)
- Cache local execute cache copies of block inputs and fields to know that the keys are their own properties and to remove input.custom_block if its there
- Cache fields VARIABLE, LIST, or BROADCAST_OPTION and booleans to check if that is what fields contains to assign these regular fields to argValues quickly

### Numbers

Here are some gathered metrics from the benchmark suite using blocks performed per second.

Samsung Chromebook Plus and iPhone 6S+ see big gains.

<del>Projects on a Samsung Chromebook Plus are 25% to 50% faster.</del>

<details>
<summary>Samsung Chromebook Pro</summary>

Project ID          | before |  after |    %
--------------------|--------|--------|-----
Floating Blocks     | 252539 | 278216 | 10.1
Scratch Cat         | 292586 | 319532 |  9.2
Bouncy Heros        | 146079 | 153622 |  5.1
Stacky Towers       | 258688 | 296762 | 14.7 
3D Solar System     | 245551 | 291313 | 18.6
Pixel Art Maker     | 284832 | 323312 | 13.5
Dynamic Maurer Rose |  15780 |  16842 |  6.7

</details>

<details>
<summary>Samsung Chromebook Plus</summary>

Project ID          | before |  after |    %
--------------------|--------|--------|-----
Floating Blocks     |  57723 |  65991 | 14.3
Scratch Cat         |  76960 |  81804 |  6.2
Bouncy Heros        |  38491 |  42662 | 10.8
Stacky Towers       |  71616 |  76758 |  7.1
3D Solar System     |  64093 |  79020 | 23.2
Pixel Art Maker     |  69604 |  81340 | 16.8
Dynamic Maurer Rose |  17550 |  18373 |  4.6

</details>

<details>
<summary>iPhone 6S+</summary>

Project ID          | before |  after |    %
--------------------|--------|--------|-----
Floating Blocks     | 146089 | 173187 | 18.5
Scratch Cat         | 233421 | 248157 |  6.3
Bouncy Heros        |  65755 |  65755 |  0.0
Stacky Towers       | 123809 | 191943 | 55.0
3D Solar System     | 306352 | 476707 | 55.6
Pixel Art Maker     | 211448 | 291472 | 37.8
Dynamic Maurer Rose |  28037 |  37821 | 34.8

</details>

<details>
<summary>15" 2017 MacBookPro</summary>

Project ID          | before |  after |    %
--------------------|--------|--------|-----
Floating Blocks     | 412900 | 445739 |  7.9
Scratch Cat         | 482179 | 520419 |  7.9
Bouncy Heros        | 181250 | 208117 | 14.8
Stacky Towers       | 485965 | 529686 |  8.9
3D Solar System     | 458371 | 491895 |  7.3
Pixel Art Maker     | 467683 | 501665 |  7.2
Dynamic Maurer Rose |  79143 |  81413 |  2.8

</details>

### Test Coverage

Update affected thread and sequencer tests.

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -113,14 +113,6 @@ const RECURSIVE = true;
  */
 const execute = function (sequencer, thread, recursiveCall) {
     const runtime = sequencer.runtime;
-    const target = thread.target;
-
-    // Stop if block or target no longer exists.
-    if (target === null) {
-        // No block found: stop the thread; script no longer exists.
-        sequencer.retireThread(thread);
-        return;
-    }
 
     // Current block to execute is the one on the top of the stack.
     const currentBlockId = thread.peekStack();
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -1,4 +1,5 @@
 const BlockUtility = require('./block-utility');
+const BlocksExecuteCache = require('./blocks-execute-cache');
 const log = require('../util/log');
 const Thread = require('./thread');
 const {Map} = require('immutable');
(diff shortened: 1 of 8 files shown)


mzgoddard:
Lots more info on the changes in the commit messages too.

src/engine/execute.js
```diff
@@ -98,32 +99,27 @@ const handleReport = function (
     }
 };
 
+/**
+ * A convenienve constant to hide that the recursiveCall argument to execute is
```
kchadha, line 103: Typo in spelling of 'convenience'/'convenient'

kchadha (commented):
This looks good to me!
There are a lot of changes, so someone who is more familiar with the execute function should also take a look. The parts of it that changed that are related to things I wrote (e.g. variable/list/broadcast stuff, and some of the shadow stuff) looks good to me.

Thanks for all of this work!

src/engine/sequencer.js
```diff
@@ -197,7 +197,11 @@ class Sequencer {
                 this.runtime.profiler.records.push(
                     this.runtime.profiler.START, executeProfilerId, null, performance.now());
             }
-            execute(this, thread);
+            if (thread.target === null) {
+                this.retireThread(thread);
+            } else {
+                execute(this, thread);
+            }
```
cwillisf, line 204: Ooh, I like this change! 👍

cwillisf (approved):
Nice work!

At first the `recursiveCall` flag felt a bit hacky to me, but now that I see how it's being used it makes sense. In a future version it might make sense to rename the flag to something indicating its meaning in a block-centric sense. I can't think of a snappy way to say "this is/isn't the outermost block in this execution step" without running into naming collisions or too-vague names using "root" or something. Just thinking out loud :)

Issue #865, opened 2017-12-15
Scroll Wheel runtime detection is broken


griffpatch:
### Expected Behavior

Moving the scroll wheel while a project is running should trigger the "when up and down arrow pressed?"
hat blocks (but not the <key up/down pressed?> detection blocks. (See: https://scratch.mit.edu/projects/229142/ or https://llk.github.io/scratch-gui/develop/#193724034).

This functionality was also recently broken in the Scratch 2 (most likely when the keyboard code was changed around to support the edge browser).

### Actual Behavior

The  "when up and down arrow pressed?" hat block does not fire.

### Steps to Reproduce

Fire up this project: https://llk.github.io/scratch-gui/develop/#193724034 and try it out. (I remixed: https://scratch.mit.edu/projects/229142/ as it was a Scratch 1.3 project so doesn't load in scratch 3 yet). 

### Operating System and Browser

Windows PC running Chrome

cwillisf:
Good news: this will be fixed in version 459 and above, which we are planning to release very soon.

Edit: oops, I didn't realize this was in the 3.0 VM repository. Anyway, the Scratch 2.0 bug will be fixed very soon ;)

PR #920, opened 2018-01-26
Scroll wheel detection for "when key (up/down) pressed" blocks

towerofnix:
### Resolves

Towards #865. LLK/scratch-gui#1352 is needed to actually fix the issue and expose this feature.

### Proposed Changes

This adds an IO class for detecting the mouse wheel being scrolled.

**Note the following conditions are for if the "natural scrolling" (or similar) system setting is disabled.**

* Scrolling the mouse wheel down activates "when down arrow key pressed" blocks.
* Scrolling the mouse wheel up activates "when up arrow key pressed" blocks.

If "natural scrolling" is enabled, these are flipped; scrolling the mouse wheel down activates up-key-pressed, and vice versa. This is consistent with how Scratch 2.0 and 1.4 behave, and usually is wanted; theoretically, this scrollwheel-detection is most likely to be used to implement actual scrolling systems within Scratch. So "when up key pressed" would be used to move content up, and if the user has "natural scrolling" enabled, then this would be when the user turns the mouse wheel *down* - and the user would already be used to moving the wheel down meaning moving the content up.

(All this is more intuitive if you try it out yourself! Project [`66833944`](https://scratch.mit.edu/projects/66833944/) is a good demonstration.)

### Reason for Changes

For consistency with Scratch 2.0 and 1.4. Also, this should work on all browsers (where a mouse is used); Scratch 2.0 (un-)notoriously didn't support this scroll wheel detection trick on all browsers. (I remember only Chrome would detect them on my Mac.)

### Test Coverage

Basic unit tests are included; they mock the runtime to see what blocks are activated by scrolling.

This has been manually tested with the projects mentioned in issue #865 (namely, https://scratch.mit.edu/projects/229142/).


paulkaplan:
Wow thanks @towerofnix, sorry about the delay in reviewing. @cwillisf you might be the best person to review this since you were recently dealing with these events?

thisandagain:
@cwillisf Can you please review?

cwillisf (approved):
Sorry for the delay... this looks good to me! Thanks!
I especially appreciate that it comes with tests :)

PR #921, opened 2018-01-29
Playground: load renderer as UMD

cwillisf:
### Resolves

The recent changes to the renderer's build output packaging broke this repository's playground / benchmark suite. These changes aren't the only way to fix the issue, but this is consistent with the way that the storage module was already being loaded.

### Proposed Changes

Load `ScratchRender` as a separate `<script>` tag in the playground's `index.html` file. This exposes the renderer constructor as `ScratchRender` on the global object, instead of `RenderWebGL` as before.

### Reason for Changes

See LLK/scratch-render#227

### Test Coverage

I tested both the benchmark playground (`index.html`) and the full benchmarking suite (`suite.html`) locally.

### DIFF
--- a/src/playground/benchmark.js
+++ b/src/playground/benchmark.js
@@ -476,7 +476,7 @@ const runBenchmark = function () {
 
     // Instantiate the renderer and connect it to the VM.
     const canvas = document.getElementById('scratch-stage');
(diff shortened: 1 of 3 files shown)


paulkaplan (approved):
LG

Issue #899, opened 2018-01-17
Clamping sound pitch effect


kchadha:
### Expected Behavior

Executing blocks with valid inputs should most often have perceivable effects to the user.

### Actual Behavior

Setting the pitch to a really big negative number takes a long time to execute with no perceivable effect to the user.

The pitch effect should possibly be clamped to reasonable bounds so that effects can be perceived and code doesn't just seem to be hanging.

### Operating System and Browser

_e.g. Mac OS 10.11.6 Safari 10.0_

ericrosenbaum:
Currently the pitch effect is clamped from -600 to 600. It is 120 steps per octave, so that's from -5 octaves to +5 octaves of pitch shifting. Maybe that's too much. :D

The issues at extreme parts of the range include:

- At extreme low values, you can't hear anything (mentioned above)
- At extreme low values, the sound is too long (also mentioned above). Pitch effect of -600 is shifted down 5 octaves, which will double the duration 5 times. So a sound that's 1 second long will last 32 seconds!
- At extreme high values, the sound can be become so short it's just a tiny click.
- At both high and low values, the sound becomes increasingly distorted (I think this is mostly interpolation and aliasing artifacts)

I did some experimenting with a variety of sounds, and I think a range of three octaves in either direction, -360 to 360, would work well. Generally, sounds are still audible at both -360 and 360, and the distortion is not terrible. At -360, the duration is doubled three times, so a 1 second sound lasts 8 seconds, which is long but not crazy.

PR #926, opened 2018-01-31
Clamp pitch effect range to -360 to 360

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/899

### Proposed Changes

Reduce the range of the pitch effect so that it is clamped from -360 to 360 (from three octaves down to three octaves up).

### Reason for Changes

Previously the pitch effect was clamped from -600 to 600. It is 120 steps per octave, so that's from -5 octaves to +5 octaves of pitch shifting. Maybe that was too much. :D

The issues at extreme parts of the range include:

- At extreme low values, you can't hear anything.
- At extreme low values, the sound is too long. Pitch effect of -600 is shifted down 5 octaves, which will double the duration 5 times. So a sound that's 1 second long will last 32 seconds!
- At extreme high values, the sound can be become so short it's just a tiny click.
- At both high and low values, the sound becomes increasingly distorted (I think this is mostly interpolation and aliasing artifacts)

I did some experimenting with a variety of sounds, and I think a range of three octaves in either direction, -360 to 360, will work well. Generally, sounds are still audible at both -360 and 360, and the distortion is not terrible. At -360, the duration is doubled three times, so a 1 second sound lasts 8 seconds, which is long but not crazy.

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -74,7 +74,7 @@ class Scratch3SoundBlocks {
      */
     static get EFFECT_RANGE () {
         return {
-            pitch: {min: -600, max: 600}, // -5 to 5 octaves
+            pitch: {min: -360, max: 360}, // -3 to 3 octaves
             pan: {min: -100, max: 100} // 100% left to 100% right
         };
     }


kchadha (approved):
I think this sounds way better in comparison to the old range.

Some follow-up questions: 
1) Can/should the same kind of clamping behavior apply to the 'faster'/'slower' effects in the sound editor?

2) Is there a way we can communicate this range to the user, either through on-boarding material, documentation, through the blocks, or all of the above? For an example of 'through the blocks', in Scratch 2.0, the tempo seems to have a limit, of 20 to 500, which is implicitly communicated to the user via the tempo reporter block monitor (e.g. trying to set the tempo to a really big number results in it being set to 500). This brings me to my next question.

3) Should the pitch effect (and others) have reporters/monitors associated with them?

ericrosenbaum:
Thanks @kchadha! 

1. Good question- I'll file an issue with 'needs-discussion' for clamping the effects in the sound editor.

2. To communicate the range I think our best option right now (as discussed in the meeting) is to use the visible range on slider fields for numeric inputs (which we will add later).

3. I think we're not planning to add reporters for the sound effect values- this keeps them parallel to the graphic effects blocks, and helps keep the blocks menu small.

PR #930, opened 2018-02-02
Only measure loudness once per step

### LINKED ISSUE #72: Loudness measurement should be rate-limited
### Expected behavior

If you use several loudness blocks, they should not cause the loudness analysis code to run multiple times in a single scratch animation frame. 

### Actual behavior

Each time a loudness reporter is called it runs the loudness analysis.

### Steps to reproduce

This program uses a run-without-screen-refresh custom block to check the loudness 100,000 times, and measure the time it takes:

screen shot 2018-01-16 at 6 05 53 pm

On Scratch 2.0 on my machine, this reports 0.08 seconds, and on Scratch 3.0 it reports about 0.45 seconds.

### Notes

One strategy for improving this would be store a timestamp each time the loudness is requested, and also store the loudness value itself. Then, each time, if the current time is long enough after the timestamp (e.g. one frame at 30fps- since we are not measuring the platform's audio buffer duration), then make a new measurement- otherwise just return the stored value. For the program above, this should result in the loudness analysis being done once each time the custom block is run, instead of 100,000 times.

### Operating system and browser

Mac OS 10.12.6 Chrome 63

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-audio/issues/72

### Proposed Changes

This change stores the calculated loudness value and the time. When loudness is requested again, it uses the stored value unless more than a frame's duration has passed. 

### Reason for Changes

Calculating the loudness in the audio engine is fairly expensive, and the audio buffers update at a rate close to the frame rate, so we should calculate the loudness value ideally at most once per frame.

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -1,4 +1,5 @@
 const Cast = require('../util/cast');
+const Timer = require('../util/timer');
 
 class Scratch3SensingBlocks {
     constructor (runtime) {
@@ -23,6 +24,8 @@ class Scratch3SensingBlocks {
         this.runtime.on('ANSWER', this._onAnswer.bind(this));
         this.runtime.on('PROJECT_START', this._resetAnswer.bind(this));
         this.runtime.on('PROJECT_STOP_ALL', this._clearAllQuestions.bind(this));
+
+        this.timer = new Timer();
     }
 
     /**
@@ -217,7 +220,19 @@ class Scratch3SensingBlocks {
 
     getLoudness () {
         if (typeof this.runtime.audioEngine === 'undefined') return -1;
-        return this.runtime.audioEngine.getLoudness();
+        if (typeof this.runtime.currentStepTime === 'undefined') return -1;
+
+        // Only measure loudness once per step
+        if (this.loudnessTime) {
+            const timeSinceLoudness = this.timer.time() - this.loudnessTime;
+            if (timeSinceLoudness < this.runtime.currentStepTime) {
+                return this.loudness;
+            }
+        }
+
+        this.loudnessTime = this.timer.time();
+        this.loudness = this.runtime.audioEngine.getLoudness();
+        return this.loudness;
     }
 
     getAttributeOf (args) {
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -236,7 +236,7 @@ class Scratch3SensingBlocks {
 
     getLoudness () {
         if (typeof this.runtime.audioEngine === 'undefined') return -1;
-        if (typeof this.runtime.currentStepTime === 'undefined') return -1;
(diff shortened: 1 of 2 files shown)


src/blocks/scratch3_sensing.js
```diff
@@ -23,6 +24,8 @@ class Scratch3SensingBlocks {
         this.runtime.on('ANSWER', this._onAnswer.bind(this));
         this.runtime.on('PROJECT_START', this._resetAnswer.bind(this));
         this.runtime.on('PROJECT_STOP_ALL', this._clearAllQuestions.bind(this));
+
+        this.timer = new Timer();
```
paulkaplan: A few things here:
1. Can you put this up with the other properties that are initialized (below _answer ^)
2. Can you change this to be `_` prefixed if it isn't going to be accessed externally.
3. There are two other private properties you added in `getLoudness`, (`this.loudness` and `this.loudnessTime`), those should also be initialized here, use the `_` prefix, and can you update the names to be more clear (`this.loudness` should probably be `this._cachedLoudness` and `this.loudnessTime` should probably be `this._loudnessLastMeasuredTime` or something similarly explicit).

src/blocks/scratch3_sensing.js
```diff
@@ -217,7 +220,19 @@ class Scratch3SensingBlocks {
 
     getLoudness () {
         if (typeof this.runtime.audioEngine === 'undefined') return -1;
-        return this.runtime.audioEngine.getLoudness();
+        if (typeof this.runtime.currentStepTime === 'undefined') return -1;
+
+        // Only measure loudness once per step
+        if (this.loudnessTime) {
```
paulkaplan: Once you initialize loudnessTime in the constructor, you won't need this outer `if`, because you can initialize loudness time to zero, and timer.time() - 0 will always be larger than the step time on the first step.

paulkaplan (dismissed):
I left a comment about a couple of minor name/initialization things. 

Because this PR adds a level of logic to getting the loudness this a good place to add a unit test for that logic. I can show you how to add a unit test for this if you want. Basically, I would expect a test to stub out the `audioEngine.getLoudness` and check that (1) the two early return conditions operate as expected (2) the stubbed audio engine is used to get loudness in a few edge cases where the cache would be empty and (3) does not use the audio engine to get loudness if it already has been gotten.

src/blocks/scratch3_sensing.js
ericrosenbaum: all set

src/blocks/scratch3_sensing.js
ericrosenbaum: all set

ericrosenbaum:
Okay, I made the fixes and added a unit test. The unit test required some trickery to simulate time passing- not sure I did it in a good way (e.g. I had to stub the timer, which is private), so it may need further work.

paulkaplan (approved):

PR #934, opened 2018-02-13
Extensions dynamic menu

picklesrus:
### Resolves

I didn't see a feature request for this. Let me know if there is one!

### Proposed Changes

Add support for dynamic menus in extensions.  

### Reason for Changes

Resolving a TODO.

### Test Coverage

I added integration level tests for both 2 types of menus.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -551,22 +551,23 @@ class Runtime extends EventEmitter {
      */
     _buildMenuForScratchBlocks (menuName, menuItems, categoryInfo) {
         const menuId = this._makeExtensionMenuId(menuName, categoryInfo.id);
-
-        /** @TODO: support dynamic menus when 'menuItems' is a method name string (see extension spec) */
-        if (typeof menuItems === 'string') {
-            throw new Error(`Dynamic extension menus are not yet supported. Menu name: ${menuName}`);
-        }
-        const options = menuItems.map(item => {
-            switch (typeof item) {
-            case 'string':
-                return [item, item];
-            case 'object':
-                return [item.text, item.value];
-            default:
-                throw new Error(`Can't interpret menu item: ${item}`);
+        var options = null;
+        if (typeof menuItems === 'function') {
+             options = function () {
+                return menuItems();
             }
-        });
-
+        } else {
+            options = menuItems.map(item => {
+                switch (typeof item) {
+                case 'string':
+                    return [item, item];
+                case 'object':
+                    return [item.text, item.value];
+                default:
+                    throw new Error(`Can't interpret menu item: ${item}`);
+                }
+            });
+        }
         return {
             json: {
                 message0: '%1',
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -551,11 +551,11 @@ class Runtime extends EventEmitter {
      */
     _buildMenuForScratchBlocks (menuName, menuItems, categoryInfo) {
         const menuId = this._makeExtensionMenuId(menuName, categoryInfo.id);
-        var options = null;
+        let options = null;
         if (typeof menuItems === 'function') {
-             options = function () {
+            options = function () {
                 return menuItems();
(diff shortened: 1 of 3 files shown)


picklesrus:
I figured this was far enough along that a PR was right way to continue the conversation.  I also wanted to get travis to run and didn't know how to do that otherwise :) (I'll fix the lint errors from Travis when I get back on Thursday - I have the wrong eslint config right now)

I'll essentially be out the next 2 days so no super hurry.

src/engine/runtime.js
```diff
@@ -551,22 +551,23 @@ class Runtime extends EventEmitter {
...
         const menuId = this._makeExtensionMenuId(menuName, categoryInfo.id);
-
-        /** @TODO: support dynamic menus when 'menuItems' is a method name string (see extension spec) */
-        if (typeof menuItems === 'string') {
-            throw new Error(`Dynamic extension menus are not yet supported. Menu name: ${menuName}`);
+        let options = null;
+        if (typeof menuItems === 'function') {
+            options = function () {
+                return menuItems();
+            };
```
cwillisf: Could this just be `options = menuItems`?

src/extension-support/extension-manager.js
```diff
@@ -242,8 +242,34 @@ class ExtensionManager {
...
+        const menuNames = Object.getOwnPropertyNames(menus);
+        for (let i = 0; i < menuNames.length; i++) {
+            const item = menuNames[i];
+            // If the value is a string, it should be the name of a function in the
+            // extension object to call to populate the menu whenever it is opened.
+            // Set up the binding for the function object here so
+            // we can use it later when converting the menu for Scratch Blocks.
+            if (typeof menus[item] === 'string') {
+                const serviceObject = dispatch.services[serviceName];
+                const menuFunc = serviceObject[menus[item]].bind(serviceObject);
```
cwillisf: Ideally this should be something like `dispatch.call.bind(dispatch, serviceName, menus[item])` but that's only OK if you can consume a promise for the menu items. If that's not easy then I think this is fine for now. We'll need to figure out how to deal with worker extensions later, though.

cwillisf (changes requested):
Looks good! I marked a couple points for discussion but it wouldn't surprise me if we end up merging this as-is :)

picklesrus:
Figured out eslint configs so lint is all fixed now and Travis is happy. Ready for a review.

src/extension-support/extension-manager.js
picklesrus: I dug into this a little and the problem I'm having right now is that it actually errors out in the blocks code while trying to replace the menu (whether I return a promise from the extension's menu build function or not).  I tried returning a few different formats for the menu but haven't had any luck yet.  It is tricky to debug since the blocks code is compressed into a giant blob. Is there a way to run the gui with uncompressed blocks?  That would make it much easier to tell if I'm on the right track or in a rabbit hole :)

src/engine/runtime.js
picklesrus: Yes :) Fixed!

src/extension-support/extension-manager.js
picklesrus: As discussed, punting on this for now since scratch-blocks doesn't currently accept the promise or have a mechanism to update the menu outside of the function called when the menu is clicked. Added a TODO.

cwillisf (approved):
@picklesrus and I talked elsewhere and came to the conclusion that fully-dynamic menus (calling the menu function at menu-open time) are not plausible just yet, so with the most recent changes this looks good for now. We'll need to make significant changes once we bring back worker-based extensions, as the TODO indicates, but that shouldn't block this change.

cwillisf:
Oh, I guess I should double-check: @picklesrus, is this ready to merge? It looks good to me...

picklesrus:
Should be! Tests passed and it worked for me on Friday :).

PR #936, opened 2018-02-14
Feature/require music assets

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-storage/issues/27

### Proposed Changes

Instead of loading the static sound assets for the music extension via scratch-storage, require them via a manifest that causes them to download at the time the editor loads. When the extension loads, proceed to decoding them.

### Reason for Changes

Scratch storage wasn't designed for extension static assets in the first place. 

Additional extensions created by our team that require static assets will be able to load them using this technique.


rschamp (dismissed):
Tests are failing because we're using loaders which don't play nicely with tap/node. The long-term fix is to change our runner to jest which is able to mock webpack loaders and is what we need to use to run React tests elsewhere.  We've talked about this for extensions in the vm before, sorry I forgot about this issue when we were pairing.

In the short term I think we could just wrap the require in a `try ... catch`. This lets tests pass for me locally:
```javascript
let assetData = {};
try {
    assetData = require('./manifest');
} catch (e) {
    // Non-webpack environment, don't worry about assets.
}
```

PR #939, opened 2018-02-20
Block drag

fsih:
### Resolves
Depends on https://github.com/LLK/scratch-blocks/pull/1384
Pass the drag over sprite event along to the GUI

### Proposed Changes

When receiving an end drag event over a  sprite in the target pane, add that code to the target sprite's workspace

### Reason for Changes

Working toward share the love feature


paulkaplan (approved):
LG! thanks for doing that refactor

PR #942, opened 2018-02-22
Make sprite.costumes private so that it's enforceable that all costumes are given a unique ID

fsih:
### Resolves
Should go in after https://github.com/LLK/scratch-gui/pull/1480

This resolves an issue that needs to be resolved before https://github.com/LLK/scratch-gui/issues/923
The paint editor needs to be given a unique ID associated with each costume in order to recognize when the costume it's editing has changed. Different costumes should have different undo stacks, so that you can't undo into a different costume's stack.

Right now, the ID is not uniquely identifying enough, and the duplicate costume function gives the exact same ID for both the old and new costume. This change "privatizes" target.sprite.costumes so that sprites can enforce that each costume has a unique ID.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -332,7 +332,7 @@ class Scratch3LooksBlocks {
         }
         if (target === this.runtime.getTargetForStage()) {
             // Target is the stage - start hats.
-            const newName = target.sprite.costumes[target.currentCostume].name;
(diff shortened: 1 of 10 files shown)


fsih:
Hang on, I'm going to switch this to a setter on Ray's suggestion

fsih:
I have changed costumes to use a setter and getter

fsih:
Nooo it's doing the wrong thing in the renderer

PR #943, opened 2018-02-23
Make sprite.costumes private

fsih:
I broke this out of https://github.com/LLK/scratch-vm/pull/942 since it turned out we didn't need to add IDs, but I think it's still a good organizational change that accessors are not constantly reaching through target and sprite to manipulate costumes, and sprite manages its own costume list.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -332,7 +332,7 @@ class Scratch3LooksBlocks {
         }
         if (target === this.runtime.getTargetForStage()) {
             // Target is the stage - start hats.
-            const newName = target.sprite.costumes[target.currentCostume].name;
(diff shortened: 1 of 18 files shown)


cwillisf (approved):
I think this is definitely a step in the right direction.

It does feel a little odd to me that the "client" is fetching the whole list of costumes instead of just one at a time: this means they could mess with the array if they really wanted to, but the fact that it came from an accessor will probably discourage that.

I do like the automatic resolution of name collisions :)

cwillisf:
Marking as "needs work" due to the merge conflict, but otherwise LGTM

fsih:
the getter for costumes returns a clone of costumes, so accessors now shouldn't be able to change what's in the list without going through addCostumeAt and deleteCostumeAt

PR #944, opened 2018-02-23
Duplicate costume

fsih:
Work toward https://github.com/LLK/scratch-gui/issues/923
Add functions to allow duplicating a costume or sound

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -409,6 +409,17 @@ class RenderedTarget extends Target {
     }
 
     /**
+     * Add a costume at the given index, taking care to avoid duplicate names.
+     * @param {!object} costumeObject Object representing the costume.
+     * @param {!int} index Index at which to add costume
+     */
+    addCostumeAt (costumeObject, index) {
+        const usedNames = this.sprite.costumes.map(costume => costume.name);
+        costumeObject.name = StringUtil.unusedName(costumeObject.name, usedNames);
+        this.sprite.costumes.splice(index, 0, costumeObject);
+    }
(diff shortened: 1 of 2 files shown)


src/sprites/rendered-target.js
```diff
@@ -458,6 +469,17 @@ class RenderedTarget extends Target {
         this.runtime.requestTargetsUpdate(this);
     }
 
+    /**
+     * Add a sound, taking care to avoid duplicate names.
+     * @param {!object} soundObject Object representing the sound.
+     * @param {!int} index Index at which to add costume
+     */
+    addSoundAt (soundObject, index) {
```
paulkaplan: instead of a new method that duplicates some of the code in `addSound`, could we just add an optional param to addSound that lets you specify the index, defaulting to adding the sound to the end? It would reduce the duplication and lower the surface are for bugs. (ditto for costumes methods)

paulkaplan (changes requested):
Minor refactoring requested to reduce duplication. Let me know if that makes sense?

paulkaplan (approved):

paulkaplan (approved):
LG! going to merge this because the GUI PR depends on it. But @cwillisf since you were tagged as a reviewer you might still want to look over this.

cwillisf:
Sorry I missed this one, but it looks good to me.

Running the duplication through `loadCostume` / `loadSound` isn't really the right long-term solution but it's the best we have at the moment. Once we have a better way to create new assets (whether by cloning or by creating an empty asset then filling it) we should revisit this code to use those new features.

PR #945, opened 2018-02-24
Translate extension block

picklesrus:
### Proposed Changes

Adding an extension with a translate block.

Note that this depends on https://github.com/LLK/scratch-vm/pull/934.  I have those changes in this branch so they show up in the diff. The only file actually changing is the new one: scratch3_translate/index.js which is in the most recent commit.


picklesrus:
I rebased on a more recent version of develop with #934 in it so this only has the translate extension changes in it.  Should be ready for a review!

thisandagain (approved):
Thanks @picklesrus !

PR #950, opened 2018-02-26
Fix music extension menu import 

### LINKED ISSUE #835: Drum and instrument blocks load from sb2 with numbers instead of strings
### Expected Behavior

Drum and instrument menus should have a consistent appearance.

### Actual Behavior

The music extension blocks "play drum" and "set instrument" have menus with strings containing the instrument names that looks like this ("strings only"):

screen shot 2017-11-30 at 3 02 29 pm

But when you import an SB2 with play drum and play instrument blocks, the blocks look like this ("numbers only"):

screen shot 2017-11-30 at 3 04 47 pm

In 2.0, "play drum" and "set instrument" blocks are a hybrid type (like the current 3.0 "point in direction" block) like this ("number input with menu"): 

screen shot 2017-11-30 at 4 46 18 pm

So what should we do in 3.0?

1) Change the 3.0 blocks to behave like 2.0, using a "number input with menu". This would require creating a new field type for extensions to use I think?

2) Keep the "strings only" style, and convert sb2 blocks to this type on import. Might require a hacky fix.

3) Leave as is: imported sb2 blocks have "number only" inputs, and new blocks have "string only" inputs.

I lean toward #2 but I'm not sure how to implement it. Curious what others think. @carljbowman @thisandagain @cwillisf

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/835

### Proposed Changes

In order for the menus on the music blocks to import and function correctly, make all of the following strings match: menu name (DRUM or INSTRUMENT), arg name (DRUM or INSTRUMENT), and inputOp name (music.menu.DRUM or music.menu.INSTRUMENT).  

Also, make sure the default values for these menus are set when they are imported with shadows obscured.

### Reason for Changes

If these strings do not match, the menus do not import correctly- the menu does not display the saved value. Making them match lets them import correctly.

Ideally, the menu could be named something different from the arg in the extension definition. So this may be a temporary fix. 

I believe there is an underlying bug in the system that generates the menus.  @cwillisf may be able to investigate further when we encounter this issue again as we implement more blocks.

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -464,7 +464,7 @@ class Scratch3MusicBlocks {
                     arguments: {
                         DRUM: {
                             type: ArgumentType.NUMBER,
-                            menu: 'drums',
+                            menu: 'DRUM',
                             defaultValue: 1
                         },
                         BEATS: {
@@ -506,7 +506,7 @@ class Scratch3MusicBlocks {
                     arguments: {
                         INSTRUMENT: {
                             type: ArgumentType.NUMBER,
-                            menu: 'instruments',
+                            menu: 'INSTRUMENT',
                             defaultValue: 1
                         }
                     }
@@ -540,8 +540,8 @@ class Scratch3MusicBlocks {
                 }
             ],
             menus: {
-                drums: this._buildMenu(this.DRUM_INFO),
-                instruments: this._buildMenu(this.INSTRUMENT_INFO)
+                DRUM: this._buildMenu(this.DRUM_INFO),
+                INSTRUMENT: this._buildMenu(this.INSTRUMENT_INFO)
             }
         };
     }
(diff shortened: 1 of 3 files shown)


kchadha:
@ericrosenbaum, @cwillisf, do you have a sense of what kinds of functionality would be possible by having different names for the arg vs. the menu? In the case of broadcast messages, I had separated BROADCAST_OPTION from BROADCAST_INPUT only for ease of reasoning about what's happening, but it wasn't actually necessary. When checking for the input vs. the shadow field, I could have gotten the same information by just looking at the different sections of the block xml from scratch-blocks 
(`<fields>` vs. `<inputs>`).

This question is not meant to be a review of this PR, just trying to get a sense of what we're sacrificing or not.

kchadha (approved):
LGTM!

Issue #837, opened 2017-11-30
Saved tempo in sb2 is not used


ericrosenbaum:
### Expected Behavior

When we load an sb2 file, we should load the tempo setting stored in the project. 

It appears that the Stage object in the sb2 has a tempoBPM property. On loading the sb2, we can store this value in the stage, and then the music extension loads, when can check for this property in the stage and set it in the music extension's custom state.

### Actual Behavior

The tempo stored in the project is ignored.

### Steps to Reproduce

This project should play the cowbell at 500 BPM when you press the green flag, but it plays at 60 BPM:

https://llk.github.io/scratch-gui/develop/#190025063

ericrosenbaum:
Notes on this:

There's a related problem for "set instrument", thought tempo is global and instrument is custom state per-sprite.

We will need to load the tempo and instrument from sb2 files, store them in the VM somewhere, and then load them at the time the music extension loads

We also need to consider how to save these properties in SB3.

towerofnix:
>Some options:
>(end of message)

Ah, good!

ericrosenbaum:
oops (fixed)

PR #953, opened 2018-02-27
Load tempo from sb2

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/837

### Proposed Changes

- Add a tempo property to rendered target (used only by the stage, only for the music extension)
- Load the tempo stored in an sb2, use it to set the stage's tempo
- The extension no longer has its own tempo property, and uses the stage's tempo

### Reason for Changes

This change is necessary for compatibility. Some 2.0 projects rely on a saved tempo for correct behavior.

### Test Coverage

No tests

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -44,13 +44,6 @@ class Scratch3MusicBlocks {
         this.runtime = runtime;
 
         /**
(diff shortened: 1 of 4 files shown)


kchadha (approved):
LGTM

PR #956, opened 2018-02-28
Validate sb2

kchadha:
### Proposed Changes

Uses scratch-parser to validate a given scratch 2.0 project. Throws an error if project cannot be validated, so that something else, e.g. scratch-gui, can handle the error.

### Reason for Changes

Currently, some projects fail to load, silently, and the scratch-gui gets stuck in the loading state forever.
Here is an example project with this problem: 10003280

With these changes, scratch-vm attempts to validate the project and throws errors, the failed state gets caught by the error boundary in scratch-gui.

This is a step towards detecting invalid Scratch 2.0 projects.

Eventually, scratch-parser will be able to validate any given project as one of Legacy, Scratch 2.0, or Scratch 3.0.

### Test Coverage

Existing tests pass.

#### Additional Notes

LLK/scratch-gui#1528 should be merged before this PR.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -9,6 +9,8 @@ const sb2 = require('./serialization/sb2');
 const sb3 = require('./serialization/sb3');
 const StringUtil = require('./util/string-util');
 const formatMessage = require('format-message');
+const validate = require('scratch-parser');
+
 const Variable = require('./engine/variable');
 
 const {loadCostume} = require('./import/load-costume.js');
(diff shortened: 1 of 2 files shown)


src/virtual-machine.js
```diff
@@ -226,26 +228,40 @@ class VirtualMachine extends EventEmitter {
 
         // Validate & parse
         if (typeof json !== 'string' && typeof json !== 'object') {
-            log.error('Failed to parse project. Invalid type supplied to fromJSON.');
-            return;
+            const invalidTypeError = 'Failed to parse project. Invalid type supplied to fromJSON.';
+            log.error(invalidTypeError);
```
rschamp: If you're throwing it, then it shouldn't log at this point, since the thrown error may eventually be caught and dealt with (possibly by logging the error).

src/virtual-machine.js
```diff
@@ -226,26 +228,40 @@ class VirtualMachine extends EventEmitter {
...
         } else {
-            deserializer = sb2;
+            // scratch-parser expects a json string or a buffer
+            const possibleSb2 = typeof json === 'object' ? JSON.stringify(json) : json;
+            validate(possibleSb2, (err, project) => {
+                if (err) {
+                    const errorMessage =
+                        `The given project could not be validated, parsing failed with error: ${JSON.stringify(err)}`;
+                    log.error(errorMessage);
+                    throw new Error(errorMessage);
```
rschamp: Same issue here, re: error logging

rschamp (changes requested):

kchadha:
@rschamp touché about the unnecessary logging. Removed the log messages before the error throws.

kchadha:
@rschamp @thisandagain, Looks like tests are failing because we need to release (or pre-release) a new version of scratch-parser, so that changes I made, which are currently in develop, get pulled in by scratch-vm. How should we go about doing this?

rschamp (approved):
Looks good, thanks @kchadha!

PR #961, opened 2018-03-05
Temporarily return real costumes array. GUI currently relies on this behavior somewhere

fsih:
Fixes: Adding a costume currently fails

Eventually we should fix GUI to not access the costumes array directly and revert this change.

### DIFF
--- a/src/sprites/sprite.js
+++ b/src/sprites/sprite.js
@@ -66,7 +66,7 @@ class Sprite {
      *     addCostumeAt, deleteCostumeAt, or setting costumes.
      */
     get costumes () {
-        return this.costumes_.slice(0);
+        return this.costumes_;
     }
 
     /**


kchadha:
@fsih VM relies on this behavior too. I just spent a while debugging why save-load was no longer working after rebasing on scratch-vm#develop last night, and I found that updateSvg was no longer modifying the metadata for costumes properly because of [this change](https://github.com/LLK/scratch-vm/pull/943/files#diff-8cd9925e51cd265b37b1646b1f9d163bL484). When an update to a costume is made, vm needs to be able to modify the runtime info about costumes so that they can be serialized properly on save.

kchadha:
LG as a temporary fix.

kchadha (approved):

fsih:
Holding off on merging this since it's still incompatible with the save/load change we want to test for smoketesting

fsih:
Merging since it seems the save/load change still isn't working if we revert the original PR

PR #963, opened 2018-03-05
Implement for-each block

### LINKED ISSUE #959: Implement doForLoop block
### Expected Behavior

As [per](https://scratch.mit.edu/users/thisandagain/#comments-43174070) @thisandagain, the "doForLoop" block should be implemented.

"The "doForLoop" will probably continue to be supported."

### Actual Behavior

There is no code to handle how the for loop block.

### Steps to Reproduce

Load a project that has the doForLoop block. Note that, in 2.0, doForLoop is an obsolete block; it cannot be found in the block palette. The only projects which use them are "hacked" projects (that is, their JSON has been modified to include doForLoop). However, the block does exist and function (mostly) as expected.

(I'd make an example project, but Flash isn't running well on my computer lately!)

---

[Here is the relevant code in scatch-flash.](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/interpreter/Interpreter.as#L503-L540) The main takeaway points:

* If listArg is a string, and `Number(listArg)` is not NaN, then treat listArg as if it is a number (per below). (Test case: passing "join [1] [2]", a string, to doForLoop should behave the same way as passing (10 + 2), a number.)
* If listArg is a number (or the above case), use the range from 1 to listArg (inclusive) as the list. (For example, if listArg is 5, list is [1, 2, 3, 4, 5]).
* doForLoop otherwise works exactly like a normal "for" loop: On each tick, set the value of the passed loop variable to the item of the list at the current index and increment the index by one, then run the substack.

This will probably also take implementation in the -gui and -blocks repositories.

towerofnix:
### Resolves

Resolves #959. Should not be merged before LLK/scratch-blocks#1394.

### Proposed Changes

(Edit note: the below changes are irrelevant; the code now behaves identically to Scratch 2.0, no added features that better fit an extension.)

Adds the implementation for the "for each" block. A code comment that explains how it works is included in the implementation, and is quoted below:

>The for-each block has two different but similar modes of operation, based on the value to be iterated:
>
> 1. If the value is a string, keep INTERNAL track of the index. The value assigned to the Scratch variable will be the letter at the current index, which will increase by 1 each tick. For example:
>
>    ```
>    for each (x) in [abcdef]:
>      say (x)
>    ```
>
>    ..is equivalent to:
>
>    ```
>    set (i_) to 0
>    repeat until (i_) > length of [abcdef]:
>      change (i_) by 1
>      set (x) to letter (i_) of [abcdef]
>      say (x)
>    ```
>
>    In this example, i_ is effectively invisible to the Scratch code: there is no way to view or change it. Also, changing (x) has no effect on the next iteration.
>
> 2. If the value is a number, DO NOT keep internal track of the index. When for-each is passed a number, it iterates over all the numbers from 1 to that number (inclusive). For example:
>
>    ```
>    for each (x) in [10]:
>      say (x)
>    ```
>
>    ..is equivalent to:
>
>    ```
>    set (x) to 0
>    repeat until (x) > 10:
>      change (x) by 1
>      say (x)
>    ```
>
>    The critical difference is that (x) is just a normal Scratch variable. Changing it DOES have an effect on the next iteration. This makes it possible to "skip" items, or to go back, et cetera. For example:
>
>    ```
>    for each (x) in [10]:
>      if (x) mod 4 = 0:
>        say [Mod 4 - Skip the next.]
>        change (x) by 1
>      else:
>        say x
>    ```
>
>    ..would say 1, 2, 3, "Mod 4", 6, 7, "Mod 4", 10.
>
>The mode of operation above is decided at the beginning of the for-each loop, and changing the value that is being iterated *while* it is being iterated does not change the mode. 

### Reason for Changes

For compatibility with Scratch 2.0 projects that use the hacked "for each" block. Also, as a powerful and convenient block. What it can do is described above, but here are some examples:

Basic fizz-buzz - no special usage of "for" loop, it simply iterates a range of numbers:

Fizz-buzz script

This says "1, 2, fizz, 4, buzz, fizz, 7, 8, fizz, buzz, 11, fizz, 13, 14, fizz buzz".

Changing the iterator variable while running - this skips numbers that are immediately after numbers divisible by 4:

Skip past (divisible by 4)+1

This says "1, 2, 3, mod 4, 6, 7, mod 4, 10".

Manipulating a list while it is being iterated - this is a bit more complicated! When the loop reaches items equal to "baddy", those items are removed, and the loop continues from the item that appears in baddy's place. When the loop reaches items equal to "boom", the "boom" is removed, and four items ("wow!!!", "watch out for the sneaky", "baddy", and "cool!!!") are added in its place. The loop then continues from just past those items.

baddyboom

This prints:

* 1: a
* 2: b
* 3: baddy
* 3: boom
* 7: baddy
* 7: c

And the resulting modified list looks like "a, b, wow!!!, watch out for the sneaky, baddy, cool!!!, c".

All of the above changes maintain compatibility with 2.0 projects, assuming those 2.0 projects did not modify the iterator variable, since that would not have an effect in 2.0 (since the index was entirely stored and counted internally, then copied into the iterator variable). Since the "for each" block was used so rarely in 2.0, I think it's not unreasonable to break those for the sake of making the block significantly more versatile.

### Test Coverage

Besides making the examples above, I manually tested importing and running some 2.0 projects:

* https://scratch.mit.edu/projects/21950622/ - Imported. Ran correctly ("1, 2, 3, 4, 5, 6, 7, 8, 9, 10, Look inside!").
* https://scratch.mit.edu/projects/48023824/ - Imported. Ran correctly (replacing the block with "repeat (32)" has the same effect).
* https://scratch.mit.edu/projects/92051100/ - **Does not import correctly.** This is because there's a reporter in the variable field. I think this isn't actually a problem for this PR; putting a reporter inside a "set variable" block probably has the same issue.
  ![Imports as "for each [concatenate:with:,hello ,world] in (10)"](https://user-images.githubusercontent.com/9948030/37001361-a4dee3ce-209c-11e8-8f1e-2963e44b5e3c.png)
* https://scratch.mit.edu/projects/63313634/ - Imported. Ran correctly (the sprite moves across the screen, saying the numbers 1 through 10 each time it steps). Note that this project uses nested for loops.

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -121,7 +121,8 @@ class Scratch3ControlBlocks {
         // is being iterated does not change the mode.
 
         if (typeof util.stackFrame.iterationMode === 'undefined') {
-            let value = args.VALUE, mode;
+            const value = args.VALUE;
+            let mode;
             if (typeof value === 'string' && !isNaN(Number(value))) {
                 mode = 'number';
             } else if (typeof value === 'number') {
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -66,92 +66,14 @@ class Scratch3ControlBlocks {
         const variable = util.target.lookupOrCreateVariable(
             args.VARIABLE.id, args.VARIABLE.name);
 
-        // The for-each block has two different but similar modes of operation,
-        // based on the value to be iterated:
-        //
-        // 1. If the value is a string, keep INTERNAL track of the index.
-        //    The value assigned to the Scratch variable will be the letter at
-        //    the current index, which will increase by 1 each tick. For example:
-        //
-        //    for each (x) in [abcdef]:
-        //      say (x)
-        //
-        //    ..is equivalent to:
-        //
-        //    set (i_) to 0
-        //    repeat until (i_) > length of [abcdef]:
-        //      change (i_) by 1
-        //      set (x) to letter (i_) of [abcdef]
-        //      say (x)
-        //
-        //    In this example, i_ is effectively invisible to the Scratch code:
-        //    There is no way to view or change it. Also, changing (x) has no
-        //    effect on the next iteration.
-        //
-        // 2. If the value is a number, DO NOT keep internal track of the index.
-        //    When for-each is passed a number, it iterates over all the numbers
-        //    from 1 to that number (inclusive). For example:
-        //
-        //    for each (x) in [10]:
-        //      say (x)
-        //
-        //    ..is equivalent to:
-        //
-        //    set (x) to 0
-        //    repeat until (x) > 10:
-        //      change (x) by 1
-        //      say (x)
-        //
-        //    The critical difference is that (x) is just a normal Scratch
-        //    variable. Changing it DOES have an effect on the next iteration.
-        //    This makes it possible to "skip" items, or to go back, et cetera.
-        //    For example:
-        //
-        //    for each (x) in [10]:
-        //      if (x) mod 4 = 0:
-        //        say [Mod 4 - Skip the next.]
-        //        change (x) by 1
-        //      else:
-        //        say x
-        //
-        //    ..would say 1, 2, 3, "Mod 4", 6, 7, "Mod 4", 10.
-        //
-        // The mode of operation above is decided at the beginning of the for-
-        // each loop, and changing the value that is being iterated *while* it
-        // is being iterated does not change the mode.
-
-        if (typeof util.stackFrame.iterationMode === 'undefined') {
-            const value = args.VALUE;
-            let mode;
-            if (typeof value === 'string' && !isNaN(Number(value))) {
-                mode = 'number';
-            } else if (typeof value === 'number') {
-                mode = 'number';
-            } else {
-                mode = 'string';
-            }
-
-            if (mode === 'number') {
-                variable.value = 0;
-            } else {
-                util.stackFrame.stringIndex = 0;
-            }
-
-            util.stackFrame.iterationMode = mode;
+        if (typeof util.stackFrame.index === 'undefined') {
+            util.stackFrame.index = 0;
         }
 
-        if (util.stackFrame.iterationMode === 'string') {
-            if (util.stackFrame.stringIndex < args.VALUE.length) {
-                variable.value = args.VALUE[util.stackFrame.stringIndex++];
-                util.startBranch(1, true);
-            }
-        }
-
-        if (util.stackFrame.iterationMode === 'number') {
-            if (variable.value < Number(args.VALUE)) {
-                variable.value++;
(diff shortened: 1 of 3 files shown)


towerofnix:
Note - Travis CI failed here:

```
./test/integration/hat-threads-run-every-frame.js ... 53/54
  edge activated hat thread runs once every frame
  not ok expect truthy value
    at:
      line: 50
      column: 15
      file: test/integration/hat-threads-run-every-frame.js
      function: vm.loadProject.then
    stack: |
      vm.loadProject.then (test/integration/hat-threads-run-every-frame.js:50:15)
    source: |
      t.assert(vm.runtime.threads[0].status === Thread.STATUS_DONE);
```

This test succeeded in a different attempt on Travis, and the error didn't happen locally, so I doubt this is related to this PR.

towerofnix:
Okay, SB2 importing and manual testing complete. All projects (besides one that fails to import for unrelated reasons) load and run well. It looks like that test case isn't failing anymore now, as well.

kchadha:
@towerofnix This is awesome work! Thank you for the thorough documentation, it was very helpful in understanding the behavior of the block and the two modes that have been implemented.

Looking through the proposed changes, I am noticing some deviations from the Scratch 2.0 behavior (they are documented below).

Since this block is being added to Scratch 3.0 specifically for maintaining compatibility with older projects, I think the behavior of the 3.0 block should have a one-to-one correspondence with the Scratch 2.0 block. Thus, I don't think we should change any behavior from the 2.0 block or add any features that the 2.0 block does not have.

Deviations I noticed: 
In Scratch 2.0, it is not possible to ‘skip’ items like you mention in the documentation for the forEach function. This makes the following program behave differently in Scratch 2.0 vs. Scratch 3.0:

image

Another difference I saw was in the behavior of this project:

image

In Scratch 2.0, running this project results in the scratch cat saying 1, then 2, then 3, 3 times. In Scratch 3.0, running this project results in the scratch cat cycling from 1-3 only once.

Finally, as far as I can tell, I don't think the scratch 2.0 for each block supports iterating over letters of strings like you've described in your documentation.

In order to get this merged into develop, the 3.0 behavior should be as close to the 3.0 behavior as possible.

kchadha (changes requested):
Changes requested in comment.

towerofnix:
Alright, I'll change those - admittedly those changes would far better fit an extension :)

kchadha:
@towerofnix I think these changes are no longer working against develop, can you investigate what's going on? It also seems that unit tests are failing.

towerofnix:
@kchadha Whoops, I forgot to remove the tests for the extra behavior. Thanks for pinging me on this; tests should be working now.

kchadha (approved):

kchadha:
@towerofnix, thanks for fixing the tests! Everything looks good.

PR #964, opened 2018-03-05
Preliminary Save and Load Work

kchadha:
### Proposed Changes

Preliminary work for saving and loading 3.0 projects. This includes:
- Serialization and deserialization of costumes and sounds on save and load respectively
- Updating asset metadata properly when making edits in sound and costume editors
- Making heavy use of storage caching (improvements to be made to this, see 'Changes yet to come' section)

### Reason for Changes

This is the first cut at a working implementation of save and load. Improvements need to be made (see 'Changes yet to come' section below) to optimize code to reduce duplication of metadata stored in the VM runtime, code duplication, and unnecessary calls to scratch-storage.

### Test Coverage

Existing tests pass. Need to make changes to how errors are handled (especially in serialize-assets and deserialize-assets). Need to add unit and integration tests for save/load functionality.

### Changes yet to come
In the future, this implementation should be improved by:
- removing unnecessary storage caching calls, e.g.
    - Implementing an indication of a dirty asset and only encoding assets upon 'save' instead of each time a sound or costume is updated.
- removing duplicate information from the sound/costume metadata stored in the runtime ('assetId' vs. 'md5')
- clean up top-level loadProject/loadProjectLocal api
- #194 and validating finalized spec
- see similar section in LLK/scratch-gui#1564

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -513,6 +513,9 @@ class VirtualMachine extends EventEmitter {
             storage.DataFormat.SVG,
             (new TextEncoder()).encode(svg)
         );
+        // If we're in here, we've edited an svg in the vector editor,
+        // so the dataFormat should be 'svg'
+        costume.dataFormat = storage.DataFormat.SVG;
         this.emitTargetsUpdate();
     }
 
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -15,6 +15,7 @@ const Variable = require('./engine/variable');
 
 const {loadCostume} = require('./import/load-costume.js');
 const {loadSound} = require('./import/load-sound.js');
+const {serializeSounds, serializeCostumes} = require('./serialization/serialize-assets');
 
 const RESERVED_NAMES = ['_mouse_', '_stage_', '_edge_', '_myself_', '_random_'];
 
@@ -206,7 +207,14 @@ class VirtualMachine extends EventEmitter {
      */
     saveProjectSb3 () {
         // @todo: Handle other formats, e.g., Scratch 1.4, Scratch 2.0.
-        return this.toJSON();
+        const soundDescs = serializeSounds(this.runtime);
+        const costumeDescs = serializeCostumes(this.runtime);
+
+        return {
+            projectJson: this.toJSON(),
+            sounds: soundDescs,
+            costumes: costumeDescs
+        };
     }
 
     /**
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -186,6 +186,27 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * Load a project from a Scratch 3.0 sb3 file containing a project json
+     * and all of the sound and costume files.
+     * @param {JSZip} sb3File The sb3 file representing the project to load.
+     * @return {!Promise} Promise that resolves after targets are installed.
+     */
+    loadProjectLocal (sb3File) {
+        // TODO need to handle sb2 files as well, and will possibly merge w/
+        // above function
+        return sb3File.file('project.json').async('string')
+            .then(json => {
+                // TODO look at promise documentation to do this on success,
+                // but something else on error
+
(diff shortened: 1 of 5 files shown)


src/virtual-machine.js
```diff
@@ -184,6 +186,27 @@ class VirtualMachine extends EventEmitter {
...
+     * @param {JSZip} sb3File The sb3 file representing the project to load.
+     * @return {!Promise} Promise that resolves after targets are installed.
+     */
+    loadProjectLocal (sb3File) {
+        // TODO need to handle sb2 files as well, and will possibly merge w/
+        // above function
+        return sb3File.file('project.json').async('string')
+            .then(json => {
+                // TODO look at promise documentation to do this on success,
+                // but something else on error
```
cwillisf: There are two main ways to handle errors with promises:
1. If it's the sort of error that you'd throw an exception for, you can call `reject` instead of `resolve`. Handling a `reject` call happens in the second arg for `then`. Note that JS debuggers tend to treat calling `reject` the same way they treat throwing an exception, so I wouldn't choose this option unless the bad thing is worthy of exception-like behavior. Here's a condensed version:
```js
new Promise(
    // There's no spec or anything that says reject must be called with an Error object, but I find it convenient.
    (resolve, reject) => { if (bad) reject(new Error("It's bad.")); else resolve("It's good."); }
  ).then(
    result => { console.log("The good result is:", result); }, // <- note the comma here
    e => { console.log("Something bad happened:", e); }
  );
```

2. Encode the error into the `resolve` result -- I imagine you already thought of this option. :)
A compliant promise gives you only one result, but that one result can be whatever works for you. For example, it could be `null` in the case of an error, or it could be a Node.js-style `err, result` pair if you stuff those into an array or an object. Some promise libraries support a [`spread`](http://bluebirdjs.com/docs/api/spread.html) call that makes this easier, but that's not (currently) standard.

In the storage engine, I originally went with option 1 for things like HTTP 404. I later switched to option 2 and just return `null` in the case of a network error, since IMO it doesn't make sense to throw an exception as part of the "try here, then try here" pattern that the storage engine uses intentionally. There may be a few remnants of the option 1 plan in the storage engine...

src/virtual-machine.js
```diff
@@ -205,8 +228,25 @@ class VirtualMachine extends EventEmitter {
...
-        // @todo: Handle other formats, e.g., Scratch 1.4, Scratch 2.0.
-        return this.toJSON();
+        const soundDescs = serializeSounds(this.runtime);
+        const costumeDescs = serializeCostumes(this.runtime);
+        const projectJson = this.toJSON();
+
+        const zip = new JSZip();
+
+        // Put everything in a zip file
+        // TODO compression?
```
cwillisf, line 239: Does JSZip default to no compression? I would expect this to default to medium-strength compression settings or something like that...

cwillisf (commented):
The code itself looks fine, but I'd like to discuss the architecture a bit further before merging this.

I feel like some of this code belongs in `scratch-storage`. In particular, I feel like `scratch-storage` should have the `jszip` dependency and should be responsible for the serialize and deserialize functions you have here. Do you think that would make sense?

If that's not a good idea, or if your answer is "yeah that's the plan, but this is a step along the way" then I think I'm fine with merging this as-is.

src/virtual-machine.js
kchadha: It defaults to no compresssion, but it's easy to add an argument, and I think you can provide a number (from 1-9) representing a speed/space trade-off:

https://stuk.github.io/jszip/documentation/api_jszip/generate_async.html#compression-and-compressionoptions-options

src/virtual-machine.js
kchadha: Yeah, Ray and I talked about this a little bit too (I think last week). Mostly that was a TODO comment to remind myself to take care of error handling later as the workflow gets more fleshed out/cleaned up.

cwillisf (approved):

PR #968, opened 2018-03-07
ensure this.workerClass exists before instanceof check

gnarf:
This avoids an error I was running into in a node environment:

```
TypeError: Right-hand side of 'instanceof' is not an object
```


thisandagain (approved):
LGTM

PR #969, opened 2018-03-08
Ensure music extension _decodeSound is promise

gnarf:
Currently if there is no audio engine detected this `_decodeSound` method returns undefined,
which in turn causes the line above to be a runtime error while trying to call `.then` on it.


thisandagain (approved):

thisandagain:
Thanks @gnarf! 👍

PR #979, opened 2018-03-15
SB3 Serialization & Load Project

kchadha:
a.k.a Save/Load pt.2
### Resolves
Towards resolving #194.

### Proposed Changes
- SB3 Serialization cleanup (be explicit about what we're serializing, attempting to serialize only the minimum information necessary to load a project)
- Leverage changes in LLK/scratch-parser#21 to validate sb3 files
- Load .sb, .sb2, and .sb3 files (.sb files are still not supported, but they are at least passed through to scratch-parser)

#### Additional changes to come:
- More serialization work (serialize monitors, extension custom state)
- Support for sb1 import

#### Eventually (these are changes we want for the ideal world of save/load, but should not block release of this feature):
- Reducing unnecessary processes surrounding saving projects (only encode sounds/costumes when necessary instead of every time an edit is made -- even more eventually, move this to scratch-storage)
- Cleanup of unnecessary/duplicate information tracked in VM (e.g. assetId vs. md5 for sounds and costumes)
- Cleanup of relationship between scratch-vm, scratch-storage, and scratch-parser (this PR makes some progress towards this goal)

### Reason for Changes

Continuation Save/Load work.

### Test Coverage

Existing tests pass. Additional tests need to be added for sb3 round-trip testing.
(Travis build will fail until LLK/scratch-parser#21 gets merged)

### Additional Notes
This PR depends on LLK/scratch-parser#21, and is a dependency of LLK/scratch-gui#1621

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -29,12 +29,8 @@ const serializeBlock = function (block) {
     const obj = Object.create(null);
     obj.id = block.id;
     obj.opcode = block.opcode;
-    if (block.next) {
-        obj.next = block.next;
-    }
-    if (block.parent) {
-        obj.parent = block.parent;
-    }
+    obj.next = block.next;
+    obj.parent = block.parent;
     obj.inputs = block.inputs;
     obj.fields = block.fields;
     obj.topLevel = block.topLevel;
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -33,7 +33,7 @@ const serializeBlock = function (block) {
     obj.parent = block.parent;
     obj.inputs = block.inputs;
     obj.fields = block.fields;
-    obj.topLevel = block.topLevel;
+    obj.topLevel = block.topLevel ? block.topLevel : false;
     obj.shadow = block.shadow;
     if (block.topLevel) {
         if (block.x) {
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -85,9 +85,9 @@ const serializeSound = function (sound) {
     return obj;
 };
 
-const serializeTarget = function (target/* , runtime*/) {
+const serializeTarget = function (target) {
     const obj = Object.create(null);
-    obj.isStage = target.isStage; // target.id === runtime.getTargetForStage().id;
+    obj.isStage = target.isStage;
     obj.name = target.name;
     obj.variables = target.variables; // This means that uids for variables will persist across saves/loads
     obj.blocks = serializeBlocks(target.blocks);
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -62,6 +62,12 @@ const serializeCostume = function (costume) {
     obj.assetId = costume.assetId;
     obj.name = costume.name;
     obj.bitmapResolution = costume.bitmapResolution;
+    // serialize this property with the name 'md5ext' because that's
+    // what it's actually referring to. TODO runtime objects need to be
+    // updated to actually refer to this as 'md5ext' instead of 'md5'
+    // but that change should be made carefully since it is very
+    // pervasive
+    obj.md5ext = costume.md5;
     obj.dataFormat = costume.dataFormat;
     obj.rotationCenterX = costume.rotationCenterX;
     obj.rotationCenterY = costume.rotationCenterY;
@@ -76,8 +82,12 @@ const serializeSound = function (sound) {
     obj.format = sound.format;
     obj.rate = sound.rate;
     obj.sampleCount = sound.sampleCount;
-    // TODO eventually want to get rid of this
-    obj.md5 = sound.md5;
+    // serialize this property with the name 'md5ext' because that's
+    // what it's actually referring to. TODO runtime objects need to be
+    // updated to actually refer to this as 'md5ext' instead of 'md5'
+    // but that change should be made carefully since it is very
+    // pervasive
+    obj.md5ext = sound.md5;
     // TODO do we need this soundID
     // (not to be confused with soundId which is a uid for sounds)
     // obj.soundID = sound.soundID;
@@ -176,6 +186,7 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
     const costumePromises = (object.costumes || []).map(costumeSource => {
         // @todo: Make sure all the relevant metadata is being pulled out.
         const costume = {
+            assetId: costumeSource.assetId,
             skinId: null,
             name: costumeSource.name,
             bitmapResolution: costumeSource.bitmapResolution,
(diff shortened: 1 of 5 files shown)


src/serialization/sb3.js
```diff
@@ -25,6 +25,87 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
  * @property {Map.<string, string>} extensionURLs - map of ID => URL from project metadata. May not match extensionIDs.
  */
 
+const serializeBlock = function (block) {
+    const obj = Object.create(null);
```
cwillisf, line 29: Interesting - I hadn't seen (noticed?) this before. Based on what I read [here](https://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull) maybe we should use this more often...

src/serialization/sb3.js
```diff
@@ -25,6 +25,87 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
+const serializeTarget = function (target) {
+    const obj = Object.create(null);
+    obj.isStage = target.isStage;
+    obj.name = target.name;
+    obj.variables = target.variables; // This means that uids for variables will persist across saves/loads
+    obj.blocks = serializeBlocks(target.blocks);
+    obj.currentCostume = target.currentCostume;
+    obj.costumes = target.costumes.map(serializeCostume);
+    obj.sounds = target.sounds.map(serializeSound);
+    if (!obj.isStage) {
```
cwillisf, line 104: 👍

src/virtual-machine.js
```diff
@@ -177,34 +177,38 @@ class VirtualMachine extends EventEmitter {
...
+        const validationPromise = new Promise((resolve, reject) => {
+            validate(input, (error, res) => {
+                if (error) {
+                    reject(error);
+                }
+                if (res) {
+                    resolve(res);
+                }
+            });
+        });
```
cwillisf, line 205: If we end up doing this Node-style callback <=> Promise conversion more often it might make sense to make a utility function or two for it. I think this is the first one I've seen in VM, though, so I think this is fine for now.

cwillisf (approved):
LGTM, so long as you resolve the test failures of course

cwillisf:
(marking as needs work for test failure)

src/serialization/sb3.js
```diff
@@ -25,6 +25,87 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
+    if (block.mutation) {
+        obj.mutation = block.mutation;
+    }
+    return obj;
+};
+
+const serializeBlocks = function (blocks) {
+    const obj = Object.create(null);
+    for (const blockID in blocks) {
+        obj[blockID] = serializeBlock(blocks[blockID]);
```
rschamp: In `for...in` loops, we should use [hasOwnProperty](https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_sound.js#L236) to make sure we're only looping on instance-level properties and not those added to the prototype (https://stackoverflow.com/questions/12735778/for-in-and-hasownproperty).

src/serialization/sb3.js
```diff
@@ -25,6 +25,87 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
+    const obj = Object.create(null);
+    obj.assetId = sound.assetId;
+    obj.name = sound.name;
+    obj.dataFormat = sound.dataFormat;
+    obj.format = sound.format;
+    obj.rate = sound.rate;
+    obj.sampleCount = sound.sampleCount;
+    // TODO eventually want to get rid of this
+    obj.md5 = sound.md5;
+    // TODO do we need this soundID
```
rschamp: I don't think so, since it's runtime-specific.

src/serialization/sb3.js
```diff
@@ -33,7 +114,12 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
 const serialize = function (runtime) {
     // Fetch targets
     const obj = Object.create(null);
-    obj.targets = runtime.targets.filter(target => target.isOriginal);
+    const flattenedOriginalTargets = JSON.parse(JSON.stringify(
+        runtime.targets.filter(target => target.isOriginal)));
+    obj.targets = flattenedOriginalTargets.map(t => serializeTarget(t, runtime));
+    // runtime.targets.filter(target => target.isOriginal);
```
rschamp: Cruft?

src/serialization/sb3.js
```diff
@@ -114,7 +200,7 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
             // fileUrl: soundSource.fileUrl,
```
rschamp: Cruft?

src/serialization/sb3.js
```diff
@@ -114,7 +200,7 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
...
             rate: soundSource.rate,
             sampleCount: soundSource.sampleCount,
-            soundID: soundSource.soundID,
+            // soundID: soundSource.soundID,
```
rschamp: Cruft?

src/virtual-machine.js
```diff
@@ -177,34 +177,38 @@ class VirtualMachine extends EventEmitter {
...
-                sb3File.file('project.json').async('string')
-                    .then(json => {
-                        // TODO error handling for unpacking zip/not finding project.json
-                        json = JSON.parse(json); // TODO catch errors here (validation)
-                        return sb3.deserialize(json, this.runtime, sb3File)
-                            .then(({targets, extensions}) =>
-                                this.installTargets(targets, extensions, true));
-                    });
+        // Clear the current runtime
+        this.clear();
```
rschamp: Maybe this should only happen once we're sure the input is going to work?

src/virtual-machine.js
```diff
@@ -258,47 +264,36 @@ class VirtualMachine extends EventEmitter {
         return JSON.stringify(sb3.serialize(this.runtime));
     }
 
+    // TODO do we still need this function? Keeping it here so as not to introduce
+    // a breaking change.
```
rschamp, line 271: Maybe at this point we should add a `log.warning()` when it's called?

test/unit/serialization_sb3.js
```diff
@@ -5,11 +5,13 @@ const demoSb3 = require('../fixtures/demo.json');
 
 test('serialize', t => {
     const vm = new VirtualMachine();
-    vm.fromJSON(JSON.stringify(demoSb3));
-    const result = sb3.serialize(vm.runtime);
-    // @todo Analyze
-    t.type(JSON.stringify(result), 'string');
-    t.end();
+    vm.fromJSON(JSON.stringify(demoSb3))
```
rschamp: This test should move to using `loadProject`.

rschamp (changes requested):
This looks great overall, just a few nits and questions.

src/serialization/sb3.js
towerofnix: Would it be better to just use `for...of` and Object.keys?

```js
for (const blockID of Object.keys(blocks)) {
  ...
}
```

kchadha:
@rschamp @cwillisf I believe I addressed all the PR comments asking for or suggesting changes. Please take another look and let me know what you think. These changes also involve the 'md5' to 'md5ext' name change during serialization (and then translating back the other way during deserialization). This change was discussed in the LLK/scratch-parser#21.

P.S. Tests will continue to fail until LLK/scratch-parser#21 is merged, and so this branch should remain un-merged until that happens and these tests finally pass.

rschamp (approved):
LGTM!

src/virtual-machine.js
```diff
@@ -182,14 +182,16 @@ class VirtualMachine extends EventEmitter {
      * @return {!Promise} Promise that resolves after targets are installed.
      */
     loadProject (input) {
-        if (typeof input === 'object' && !(input instanceof ArrayBuffer)) {
+        if (typeof input === 'object' && !ArrayBuffer.isView(input)) {
```
cwillisf: Oh cool - I didn't know about `isView`!

cwillisf (approved):

src/virtual-machine.js
kchadha: @cwillisf Turns out ArrayBuffers are not views of themselves. Just added the instanceof check back in, but the isView check should work for Uint8Array, Float32Array, and all those good things.

kchadha:
Tests pass now that scratch-parser has a new release on npm. Merging this PR.

Issue #919, opened 2018-01-25
Should clones be able to start their own copy of a sound?


ericrosenbaum:
This [PR](https://github.com/LLK/scratch-vm/pull/891), “Audio effects state for clones,” introduced a new behavior that need discussion and possible changes. 

It is related to the behavior of a clone starting a sound when another clone or the parent sprite has already started that same sound. In 2.0, the sound restarts. After the above PR, a new copy of the sound starts playing (with the clone's effects settings applied). 

Example:
If you have a sprite with the sound "happy song", and you run "forever [start sound happy song]", it will restart the sound every frame, creating a stuttering sound. This is true in both 2.0 and 3.0. However, in 3.0, if you create a bunch of clones, and run "when I start as a clone [start happy song]", the clones will all separately start their own copy of the song. 

Upside: you can make music using clones that play the same sound at overlapping times with different audio effect settings (pitch, pan, volume). 

Downside: you get a potentially very loud sound instead of a stuttering sound if you make a lot of clones that play a sound simultaneously.

We should consider how often this downside case would occur in existing projects, if possible.

CatsAreFluffy:
I'd say that the old behavior is less intuitive, since why would all clones share one speaker?

thisandagain:
Discussed! @ericrosenbaum will follow-up with changes. :-)

Issue #949, opened 2018-02-26
Deleting a clone stops its sounds


ericrosenbaum:
### Expected Behavior

In Scratch 2.0, if you start a sound in a clone, then delete the clone, and the sound will continue to play.

### Actual Behavior

Currently in 3.0, because clones have their own audio players which get disposed on clone delete, the sound stops.

### Steps to Reproduce

The meow should play when you create a clone:

![image](https://user-images.githubusercontent.com/567844/36678658-52816ccc-1adf-11e8-9852-dadbd83ba065.png)

### Operating System and Browser

Mac OS 10.13.2 Chrome 64

ericrosenbaum:
After discussion with @thisandagain we realized that getting this to work as in 2.0 (i.e. sounds played by clones continue after the clone is deleted) will require a re-factor in scratch-audio:

- We'd merge AudioPlayer and SoundPlayer into a single SoundPlayer class which manages its own audio effects chain (volume and pan effect nodes). This may mean we're creating more audio nodes than before, but probably not (you're more likely to have a lot of clones at once than a lot of different sounds playing at once).
- Every sound would have a single SoundPlayer (created on demand, and deleted when the sound ends). 
- The sprite would manage an array of SoundPlayers, which would be accessed by a clone when it wants to play a sound (much like the current activeSoundPlayers array).
- When a clone starts a sound, or updates an effect value (volume, pan, or pitch), it would request its parent sprite to do so.
- When the parent sprite receives a request to update an effect value, it needs to apply this update to all sounds currently being played by that clone. This means we'd need some kind of ID system (do clones have IDs already?)

TheLogFather:
Just a quick note that the above behaviour in Scratch 2 is different if the sound is played "until done".

In that case, deleting the clone *does* stop the sound. (That's because the script is still running, and stopping the script, when the clone gets deleted, also stops the sound currently being played by that script.)

TheLogFather:
Actually, it's worth taking a look at this old project of mine:
https://scratch.mit.edu/projects/24414513
It uses clones to allow independent control (volume, stop/start) over sounds playing simultaneously.

Now compare with Scratch 3 preview:
https://preview.scratch.mit.edu/#24414513

(I think it'd be worth adding the above project for compatibility-testing – and not just for the sounds...)

ericrosenbaum:
Fixed!

TheLogFather:
[...]

PR #991, opened 2018-03-19
Clones share their parent’s activeSoundPlayers object

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/919

### Proposed Changes

When a clone is created, it gets a reference to the parent sprite's activeSoundPlayers object, which keeps track of which sounds are currently playing. 

### Reason for Changes

This change allows clones to have their own audio effects state, but brings back the 2.0 behavior where a sprite and its clones cannot play multiple copies of the same sound at the same time. Instead, if one of them starts playing a sound, and then another starts the same sound, it will stop and restart.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -146,6 +146,13 @@ class RenderedTarget extends Target {
         this.audioPlayer = null;
         if (this.runtime && this.runtime.audioEngine) {
             this.audioPlayer = this.runtime.audioEngine.createPlayer();
+            // If this is a clone, it gets a reference to its parent's activeSoundPlayers object.
+            if (!this.isOriginal) {
+                const parent = this.sprite.clones[0];
+                if (parent && parent.audioPlayer) {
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
Looks good to me!

I wonder if the `activeSoundPlayers` should ideally live somewhere sprite-specific, not owned by the audio player or the clone. Just something to think about for a future, low-priority refactor... I think this is fine for now, and possibly forever :)

griffpatch:
Does that mean playing the same sound with a different pitch effect would
stop one of the sounds playing of you are using clones?

CatsAreFluffy:
:(

ericrosenbaum:
Good question, @griffpatch. I wish we could have that ability! The workaround would be to add several copies of the sound to the sprite, and then code it so that each clone plays its own copy. The whole thing may need a refactor anyway (see https://github.com/LLK/scratch-vm/issues/949)

Issue #941, opened 2018-02-21
Features needed before all blocks can become extensions


cwillisf:
The extension system is missing some features that will be needed before we can convert all blocks packages into extensions.

### Blocking
- [x] Synchronous dynamic menus
  - Menu items defined by a function which must return items immediately (as opposed to calling a callback or returning a promise)
  - Related: LLK/scratch-vm#934
- [ ] Menus referring to sprites, costumes, sounds, etc. should handle object renaming
- [ ] Monitors
- [x] C Blocks
  - Must support loop icon
- [x] E Blocks
  - Must support text on the middle line
- [x] Non-droppable menus
- [ ] Event-triggered hats

### Not Blocking
- [ ] Blocks that can dynamically have/not have a "next statement" connection
  - See "Stop All" block

PR #994, opened 2018-03-20
Looks extension support

cwillisf:
### Resolves

This change represents progress toward #941

### Proposed Changes

The main functional change here is that a target ID is now passed to the function implementing a dynamic menu in an extension. This allows the menu function to collect information about its "context" target. For example, the costume menu on the `set costume to {costume1}` block can collect the list of costumes available on the target.

Currently the editing target is supplied, which seems sufficient: it shouldn't be possible to open a drop-down menu without a valid editing target, and it seems that the menu should always run in the context of the current editing target rather than any other target. **Reviewer:** if you can think of a situation where this isn't appropriate, please let me know!

Other changes:
- Add JSDoc types for extension metadata
- Add JSDoc comments for `ArgumentType` and `BlockType` enums
- Provide a nicer error message when the extension system can't find a block implementation function

### Reason for Changes

This is a step toward implementing all blocks as "extensions" -- or at least centralizing block definitions and eliminating the toolbox XML file in scratch-blocks.

### Test Coverage

None yet.

### DIFF
--- a/src/extension-support/argument-type.js
+++ b/src/extension-support/argument-type.js
@@ -1,8 +1,31 @@
+/**
+ * Block argument types
+ * @enum {string}
+ */
 const ArgumentType = {
+    /**
+     * Numeric value with angle picker
+     */
     ANGLE: 'angle',
+
+    /**
+     * Boolean value with hexagonal placeholder
+     */
     BOOLEAN: 'Boolean',
+
+    /**
+     * Numeric value with color picker
+     */
     COLOR: 'color',
+
+    /**
+     * Numeric value with text field
+     */
     NUMBER: 'number',
+
+    /**
+     * String value with text field
+     */
     STRING: 'string'
 };
 
--- a/src/extension-support/block-type.js
+++ b/src/extension-support/block-type.js
@@ -1,8 +1,31 @@
+/**
+ * Types of block
+ * @enum {string}
+ */
 const BlockType = {
+    /**
+     * Boolean reporter with hexagonal shape
+     */
     BOOLEAN: 'Boolean',
+
+    /**
+     * Command block
+     */
     COMMAND: 'command',
+
+    /**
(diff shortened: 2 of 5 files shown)


src/extension-support/argument-type.js
```diff
@@ -1,8 +1,31 @@
...
 const ArgumentType = {
+    /**
+     * Numeric value with angle picker
+     */
     ANGLE: 'angle',
+
+    /**
+     * Boolean value with hexagonal placeholder
+     */
     BOOLEAN: 'Boolean',
```
rschamp, line 14: Just noticed this... why is this the only one capitalized? Is it referring to the actual JS `Boolean`?

rschamp (approved):
LGTM!

I think eventually it will be good to try to get away from using `editingTarget` so much and just operate on explicit targets passed... but for now that's what we're doing so we should keep using it so we don't have to refactor anything.

src/extension-support/argument-type.js
cwillisf: Heh - I was wondering if anyone would balk at that. I wrote all the words the way they'd appear in the middle of a sentence, where Boolean should get a capital because it's based on the name of George Boole. My thinking is that either these value strings won't ever matter as long as they're unique, or they could be used in a message like `Invalid value for ${arg.type} argument`, where "Boolean" would be the [technically-correct](https://imgur.com/gallery/PpV0evc) text to insert.

cwillisf:
Oh, and regarding `editingTarget`: I think that's actually the correct thing to pass in here, though it should probably come from the GUI rather than the VM. I suppose, ideally speaking, menu functions would be called with the editing target's ID when the user clicks a menu in the block palette, and when the user clicks a menu in the workspace the menu function should receive the ID of the target which owns the current workspace. Right now it's not possible to view a workspace that isn't the editing target, but if it ever becomes possible then we'll need to do something like this.

PR #996, opened 2018-03-23
Position the speech bubble more tightly.

### LINKED ISSUE #1568: Say bubble appears far from sprites with empty space in them
If a sprite has space in one of the upper corners but not the other (i.e. Max has space in the upper left side of the image) and says something while on the right side of the screen the speech bubble should appear close to the sprite.

Instead the sprite is treated like a square and the speech bubble appears at the corner of the square, even if there isn't anything drawn in that region.  This makes the bubble seem detached from the sprite.

To reproduce:
*Paint a new sprite.  
*Use the brush to make a diagonal line, either from the top left of the paint area to the bottom right, or from the top right to the bottom left. 
*Add a Say Hello block to the sprite and click it.
*Move the sprite to the edge of the stage so the speech bubble moves to the side that is mostly empty space.
*note the speech bubble is far from anything drawn on the sprite.

In Scratch 2 it moves the speech bubble in so it is closer to the sprite at the top:

### LINKED ISSUE #1591: Say blocks appear further a part so the conversation between sprites looks weird
### Expected Behavior

Say blocks for each sprite should appear closer to one another
https://scratch.mit.edu/projects/56987002/#editor

### Actual Behavior
Say blocks appear further a part so the conversation between sprites looks weird
say_blocks_far_apart

### Steps to Reproduce

1. Create a project between two sprites that talk in 3.0
2. Compare that project to something similar in 2.0
3. Notice that the say blocks in 3.0 are much further a part.
4. Note: Importing an existing 2.0 project into 3.0 does not show the issue. Only using new sprites in 3.0 shows the issue.

Chrome 64

paulkaplan:
### Resolves

Fixes LLK/scratch-gui#1591 
Fixes LLK/scratch-gui#1568

### Proposed Changes

Use the new getBoundsForBubble API to position the bubble more tightly.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -119,7 +119,7 @@ class Scratch3LooksBlocks {
     _positionBubble (target) {
         const bubbleState = this._getBubbleState(target);
         const [bubbleWidth, bubbleHeight] = this.runtime.renderer.getSkinSize(bubbleState.drawableId);
-        const targetBounds = target.getBounds();
+        const targetBounds = target.getBoundsForBubble();
         const stageBounds = this.runtime.getTargetForStage().getBounds();
         if (bubbleState.onSpriteRight && bubbleWidth + targetBounds.right > stageBounds.right &&
             (targetBounds.left - bubbleWidth > stageBounds.left)) { // Only flip if it would fit
@@ -137,7 +137,7 @@ class Scratch3LooksBlocks {
                     ) : (
                         Math.max(stageBounds.left, targetBounds.left - bubbleWidth)
                     ),
-                    Math.min(stageBounds.top, targetBounds.top + bubbleHeight)
+                    Math.min(stageBounds.top, targetBounds.bottom + bubbleHeight)
                 ]
             });
             this.runtime.requestRedraw();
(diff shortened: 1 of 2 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -119,7 +119,7 @@ class Scratch3LooksBlocks {
     _positionBubble (target) {
         const bubbleState = this._getBubbleState(target);
         const [bubbleWidth, bubbleHeight] = this.runtime.renderer.getSkinSize(bubbleState.drawableId);
-        const targetBounds = target.getBounds();
+        const targetBounds = target.getBoundsForBubble();
```
cwillisf, line 122: I was going to suggest adding a `null` check in this call, since `target.getBoundsForBubble()` can return `null`, but then I noticed two things:
- `target.getBounds()` could return `null` in the same situation (no renderer) so this change isn't making things any worse, but
- the line above this calls a method on `this.runtime.renderer` without checking first, which means that the check in `target.getBoundsForBubble()` is currently unnecessary.

I guess I would low-key advocate for removing the check in `getBoundsForBubble()` but I'd also be OK with leaving this as-is. At some point I'd like to make a "discipline" pass where we decide which functions can or can't assume the presence of a renderer, audio engine, storage module, etc., if for no other reason than developer ergonomics.

cwillisf (approved):
LGTM, unless you want to do something about the `null` check thing I commented on...

PR #1000, opened 2018-03-25
Deserialize sb2 assets while loading local .sb2.

kchadha:
### Proposed Changes

Local SB2 Import should also deserialize assets from the .sb2 file.

### Reason for Changes

Supports loading an .sb2 file when the assets cannot be found in storage or when not connected to the internet).

### Test Coverage

Tests have been updated to pass in .sb2's whole instead of necessarily extracting project.json first.


rschamp (approved):
Looks good! Is there an issue associated with this change — if so could you add it to the description? Also, test cases for things that currently break that we can confirm this fixes.

src/serialization/deserialize-assets.js
```diff
@@ -64,16 +67,18 @@ const deserializeSound = function (sound, runtime, zip) {
...
  */
-const deserializeCostume = function (costume, runtime, zip) {
+const deserializeCostume = function (costume, runtime, zip, assetFileName) {
     const storage = runtime.storage;
     const assetId = costume.assetId;
-    const fileName = costume.md5 ?
-        costume.md5 :
-        `${assetId}.${costume.dataFormat}`; // The md5 property has the full file name
+    const fileName = assetFileName ? assetFileName :
+        `${assetId}.${costume.dataFormat}`;
```
cwillisf, line 81: 😎

src/serialization/sb2.js
```diff
@@ -256,9 +267,19 @@ const parseScratchObject = function (object, runtime, extensions, topLevel) {
...
+            const ext = idParts[1].toLowerCase();
+            sound.dataFormat = ext;
+            sound.assetId = md5;
+            // If there is no internet connection, or if the asset is not in storage
+            // for some reason, and we are doing a local .sb2 import, (e.g. zip is provided)
+            // the file name of the sound should be the soundID (provided from the project.json)
+            // followed by the file ext
+            const assetFileName = `${soundSource.soundID}.${ext}`;
+            soundPromises.push(deserializeSound(sound, runtime, zip, assetFileName)
+                .then(() => loadSound(sound, runtime)));
```
cwillisf, line 282: Hmm... if you can find a good way to make a function which handles both cases (sound & costume) that'd be great, but maybe it's too tricky given how many different objects & properties are flying around?

cwillisf (approved):
👍

kchadha:
@rschamp, to address your comment, no there was no open issue that this closes, but the reason loading .sb2 files has been working thus far is that storage has been looking up the assets on the server instead of getting them from the provided .sb2 file. Something that would break before this change would be an .sb2 file (with custom assets) created from the offline scratch editor (so the assets don't get stored on the server) or loading an .sb2 file locally without an internet connection. I'm not sure how to add a test other than to create a fixture for a project like the first option I mentioned...

src/serialization/sb2.js
kchadha: I agree with you and this definitely on my list of things I want to do in the future, but I think I'm going to hold off on this for now. Eventually, we want to move these functions into scratch-storage anyways.

kchadha:
@rschamp, @cwillisf, I just added a new fixture and integration test that will fail on develop but passes with these changes. The new fixture is an sb2 project from the offline editor with a custom costume and a custom sound recording. The test compares what's returned from storage with the contents of the costume/sound file in the actual .sb2 file. Without the changes proposed in this PR, the test should fail since what is returned from the server is a default asset or null.

src/serialization/deserialize-assets.js
gnarf: You could make this the default value for `assetFileName`:

```js
function (costume, runtime, zip, assetFileName = `${costume.assetId}.${costume.dataFormat}`)
```

ends up slightly more readable than a ternary "args fixing" statement IMO

test/fixtures/readProjectFile.js
```diff
@@ -0,0 +1,12 @@
+const AdmZip = require('adm-zip');
+const fs = require('fs');
+
+module.exports = {
+    readAsBuffer: function (path) {
+        return new Buffer(fs.readFileSync(path));
+    },
+    readAsString: function (path) {
```
gnarf: These method names don't seem to make much sense to me... I know they are just in tests, but it might clarify things a bit to rename them.

maybe `readAsString` => `projectJsonFromZip` and
`readAsBuffer` => `readFileToBuffer`

kchadha: @gnarf, good point! How about `readAsString` => `extractProjectJson`?

gnarf: sounds even better than my suggetion :)

test/unit/serialization_sb2.js
```diff
@@ -15,7 +15,7 @@ test('spec', t => {
 test('default', t => {
     // Get SB2 JSON (string)
     const uri = path.resolve(__dirname, '../fixtures/default.sb2');
-    const file = readAsString(uri);
+    const file = extractProjectJson(uri);
     const json = JSON.parse(file);
```
gnarf: just looking at this diff made me realize that I think every time you use this you call JSON.parse on it, you might as well make that part of the method too?

thisandagain (approved):
😎

rschamp (changes requested):
@kchadha Thanks for adding the test! For posterity, what we discussed offline:
For this test please remove all helpers from the test storage so that the test is unable to try to get assets from anywhere but the local zip. Then this should fail if it's ever unable to find those assets rather than falling back to the network.

kchadha:
@rschamp Modified the test.

rschamp (approved):

Issue #304, opened 2016-10-24
Implement video motion sensing


tmickel:
We need to do at least two things:
- Display the video on the stage (do this in scratch-render?)
- Implement the sensing primitives. Reference: https://github.com/LLK/scratch-flash/blob/master/src/primitives/VideoMotionPrims.as. Personally I might lean toward a direct reproduction of this if possible, i.e., a translation AS3->JavaScript. More documentation at https://llk.media.mit.edu/papers/tony-hwang-thesis.pdf

thisandagain:
> Personally I might lean toward a direct reproduction of this if possible, i.e., a translation AS3 -> JavaScript

Agreed.

CosmicWebServices:
Well @thisandagain there are some really good javascript libraries that can do what you want and more...

AmazingMech2418:
There are only 2 blocks currently, but on some of my projects that use pen and video, a video transparency block like what's in Scratch 2 would be very helpful.

PR #1001, opened 2018-03-26
Motion detect

mzgoddard:
### Resolves

Incremental part of #304

### Proposed Changes

- Add rough version of videoSensing extension
- Implement scratch 2 `senseVideoMotion` opcode as `videoSensing.videoOn`
- Add a debug and benchmark page for video sensing

### Test Coverage

Part of a future PR.

-----

This is an incremental implementation of the Video Sensing extension. It implements the the video motion input block but does not yet implement the video motion event block. The video motion block supports scratch 2's motion and direction attributes of either the stage or 'this sprite'.

The extension started as a clone of the music extension so there may remain some parts in the video sensing extension that may not be needed. As an incremental step to implement video sensing, we could merge this in while further PRs I'm working on add the event block and do further documentation, clean up, and testing. Review on things like names displayed in the gui or argument names would be great to get.

### DIFF
--- a/src/extensions/scratch3_video_sensing/index.js
+++ b/src/extensions/scratch3_video_sensing/index.js
@@ -1,0 +1,336 @@
+const ArgumentType = require('../../extension-support/argument-type');
+const BlockType = require('../../extension-support/block-type');
+const Clone = require('../../util/clone');
+const log = require('../../util/log');
+const Timer = require('../../util/timer');
+
+const VideoMotion = require('./lib');
+
+/**
+ * Icon svg to be displayed at the left edge of each extension block, encoded as a data URI.
+ * @type {string}
+ */
+// eslint-disable-next-line max-len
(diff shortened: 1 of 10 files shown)


src/extensions/scratch3_video_sensing/index.js
```diff
@@ -0,0 +1,336 @@
...
+            this._drawable,
+            Scratch3VideoSensingBlocks.ORDER
+        );
+        this.runtime.renderer.updateDrawableProperties(this._drawable, {
+            skinId: this._skinId
+        });
+    }
+
+    _loop () {
+        setTimeout(this._loop.bind(this), this.runtime.THREAD_STEP_INTERVAL_COMPATIBILITY);
```
cwillisf: Should this use `Scratch3VideoSensingBlocks.INTERVAL` instead? Or perhaps it should be `runtime.currentStepTime`?

src/extensions/scratch3_video_sensing/lib.js
```diff
@@ -0,0 +1,275 @@
...
+        let A2;
+        let A1B2;
+        let B1;
+        let C1;
+        let C2;
+        let u;
+        let v;
+        let uu;
+        let vv;
+        let n;
```
cwillisf: I'd suggest moving these inside the relevant scopes, where possible. For example, it looks like the `A`/`B`/`C` family could move into the `j` loop, to where they get initialized each loop. Unless JS is silly enough to cause a performance hit in a case like that...?

src/extensions/scratch3_video_sensing/view.js
```diff
@@ -0,0 +1,328 @@
...
+
+        this.output = output;
+        this.buffer = new Uint32Array(WIDTH * HEIGHT);
+    }
+
+    static get OUTPUT () {
+        return OUTPUT;
+    }
+
+    draw () {
```
cwillisf: There's a lot of code duplicated here... is that intentional? Could they share code, or is one of the copies obsolete?

webpack.config.js
```diff
@@ -86,7 +86,8 @@ module.exports = [
                 'scratch-storage',
                 // Renderer
                 'scratch-render'
-            ]
+            ],
+            'motion-extension': './src/extensions/scratch3_video_sensing/debug'
```
cwillisf, line 90: It shouldn't be necessary to explicitly mention the sensing extension in the webpack config as long as it's `require`d as you do in `extension-manager.js`. If you want to expose debug stuff to the JS console, I would use `global` (which Webpack turns into `self` or `window`). I imagine you're familiar, but if not, you can find an example in `src/extension-support/extension-worker.js`.

cwillisf (changes requested):
I have a few nitpicks but it looks like you're making great progress! 📹

src/extensions/scratch3_video_sensing/index.js
thisandagain: 👍 That is actually my bad from the Vision extension. Sorry @mzgoddard !

src/extensions/scratch3_video_sensing/source.js
```diff
@@ -0,0 +1 @@
+
```
thisandagain: Can this be removed?

webpack.config.js
mzgoddard: This is used by `src/playground/motion.html`.

src/extensions/scratch3_video_sensing/source.js
mzgoddard: Yes.

src/extensions/scratch3_video_sensing/index.js
mzgoddard: That's a good point @cwillisf. Eventually we probably want to use [`timeupdate`](https://developer.mozilla.org/en-US/docs/Web/Events/timeupdate) or some thing to execute a step of the loop approximately at the hertz of the camera.

src/extensions/scratch3_video_sensing/lib.js
mzgoddard: I think that'll be better for readability.

src/extensions/scratch3_video_sensing/view.js
mzgoddard: Its duplicated intentionally. A lot of it was finding what needed to be copied to get the output renders to help understand the algorithm output in real time. The gradX, gradY, gradT calculations could be deduped. Some of the copies are out of date in minor ways.

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -0,0 +1,336 @@
...
+        return 'Scratch.videoSensing';
+    }
+
+    /**
+     * The default music-related state, to be used when a target has no existing music state.
+     * @type {MusicState}
+     */
+    static get DEFAULT_MOTION_STATE () {
+        return {
+            currentInstrument: 0
```
gnarf, line 197: ?? copy/pasta leftover from music?

mzgoddard:
@cwillisf Do you see any more improvements this needs?

cwillisf (approved):
Looks good, assuming that future work will clean up the remaining commented and/or music-related bits.

PR #1007, opened 2018-03-30
Motion detect 2

mzgoddard:
### Resolves

Incremental part of #304

### Proposed Changes

- Adds missing `whenMotionGreaterThan` (scratch 3 video motion version of scratch 2's whenSensorGreaterThan) block
- Reduces code duplication in VideoSensing extension
- Documents current VideoSensing code surface

### Reason for Changes

#1001 Adds the working extension. This PR builds on that improving the code quality and documentation so the extension can have a better maintenance experience and as a code area for learning about Scratch.

### Further Work

`whenMotionGreaterThan` presents two problems one in the current extension interface and another in the sb2_specmap translation.

1. The current extension interface uses a fixed meta object (`{edgeActivated: true}`) for hats. `whenMotionGreaterThan` works as an edgeActivated hat but to maintain Scratch 2 compatibility it also needs to define `restartsExistingThreads: true` in its hat meta information.
2. The sb2_specmap (to my knowledge) does not handle splitting a Scratch 2 opcode into multiple Scratch 3 opcodes. One avenue I see for fixing this is leaving the existing opcode translation for Scratch 2's `whenSensorGreaterThan` to `event_whengreaterthan` and doing more specified logic in `sb2.js` where it works on input values to differentiate `event_whengreaterthan` into `videoSensing.whenMotionGreaterThan` when the Scratch 2's block is for `'video motion'`.


mzgoddard:
It will probably be difficult to review this until #1001 is merged.

cwillisf:
Regarding `restartsExistingThreads`: I'm adding support for that in extensions right now :)

I'm not sure I understand what you were saying about `event_whengreaterthan` - could you explain the problem a little more? As part of my current extension work I'm also adding support for "event" hats which run when an event is emitted and have no implementation function... would that help?

mzgoddard:
That could be a way to fix the issue. I don't think its the one we want? The event is known and the implementation is known.

@cwillisf So the problem is deserializing a Scratch 2 project, when a `whenSensorGreaterThan` block is encountered in Scratch 3 it'll be multiple possible blocks depending on the first argument stored in the Scratch 2 block. One of those opcodes is the `videoSensing.whenMotionGreaterThan` in this PR. The deserializer can deterministically know when to transform `whenSensorGreaterThan` opcode into `videoSensing.whenMotionGreaterThan`. There is not yet a defined pattern to do this. The `sb2_specmap.js` lookup can do 1 to 1 translations. After that lookup is some if/else handling to transform special case arguments.

Example serialized Scratch 2 block:
```json
["whenSensorGreaterThan", "video motion", 10]
```

The solution could be to add if/else handling for 1 to one of many opcode transformations around there. Should that be the pattern or should it be something else? I'm looking for advice on how to solve this. I know a way I could solve it but I'm not certain that is the way it should be solved.

cwillisf:
Oh! I see. My first impulse is to allow functions in the SB2 specmap, and then do something like this in the code that uses the specmap:
```js
let sb3block = specmap[sb2block.opcode];
if (typeof specMapEntry === 'function') {
  sb3block = sb3block(sb2block);
}
```
Since I expect this to be extremely rare -- this may be the only case, in fact -- I would also be OK with a special-case if/else to handle it. Just depends on how YAGNI you're feeling, I guess :)

mzgoddard:
I'll make a PR with that (https://github.com/LLK/scratch-vm/pull/1007#issuecomment-378099600) or whichever so it can have any specific review to that desired.

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -81,6 +173,11 @@ class Scratch3VideoSensingBlocks {
         });
     }
 
+    /**
+     * Create a campus to render the user media video to down/up sample to the
```
rschamp, line 177: campus!

rschamp (approved):
@mzgoddard Thank you for all of this documentation!

I noticed is that if the hat block is ever used, the project is always shown as running. This is also true for the "when timer > _" hat so probably due to the way we determine the running state. It's not this way in Scratch 2.  Not anything for you to act on, just something that came up while I was testing it out.

PR #1008, opened 2018-03-31
Implement "item # of (thing) in (list)" block

towerofnix:
### Resolves

Resolves LLK/scratch-gui#600. Should be merged alongside LLK/scratch-blocks#1425, which adds the actual block.

### Proposed Changes

Adds the functionality for the "item # of (thing) in (list)" block, as described in the above issue.

Note that this follows @mrjacobbloom's [comment](https://github.com/LLK/scratch-gui/issues/600#issuecomment-321864666):

> Since scratch generally ignores case, I'd expect this block to be completely case-insensitive.

The algorithm in this block uses `Cast.compare`, which is the same function that the `=` operators block uses.

### Reason for Changes

To implement the help-wanted feature.

### Test Coverage

I added a new unit test file, `blocks_data.js`. It's a bit empty since it only contains tests for this new block, but I figured I'd get it started. The tests include:

```
item # of "burrito" in ["apple", "taco", "burrito", "extravaganza"]` is 3.
item # of "jump" in ["aaaaapple", "burrito"]` is 0.

item # of "Jump" in ["jump", "Jump", "123", 123, 800]` is 1. ("jump" = "Jump")
   item # of 123 in ["jump", "Jump", "123", 123, 800]` is 3. ("123" = 123)
 item # of "800" in ["jump", "Jump", "123", 123, 800]` is 5. (800 = "800")
```

Lint says this looks good.

### DIFF
--- a/src/blocks/scratch3_data.js
+++ b/src/blocks/scratch3_data.js
@@ -24,6 +24,7 @@ class Scratch3DataBlocks {
             data_insertatlist: this.insertAtList,
             data_replaceitemoflist: this.replaceItemOfList,
             data_itemoflist: this.getItemOfList,
+            data_itemnumoflist: this.getItemNumOfList,
             data_lengthoflist: this.lengthOfList,
             data_listcontainsitem: this.listContainsItem
         };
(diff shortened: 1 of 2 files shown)


towerofnix:
> The algorithm in this block uses `Cast.compare`, which is the same function that the `=` operators block uses.

Adding onto that, this block is basically equivalent to the following Scratch (pseudo-)script:

Scratchblocks screenshot

src/blocks/scratch3_data.js
```diff
@@ -122,6 +123,34 @@ class Scratch3DataBlocks {
...
+        // Go through the list items one-by-one using Cast.compare. This is for
+        // cases like checking if 123 is contained in a list [4, 7, '123'] --
+        // Scratch considers 123 and '123' to be equal.
+        for (let i = 0; i < list.value.length; i++) {
+            if (Cast.compare(list.value[i], item) === 0) {
+                return i + 1;
+            }
+        }
+
+        // We don't bother using .indexOf() at all, because it would end up with
```
gnarf, line 140: It wouldn't be any "faster" performance wise, the for loop will win here,  but you could probably use `findIndex` to simplify the code here:

```js
return 1 + list.value.findIndex( cmp => Cast.compare(cmp, item) === 0 );
```

towerofnix: Oh, that's definitely clever. A related change would be below, for `listContainsItem`:

```js
return list.value.some(cmp => Cast.compare(cmp, item) === 0);
```

I'll probably leave this as is for the (probably very minor) performance boost (and so I don't modify the other function's code in this PR), but I can make that change if wanted. Array functions like `findIndex` and `some` are very nice!

ericrosenbaum:
This looks great!

Issue #954, opened 2018-02-27
Support saving and loading state from extensions


ericrosenbaum:
Should the SB3 file format include state information used by extensions? 

There are some examples of "global" (i.e. non-sprite-specific) state used by extensions that we will need to save and load, for compatibility reasons. This includes:

- tempo (music extension)
- videoAlpha (video sensing extension)

For sprite-specific state (now handled by the "custom state" mechanism), it seems to generally not be saved in 2.0 (such as pen color and size, instrument, volume, etc).

So the question is, do we need SB3 to be able to save and load global or sprite-specific state? 

We can handle the legacy saved state for tempo and videoAlpha by adding them to the stage (I did this for the [tempo](https://github.com/LLK/scratch-vm/pull/953)). The SB3 format will presumably need to include these. 

Will there be future cases where an extension will need to save global or sprite-specific state info?

thisandagain:
Thanks for documenting this @ericrosenbaum. Given @kchadha's work on SB3 save / load we should probably chat about this soon. :-)

mrjacobbloom:
>Will there be future cases where an extension will need to save global or sprite-specific state info?

Probably! I say make the extension system as powerful as possible. One example use case I can think of is for a physics extension, remembering a sprite's gravity. Or for an extension that adds custom reporters, there would be plenty of state needed to keep everything linked together correctly. Or for an extension that types text on the stage on a pen-like layer, remembering what text was last left on the stage

thisandagain:
From discussion this is something we would like to support, but this will not be ready for initial release. Moving to backlog.

PR #1014, opened 2018-04-03
Import and store video state

ericrosenbaum:
### Resolves

Related to https://github.com/LLK/scratch-vm/issues/954

### Proposed Changes

- Add videoState and videoTransparency properties to rendered target
- Re-arrange rendered target's properties a tiny bit, with comments clarifying which ones are global to the project, and stored in the stage
- Load videoState and videoTransparency from SB2 files

Note that SB2 files store video state as a boolean, so it does not save a difference between "on" and "on-flipped". In Scratch 3.0, we should store a string, so we can save any of the three states (and possibly others in the future, e.g. front vs rear camera).

cc @mzgoddard

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -118,16 +118,31 @@ class RenderedTarget extends Target {
         this.rotationStyle = RenderedTarget.ROTATION_STYLE_ALL_AROUND;
 
         /**
-         * Current tempo (used by the music extension)
+         * Loudness for sound playback for this target, as a percentage.
+         * @type {number}
+         */
+        this.volume = 100;
+
+        /**
+         * Current tempo (used by the music extension).
+         * This property is global to the project and stored in the stage.
          * @type {number}
          */
         this.tempo = 60;
 
         /**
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LGTM, thanks for adding these!

src/sprites/rendered-target.js
```diff
@@ -194,6 +209,18 @@ class RenderedTarget extends Target {
...
 
+    /**
+     * Available states for video input.
+     * @type {object}
+     */
+    static get VIDEO_STATE () {
+        return {
+            'OFF': 'off',
+            'ON': 'on',
+            'ON-FLIPPED': 'on-flipped'
```
mzgoddard, line 220: `ON-FLIPPED` cannot be used like "normal" enums. `VIDEO_STATE.ON-FLIPPED` will be a ReferenceError.

kchadha: @mzgoddard Is this because of the hyphen? or am I missing something?

kchadha: has patch in #1031

PR #1016, opened 2018-04-05
Video IO Device Implementation

gnarf:
### Resolves

Incremental part of #304

### Proposed Changes

* Implements an IO device interface to share access to the video stream with other parts
* Implements remaining blocks `Turn video [off/on/on-flipped]` and `Set video transparency to [N]`

### DIFF
--- a/src/extensions/scratch3_video_sensing/index.js
+++ b/src/extensions/scratch3_video_sensing/index.js
@@ -339,14 +339,13 @@ class Scratch3VideoSensingBlocks {
     }
 
     videoToggle (args) {
-        // imported blocks have VIDEO_STATE "off", "on", "on-flipped" as opposed to the numerics?
         const state = Number(args.VIDEO_STATE);
         // 1 == off, 2 & 3 are on (3 is flipped)
-        if (args.VIDEO_STATE === 'off' || state === 1) {
+        if (state === 1) {
             this.runtime.ioDevices.video.disableVideo();
         } else {
             this.runtime.ioDevices.video.enableVideo();
-            this.runtime.ioDevices.video.mirror = args.VIDEO_STATE === 'on' || state === 2;
+            this.runtime.ioDevices.video.mirror = state === 2;
         }
     }
 
--- a/src/extensions/scratch3_video_sensing/index.js
+++ b/src/extensions/scratch3_video_sensing/index.js
@@ -41,11 +41,7 @@ class Scratch3VideoSensingBlocks {
             // Boot up the video, canvas to down/up sample the video stream, the
             // preview skin and drawable, and kick off looping the analysis
             // logic.
-            this.runtime.ioDevices.video.requestVideo()
-                .then(({release}) => {
-                    this.releaseVideo = release;
-                    this._loop();
-                });
+            this._loop();
         }
     }
 
@@ -218,6 +214,14 @@ class Scratch3VideoSensingBlocks {
         return 2;
     }
 
+    get VIDEO_STATE_INFO () {
+        return [
+            {name: 'off'},
+            {name: 'on'},
+            {name: 'on-flipped'}
+        ];
+    }
+
     /**
      * @returns {object} metadata for this extension and its blocks.
      */
@@ -255,11 +259,33 @@ class Scratch3VideoSensingBlocks {
                             defaultValue: 10
                         }
                     }
+                },
+                {
+                    opcode: 'videoToggle',
+                    text: 'turn video [VIDEO_STATE]',
+                    arguments: {
+                        VIDEO_STATE: {
+                            type: ArgumentType.NUMBER,
+                            menu: 'VIDEO_STATE',
+                            defaultValue: 1
+                        }
+                    }
+                },
+                {
+                    opcode: 'setVideoTransparency',
+                    text: 'set video transparency to [TRANSPARENCY]',
+                    arguments: {
+                        TRANSPARENCY: {
+                            type: ArgumentType.NUMBER,
+                            defaultValue: 0
+                        }
+                    }
                 }
             ],
             menus: {
                 MOTION_DIRECTION: this._buildMenu(this.MOTION_DIRECTION_INFO),
-                STAGE_SPRITE: this._buildMenu(this.STAGE_SPRITE_INFO)
+                STAGE_SPRITE: this._buildMenu(this.STAGE_SPRITE_INFO),
+                VIDEO_STATE: this._buildMenu(this.VIDEO_STATE_INFO)
             }
         };
     }
@@ -311,6 +337,20 @@ class Scratch3VideoSensingBlocks {
         this.detect.analyzeFrame();
         const state = this._analyzeLocalMotion(util.target);
         return state.motionAmount > Number(args.REFERENCE);
+    }
+
+    videoToggle (args) {
+        const state = Number(args.VIDEO_STATE);
+        // 1 == off, 2 & 3 are on (3 is flipped)
+        if (state > 1) {
+            this.runtime.ioDevices.video.requestVideo();
+        } else {
+            this.runtime.ioDevices.video.disableVideo();
+        }
+    }
+
+    setVideoTransparency (args) {
+        this.runtime.ioDevices.video.setPreviewGhost(Number(args.TRANSPARENCY));
     }
 }
 
--- a/src/extensions/scratch3_video_sensing/index.js
+++ b/src/extensions/scratch3_video_sensing/index.js
@@ -117,7 +117,6 @@ class Scratch3VideoSensingBlocks {
         const offset = time - this._lastUpdate;
         if (offset > Scratch3VideoSensingBlocks.INTERVAL) {
             const frame = this.runtime.ioDevices.video.getFrame({
-                mirror: true,
                 format: Video.FORMAT_IMAGE_DATA,
                 dimensions: Scratch3VideoSensingBlocks.DIMENSIONS
             });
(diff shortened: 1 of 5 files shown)


src/io/video.js
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        // if we haven't already created and started a preview frame render loop, do so
+        if (!this._renderPreviewFrame) {
+            renderer.updateDrawableProperties(this._drawable, {
+                visible: true
+            });
+
+            this._renderPreviewFrame = () => {
+                if (!this._renderPreviewFrame) {
```
mzgoddard: Check if `this._renderPreviewFrame !== renderPreviewFrame`. `this._renderPreviewFrame` could be truthy if video was quickly toggled off and on.

src/io/video.js
```diff
@@ -0,0 +1,332 @@
...
+        if (this._skin && this._drawable) {
+            this._skin.clear();
+            this.runtime.renderer.updateDrawableProperties(this._drawable, {visible: false});
+        }
+        this._renderPreviewFrame = null;
+    }
+
+    _setupPreview () {
+        const {renderer} = this.runtime;
+        if (!renderer) return;
```
mzgoddard, line 282: Return early if `this._video` is set to null. There is a race condition around `this._video` if video is quickly enabled and disabled.

src/io/video.js
```diff
@@ -0,0 +1,332 @@
...
+     * Should probably be moved to -render or somewhere similar later
+     * @private
+     * @return {Promise} When video has been received, rejected if video is not received
+     */
+    _setupVideo () {
+        if (this._singleSetup) {
+            return this._singleSetup;
+        }
+
+        this._video = document.createElement('video');
```
mzgoddard: There is a race condition on this member `this._video`. It affects a lot of the code in here relating to enabling and disabling the video stream. There are a few ways we can fix it.

The first I can think of is to put all the enabling and disabling code into a task queue. We could use a separate promise chain from `_singleSetup` to create that queue. That queue would also help manage the race conditions in the video preview.

Another fix, would be the more thorough stage change tracking so the right parts of the video setup calls execute or don't. (Internally there is more than if the video is on or off.)

-----

An extra credit change would be to set a state value to enabled or disabled and debounce the enabling/disabling work. When the debounce'd work starts, check the enabled or disabled state and kick off that work. That way quick camera toggling doesn't create multiple video tags or getUserMedia calls.

mzgoddard (commented):
I think there are race conditions around enabling and disabling the camera if they are quickly toggled back and forth.

src/io/video.js
gnarf: I'm not sure that this one matters, we'd just end up with a empty pen skin drawable ready for the video to toggle back on.

src/io/video.js
gnarf: I think I'd rather guard by clearing any previous/double setTimeout's -- doing a double render is fairly inexpensive here

src/io/video.js
gnarf: What is the race you're seeing? enable/disable/enable causing two getUserMedia calls to overwrite each other's stream?  I think the only race-blocking needed would be stopping the old `_track` if we had one.

src/io/video.js
```diff
@@ -0,0 +1,332 @@
...
+                    height: {min: 360, ideal: 480}
+                }
+            }, stream => {
+                this._video.src = window.URL.createObjectURL(stream);
+                // Hint to the stream that it should load. A standard way to do this
+                // is add the video tag to the DOM. Since this extension wants to
+                // hide the video tag and instead render a sample of the stream into
+                // the webgl rendered Scratch canvas, another hint like this one is
+                // needed.
+                this._track = stream.getTracks()[0];
```
gnarf: if _track was setup by the "lost" promise, we could even bail early here...

src/io/video.js
gnarf: I did a bunch of extra checks to make sure the state was what I expected...

I also tested in my gui with some blocks like 

image 1

It seems to do the "right thing" which is end in whatever state you asked for.

gnarf: ^^^ in this state there might end up being like 3 or 4 getUserMedia calls going at once, but they all clean themselves up and only one stream is kept

src/io/video.js
mzgoddard: It'll leave an empty pen skin but the next enable won't start the preview loop because _renderPreviewFrame is still set. Another disable followed by an extra enable would be needed.

gnarf: if `_renderPreviewFrame` is set - it is running... `getFrame` just exits early if `!this.videoReady` -- so it will get called once per frame forever once we set it up, unless we remove it.

src/io/video.js
mzgoddard: The race condition here involved the video tag and preview work without the checks that have since been added.

src/io/video.js
```diff
@@ -91,25 +91,33 @@ class Video {
...
-        // by clearing refs to video and track, we should lose our hold over the camera
-        this._video = null;
-        if (this._track) {
-            this._track.stop();
+        this.enabled = false;
+        // If we have begun a setup process, wait for it to complete
+        if (this._singleSetup) {
+            this._singleSetup
+                .then(() => {
+                    // we might be asked to re-enable before setup completes
```
mzgoddard: Wicked.

src/io/video.js
mzgoddard: Yes. You're right.

src/io/video.js
mzgoddard: Wish I could mark this as resolved.

gnarf:
The error state in the travis-CI seems totally unrelated / random failure?

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -380,11 +254,33 @@ class Scratch3VideoSensingBlocks {
...
+                },
+                {
+                    opcode: 'videoToggle',
+                    text: 'turn video [VIDEO_STATE]',
+                    arguments: {
+                        VIDEO_STATE: {
+                            type: ArgumentType.NUMBER,
+                            menu: 'VIDEO_STATE',
+                            defaultValue: 1
+                        }
```
cwillisf: I think I'd prefer if this were a string argument (with values like `'on'`) rather than integers. This would make it easier for a human to understand if they read the project JSON, easier to read during debugging, and shouldn't cause a significant performance difference. It also means that if we reorder the items later, existing projects will still work (`'on'` will still mean `'on'` whereas item `1` might become item `2`). Check out the `COLOR_PARAM` menu in the pen extension for an example.

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -437,6 +333,21 @@ class Scratch3VideoSensingBlocks {
         const state = this._analyzeLocalMotion(util.target);
         return state.motionAmount > Number(args.REFERENCE);
     }
+
+    videoToggle (args) {
+        const state = Number(args.VIDEO_STATE);
```
cwillisf: As a minor style point, similar to above: we tend to keep things as string values (emulating `enum` values) rather than using integers in situations like this. For example, I would replace `VIDEO_STATE_INFO` with something like `{OFF: 'off', ON: 'on', FLIPPED: 'on-flipped'}` (your choice on the exact wording). Since JS interns all strings there's no extra comparison cost, and this means both the code and the runtime values are self-documenting.

If there's specific reason to use numbers, I'd still prefer `if (state === VideoState.OFF)` or something like that -- at least the code is self-documenting, then, even if the runtime values aren't :)

src/io/video.js
```diff
@@ -0,0 +1,371 @@
...
+            if (this._track) {
+                this._track.stop();
+            }
+            this._track = null;
+        }
+    }
+
+    /**
+     * Return frame data from the video feed in a specified dimensions, format, and mirroring.
+     * @return {ArrayBuffer|Canvas|string|null} Frame data in requested format, null when errors.
```
cwillisf: Since this is a public method, please add `@param` documentation

cwillisf (dismissed):
Overall this looks great! I have a couple of nits to pick but it's great to see so much progress so quickly!

If this is intended as a WIP review then my comments might be overly picky - let me know if that's the case. Thanks!

src/extensions/scratch3_video_sensing/index.js
gnarf: :+1: I had just copied the pattern used for the "stage / sprite" menu, should that also be using text instead of numbers?

src/extensions/scratch3_video_sensing/index.js
gnarf: yeah - numbers suck for this - I like the idea of strings better too! :)

src/io/video.js
gnarf: of course! Is there an example you can think of when using a destructured object for params?  Like how do I document `dimensions` and etc as props of the param?

src/extensions/scratch3_video_sensing/index.js
gnarf: I was just looking at the pen extension to get some ideas how to do this in the menus system, I have to say - It's a bit confusing with how much abstraction is happening in these menu/block definitions... should I be using `format-message` for these things too?  @mzgoddard do you have a better idea how to do this than I?

src/io/video.js
gnarf: d682555

src/extensions/scratch3_video_sensing/index.js
mzgoddard: @gnarf I'm not immediately sure. I used numbers following what I thought was the idiom in the music extension.

mzgoddard:
@cwillisf How's it look now?

rschamp (changes requested):
A couple of things I caught while testing (built this pr merged into develop, and linked that into the GUI):

After using the turn video on block, I can't change the state of the video — no errors, it just seems to no-op.

The `turn video ___v` block should have `on` as the default option. Instead the default is an empty state:
image

That dropdown should also be non-droppable, but I don't think we have the capability to do that yet in the extension system?

mzgoddard:
@rschamp I fixed the no-op bug and the dropdown default.

rschamp (approved):
Looks great!

Issue #987, opened 2018-03-16
Clones don't inherit the parent's instrument


towerofnix:
### Expected Behavior

Clones created of a sprite should be spawned with the same current instrument as that sprite.

### Actual Behavior

Clones don't inherit the instrument; clones are always spawned with the default instrument, Piano.

### Steps to Reproduce

Try making a script like this:

![Set instrument to guitar; create clone of myself. When I start as a clone, play a note.](https://user-images.githubusercontent.com/9948030/37542137-f0a44144-293b-11e8-9e52-aae571065fb1.png)

Compare the result with a script like this:

![Create clone of myself. When I start as a clone, set instrument to guitar, then play a note.](https://user-images.githubusercontent.com/9948030/37542171-0bc82198-293c-11e8-9dd2-3cd2c9868619.png)

### Operating System and Browser

Debian (testing), Firefox 60.0b3.

towerofnix:
Here's [a project](https://llk.github.io/scratch-gui/develop/#110203460) I noticed this issue in.

PR #1017, opened 2018-04-05
Clones inherit music custom state

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/987

### Proposed Changes

Listen for the targetWasCreated event, and copy any custom state for the music extension (just the instrument setting) into the new clone.

### Reason for Changes

Compatibility with Scratch 2.0, in which clones inherit the instrument set by the parent.


cwillisf (approved):
LGTM

PR #1019, opened 2018-04-05
Test VideoMotion and VideoSensing blocks

mzgoddard:
### Resolves

Incremental part of #304

### Proposed Changes

- Add pngjs dev dependency to read test images
- Test VideoMotion and VideoSensing blocks

### Test Coverage

Add tests for VideoMotion and VideoSensing blocks


mzgoddard:
Sibling to #1016. Tiny code conflict that will need to be changed after #1016 goes in. (Or #1016 would need after this one.)

test/unit/extension_video_sensing.js
```diff
@@ -0,0 +1,408 @@
...
+ */
+const readPNG = name => (
+    new Promise((resolve, reject) => {
+        const png = new PNG();
+        createReadStream(join(__dirname, `${pngPrefix}${name}.png`))
+            .pipe(png)
+            .on('parsed', () => {
+                // Copy the RGBA pixel values into a separate typed array and
+                // cast the array to Uint32, the array format VideoMotion takes.
+                resolve(new Uint32Array(new Uint8ClampedArray(png.data).buffer));
```
cwillisf, line 43: Is this different from `resolve(new Uint32Array(png.data.buffer))`? My reading of the `pngjs` code suggests that `png.data` is a `Buffer`, which means that in Node 4.x+ it's also a `Uint8Array` already, so it should just be a view on the same actual data...?

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -359,12 +378,12 @@ class Scratch3VideoSensingBlocks {
                         MOTION_DIRECTION: {
                             type: ArgumentType.NUMBER,
                             menu: 'MOTION_DIRECTION',
-                            defaultValue: 1
+                            defaultValue: Scratch3VideoSensingBlocks.MOTION
```
cwillisf, line 381: 😄

test/unit/extension_video_sensing.js
```diff
@@ -0,0 +1,408 @@
...
+ * @param {number} actual - actual angle in degrees
+ * @param {number} expect - expected angle in degrees
+ * @param {number} optMargin - allowed margin between actual and expect in degrees
+ * @returns {boolean} true if actual is close to expect
+ */
+const isNearAngle = (actual, expect, optMargin = 10) => (
+    ((actual < expect + optMargin) && (actual > expect - optMargin)) ||
+    ((actual + 360 < expect + optMargin) && (actual + 360 > expect - optMargin)) ||
+    ((actual - 360 < expect + optMargin) && (actual - 360 > expect - optMargin))
+);
```
cwillisf: Super minor: the wrapping part of this might be cleaner to implement in terms of modulo (`%`), or in terms of `wrapClamp`(*) from `math-util.js`. I could also see `isNearAngle` moving into `math-util.js` in case anyone else wants to use it.

(*) That said, I was surprised to see how our current implementation of `wrapClamp` works just now -- I'm not confident that it's retaining as much precision as it could.

cwillisf (approved):
Looks good! I have a couple questions / extra-credit suggestions, but I'm also OK with this as-is.

test/unit/extension_video_sensing.js
mzgoddard: It's the use of `new Uint8ClampedArray(png.data)` to first make a copy as a object ownership safety in case `png.data` was externally modified. `Uint8Array` would be fine, using `Uint8ClampedArray` since its more canonical for pixel data manipulation.

I can toss the copy. I suppose its a personal default behaviour of mine for this kind of scenario (loading an image) when high performance isn't a priority, like in a test.

src/extensions/scratch3_video_sensing/index.js
mzgoddard: I followed up in a new PR with better enums!

mzgoddard: https://github.com/LLK/scratch-vm/pull/1028

test/unit/extension_video_sensing.js
cwillisf: Oh, I see. That's cool - no objection now that I understand what you're doing, especially in a test like you said :)

rschamp (approved):

PR #1020, opened 2018-04-05
Allow specmap to have methods to branch into multiple sb3 opcodes

gnarf:
### Resolves

Incremental part of #304

### Proposed Changes

* allow sb2 "whenSensorGreaterThan" to have multiple output opcodes 
* sb2_specmap now can have functions that take the block and return multiple different specmaps

### Reason for Changes

* Something needs to happen to allow this old opcode to get converted to a videoSensing opcode, but only when "video motion" is the used sensor.

You can test this on 213408319 with a locally running gui

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -448,6 +448,24 @@ const sb2import = function (json, runtime, optForceSprite) {
 };
 
 /**
+ * Given the sb2 block, inspect the specmap for a translation method or object.
+ * @param {!object} block a sb2 formatted block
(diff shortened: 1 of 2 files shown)


mzgoddard:
This is related to this conversation https://github.com/LLK/scratch-vm/pull/1007#issuecomment-378099600

cwillisf (approved):
:D
I'm really happy with how this turned out. Thanks!

(Oh, and it turns out that I have another case for this already, in the control blocks...)

PR #1021, opened 2018-04-05
Features for control extension

cwillisf:
### Resolves

Progress toward #941

### Proposed Changes

This PR contains a few changes aimed at supporting control blocks as an extension and fleshing out the extension system in general. Specifically:
- Add a `LOOP` block type, which is like `CONDITIONAL` but re-evaluates the parent block after running a substack.
- Implement `LOOP` and `CONDITIONAL` block types.
- Add support for calling `formatMessage` on user-visible text in extensions.
  - Extension text which needs to change depending on whether the target is a sprite or stage can use `formatMessage`'s `select` feature: `{targetType, select, stage {text for stage} sprite {text for
sprite} other {text for other}`
- Add support for loading "core" extensions -- the list is empty right now, but shows how it would be done.

Note that this PR does not include actually converting control blocks into an extension: that will be merged separately later.

### Reason for Changes

See #941 

### Test Coverage

This PR adds two new sets of tests:
- `test/unit/maybe_format_message.js` tests the new `maybeFormatMessage` function, including the use of the `select` feature, and
- `test/unit/extension_conversion.js` tests the Runtime's conversion of extension metadata into scratch-blocks-ready formats, covering several types of blocks as well as separators.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -8,6 +8,7 @@ const BlockType = require('../extension-support/block-type');
 const Sequencer = require('./sequencer');
 const Thread = require('./thread');
 const Profiler = require('./profiler');
+const log = require('../util/log');
 
 // Virtual I/O devices.
 const Clock = require('../io/clock');
@@ -54,6 +55,16 @@ const ArgumentTypeMap = (() => {
     };
     return map;
 })();
+
+/**
+ * Predefined "Converted block info" for a separator between blocks in a block category
+ * @type {ConvertedBlockInfo}
+ */
+const ConvertedSeparator = {
+    info: {},
+    json: null,
+    xml: '<sep gap="36"/>'
+};
 
 /**
  * These constants are copied from scratch-blocks/core/constants.js
@@ -486,7 +497,7 @@ class Runtime extends EventEmitter {
 
     /**
      * Register the primitives provided by an extension.
-     * @param {ExtensionInfo} extensionInfo - information about the extension (id, blocks, etc.)
+     * @param {ExtensionMetadata} extensionInfo - information about the extension (id, blocks, etc.)
      * @private
      */
     _registerExtensionPrimitives (extensionInfo) {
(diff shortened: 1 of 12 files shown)


cwillisf:
Marked as "needs work" pending better test coverage...

src/engine/runtime.js
```diff
@@ -536,29 +521,52 @@ class Runtime extends EventEmitter {
...
+                        edgeActivated: blockInfo.isEdgeActivated,
+                        restartExistingThreads: blockInfo.shouldRestartExistingThreads
+                    };
+                }
+            } catch (e) {
+                log.error('Error parsing block: ', {block: blockInfo, error: e});
+            }
+        }
+    }
+
```
rschamp, line 569: If we can avoid functions that modify state external to them when we can, I think it will be easier to understand how everything works in general. It is a little unavoidable in the blocks loop since we have to register the hats and block implementations, but could we at least break this up into two, one to get the menus and one to get the blocks?

src/engine/runtime.js
```diff
@@ -484,15 +472,28 @@ class Runtime extends EventEmitter {
...
      * Register the primitives provided by an extension.
-     * @param {ExtensionInfo} extensionInfo - information about the extension (id, blocks, etc.)
+     * @param {ExtensionMetadata} extensionInfo - information about the extension (id, blocks, etc.)
      * @private
      */
     _registerExtensionPrimitives (extensionInfo) {
         const categoryInfo = {
             id: extensionInfo.id,
-            name: extensionInfo.name,
+            name: maybeFormatMessage(extensionInfo.name),
```
rschamp, line 496: lol

rschamp (approved):
My only comment is a nice-to-have style thing that shouldn't block this. LGTM but I defer to @thisandagain since you guys will be using this most immediately.

src/engine/runtime.js
thisandagain: Thanks for catching this @rschamp. Good point. 👍

thisandagain (approved):
The test coverage looks great. Thanks @cwillisf. This looks good enough to land and iterate on.

src/engine/runtime.js
thisandagain: @cwillisf This seems ok to do in follow-up. Do you mind filing an issue?

cwillisf: 👍

cwillisf: Filed as #1043

PR #1022, opened 2018-04-05
Add back missing Scratch3VideoSensingBlocks.INTERVAL

mzgoddard:
The INTERVAL time defines how often to analyze a frame. Without the value the
test whether a frame should be analyzed will always be false.

### Resolves

A small error in #1007 removed this member.

### Reason for Changes

The INTERVAL is used in one of the methods for VideoSensingBlocks. Without it a test to see if the newest frame should be analyzed is always false. Effectively keeping the VideoSensing extension from executing its intended purpose.

### Test Coverage

I'm trying to think about how to test this for regression, but its tricky. Need to separate some part of `_loop` but that could in theory cause a different regression.

### DIFF
--- a/src/extensions/scratch3_video_sensing/index.js
+++ b/src/extensions/scratch3_video_sensing/index.js
@@ -93,6 +93,15 @@ class Scratch3VideoSensingBlocks {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


rschamp (approved):
It works!

PR #1031, opened 2018-04-06
Block serialization compression

### LINKED ISSUE #194: Design specification for "SB3"

kchadha:
### Resolves

#194

### Proposed Changes

This is the last of the save/load work. This pull request covers the following changes:

* New list limit of 200k items per list.
* Make sb3 format more compact. This includes: 
  * serializing the following block opcodes into a more compact form ([`number representing opcode`, `info about the opcode`, `(in one or more elements)`]:
    * `math_number`
    * `math_positive_number`
    * `math_whole_number`
    * `math_angle`
    * `math_integer`
    * `colour_picker`
    * `text`
    * `event_broadcast_menu` (this is the shadow block used in broadcast menus)
    * `data_variable`
    * `data_listcontents`
  * reducing duplication in how our data is structured, for example removing serialization of the `id`-uid pair in the  blocks, which were previously stored as:
    ```
    blocks: {
        'this is my long uid for my block' : {
             opcode: 'a cool block'
             id: 'this is my long uid for my block'
             ...  more stuff here ...
    }
    ```
* Added serialization and deserialization for custom state such as `tempo`, `volume`, and video related state

**Note:** Monitor serialization/deserialization is not covered by this work.

### Reason for Changes

Making serialized 3.0 format smaller.

### Test Coverage

Existing save/load tests pass. New test cases for the 3.0 format to be added in the near future.

#### Related PRs
LLK/scratch-parser#32
LLK/scratch-gui#1738

### DIFF
--- a/src/serialization/deserialize-assets.js
+++ b/src/serialization/deserialize-assets.js
@@ -106,14 +106,11 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
         log.error(`Could not find costume file associated with the ${costume.name} costume.`);
         return Promise.resolve(null);
     }
-    let dataFormat = null;
     let assetType = null;
     const costumeFormat = costume.dataFormat.toLowerCase();
     if (costumeFormat === 'svg') {
-        dataFormat = storage.DataFormat.SVG;
         assetType = storage.AssetType.ImageVector;
-    } else if (costumeFormat === 'png') {
-        dataFormat = storage.DataFormat.PNG;
+    } else if (['png', 'bmp', 'jpeg', 'jpg', 'gif'].indexOf(costumeFormat) >= 0) {
         assetType = storage.AssetType.ImageBitmap;
     } else {
         log.error(`Unexpected file format for costume: ${costumeFormat}`);
@@ -126,7 +123,7 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
     return costumeFile.async('uint8array').then(data => {
         storage.builtinHelper.cache(
             assetType,
-            dataFormat,
+            costumeFormat,
             data,
             assetId
         );
--- a/src/serialization/deserialize-assets.js
+++ b/src/serialization/deserialize-assets.js
@@ -30,11 +30,7 @@ const deserializeSound = function (sound, runtime, zip, assetFileName) {
         // This sound has already been cached.
         return Promise.resolve(null);
     }
-    if (!zip) {
-        // TODO adding this case to make integration tests pass, need to rethink
-        // the entire structure of saving/loading here (w.r.t. differences between
-        // loading from local zip file or from server)
-        log.error('Zipped assets were not provided.');
+    if (!zip) { // Zip will not be provided if loading project json from server
         return Promise.resolve(null);
     }
     const soundFile = zip.file(fileName);
@@ -93,11 +89,7 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
         return Promise.resolve(null);
     }
 
-    if (!zip) {
-        // TODO adding this case to make integration tests pass, need to rethink
-        // the entire structure of saving/loading here (w.r.t. differences between
-        // loading from local zip file or from server)
-        log.error('Zipped assets were not provided.');
+    if (!zip) { // Zip will not be provided if loading project json from server
         return Promise.resolve(null);
     }
 
--- a/src/serialization/deserialize-assets.js
+++ b/src/serialization/deserialize-assets.js
@@ -123,6 +123,7 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
     return costumeFile.async('uint8array').then(data => {
         storage.builtinHelper.cache(
             assetType,
+            // TODO eventually we want to map non-png's to their actual file types?
             costumeFormat,
             data,
             assetId
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -31,6 +31,53 @@ const INPUT_BLOCK_NO_SHADOW = 2;
 const INPUT_DIFF_BLOCK_SHADOW = 3;
 // haven't found a case where block = null, but shadow is present...
 
+// Constants referring to 'primitive' types, e.g.
+// math_number
+// text
+// event_broadcast_menu
+// data_variable
+// data_listcontents
+const MATH_PRIMITIVE = 4; // there's no reason these constants can't collide
+const TEXT_PRIMITIVE = 5; // with the above, but removing duplication for clarity
+const BROADCAST_PRIMITIVE = 6;
+const VAR_PRIMITIVE = 7;
+const LIST_PRIMITIVE = 8;
+
+const serializePrimitiveBlock = function (block) {
+    // Returns an array represeting a primitive block or null if not one of
+    // the primitive types above
+    if (block.opcode === 'math_number') {
+        const numField = block.fields.NUM;
+        // If the primitive block has already been serialized, e.g. serializeFields has run on it
+        // then the value of its NUM field will be an array with the value we want
+        // if (Array.isArray(numField)) return [MATH_PRIMITIVE, numField[0]];
+        // otherwise get the num out of the unserialized field
+        return [MATH_PRIMITIVE, numField.value];
+    }
+    if (block.opcode === 'text') {
+        const textField = block.fields.TEXT;
+        // if (Array.isArray(textField)) return [TEXT_PRIMITIVE, textField[0]];
+        return [TEXT_PRIMITIVE, textField.value];
+    }
+    if (block.opcode === 'event_broadcast_menu') {
+        const broadcastField = block.fields.BROADCAST_OPTION;
+        // if (Array.isArray(broadcastField)) return [BROADCAST_PRIMITIVE, broadcastField[0], broadcastField[1]];
+        return [BROADCAST_PRIMITIVE, broadcastField.value, broadcastField.id];
+    }
+    if (block.opcode === 'data_variable') {
+        const variableField = block.fields.VARIABLE;
+        // if (Array.isArray(variableField)) return [VAR_PRIMITIVE, variableField[0], variableField[1]];
+        return [VAR_PRIMITIVE, variableField.value, variableField.id];
+    }
+    if (block.opcode === 'data_listcontents') {
+        const listField = block.fields.LIST;
+        // if (Array.isArray(listField)) return [LIST_PRIMITIVE, listField[0], listField[1]];
+        return [LIST_PRIMITIVE, listField.value, listField.id];
+    }
+    // If none of the above, return null
+    return null;
+};
+
 const serializeInputs = function (inputs) {
     const obj = Object.create(null);
     for (const inputName in inputs) {
@@ -73,6 +120,9 @@ const serializeFields = function (fields) {
 };
 
 const serializeBlock = function (block) {
+    const serializedPrimitive = serializePrimitiveBlock(block);
+    if (serializedPrimitive) return serializedPrimitive;
+    // If serializedPrimitive is null, proceed with serializing a non-primitive block
     const obj = Object.create(null);
     // obj.id = block.id; // don't need this, it's the index of this block in its containing object
     obj.opcode = block.opcode;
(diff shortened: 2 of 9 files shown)


thisandagain (approved):
This looks really good. Thanks for documenting this so thoroughly. 👍

towerofnix:
Out of curiosity's sake, what's the purpose of limiting a list to 200k items? Sure, that's an absurdly huge amount that I doubt will ever cause issues! But I'm curious anyways. Were gigantic lists causing problems?

kchadha:
@towerofnix That's a good question! In 3.0, we are changing the way we store projects on the server to allow bigger projects. There are currently some projects in 2.0 that have very long lists (some for games, but others just to test the limits of a project's size). We want to continue to support these projects, but given that we'll be able to save and load much bigger projects in the future, we don't want to also allow even bigger lists because this will increase browser memory usage while attempting to run one of these projects (or display a monitor for these lists, or scroll through the blocks flyout, etc.). Ultimately, this would cause more browser crashes, which we want to avoid. We're hoping that a 200k limit is a good balance where most projects that currently use large lists (games, etc.) still run well, but we're not actively giving way to crashing more user's browsers.

src/serialization/sb3.js
```diff
@@ -25,16 +27,157 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
+ */
 const serializeBlock = function (block) {
+    const serializedPrimitive = serializePrimitiveBlock(block);
+    if (serializedPrimitive) return serializedPrimitive;
+    // If serializedPrimitive is null, proceed with serializing a non-primitive block
     const obj = Object.create(null);
-    obj.id = block.id;
     obj.opcode = block.opcode;
+    // NOTE: this is extremely important to serialize even if null;
+    // not serializing `next: null` results in strange behavior
```
rschamp: This kind of comment is ominous — it'd be helpful to explain more specifically in case we we encounter it and need to know what to look for. Isn't the strange behavior that without `null`, the block ends up without a next connection?

thisandagain (approved):

src/serialization/sb3.js
```diff
@@ -25,16 +27,157 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
     obj.next = block.next;
     obj.parent = block.parent;
-    obj.inputs = block.inputs;
-    obj.fields = block.fields;
+    obj.inputs = serializeInputs(block.inputs);
+    obj.fields = serializeFields(block.fields);
     obj.topLevel = block.topLevel ? block.topLevel : false;
-    obj.shadow = block.shadow;
+    obj.shadow = block.shadow; // I think we don't need this either..
```
rschamp: Can we remove it? Otherwise it's gonna be here forever (see scratchr2 with comments posed as questions from like 5 years ago and now they're still there like creepy ghosts).  Also, curious — what's the other thing the "either" is referring to?

src/serialization/sb3.js
```diff
@@ -142,6 +412,227 @@ const serialize = function (runtime) {
...
+    case MATH_NUM_PRIMITIVE: {
+        primitiveObj.opcode = 'math_number';
+        primitiveObj.fields = {
+            NUM: {
+                name: 'NUM',
+                value: inputDescOrId[1]
+            }
+        };
+        primitiveObj.topLevel = false;
+        // what should we do about shadows
```
rschamp: A question for the ages!

src/sprites/rendered-target.js
```diff
@@ -215,9 +215,9 @@ class RenderedTarget extends Target {
      */
     static get VIDEO_STATE () {
         return {
-            'OFF': 'off',
-            'ON': 'on',
-            'ON-FLIPPED': 'on-flipped'
+            OFF: 'off',
+            ON: 'on',
+            ON_FLIPPED: 'on_flipped'
```
rschamp: I don't think we want this change, since we display `on-flipped` in the dropdown and don't want it to read `on_flipped`.

rschamp (changes requested):
Looks great overall, thanks for all the explanatory comments!

src/serialization/sb3.js
kchadha: Oops, forgot about that comment. We do need it, so I'm just removing the comment. =]

src/serialization/sb3.js
kchadha: hah! Another out of date comment ^_^

src/serialization/deserialize-assets.js
```diff
@@ -106,14 +98,11 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
...
     }
-    let dataFormat = null;
     let assetType = null;
     const costumeFormat = costume.dataFormat.toLowerCase();
     if (costumeFormat === 'svg') {
-        dataFormat = storage.DataFormat.SVG;
         assetType = storage.AssetType.ImageVector;
-    } else if (costumeFormat === 'png') {
-        dataFormat = storage.DataFormat.PNG;
+    } else if (['png', 'bmp', 'jpeg', 'jpg', 'gif'].indexOf(costumeFormat) >= 0) {
```
cwillisf, line 105: Suggestion for the future: use the `DataFormat` enum for this check. Possibly move it into a helper, depending on how often we end up needing something like this.

src/serialization/sb3.js
```diff
@@ -1,13 +1,15 @@
 /**
  * @fileoverview
- * Partial implementation of a SB3 serializer and deserializer. Parses provided
+ * An SB3 serializer and deserializer. Parses provided
```
cwillisf, line 3: 🎉

src/serialization/sb3.js
```diff
@@ -25,14 +27,156 @@ const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js'
...
+    // not serializing `next: null` results in strange behavior with block
+    // execution
     obj.next = block.next;
     obj.parent = block.parent;
-    obj.inputs = block.inputs;
-    obj.fields = block.fields;
+    obj.inputs = serializeInputs(block.inputs);
+    obj.fields = serializeFields(block.fields);
     obj.topLevel = block.topLevel ? block.topLevel : false;
     obj.shadow = block.shadow;
```
cwillisf, line 181: There are a few fields here that seem like they could be reconstructed at load-time. For example, it seems like we should only need one of inputs / fields / shadow -- although it also seems likely that I don't fully understand what each of these is for ;). Anyway... is the plan to pare this down later as we have time to make deserialization do more work?

src/serialization/sb3.js
```diff
@@ -69,17 +306,22 @@ const serializeCostume = function (costume) {
     // but that change should be made carefully since it is very
     // pervasive
     obj.md5ext = costume.md5;
-    obj.dataFormat = costume.dataFormat;
+    obj.dataFormat = costume.dataFormat.toLowerCase();
```
cwillisf, line 309: 👍

cwillisf (approved):
Looks good!

rschamp (approved):

src/serialization/sb3.js
kchadha: That was not the plan, and that's largely due to our timeline, the current state of the VM/blocks/gui, and the nature of the Save/Load work in general (i.e. that it is a high risk feature, where once it is released we can only make changes that are strictly more permissive; supporting a plan to changing how we serialize fields and inputs in the future would necessitate supporting both versions of sb3 serialization during validation).

In my mind, this is part of the decision to have the little bit of pre-compression work as proposed in this PR, but other than that, avoid a lot of extra (and error-prone) work to piece together info that is currently still scattered across scratch-vm, scratch-blocks, and scratch-gui  about how each block is defined and what kinds of inputs and fields it expects in order to be able to reconstruct it properly during deserialization.

The rest of this post is an attempt to explain the differences between inputs, fields, and shadows, and what kind of information they represent. You probably know most if not all of this already given extension-ification work, but it helps to have it all written down in one place. 

### Inputs, Fields, and Shadows Galore

A block's 'inputs' property describes each of the arguments on the block, naming them, and what is currently connected to those inputs. A block's field describes the value of a named field in that block.

#### Side note: 
In both cases, these properties are named, as opposed to positioned in the order they appear on the block. Information about what input/field names can exist in a given block, in what order, is scattered in several places. The extension-ification work will make it so all of that information is unified and can be found in one place accessible using the block's opcode, but unfortunately to access that information without the extension-ification work would require hand writing something like the sb2_specmap, but one which maps sb3 opcodes to its argmap.

#### Example of inputs and fields:

For example, a broadcast block has one input with the key 'BROADCAST_INPUT'. When a broadcast block's input is unobscured, the `inputs` field on the block will look something like this (in the runtime, unserialized):

```inputs: {
    'BROADCAST_INPUT': {
        'name': 'BROADCAST_INPUT',
        'block': 'this is a uid for a shadow block which contains the broadcast dropdown menu'
        'shadow: 'this is a uid for a shadow block which contains the broadcast dropdown menu'
    }
}
```

#### Block vs. Shadow in an Input

The fact that block and shadow in the example above have the same uid indicates that there the broadcast block has a pluggable input and that there is nothing plugged into the input.

- If this was a block without a droppable field, the 'shadow' property would be null.

- If there was something plugged in, the 'block' and 'shadow' properties would have different uids. The block uid would refer to whatever was plugged in, and the shadow in both this new case and the case in the example above refers to the shadow block which contains the drop down menu as a *field* (described below).

#### Fields

The shadow block with the uid `this is a uid for a shadow block which contains the broadcast dropdown menu` has opcode 'event_broadcast_menu' and has a fields property which contains a single field describing the currently selected value in the dropdown menu. It looks like the following:

```fields: {
        'BROADCAST_OPTION': {
            'name': 'BROADCAST_OPTION',
            'value': 'this is the name of the broadcast message I have selected in the dropdown menu',
            'variableType': 'broadcast_msg',
            'id': 'this is the uid for the variable that represents this broadcast message'
    }
}
```

This field, because it refers to a broadcast menu dropdown has special properties like id and variableType required for connecting the selected value to the correct broadcast message variable). The field in a 'math_number' block for example would have just the  'name' and 'value' properties.

In the example above, having both the input and the field are important, because the input describes the current state of the inputs on the block, and allows us to preserve the contents of underlying shadows e.g. what was selected (via a dropdown) or typed into a field before the user plugged something into it.

This was really long, but I hope it helps to clarify some of the concerns and also what information is available where.

src/serialization/sb3.js
kchadha: 😎

src/serialization/sb3.js
cwillisf: Thanks - that's actually very helpful! Most of my extension work focuses on defining a block at a time, in the abstract: I usually don't have to worry about blocks which are connected to each other or even instantiated. Basically, if there's a UID involved then I haven't worked with it much ;)

cwillisf (approved):
👍 to the sound change, too!

src/serialization/sb3.js
kchadha: ah, gotcha! That makes sense.

Honyant:
> @towerofnix That's a good question! In 3.0, we are changing the way we store projects on the server to allow bigger projects. There are currently some projects in 2.0 that have very long lists (some for games, but others just to test the limits of a project's size). We want to continue to support these projects, but given that we'll be able to save and load much bigger projects in the future, we don't want to also allow even bigger lists because this will increase browser memory usage while attempting to run one of these projects (or display a monitor for these lists, or scroll through the blocks flyout, etc.). Ultimately, this would cause more browser crashes, which we want to avoid. We're hoping that a 200k limit is a good balance where most projects that currently use large lists (games, etc.) still run well, but we're not actively giving way to crashing more user's browsers.

My problem with this is that my only way past this limit, which I need for like a video playback thing is to use a method that combines multiple lists, which is harder to manage and expand. So either way, it will crash the browser, so I think lists should still be given infinite length or at least be given an option to do so.

kchadha:
@Honyant, this pull request was merged almost a year ago, so any new conversation on here will not be easy to track. Please feel free to create an issue for your request.

Issue #355, opened 2016-12-17
Add support for deprecated blocks from 2.0 and 1.4


TheBrokenRail:
(Edited by @thisandagain on 12/24/2016)

https://wiki.scratch.mit.edu/wiki/List_of_Obsolete_Blocks

CosmicWebServices:
Umm Arent they trying to get rid of hacked block usage...

TheBrokenRail:
Yes, but just letting the blocks disappear (like they do currently), I don't believe is a good option, may be we should convert the hacked blocks into normal blocks (like forever if turned into a if block inside a forever block)

mrjacobbloom:
I like that. And counter could be named `counter1` or `counter2` etc. using the same method they use to resolve collisions with costume names

ed-cooper:
This will also need to include the old 1.4 style sound and drum blocks

CosmicWebServices:
I feel like if they add features that ATers usually have to hack the blocks for that will be key.

thisandagain:
I'm going to modify the scope the scope of this to be more clear. We need to make sure we support the deprecated opcodes from 1.X to 2.0:
https://wiki.scratch.mit.edu/wiki/List_of_Obsolete_Blocks

For "hacked" blocks, we will support a more robust extension paradigm in 3.0 to be discussed soon.

towerofnix:
See also #1030.

@thisandagain I assume the particular blocks wanted are those titled "Pseudo-Legacy" in [the "Opcode Hacking" section of the slideshow](https://docs.google.com/presentation/d/1z-g6p_oOcTXFeFWfTFo7HQVsi0Z7bnqVUl-mQK9YgQI/edit#slide=id.g3251eded0d_0_134)? Are there any others? I've noted a couple extras below anyhow.

---

For reference:

- [x] `doForLoop`: PR #963.
- [x] `doWhile`: PR #1032. [Its behavior](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/interpreter/Interpreter.as#L456) is exactly the opposite of `doUntil` (i.e. "repeat until"): while the condition is true (rather than false), run the contained blocks.
[...]

PR #1032, opened 2018-04-07
Implement "while" block

towerofnix:
### Resolves

Towards #355 (implements "doWhile"). Should be merged alongside LLK/scratch-blocks#1430.

### Proposed Changes

Adds a spec mapping from `doWhile` (2.0) to `control_while` (3.0) and implements the while block.

The `while` block behaves exactly opposite of `repeat until`: instead of looping *until* a condition is true, it loops *while* a condition is true, or, instead of looping while a condition is *false*, it loops while a condition is *true*.

(There was an incorrect comment in `repeatUntil` which I fixed in this PR.)

### Reason for Changes

Compatibility with Scratch 2.0 projects. This is a hacked block; in total, [1505 projects use the `while` loop](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752).

### Test Coverage

A new unit test (the same test as "repeat until" but checking that it repeats until a condition is false instead of true).

Also manually tested with this script (I got the "while" block from importing [this project](https://scratch.mit.edu/projects/49037046)):

Start a sound; while the timer is less than one, set the audio pitch proportional to the timer value

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -17,6 +17,7 @@ class Scratch3ControlBlocks {
         return {
             control_repeat: this.repeat,
             control_repeat_until: this.repeatUntil,
+            control_while: this.repeatWhile,
             control_for_each: this.forEach,
             control_forever: this.forever,
             control_wait: this.wait,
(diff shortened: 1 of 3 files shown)


kchadha (approved):
Looks good! Thanks for fixing the comment on repeatUntil. =]

PR #1033, opened 2018-04-07
Implement counter blocks

towerofnix:
### Resolves

Towards #355 (implements counter blocks). Should be merged alongside LLK/scratch-blocks#1431.

### Proposed Changes

Adds spec mappings from `COUNT`, `INCR_COUNT`, and `CLR_COUNT` to `control_get_counter`, `control_incr_counter`, and `control_clear_counter`, respectively.

Adds implementations for those blocks:

* `control_get_counter`: Return the counter value.
* `control_incr_counter`: Add one to the counter value.
* `control_clear_counter`: Set the counter value to zero.

### Reason for Changes

Compatibility with Scratch 2.0 projects. These are hacked blocks; in total, [around 1000 projects use the counter blocks](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752).

### Test Coverage

A new unit test for the blocks. Tested manually with [this project](https://scratch.mit.edu/projects/25344202/). (That project also uses the "while" block, which might cause issues opening it in this PR; I had my changes from #1032 active while working on this.)

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -7,6 +7,12 @@ class Scratch3ControlBlocks {
          * @type {Runtime}
          */
         this.runtime = runtime;
+
+        /**
+         * The "counter" block value. For compatibility with 2.0.
+         * @type {number}
+         */
+        this._counter = 0;
     }
 
     /**
@@ -25,7 +31,10 @@ class Scratch3ControlBlocks {
             control_if_else: this.ifElse,
             control_stop: this.stop,
             control_create_clone_of: this.createClone,
-            control_delete_this_clone: this.deleteClone
+            control_delete_this_clone: this.deleteClone,
+            control_get_counter: this.getCounter,
+            control_incr_counter: this.incrCounter,
+            control_clear_counter: this.clearCounter
         };
     }
 
@@ -145,6 +154,18 @@ class Scratch3ControlBlocks {
         if (util.target.isOriginal) return;
         this.runtime.disposeTarget(util.target);
         this.runtime.stopForTarget(util.target);
+    }
+
+    getCounter () {
+        return this._counter;
+    }
+
+    clearCounter () {
+        this._counter = 0;
+    }
+
+    incrCounter () {
+        this._counter++;
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_control.js
```diff
@@ -7,6 +7,12 @@ class Scratch3ControlBlocks {
          * @type {Runtime}
          */
         this.runtime = runtime;
+
+        /**
+         * The "counter" block value. For compatibility with 2.0.
+         * @type {number}
+         */
+        this._counter = 0;
```
kchadha, line 15: @cwillisf Does adding this here complicate extension-ifying control blocks? Having this stored here all the time seems weird to me given that these are not official blocks and [are not used very often](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752), but I'm not sure what's a better way to keep track of this value. Having it be a private variable is definitely a good step towards eliminating confusion, but is there a better place to store this info? Maybe somewhere in the runtime?

cwillisf: I wonder if it would make sense to define these blocks in a separate package, like "Legacy Control" or something. Then this could be a private variable in that blocks package and, in the future, that extension.
The good news is that all of this is fairly self contained, in that `_counter` is only manipulated by these new block implementation functions, so we can move it fairly easily later. My inclination would be to go with this as-is but talk about our options SOON (next product meeting?) and, if we decide to move these opcodes, try to get that done before we have people saving projects with these blocks :)
Another option is to add support for extensions which add blocks to existing categories. I don't think that would be difficult but I'm not sure it's a good idea.

kchadha: @cwillisf Thanks for the response! I think your first inclination of merge this as is but discuss the possibility of moving these SOON sounds like the best option (especially the bit about saving projects).

kchadha (approved):
Looks good!

src/blocks/scratch3_control.js
towerofnix: For reference, I noticed that the "sensing" blocks also store data like this. In particular, the "answer" block's output is [stored](https://github.com/LLK/scratch-vm/blob/d177f4cd0c316653b83c7547c001566523d7f374/src/blocks/scratch3_sensing.js#L12-L16) (and set by "ask"), which I figured is pretty much the same thing as storing the counter value.

PR #1038, opened 2018-04-10
Attach the costume size to the costume when loaded and modified

paulkaplan:
### Resolves

Required to complete LLK/scratch-gui#1754

### Proposed Changes

Attach the costume size to the costume when loaded and modified.

Also https://github.com/LLK/scratch-render/pull/254 includes an API change, changing the old `getSkinSize(drawableId)` to `getCurrentSkinSize(drawableId)` so that the new `getSkinSize(skinId)` could be used.

### Reason for Changes

The costume size is needed by the GUI in order to display costume meta data on the costume tiles.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -26,6 +26,7 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
     ];
     if (costumeAsset.assetType === AssetType.ImageVector) {
         costume.skinId = runtime.renderer.createSVGSkin(costumeAsset.decodeText(), rotationCenter);
+        costume.size = runtime.renderer.getSkinSize(costume.skinId);
         return costume;
     }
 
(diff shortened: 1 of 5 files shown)


kchadha (approved):
LG.

src/import/load-costume.js
```diff
@@ -50,6 +51,7 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
         imageElement.src = costumeAsset.encodeDataURI();
     }).then(imageElement => {
         costume.skinId = runtime.renderer.createBitmapSkin(imageElement, costume.bitmapResolution, rotationCenter);
+        costume.size = runtime.renderer.getSkinSize(costume.skinId);
```
cwillisf, line 54: Does this reliably get the right size when the image file isn't cached? I would expect the async nature of the `image` tag used to load the file would mean that, while the skin ID can reliably be allocated before the `create*Skin` methods return, the renderer doesn't necessarily know the image size just yet.

cwillisf (dismissed):
Marking this as "request changes" to discuss the async issue I commented on below. If that's not an issue then everything else looks good to me!

src/import/load-costume.js
paulkaplan: @cwillisf I think that because the imagedata has already been loaded in the previous promise, and the `_getBitmapSize` function in the renderer just uses `image.width/image.height`, I think the size will always be available at this time. I tried with a few bitmap skins and it looks like it is working right.

PR #1045, opened 2018-04-10
Fix import of broadcast block empty string obscured shadow import

### LINKED ISSUE #1044: Broadcast shadow import issue
This bug is best described with it's repro.

1) In Scratch 2.0, create a new project
2) Drag out a broadcast block
3) Plug something into the broadcast block and unplug it so that it creates an empty string message name
4) Plug something back into the broadcast block
5) Make sure this project is saved and import it into 3.0
6) Unplug the input from the broadcast block
7) Observe that the field reads `i` but it should read `message1`

This is an issue with sb2 import, and creates a problem with save and load because a variable is not created for the underlying broadcast shadow until the shadow is obscured. So if we saved the project before step 6 above, the project would not be able to load.

kchadha:
### Resolves

#1044

### Proposed Changes

Fix issue with importing broadcast block with obscured empty string shadow. (#1044)

### Reason for Changes

#1044 and blocker for save/load

### Test Coverage

Existing tests pass.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -639,18 +639,16 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             // event_broadcast_menus have some extra properties to add to the
             // field and a different value than the rest
             if (expectedArg.inputOp === 'event_broadcast_menu') {
-                if (!shadowObscured) {
-                    // Need to update the broadcast message name map with
-                    // the value of this field.
-                    // Also need to provide the fields[fieldName] object,
-                    // so that we can later update its value property, e.g.
-                    // if sb2 message name is empty string, we will later
-                    // replace this field's value with messageN
-                    // once we can traverse through all the existing message names
-                    // and come up with a fresh messageN.
-                    const broadcastId = addBroadcastMsg(fieldValue, fields[fieldName]);
-                    fields[fieldName].id = broadcastId;
-                }
+                // Need to update the broadcast message name map with
+                // the value of this field.
+                // Also need to provide the fields[fieldName] object,
+                // so that we can later update its value property, e.g.
+                // if sb2 message name is empty string, we will later
+                // replace this field's value with messageN
+                // once we can traverse through all the existing message names
+                // and come up with a fresh messageN.
+                const broadcastId = addBroadcastMsg(fieldValue, fields[fieldName]);
+                fields[fieldName].id = broadcastId;
                 fields[fieldName].variableType = expectedArg.variableType;
             }
             activeBlock.children.push({


cwillisf (approved):
The code looks fine, but I don't feel like I understand why or how this fixes the bug. Is it really OK to ignore the `shadowObscured` flag here? 🤷

rschamp (commented):
I think this makes sense — basically even if there's something in the input, we still want to import the underlying broadcast message.  My only concern would be if there are multiple blocks with obscured shadows, would this create too many messages?  It doesn't look like the addBroadcastMsg function has any guard against that?

kchadha:
@cwillisf That's a good question (and actually made me realize a side effect of this change which I talked through with @thisandagain -- documented below)!

#### Empty String Message Names

In the Scratch 2.0 editor, users can create broadcast messages with empty string names even though that's not explicitly allowed through the new message name prompt by adding broadcast block to their workspace, plugging something into the input, and then taking that block back out.

In 3.0, we decided to translate all empty string message names to a fresh message name messageN.

#### The Issue at Hand

The code block I've changed here, was originally checking all unobscured broadcast messages, and translating all empty string message names (if present) to a single new message with a fresh name messageN.

A new variable was created for each unobscured message name (and the translated empty string message name). The idea here is that in order for a broadcast message to actually do anything useful, you need to have a corresponding hat on the workspace that's waiting for that message. So it makes sense to only create variables for the unobscured broadcast fields (these appear on all three broadcast blocks).

However, since we never translated obscured message names, this results in a fresh name generated from scratch-blocks (hence seeing a broadcast message named `i` as documented in #1044), but no variable associated with that name. This new name shows up in all the dropdowns, but trying to click a broadcast block with that name (without a hat on the workspace) results in a non-fatal error in the console because there is no variable associated with that message.

#### The Proposed Solution in this PR

The changes in this PR make it so that we are translating all obscured shadows to a single fresh message name. Thus, when you uncover any of these broadcast fields in 3.0, there's a fresh messageN in the field (and there is an actual variable associated with it, so the block should work just fine).

#### The Side Effect

The side effect of this change is that if we have a 2.0 project like the following:

image

We only see one message in the dropdown (as expected):

image

When importing this project into 3.0 however, we get the following:

image

The extra message2 is there because of the obscured shadows getting there own variable, and there already existing a message1.

#### Conclusion

After talking all of this through with @thisandagain, we decided that this behavior is okay, and if we end up moving all variable control solely into the vm, we'll probably be able to fix this issue.

kchadha:
@rschamp The above comment should answer your question about obscured shadows (particularly the screenshots). The addBroadcastMsg function makes it so that all things with one message (even `''`) get treated as a single broadcast message. So multiple obscured shadows should only result in one total new message getting added.

cwillisf:
👍 thanks for the thorough explanation!

Issue #1039, opened 2018-04-10
Dragging behavior with touching sprite should match 2.0


paulkaplan:
### Expected Behavior

_Please describe what should happen_

Both of these sprites have a "forever if touching other sprite say hello" stack running
![dragging-behavior](https://user-images.githubusercontent.com/654102/38575985-c7645588-3cca-11e8-8e44-3c0cce58a653.gif)

### Actual Behavior

_Describe what actually happens_

![dragging-behavior--3-develop](https://user-images.githubusercontent.com/654102/38575993-d476fcbc-3cca-11e8-95d0-7ef3d9937ae7.gif)

It appears that the correct behavior (from 2.0) is that when a sprite is dragging, it is filtered from the other sprites "touching sprite" candidate list. This explains why abby says "hello!" but the cat does not, until you let go of abby.

AmazingMech2418:
This is because of the code for dragging. The drag function removes event listeners from scripts in the project from executing during the drag feature.

AmazingMech2418:
Oh. Nevermind. The dragged one shouldn't cause the other to say "Hello!".

AmazingMech2418:
I'm working on fixing it right now!

paulkaplan:
Fixed in https://github.com/LLK/scratch-vm/pull/1040, @amazinigmech2418 remember we are looking for PRs only on issues with the `help-wanted` tag. I didn't add the tag because i was already working on the fix.

AmazingMech2418:
OK. Sorry.

PR #1047, opened 2018-04-10
Drag Fix Part 2

AmazingMech2418:
### Resolves

https://github.com/LLK/scratch-vm/issues/1039

### Proposed Changes

This will apply the new array to the scratch-render.js file in the gh-pages branch.

### Reason for Changes

To only allow touching []? booleans if the candidate drawable isn't being dragged.
 
### Test Coverage

I unfortunately couldn't test it due to downloading errors for the vm.


paulkaplan:
See https://github.com/LLK/scratch-vm/issues/1039#issuecomment-380228904

Issue #1056, opened 2018-04-13
Pen "clear" may be mistaken to mean "transparent"


thisandagain:
The word "clear" for pen is even more ambiguous than it already was. One possibility is to use  "clear all" within these blocks:

1) clear graphic effects, 2) clear sound effects, 3) clear 
would be:
1) clear all graphic effects, 2) clear all sound effects, 3) clear all [within Pen]

![image](https://user-images.githubusercontent.com/4721032/38574439-d18e6f34-3cc6-11e8-8476-24aaeef3d1c8.png)

---
Moved from LLK/scratch-gui#1756
/cc @ntlrsk

PR #1057, opened 2018-04-13
Rename pen 'clear' to 'erase all'

thisandagain:
### Resolves

Resolves GH-1056

### Proposed Changes

- Changes pen "clear" to "erase all"

### Reason for Changes

- There is a sense that the word "clear" may be confusing / ambiguous. As per discussion we decided to narrow the scope of this change to just the pen block.

### Note

- I only changed the block text and not the underlying opcode naming. This was intentional until we are sure that this change will "stick".

/cc @ntlrsk


ericrosenbaum (approved):
LG!

Issue #929, opened 2018-02-02
Figure out why push reported work is causing bugs


kchadha:
This is a follow up issue from #924 to figure out what is causing the regression from the push reported work. This work is being reverted in #928.

Below are examples of projects that fail due to this work, specifically [this commit](https://github.com/LLK/scratch-vm/commit/6b7582f1c70ce4f90a0ca5eb986d96771414d86b). 

https://llk.github.io/scratch-gui/develop/#10128407
https://llk.github.io/scratch-gui/develop/#10761768
https://llk.github.io/scratch-gui/develop/#56667360
https://llk.github.io/scratch-gui/develop/#21554369

mzgoddard:
I found what was not working. Procedures and probably other blocks that pass booleans without a block like
![screen shot 2018-04-16 at 5 30 01 pm](https://user-images.githubusercontent.com/833298/38836402-01b35240-419c-11e8-9bb4-aec17df293bf.png)
return undefined from the empty boolean argument. This and some behaviour in the pushReported optimization caused procedures with undefined boolean arguments to get a reported value from another previously executed block. That value being truthy would case the procedure to behaviour differently.

The above example would pass 7 instead of undefined to the procedure call. One way to look at how this result came about may be that booleans don't have a shadow block or return false as a implied shadow value.

Fixing this in the pushReported optimization isn't going to be a lot of code. There is not a clear testing idiom though to test for another regression against this.

PR #1065, opened 2018-04-16
Push reported again. Now with empty boolean support!

mzgoddard:
### Resolves

Fix  https://github.com/LLK/scratch-vm/issues/929

### Proposed Changes

Clear pushReported's optimization `justReported` member after reading it in `execute.js`. With `justReported` cleared, the value can't leak to other members of argValues to execute this stack frame's blockFunction with.

### Reason for Changes

Previously reported values leaked into boolean arguments that did not define a block. If those values were truthy that would change the execution behaviour when the scratch project's author intended it to be falsey.

### Test Coverage

There isn't a clear idiom how to test that the vm executes a block or procedure correctly. There are tests that test the specific block code in `src/blocks` and integration tests to see that the vm runs with those blocks. But not that the end or partial state of the integration test matches expected values. Anyways test coverage will be provided by another PR that will be easier to merge than this one as it'll just test the existing behaviour to ensure this specific regression doesn't occur in the future. (And this PR can provide the oppurtunity to discuss how to test this.)

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -1,4 +1,5 @@
 const BlockUtility = require('./block-utility');
+const BlocksExecuteCache = require('./blocks-execute-cache');
 const log = require('../util/log');
 const Thread = require('./thread');
 const {Map} = require('immutable');
@@ -99,31 +100,26 @@ const handleReport = function (
 };
 
 /**
+ * A convenienve constant to hide that the recursiveCall argument to execute is
+ * a boolean trap.
+ * @const {boolean}
+ */
+const RECURSIVE = true;
+
+/**
  * Execute a block.
  * @param {!Sequencer} sequencer Which sequencer is executing.
  * @param {!Thread} thread Thread which to read and execute.
+ * @param {boolean} recursiveCall is execute called from another execute call?
  */
-const execute = function (sequencer, thread) {
+const execute = function (sequencer, thread, recursiveCall) {
     const runtime = sequencer.runtime;
-    const target = thread.target;
-
-    // Stop if block or target no longer exists.
-    if (target === null) {
-        // No block found: stop the thread; script no longer exists.
-        sequencer.retireThread(thread);
-        return;
-    }
 
     // Current block to execute is the one on the top of the stack.
     const currentBlockId = thread.peekStack();
     const currentStackFrame = thread.peekStackFrame();
 
-    let blockContainer;
-    if (thread.updateMonitor) {
-        blockContainer = runtime.monitorBlocks;
-    } else {
-        blockContainer = target.blocks;
-    }
+    let blockContainer = thread.blockContainer;
     let block = blockContainer.getBlock(currentBlockId);
     if (typeof block === 'undefined') {
         blockContainer = runtime.flyoutBlocks;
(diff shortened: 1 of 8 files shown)


mzgoddard:
One thought I have for testing that block execution happens as expect could be to connect to the runtime `SAY` event in integration tests. Says that start with `pass` or `fail` run `t.pass` or `t.fail`.

Using the procedure integration as an example:
screen shot 2018-04-16 at 5 58 38 pm

Integrations that have more assertions could wrap that as they want with procedures if the integration isn't on procedure behaviour itself.
screen shot 2018-04-16 at 5 56 26 pm

I figure integration tests like this would be a new folder in the test folder and would have a js file automatically run all the sb2 and sb3 files in that folder perhaps.

kchadha (commented):
The new proposed changes (compared to the old push reported work) look good.

It did take a while to re-understand what's going on in execute. I'm wondering if there should be a high level comment explaining what's happening in execute in general. 

For example, I _think_ we know that there's never going to be an input and a field together, a block always has either/or, so it would help to have a description of 1) what's happening at a high level -- e.g. what information we need to look for/check for and 2) what is expected to be cached in any given run of execute (recursive or not).

kchadha:
As for testing, I do think we need to have some - are the blocks being interpreted as we think they should be - tests. Sort of going along with your 'say' idea, we could also have tests where stacks of blocks change one or more variables, and the test makes assertions about the value of the variable(s) at the end of running a stack. Of course, this would necessitate a test that ensures that variables are working correctly.

mzgoddard:
@kchadha I added a comment https://github.com/LLK/scratch-vm/pull/1065/commits/19a32cd4898a9e9193362cd5a2fed8694280e543 for the BlocksExecuteCache area in execute and some other new comments.

mzgoddard:
I'm going to work on and make a separate PR for the test that could be reviewed and merged separately.

src/engine/execute.js
```diff
@@ -98,32 +99,27 @@ const handleReport = function (
     }
 };
 
+/**
+ * A convenienve constant to hide that the recursiveCall argument to execute is
```
cwillisf: s/convenienve/convenience/

src/engine/execute.js
```diff
@@ -136,33 +132,97 @@ const execute = function (sequencer, thread) {
...
+        blockCached._definedBlockFunction = typeof blockCached._blockFunction !== 'undefined';
 
-    if (!opcode) {
-        log.warn(`Could not get opcode for block: ${currentBlockId}`);
-        return;
+        const fieldKeys = Object.keys(fields);
+
+        // Store the current shadow value if there is a shadow value.
+        blockCached._isShadowBlock = fieldKeys.length === 1 && Object.keys(inputs).length === 0;
+        blockCached._shadowValue = fieldKeys.length === 1 && fields[fieldKeys[0]].value;
```
cwillisf: Should this be:
```js
blockCached._shadowValue = blockCached._isShadowBlock && fields[fieldKeys[0]].value;
```
?

src/engine/execute.js
```diff
@@ -173,37 +233,51 @@ const execute = function (sequencer, thread) {
...
-            fieldName === 'BROADCAST_OPTION') {
-            argValues[fieldName] = {
-                id: fields[fieldName].id,
-                name: fields[fieldName].value
-            };
-        } else {
+    // Add all fields on this block to the argValues. Some known fields may
+    // appear by themselves and can be set to argValues quicker by setting them
+    // explicitly.
+    if (blockCached._isFieldKnown) {
```
cwillisf: It looks like, with minor modification above, this could be simplified to something like:
```js
switch (blockCached._fieldKind) {
case foo.FIELD_VARIABLE:
    argValues.VARIABLE = blockCached._fieldVariable;
    break;
// similar cases for list & broadcast
default:
    for (const fieldName in fields) {
...
```

src/engine/execute.js
```diff
@@ -212,7 +286,26 @@ const execute = function (sequencer, thread) {
             currentStackFrame.waitingReporter = null;
             thread.popStack();
         }
-        const inputValue = currentStackFrame.reported[inputName];
+
+        let inputValue;
+        if (
+            currentStackFrame.waitingReporter === null
+        ) {
```
cwillisf: extra newlines above

cwillisf (changes requested):
I'm marking as "request changes" just for some weird formatting and a typo; the code looks ok to me. I mean, it does make me a bit nervous but I'm glad to see the updated tests as well as #1067. ;)

I did have a couple other suggestions & questions, but I don't think any of them are blocking.

mzgoddard:
@cwillisf I modified code in response to your review.

cwillisf (approved):

PR #1067, opened 2018-04-18
Boolean shadow regression test

mzgoddard:
### Resolves

How to regression test the issue that lead to the push reported optimization being reverted. https://github.com/LLK/scratch-vm/pull/1065#issuecomment-381764465

### Proposed Changes

Use a group of integration tests with the test work specified in scratch projects using SAY blocks.

- SAY "plan NUMBER_OF_TESTS" calls `t.plan(Number(NUMBER_OF_TESTS))` in javascript
- SAY "pass MESSAGE" calls `t.pass(MESSAGE)`
- SAY "fail MESSAGE" calls `t.fail(MESSAGE)`
- SAY "end" calls `t.end()`

The integration test handling supports mapping these SAY messages to tap api calls to record the test result. Each test runs until the runtime has no running threads. Tap's `tearDown` is used to clean up the virtual machine interval so the test and later the process running the tests may naturally exit.

### Reason for Changes

The block `test/unit` tests directly test the javascript functions for each blocks. As a unit test they do not (and should not) test the block function by calling it through `execute`.

The `test/integration` tests are less specific than what may eventually be a 1-to-1 set of tests to block combinations testing execution of block sets. The `test/integration` can handle larger scope tests outside of making sure block results are what is intended.

### Test Coverage

This PR is test coverage! And test coverage for the push reported optimization regression.


package.json
```diff
@@ -20,9 +20,10 @@
     "lint": "eslint . && format-message lint src/**/*.js",
     "prepublish": "in-publish && npm run build || not-in-publish",
     "start": "webpack-dev-server",
-    "tap": "tap ./test/{unit,integration}/*.js",
+    "tap": "tap ./test/{unit,integration,execute}/*.js",
     "tap:unit": "tap ./test/unit/*.js",
-    "tap:integration": "tap ./test/integration/*.js",
+    "tap:execute": "tap ./test/execute/*.js",
```
mzgoddard: I'm not sure if execute should have its own `tap:*` command. Should execute tests be inside the integration folder somehow?

test/execute/index.js
```diff
@@ -0,0 +1,151 @@
...
+                    }
+                });
+            };
+
+            // Listen to SAY events and match the beginning of SAY text's to
+            // basic tap methods.
+            let didPlan;
+            let didEnd;
+            vm.runtime.on('SAY', (target, type, text) => {
+                if (text.toLowerCase().startsWith('pass')) {
```
gnarf: random thoughts, but after I'm done writing this, I don't think it's any more/less readable -- I kind of hate the repeat of toLowerCase for every one of the ifs here.

```js
const sayCommands = {
  comment(reason) { t.comment(reason); },
  pass(reason) { t.pass(reason); },
  fail(reason) { t.fail(reason); },
  plan(reason) {
    didPlan = true;
    t.plan(Number(reason) + 1);
  },
  end() {
    didEnd = true;
    if (!didPlan) {
      testVmError();
    }
    t.end();
  },
};

const sayHandler = (target, type, text) => {
  const [command] = text.toLowerCase().split(/\s+/, 1);
  if (sayCommands[command]) {
    return sayCommands[command](text.substring(command.length));
  }

  // default to a comment with the full text if we didn't match any command prefix
  return sayCommands.comment(text);
};

File: test/execute/index.js [hunk 3]
```diff
@@ -0,0 +1,151 @@
...
+                vm.setTurboMode(false);
+
+                // Stop the runtime interval once the test is complete so the
+                // test process may naturally exit.
+                t.tearDown(() => {
+                    clearInterval(vm.runtime._steppingInterval);
+                });
+
+                loadPromise = vm.loadProject(project)
+                    .then(() => vm.greenFlag())
```
gnarf: I think we could move this down to after the `try` block and attach it to the `return loadPromise.` because this is really `wholeTestDonePromise` now...

test/execute/index.js
```diff
@@ -0,0 +1,151 @@
...
+                    .then(() => (
+                        // When the number of threads reaches 0 the test is
+                        // expected to be complete.
+                        new Promise((resolve, reject) => {
```
gnarf: I almost want to see this turned into a helper, it feels super deep in the indentation at this point.

```js
.then(() => whenThreadsComplete(vm, t))
```

test/execute/index.js
mzgoddard: In that case we could just get rid of the try block and t.doesNotThrow tests and use the tap's default error reporting I guess.

mzgoddard:
Thanks for the review @gnarf

test/execute/index.js
```diff
@@ -0,0 +1,136 @@
...
+    .filter(uri => uri.endsWith('.sb2'))
+    .forEach(uri => {
+        test(uri, t => {
+            // Disable logging during this test.
+            log.suggest.deny('vm', 'error');
+            t.tearDown(() => log.suggest.clear());
+
+            // Map string messages to tap reporting methods. This will be used
+            // with events from scratch's runtime emitted on block instructions.
+            let didPlan;
```
gnarf: nit - can we default to `false` instead of falsey for these two?

thisandagain (commented):
This looks great @mzgoddard. I agree that it may be better to integrate these into the existing `integration` test directory if reasonable.

kchadha (approved):
LG

rschamp (changes requested):
This PR is crazy and I love it.

Even though they are different from most of the existing integration tests, I still think the execute tests should be run as part of the set of integration tests if possible.  Good to go after that I think.

package.json
```diff
@@ -20,7 +20,7 @@
     "lint": "eslint . && format-message lint src/**/*.js",
     "prepublish": "in-publish && npm run build || not-in-publish",
     "start": "webpack-dev-server",
-    "tap": "tap ./test/{unit,integration}/*.js",
+    "tap": "tap ./test/{unit,integration,execute}/*.js",
```
kchadha: looks like that extra reference to execute should go away now

rschamp (approved):

PR #1069, opened 2018-04-20
Replace extension delimiter with '_'

thisandagain:
### Proposed Changes

Replaces the `.` delimiter for all extensions with `_`.

### Reason for Changes

To ensure compatibility as blocks migrate from their current implementation to "core" extensions in the future.

### Test Coverage

Tests updated to reflect the change.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -19,14 +19,12 @@ const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
 const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js');
 
-// Constants used during deserialization of an SB3 file
+// Constants used during deserialization of an SB2 file
 const CORE_EXTENSIONS = [
-    'argument',
     'control',
     'data',
     'event',
     'looks',
-    'math',
     'motion',
     'operator',
     'procedures',
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -521,7 +521,7 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
     const index = blockMetadata.opcode.indexOf('_');
     const prefix = blockMetadata.opcode.substring(0, index);
     if (CORE_EXTENSIONS.indexOf(prefix) === -1) {
-        extensions.extensionIDs.add(prefix);
+        if (prefix !== '') extensions.extensionIDs.add(prefix);
     }
 
     // Block skeleton.
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -19,6 +19,21 @@ const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
 const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js');
 
+// Constants used during deserialization of an SB3 file
+const CORE_EXTENSIONS = [
+    'argument',
+    'control',
+    'data',
+    'event',
+    'looks',
+    'math',
+    'motion',
+    'operator',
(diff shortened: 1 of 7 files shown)


src/serialization/sb2.js
```diff
@@ -19,6 +19,21 @@ const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
 const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js');
 
+// Constants used during deserialization of an SB3 file
+const CORE_EXTENSIONS = [
```
thisandagain: Ideally this would live outside of serialization and look at the actual "core" opcodes, but that would require inspecting the runtime and exposing `defaultBlockPackages` which would be fairly invasive.

src/serialization/sb2.js
```diff
@@ -19,6 +19,21 @@ const {loadCostume} = require('../import/load-costume.js');
...
+    'argument',
+    'control',
+    'data',
+    'event',
+    'looks',
+    'math',
```
thisandagain, line 29: `math` is required for shadow blocks

src/serialization/sb2.js
```diff
@@ -19,6 +19,21 @@ const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
 const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js');
 
+// Constants used during deserialization of an SB3 file
+const CORE_EXTENSIONS = [
+    'argument',
```
thisandagain: `argument` is required for custom procedures

src/serialization/sb2.js
```diff
@@ -19,6 +19,21 @@ const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
 const {deserializeCostume, deserializeSound} = require('./deserialize-assets.js');
 
+// Constants used during deserialization of an SB3 file
```
kchadha: SB2 File?

src/serialization/sb2.js
```diff
@@ -501,12 +516,14 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
...
+
     // If the block is from an extension, record it.
-    const dotIndex = blockMetadata.opcode.indexOf('.');
-    if (dotIndex >= 0) {
-        const extension = blockMetadata.opcode.substring(0, dotIndex);
-        extensions.extensionIDs.add(extension);
+    const index = blockMetadata.opcode.indexOf('_');
+    const prefix = blockMetadata.opcode.substring(0, index);
+    if (CORE_EXTENSIONS.indexOf(prefix) === -1) {
+        extensions.extensionIDs.add(prefix);
```
kchadha: This seems like the opposite of what this piece was previously doing -- am I misunderstanding something? We're only adding the extension if it's not one of the core?

thisandagain:
A good project for testing:
http://localhost:8601/#217651758

src/serialization/sb2.js
thisandagain: From conversation: This checks to see if the opcode prefix is core. If it is *not* core, then it attempts to load an extension using the opcode prefix.

kchadha:
LG!

cwillisf:
As a follow-up I'd eventually like to combine the (now) 3 lists of core extensions, but this looks fine to me for now -- especially as a way to stop making the problem worse :)

Once we start working on extensionification again we'll probably need two lists: one for extensions that we want to auto-load, and another for "extensions" which should be skipped because they're not extensions yet. Ideally these two lists should be in one place, shared by both SB2 and SB3. I'm OK with making this part of the extensionification (re)startup cost.

PR #1071, opened 2018-04-20
Fix issue with sound upload and save/load

kchadha:
### Resolves

Uploading an mp3 sound, editing it in the sound editor, and then saving the project and uploading that saved project wasn't working.

### Proposed Changes

Fix for the issue described above.

### Test Coverage

Existing tests pass.


paulkaplan (approved):
LG

PR #1075, opened 2018-04-20
Load costume without metadata

kchadha:
### Resolves

Towards resolving LLK/scratch-gui#1777

### Proposed Changes

Allows for loading costume data without having associated metadata like resolutionCenterX/Y and bitmapResolution. If resolutionCenter is not provided, gets it from the renderer after the costume has been loaded there. This makes it so that the paint editor can also make use of this value in determining where to place the freshly loaded costume.

### Reason for Changes

LLK/scratch-gui#1777, also in support of: LLK/scratch-gui#1836 and depends on LLK/scratch-render#260

### Test Coverage

Existing tests pass.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -69,8 +69,8 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
 
         if (!rotationCenter) {
             rotationCenter = renderer.getSkinRotationCenter(costume.skinId);
-            costume.rotationCenterX = rotationCenter[0];
-            costume.rotationCenterY = rotationCenter[1];
+            costume.rotationCenterX = rotationCenter[0] * 2;
+            costume.rotationCenterY = rotationCenter[1] * 2;
         }
         return costume;
     });
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -15,7 +15,8 @@ const log = require('../util/log');
  */
 const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
     costume.assetId = costumeAsset.assetId;
-    if (!runtime.renderer) {
+    const renderer = runtime.renderer;
+    if (!renderer) {
         log.error('No rendering module present; cannot load costume: ', costume.name);
         return costume;
     }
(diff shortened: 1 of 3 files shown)


src/import/load-costume.js
```diff
@@ -15,18 +15,31 @@ const log = require('../util/log');
...
         log.error('No rendering module present; cannot load costume: ', costume.name);
         return costume;
     }
     const AssetType = runtime.storage.AssetType;
-    const rotationCenter = [
-        costume.rotationCenterX / costume.bitmapResolution,
-        costume.rotationCenterY / costume.bitmapResolution
-    ];
+    let rotationCenter;
+    if (costume.rotationCenterX && costume.rotationCenterY && costume.bitmapResolution) {
```
cwillisf, line 25: Is there actually a way to get in here without a valid `bitmapResolution`? If so, see below; if not, I would remove it from this test to avoid confusion.

src/import/load-costume.js
```diff
@@ -15,18 +15,31 @@ const log = require('../util/log');
...
+        // createSVGSkin does the right thing if rotationCenter isn't provided, so it's okay if it's
+        // undefined here
+        costume.skinId = renderer.createSVGSkin(costumeAsset.decodeText(), rotationCenter);
+        costume.size = renderer.getSkinSize(costume.skinId);
+        // Now we should have a rotationCenter even if we didn't before
+        if (!rotationCenter) {
+            rotationCenter = renderer.getSkinRotationCenter(costume.skinId);
+            costume.rotationCenterX = rotationCenter[0];
+            costume.rotationCenterY = rotationCenter[1];
+        }
```
cwillisf, line 41: If it's possible to get into this function without a `bitmapResolution`, should we read back the bitmap resolution here too? Alternatively, should we `delete costume.bitmapResolution` before leaving this function? Right now it looks like some costumes will leave this function with a `bitmapResolution` property and some will leave this function without. (Same comment applies to the other case, below.)

cwillisf (changes requested):
I like the change overall, but I do have a minor concern about the `bitmapResolution` property.

src/import/load-costume.js
kchadha: It is possible, via file upload.

src/import/load-costume.js
kchadha: Isn't bitmap resolution inapplicable for SVG images? That's why it doesn't get read back out here. IMO, it's okay to leave without the bitmapResolution if it wasn't passed in. The renderer and other places in the VM handle this correctly, and the paint editor is doing away with bitmap resolution... Eventually, I want a refactor where we don't do things with bitmapResolution anywhere except in sb2/.sprite2 import...

cwillisf: Oh, right! Ha :p
I guess my comment only applies to the bitmap branch below, then.

> The renderer and other places in the VM handle this correctly, and the paint editor is doing away with bitmap resolution... Eventually, I want a refactor where we don't do things with bitmapResolution anywhere except in sb2/.sprite2 import...

👍

cwillisf (approved):

PR #1082, opened 2018-04-24
Add updateBitmap function. getCostume can now serve bitmaps as well as svgs

fsih:
### Resolves
Depends on https://github.com/LLK/scratch-render/pull/264
Good news, this can now go in as soon as it's ready!

Progress towards support for exporting bitmaps from the paint editor.

### Proposed Changes
Adds updateBitmap, which is analogous to updateSVG. It sends an updated bitmap from the paint editor to the renderer and storage.

Also adds support for retrieving bitmap costumes to getCostume


src/virtual-machine.js
```diff
@@ -542,17 +543,59 @@ class VirtualMachine extends EventEmitter {
...
+     */
+    updateBitmap (costumeIndex, bitmap, rotationCenterX, rotationCenterY) {
+        const costume = this.editingTarget.getCostumes()[costumeIndex];
+        if (costume && this.runtime && this.runtime.renderer) {
+            costume.rotationCenterX = rotationCenterX;
+            costume.rotationCenterY = rotationCenterY;
+            this.runtime.renderer.updateBitmapSkin(costume.skinId, bitmap, [rotationCenterX, rotationCenterY]);
+        }
+
+        bitmap.toBlob(blob => {
```
fsih: toBlob is not supported in Safari and Edge

fsih: Fixed with polyfill

fsih: Soooo the polyfill broke all tests because it references window

fsih: I found a better polyfill

fsih: 05-pet-bird-001

src/virtual-machine.js
```diff
@@ -542,17 +544,59 @@ class VirtualMachine extends EventEmitter {
...
-    getCostumeSvg (costumeIndex) {
+    getCostume (costumeIndex) {
         const id = this.editingTarget.getCostumes()[costumeIndex].assetId;
-        if (id && this.runtime && this.runtime.storage &&
-                this.runtime.storage.get(id).dataFormat === 'svg') {
+        if (!id || !this.runtime || !this.runtime.storage) return null;
+        const format = this.runtime.storage.get(id).dataFormat;
+        if (format === this.runtime.storage.DataFormat.SVG) {
             return this.runtime.storage.get(id).decodeText();
+        } else if (format === this.runtime.storage.DataFormat.PNG) {
```
kchadha: The image can also be a .jpg, so you can just have an `else` here, and then not override the MIME type in your call to `storage.get` below.

src/virtual-machine.js
```diff
@@ -542,17 +544,59 @@ class VirtualMachine extends EventEmitter {
...
+        bitmap.toBlob(blob => {
+            const reader = new FileReader();
+            reader.addEventListener('loadend', () => {
+                const storage = this.runtime.storage;
+                costume.assetId = storage.builtinHelper.cache(
+                    storage.AssetType.ImageBitmap,
+                    storage.DataFormat.PNG,
+                    Buffer.from(reader.result)
+                );
+                // @todo is there a better way to make sure all info is up to date on the costume?
```
kchadha: I think you can use `bitmap.getContext('2d').getImageData(0, 0, width, height).data` where width and height are the width and height of the expected image -- instead of calling toBlob and using a FileReader. At any rate, instead of using `Buffer.from` above, you can just use `new Uint8Array` above.

kchadha (changes requested):
getCostume should be able to support jpeg images, and some suggestions for updateBitmap...

src/virtual-machine.js
fsih: <strike>I'm not sure what you mean by overriding a mime type?</strike>

Oh wait, I see encodeDataURI's argument is optional. Sorry, I'm not familiar with these data terms. I've had quite a time trying to understand the difference between a buffer and an array buffer and a uint 8 array that is also a buffer.

fsih: I think I'd rather be explicit about which file types I handle here, so that there is a nice error if there is a type I haven't handled.

kchadha: Sounds good, maybe just add an || then and take out the optional argument in the storage function call.

src/virtual-machine.js
fsih: Thanks!! That's much easier

kchadha (approved):
The changes look good to me!

src/virtual-machine.js
```diff
@@ -543,17 +543,57 @@ class VirtualMachine extends EventEmitter {
     /**
      * Get an SVG string from storage.
```
cwillisf: This comment needs a minor update :)

src/virtual-machine.js
```diff
@@ -543,17 +543,57 @@ class VirtualMachine extends EventEmitter {
...
      * @param {int} costumeIndex - the index of the costume to be got.
-     * @return {string} the costume's SVG string, or null if it's not an SVG costume.
+     * @return {string|HTMLImageElement} the costume's SVG string if it's SVG,
+     *     an HTMLImageElement if it's a PNG, or null if it couldn't be found or decoded.
```
cwillisf: Is `HTMLImageElement` the best return type? `ImageData` is another option, for example, but I'm not sure which would be "best" to return from this method.

src/virtual-machine.js
```diff
@@ -543,17 +543,57 @@ class VirtualMachine extends EventEmitter {
...
+        if (!id || !this.runtime || !this.runtime.storage) return null;
+        const format = this.runtime.storage.get(id).dataFormat;
+        if (format === this.runtime.storage.DataFormat.SVG) {
             return this.runtime.storage.get(id).decodeText();
+        } else if (format === this.runtime.storage.DataFormat.PNG ||
+                format === this.runtime.storage.DataFormat.JPG) {
+            const data = this.runtime.storage.get(id).encodeDataURI();
+            const image = new Image();
+            image.src = data;
+            return image;
```
cwillisf: Technically speaking, the image might not be ready by the time this function returns. In order to guarantee compatibility and avoid race conditions, I would recommend adjusting `getCostume` to return a `Promise` which you'd `resolve` with the return value in the image's `onload` callback. This would also give you a chance to `reject` the promise in the image's `onerror` callback. That'll probably make callers more complex, but I don't think the current code is guaranteed to be safe.

cwillisf (changes requested):
Most of this looks good to me, but I'm concerned about a possible race condition in the bitmap path of `getCostume`. See comments below...

src/virtual-machine.js
kchadha: Which one of these calls is async? encodeDataURI?

paulkaplan: That would be fine, but it also seems like just returning the encoded data URI, which is sync, would be fine too. It would also more closely parallel the SVG return (also string, also needs to be applied to an image tag to view, etc.)

src/virtual-machine.js
```diff
@@ -543,17 +543,57 @@ class VirtualMachine extends EventEmitter {
...
         }
+        log.error(`Unhandled format: ${this.runtime.storage.get(id).dataFormat}`);
         return null;
     }
 
+    /**
+     * Update a costume with the given bitmap
+     * @param {int} costumeIndex - the index of the costume to be updated.
+     * @param {HTMLCanvasElement} bitmap - new bitmap for the renderer.
```
paulkaplan: If possible I think limiting the amount of DOM stuff in the VM would be good. Passing in an [`ImageData`](https://developer.mozilla.org/en-US/docs/Web/API/ImageData) object (or a vanilla JS object with the same properties) is how this type of data is interfaced from the renderer. 

Is it possible to have the caller do the `getImageData` and passing the result instead of the canvas element? The ImageData also includes width/height properties.

src/virtual-machine.js
cwillisf: `image.src = data;` is async in that, in general, it may make a network request before the image becomes populated. I'm not aware of a special-case guarantee that a data URI will be decoded synchronously, but if that's guaranteed then there's no problem here. I assume this code has been tested successfully at least once, which suggests that I might be missing something ;)

Other possibilities: it might be that decoding a data URI is unofficially synchronous on some browsers, or it might just be quick enough for smallish images on faster computers.

kchadha: @cwillisf that makes sense, had a similar conversation with @paulkaplan offline as well, thank you both for explaining =]

src/virtual-machine.js
fsih: Actually, context.getImageData(0, 0, width, height).data isn't working for me. When I load the resulting asset and call encodeDataURI on it, and put that in an Image, the image's onload function is never called and its onerror function is always called.

fsih: Here are the data URIs that come out in case that helps in figuring out what went wrong. The one coming from imagedata.data seems to be infested with 'A's and forward slashes.

[imagedatadata.txt](https://github.com/LLK/scratch-vm/files/1961189/imagedatadata.txt)
[readerbufferdata.txt](https://github.com/LLK/scratch-vm/files/1961191/readerbufferdata.txt)

kchadha: @fsih I think `<all that canvas context stuff>.data.buffer` will work. 
If you're getting the uri anyways, you can use canvas.toDataURL('image/png') instead.

src/virtual-machine.js
```diff
@@ -541,19 +541,64 @@ class VirtualMachine extends EventEmitter {
...
+        const storage = this.runtime.storage;
+        costume.assetId = storage.builtinHelper.cache(
+            storage.AssetType.ImageBitmap,
+            storage.DataFormat.PNG,
+            bitmap.data
+        );
+        costume.dataFormat = storage.DataFormat.PNG;
+        costume.bitmapResolution = bitmapResolution;
+        // costume.size = [bitmap.width, bitmap.height];
+        // costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
```
kchadha: These should not be commented out.

src/virtual-machine.js
fsih: imageData.data.buffer isn't working for me either u_u the entire returned dataURI is "data:image/png;base64"

Every method I've tried so far via image data has given me back an unloadable dataURI including:
Buffer.from(imagedata.data)
new Uint8Array(imagedata.data)
imagedata.data

src/virtual-machine.js
fsih: Done

src/virtual-machine.js
fsih: Passed in a data URI

src/virtual-machine.js
fsih: Passed out ImageData. Couldn't get the image data to go into storage though without converting to canvas first still

src/virtual-machine.js
fsih: Done

src/virtual-machine.js
```diff
@@ -541,19 +543,77 @@ class VirtualMachine extends EventEmitter {
...
+            reader.addEventListener('loadend', () => {
+                const storage = this.runtime.storage;
+                costume.assetId = storage.builtinHelper.cache(
+                    storage.AssetType.ImageBitmap,
+                    storage.DataFormat.PNG,
+                    Buffer.from(reader.result)
+                );
+                costume.dataFormat = storage.DataFormat.PNG;
+                costume.bitmapResolution = bitmapResolution;
+                costume.size = [bitmap.width, bitmap.height];
```
paulkaplan, line 609: This `size` is used by the GUI to display the width/height of the costume. Do you think it should be scaled here by the bitmap resolution, or should the GUI scale it by bitmap resolution when displaying that info?

paulkaplan: I'm fine with having the GUI deal with it, just wondering if you had an opinion. This behavior _seems_ wrong (watch the size in the costume tile)

size-switching

fsih: Ohh yes you're right, I should fix that

So this is something @kchadha and I have been talking about a lot recently. Here is what we found: https://docs.google.com/document/d/1NFWOLRCgki9JKZkRVO8XUhyFR783fW0UTevUh8Usvts/edit?usp=sharing

VM, GUI, and paint generally expect to be working with true sizes and rotation centers. Should we change it so that the GUI expects to be getting its information manipulated by the VM the same way we do for the renderer? GUI and VM share the costume object, so it would be messy to try to switch. I think GUI should probably halve the size it displays in the container, so that the costume js object is internally consistent.

kchadha: Eeeeek. 
If I'm understanding the discussion above correctly, I say keep VM behavior the way it is right now, and have GUI do the math explicitly for display purposes.

kchadha (commented):
This looks good pending the sizing on the thumbnail changing when converting back and forth between svg and bitmap. I think there should be a follow-up issue to figure out what's going on with the buffers and scratch-storage.Asset.encodeDataURI and things in updateBitmap.

fsih:
Filed https://github.com/LLK/scratch-vm/issues/1103 to follow up

Issue #980, opened 2018-03-15
"Key pressed" block droppability behavior questions


ericrosenbaum:
@kchadha and I noticed some questions that will come up again, once we re-implement the "key pressed" hat and boolean block with droppable inputs:

- How should we support key inputs that are not in the list, such as modifier keys, if at all? How would you enter them?  
- What about non-english letters- should the key pressed block be able to detect those too if they get passed in?
- Currently you can pass in "left arrow" to detect the left arrow key. What if you're using scratch in another language- should this only use the english language string, or also match the localized one?
- What should it do if you pass in a longer string? Currently it matches the first character, so if you pass in "left arow" it matches the "l" key.

thisandagain:
Strawcat Proposal:
- Do not support modifier keys (particularly as these may impact the use of screen readers / accessibility features)
- Support non-english unicode characters (e.g. "when [æ] key pressed", "key [ø] pressed?")
- Only support the "special" cases (such as the arrow keys) in english (as per the dropdown menu values). File a feature request issue to extend support on this to other languages (backlog).
- Only match the first character

towerofnix:
What are opinions on detecting case-sensitive characters? For example, passing "key [a] pressed?" and "key [A] pressed?" would detect differently. I think this is okay since custom keyboard inputs are a pretty big possibility with this feature, but can't *really* be implemented as the "key pressed?" block is now, since we can't distinguish between upper- and lower-case keys being pressed.

I doubt this would have a negative effect on screen readers or similar tools, because I'm sure most would allow you to type capital letters, even if (they don't let let you specifically / it's difficult to) hold down shift.

This also falls in line a bit with @thisandagain's "support non-english unicode special characters" suggestion.

thisandagain:
I think I'd rather detect both "a" and "A" as I'm worried about breaking compatibility / unexpected behavior when someone forgets they have the Caps Lock key enabled.

PR #1085, opened 2018-04-25
 Re-write keyboard IO for droppability support

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/980
Depends on https://github.com/LLK/scratch-gui/pull/1869

### Proposed Changes

Update the keyboard IO system to support new use cases for the droppable input to the `key pressed?` boolean block.

A "scratch key" is defined internally as: 
- A key you can press on a keyboard, excluding modifier keys.
- It is an uppercase string of length one;
- Except for special key names for arrow keys and space (e.g. 'left arrow').
- Can be a non-english unicode letter like: æ ø ש נ 手 廿.

Generally, the key pressed hat and key pressed boolean should work as before, when their menu items are used.

But now, you can drop a variable onto the input of the key pressed boolean, with these behaviors:

- If you drop in a block that reports a string matching any of the standard scratch keys (such as 'space', 'left arrow', '1' or 'a') it should work.
- If you drop in a block that reports a number that matches the numeric key code of any of the standard scratch keys, it should work (32 for space, 37 for left arrow, 65 for A, etc).
- If you drop in a number that does not match a scratch key, it should detect the key matching the first digit of the number (e.g. dropped-in 99 should detect pressing the key '9'). 
- Similarly, if you drop in a multi-letter string, it should detect the key matching the first letter.
- Case is ignored: dropped-in 65, 'A' or 'a' should report true for pressing 'A' or 'a'.
- Dropped-in single letter non-english unicode characters should work as expected. Typing them generally requires switching to a non-english keyboard (I tried a few characters in Danish, Hebrew and Cangjie).

Note that I made a little test project that lets you play with many of these behaviors:
#218695578

**'any' key option does not respond to modifiers**
Note that this change causes a difference in the behavior when the 'any' key option is used: in Scratch 2.0, it returns true for modifier keys, but after this change it does not. I think this behavior matches the intent of @thisandagain's recommendation to not allow blocks to respond to modifier keys.

**Some unicode characters don't work on Mac Firefox**
Using the Cangjie keyboard on Mac Chrome, for example, you can set a variable to the character 日 (by pressing the key labeled 'a' on an English keyboard), drop it into the key pressed block, and it will return true if you hold down that same key. On Mac Firefox, you can set the variable by typing the character in the same way, but pressing the same key does not make the key pressed block return true. Instead, the keyboard event.key comes through as an 'a'. 

### Reason for Changes

- Maintaining support for some projects that used hacked key pressed blocks.
- Expanding support for dropped-in unicode characters.

### Test Coverage

I've updated the tests and added a couple additional ones.

### DIFF
--- a/src/io/keyboard.js
+++ b/src/io/keyboard.js
@@ -85,7 +85,7 @@ class Keyboard {
         keyArg = Cast.toString(keyArg);
 
         // If the arg matches a special key name, return it.
-        const keyNameList = Object.values(Keyboard.KEY_NAMES);
+        const keyNameList = Object.keys(Keyboard.KEY_NAMES).map(name => Keyboard.KEY_NAMES[name]);
         if (keyNameList.includes(keyArg)) {
             return keyArg;
         }
--- a/src/io/keyboard.js
+++ b/src/io/keyboard.js
@@ -22,21 +22,15 @@ class Keyboard {
      * @private
      */
     _scratchKeyToKeyCode (keyName) {
-        if (typeof keyName === 'number') {
-            // Key codes placed in with number blocks.
-            return keyName;
-        }
         const keyString = Cast.toString(keyName);
         switch (keyString) {
-        case 'space': return 32;
-        case 'left arrow': return 37;
-        case 'up arrow': return 38;
-        case 'right arrow': return 39;
-        case 'down arrow': return 40;
-        // @todo: Consider adding other special keys here.
+        case ' ': return 'space';
+        case 'ArrowLeft': return 'left arrow';
+        case 'ArrowRight': return 'right arrow';
+        case 'ArrowUp': return 'up arrow';
+        case 'ArrowDown': return 'down arrow';
         }
-        // Keys reported by DOM keyCode are upper case.
-        return keyString.toUpperCase().charCodeAt(0);
+        return keyName;
     }
 
     /**
@@ -46,18 +40,7 @@ class Keyboard {
      * @private
      */
     _keyCodeToScratchKey (keyCode) {
-        if (keyCode >= 48 && keyCode <= 90) {
-            // Standard letter.
-            return String.fromCharCode(keyCode).toLowerCase();
-        }
-        switch (keyCode) {
-        case 32: return 'space';
-        case 37: return 'left arrow';
-        case 38: return 'up arrow';
-        case 39: return 'right arrow';
-        case 40: return 'down arrow';
-        }
-        return '';
+        return keyCode;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/io/keyboard.js
```diff
@@ -16,87 +21,116 @@ class Keyboard {
...
+        keyString = Cast.toString(keyString);
+        // Convert space and arrow keys to their Scratch key names.
         switch (keyString) {
-        case 'space': return 32;
-        case 'left arrow': return 37;
-        case 'up arrow': return 38;
-        case 'right arrow': return 39;
-        case 'down arrow': return 40;
-        // @todo: Consider adding other special keys here.
+        case ' ': return 'space';
```
kchadha: We might want to use a regular expression here... but I'm not sure.

ericrosenbaum:
@kchadha ready for re-review- thanks!

thisandagain (approved):
From discussion / review yesterday I think this looks good. It would be nice to have @kchadha take one more look before merging though.

src/io/keyboard.js
```diff
@@ -16,87 +39,111 @@ class Keyboard {
...
+        case 'ArrowLeft':
+        case 'Left': return KEY_NAME.LEFT;
+        case 'ArrowUp':
+        case 'Up': return KEY_NAME.UP;
+        case 'Right':
+        case 'ArrowRight': return KEY_NAME.RIGHT;
+        case 'Down':
+        case 'ArrowDown': return KEY_NAME.DOWN;
+        }
+        // Ignore modifier keys
```
kchadha, line 60: Are modifier keys always > length 1 and all things > length 1 are necessarily modifier keys?

kchadha (approved):
The changes look good! Just had a question in the comments. Arrow keys, numbers, letters, etc. work on Edge, but I did not test language support there.

src/io/keyboard.js
ericrosenbaum: Those both seem to be generally true (except for the special cases handled just before that line). 

MDN's list of modifier keys is here: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values

There are things > 1 in other languages that "look like" single characters- the question is whether they can ever be passed as single key press events (not sure). For now we are not supporting these, but something to look out for in the future.

ericrosenbaum:
Thanks folks!

src/io/keyboard.js
kchadha: This is awesome! Thanks for the link =]

Issue #974, opened 2018-03-11
Creating clones of sprites is different with strings and integers


77Tigers:
## Expected Behavior
_**Create clone of (pick random (1) to (1))**_
should be the same as:
_**Create clone of (join (1) ())**_

## Actual Behavior

Opposite of above.

## Steps to Reproduce
Run the above code in a sprite named 1. Only one will create a clone.
Here's a demo project:
https://scratch.mit.edu/projects/209226817/#editor
https://llk.github.io/scratch-gui/develop/#209226817

## Operating System and Browser

Windows NT 10.0, Chrome 64.0.3282.186

Clark-E:
To add images, just drag and drop an image into the issue text editor. Github will automatically create an image.

77Tigers:
Thanks!

PR #1087, opened 2018-04-26
Cast 'create clone' argument to string

thisandagain:
### Resolves
Resolves GH-974

### Proposed Changes
Ensure that the `CLONE_OPTION` argument is a string before being passed to the `getSpriteTargetByName` runtime method.

### Reason for Changes
Because the `create clone of ( )` input is droppable, blocks like `pick random` may be used which return a numeric value.

### Reference
Screenshot of project as reported in the original issue after applying the patch:
image


kchadha (approved):
LGTM

PR #1088, opened 2018-04-26
Track say/thinks per target so they can maintain correct timing.

### LINKED ISSUE #1035: Say block doesn't run as long as it should
When a Second "say block for x seconds" interrupts a first "say block for y seconds", the second say block only stays on the screen until the first one would have been done.  (seems to inherit the remaining time on screen of the interrupted first say block and ignores its own "for x seconds")

### Steps to Reproduce

Create these blocks (not connected):
image

- Click on the "Say hello" block.
- Notice that the cat says "Hello"
- Notice that the "Say hello" block is glowing.
- Click on the "Say goodbye" block (within 4 seconds of clicking on the "say hello" block)
- Notice that the cat says "Goodbye"
- Notice that the "Say goodbye" and "Say hello" blocks are both glowing.
- Wait
- Notice that the speech bubble disappears early, and the "Say goodbye" block is still glowing as though it were running.

Chrome on linux.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/1035

### Proposed Changes

Add a `usageId` to the bubble state (which is target specific) which changes each time it is updated. In the "say/think for secs" blocks, check that the usage ID is the same before clearing it when the timeout finishes.

### Reason for Changes

There was a bug where interrupting a say with another timed say would keep the old timing. i.e.

interrupting-say

This fixes that problem

safely-interrupting-say

---

It is worth noting that the "disappear but keep running" behavior of the say blocks matches 2.0, which may be important for compatibility. If a say block is interrupted by another, the block stays running for its original timing, and a stack won't keep going until that original time is up. That is why this change only impacts clearing the bubble at the end of the timeout, not the promise resolution itself.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -11,6 +11,8 @@ const uid = require('../util/uid');
  *      See _renderBubble for explanation of this optimization.
  * @property {string} text - the text of the bubble.
  * @property {string} type - the type of the bubble, "say" or "think"
+ * @property {?string} usageId - ID indicating the most recent usage of the say/think bubble.
+ *      Used for comparison when determining whether to clear a say/think bubble.
  */
 
 class Scratch3LooksBlocks {
@@ -46,7 +48,7 @@ class Scratch3LooksBlocks {
             skinId: null,
             text: '',
             type: 'say',
-            usageId: null // ID for hiding a timed say/think block.
+            usageId: null
         };
     }
 
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -1,6 +1,7 @@
 const Cast = require('../util/cast');
 const Clone = require('../util/clone');
 const RenderedTarget = require('../sprites/rendered-target');
+const uid = require('../util/uid');
 
 /**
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -44,7 +45,8 @@ class Scratch3LooksBlocks {
             onSpriteRight: true,
             skinId: null,
             text: '',
-            type: 'say'
+            type: 'say',
+            usageId: null // ID for hiding a timed say/think block.
```
kchadha: Though this comment makes sense in the context of this PR, I don't think it would be clear for someone coming to this without that context. Maybe rephrase to say something along the lines of:
> ID indicating the most recent usage of the say bubble. Used for comparison when determining whether to clear a say/think bubble.

kchadha (approved):
Code looks good. Left a comment about a comment though.

cwillisf (approved):
My only comment is a +1 to @kchadha's comment about the comment. Otherwise, LGTM!

paulkaplan:
Cool thanks @kchadha @cwillisf, the `usageId` was missing from the jsdoc typedef for bubblestate anyway, so I moved the comment there and used @kchadha's recommendation.

PR #1089, opened 2018-04-26
Fix say flickering

### LINKED ISSUE #1817: Speech bubble glitch
### Expected Behavior
Speech bubble conversation should occur without any glitches
https://scratch.mit.edu/projects/56987002/#editor

### Actual Behavior
Speech bubble glitch happens when one of the two things happens:
2 sprites are talking back and forth
1 sprite has multiple "say" blocks
https://preview.scratch.mit.edu/#56987002

### Steps to Reproduce
1. Create a project in 3.0 or load in a 2.0 project that has multiple "say" blocks
2. Click the green flag
3. Observe the speech bubble glitch

Chrome 65

paulkaplan:
Only this commit https://github.com/LLK/scratch-vm/commit/3019d05181594360b0838775b9916abaf1be2f6d 

The other commit is part of https://github.com/LLK/scratch-vm/pull/1088, but i stacked them to avoid merge conflicts.

### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1817 

### Proposed Changes

There is an "optimization" in the speech bubble code that tries to hide/show the speech bubble texture instead of disposing of it completely when it is hidden. This was to prevent "flickering" with projects that use rapid "say" in a loop to get a typewriter effect. However, it was causing the previous message to flicker on projects that take turns in talking as well (issue #1817). 

Since the optimization is already in place, I just put an upper bound (200ms, arbitraryish) on whether we do a "hard clear" vs the "soft clear" that the hide/show optimization uses. The idea is that if the speech bubbles are being shown rapidly, use the hide/show optimization, but if they are being shown slowly, clear the drawable completely between says.

Honestly this doesn't feel great, but it is the only way I could think of supporting both "typewriter" style and regular conversations. I'm also ok with disabling the "optimization" and just making those typewriter projects flicker....

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -59,6 +59,14 @@ class Scratch3LooksBlocks {
     }
 
     /**
+     * Max block time we use the "soft clear" optimization to prevent flickering in "typewriter" projects.
+     * @type {string}
+     */
+    static get MAX_SOFT_CLEAR_TIME () {
+        return 0.02; // seconds
+    }
+
+    /**
      * @param {Target} target - collect bubble state for this target. Probably, but not necessarily, a RenderedTarget.
      * @returns {BubbleState} the mutable bubble state associated with that target. This will be created if necessary.
      * @private
@@ -282,12 +290,17 @@ class Scratch3LooksBlocks {
         this.say(args, util);
         const target = util.target;
         const usageId = this._getBubbleState(target).usageId;
+        const secs = args.SECS;
         return new Promise(resolve => {
             this._bubbleTimeout = setTimeout(() => {
                 this._bubbleTimeout = null;
                 // Clear say bubble if it hasn't been changed and proceed.
                 if (this._getBubbleState(target).usageId === usageId) {
-                    this._updateBubble(target, 'say', '');
+                    if (secs < this.MAX_SOFT_CLEAR_TIME) {
+                        this._updateBubble(target, 'say', '');
+                    } else {
+                        this._onTargetWillExit(target);
+                    }
                 }
                 resolve();
             }, 1000 * args.SECS);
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -277,12 +288,20 @@ class Scratch3LooksBlocks {
...
             this._bubbleTimeout = setTimeout(() => {
                 this._bubbleTimeout = null;
-                // Clear say bubble and proceed.
-                this._updateBubble(_target, 'say', '');
+                // Clear say bubble if it hasn't been changed and proceed.
+                if (this._getBubbleState(target).usageId === usageId) {
+                    if (secs < this.MAX_SOFT_CLEAR_TIME) {
+                        this._updateBubble(target, 'say', '');
+                    } else {
+                        this._onTargetWillExit(target);
```
cwillisf: I guess it's been this way for a while so I wouldn't consider this a blocker for this PR, but I would prefer if methods like `onSomething` or `handleSomething` were only called as event handlers. Since `_onTargetWillExit` is being called for other reasons than just the "target will exit" event, I'd prefer a name describing what it does rather than why it does. Something like `_destroyBubble` maybe.

Just thinking out loud -- it may not have been you who wrote this code. It may have even been me! Yesterday-me is not NEARLY as good a programmer as today-me... ;)

cwillisf (commented):
It feels wrong to me to put an arbitrary timeout on this, but if that's the best solution then I guess I'm OK with that. I wonder if there are other solutions, though.

For example: it seems like a flash of the "old" bubble is only a problem if it's not already being displayed. What if we "hard clear" the bubble on the "rising edge" of transitioning from hidden to visible? Hidden bubbles can hang out and get cleaned up when the tab closes, when the user loads a different project, etc.; typewriter-style projects almost never hard-clear; projects with large gaps between "say" or "think" blocks would hard-clear almost every time. Whatcha think?

paulkaplan:
@cwillisf I would definitely prefer a solution like that, but the way the typewriter projects work is by using "say for small time" in a loop, rather than just "say" in a loop and then "say <blank>" at the end to clear. Because of that, the bubble is actually hidden and then re-shown every time through the loop. I'm not sure how to differentiate that "very short say for secs" from the longer "say for time, wait, say for time". I'd be happy to talk about this more though, definitely looking for a better solution. 

I'm also ok with ripping out the "optimization" for now since it fixes a very small number of projects and (very mildly) breaks every project with say, and revisiting it when/if there are complaints about it

cwillisf:
> the bubble is actually hidden and then re-shown every time through the loop

Ah, I see -- that's unfortunate.

> I'm also ok with ripping out the "optimization" for now since it fixes a very small number of projects and (very mildly) breaks every project with say, and revisiting it when/if there are complaints about it

I'm starting to think this is the right approach. Or, at least, pull out the optimization for now to get back to correct behavior, and file a non-critical issue to address the flicker. If you have example projects for both "sides" of the behavior it would be nice to record project numbers, possibly after remixing them onto an account you control for archival purposes :)

paulkaplan:
@cwillisf I removed the optimization that was causing the flickering by removing the "soft clear" functionality and replacing it with the existing "hard clear". I'll submit a follow-up issue for fixing projects like this https://scratch.mit.edu/projects/52572706/

Confirmed that this does fix the flash of previous bubbles in the conversation situation.

cwillisf (approved):

PR #1092, opened 2018-04-27
Refactor Video IO

kchadha:
### Resolves

Towards resolving LLK/scratch-gu#1778.

### Proposed Changes

Changes described in LLK/scratch-gui#1880.

This PR refactors the video IO device so that the GUI acts as a video stream provider (and requests the user media stream from the browser). The functionality related to rendering a preview on the stage is left in here as the VM owns the runtime and the renderer, and should maintain ground truth that if an extension has requested video access, that the video is visible to the user on the stage (or whatever other renderer is connected).

### Reason for Changes

Ensure that video can be used by the VM for video extensions but also by the GUI for other features (e.g. camera upload).

### Test Coverage

Existing tests pass.

#### Related PRs
LLK/scratch-gui#1880. Both PRs should be reviewed together, this PR should be merged first.

### DIFF
--- a/src/io/video.js
+++ b/src/io/video.js
@@ -1,7 +1,42 @@
 class Video {
     constructor (runtime) {
         this.runtime = runtime;
+
+        /**
+         * @typedef VideoProvider
+         * @property {Function} enableVideo - Requests camera access from the user, and upon success,
+         * enables the video feed
+         * @property {Function} disableVideo - Turns off the video feed
+         * @property {Function} getFrame - Return frame data from the video feed in
+         * specified dimensions, format, and mirroring.
+         */
         this.provider = null;
+
+        /**
+         * Id representing a Scratch Renderer skin the video is rendered to for
+         * previewing.
+         * @type {number}
+         */
+        this._skinId = -1;
+
+        /**
+         * The Scratch Renderer Skin object.
+         * @type {Skin}
+         */
+        this._skin = null;
+
+        /**
+         * Id for a drawable using the video's skin that will render as a video
+         * preview.
+         * @type {Drawable}
+         */
+        this._drawable = -1;
+
+        /**
+         * Store the last state of the video transparency ghost effect
+         * @type {number}
+         */
+        this._ghost = 0;
     }
 
     static get FORMAT_IMAGE_DATA () {
@@ -30,6 +65,14 @@ class Video {
     }
 
     /**
+     * Set a video provider for this device.
+     * @param {VideoProvider} provider - Video provider to use
+     */
+    setProvider (provider) {
+        this.provider = provider;
+    }
+
+    /**
      * Request video be enabled.  Sets up video, creates video skin and enables preview.
      *
      * ioDevices.video.requestVideo()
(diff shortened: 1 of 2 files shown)


src/io/video.js
```diff
@@ -97,39 +80,18 @@ class Video {
...
-            // by clearing refs to video and track, we should lose our hold over the camera
-            this._video = null;
-            if (this._track) {
-                this._track.stop();
-            }
-            this._track = null;
-        }
+        if (!this.provider) return null;
+        this._disablePreview();
+        this.provider.disableVideo();
```
cwillisf: Since `_disablePreview()` doesn't depend on the `provider` I would recommend moving the `_disablePreview()` call above the `if`. It may never happen in practice, but I'm worried about strange situations where, I dunno, we resume from sleep mode to find that the user unplugged their camera, and now they have a preview image stuck on their stage.

kchadha: Yeah I was trying to figure out where I wanted that call.

My thought is that disable preview does still depend on the provider, because the things that are getting disabled only get set up if there was a provider... but I guess it's possible for a developer to switch providers or set their provider back to null between their calls to enableVideo and disableVideo. Perhaps switching from an existing provider to another one or to null should result in a call to disableVideo on the previous provider before that switch is allowed to happen...?

src/io/video.js
```diff
@@ -89,6 +64,14 @@ class Video {
         return 1;
     }
 
+    /**
+     * Set a video provider for this device.
+     * @param {VideoProvider} provider - Video provider to use
+     */
+    setProvider (provider) {
+        this.provider = provider;
```
gnarf: Do you think we might want to track enabled/disabled state here and enable if the provider is added late?

I don't think we should worry to much about "state fixing" in the setProvider call - the intention is that a provider would be set at creation time, and not changed while running.

src/io/video.js
gnarf: It doesn't super matter if we disable preview (it would just early exit once per frame if the video gets stopped but we don't know about it)

It still makes sense to disablePreview before the provider check though I think as @cwillisf said.

src/io/video.js
```diff
@@ -151,59 +113,11 @@ class Video {
         format = Video.FORMAT_IMAGE_DATA,
         cacheTimeout = this._frameCacheTimeout
     }) {
+        if (this.provider) return this.provider.getFrame({dimensions, mirror, format, cacheTimeout});
         if (!this.videoReady) {
```
gnarf: I don't think this block is needed here, provider does it. The early exit when not video ready will be handled above.

gnarf (approved):
<3 thanks so much for taking this on and cleaning up all these bits that made me feel dirty for having them in IO!

src/io/video.js
cwillisf: > Perhaps switching from an existing provider to another one or to null should result in a call to disableVideo on the previous provider before that switch is allowed to happen...?

That sounds right to me, but I'm also starting to feel a little YAGNI... hmm.

kchadha:
Addressed comments from @cwillisf and @gnarf, with the exception of setting enabled state/enabling video when the video provider is set. I don't think that's necessary (but @gnarf, I'm happy to talk more about it if you think it is).

thisandagain (approved):
Looks great! Thank you for cleaning this up! 😄

gnarf:
> with the exception of setting enabled state/enabling video when the video provider is set. I don't think that's necessary

:+1: 

We definitely don't need it yet -- nothing is trying to replace video provider during runtime.

PR #1093, opened 2018-04-27
Fix downloadProjectId method

gnarf:
### Reason for Changes
* I noticed while trying to benchmark my collision detect that `downloadProjectId` was assuming the project url gave us a project.json instead of a zipped sb2.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -240,7 +240,7 @@ class VirtualMachine extends EventEmitter {
         const vm = this;
         const promise = storage.load(storage.AssetType.Project, id);
         promise.then(projectAsset => {
-            vm.loadProject(projectAsset.data.buffer);
+            vm.loadProject(projectAsset.data);
         });
     }
 
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -240,7 +240,7 @@ class VirtualMachine extends EventEmitter {
         const vm = this;
         const promise = storage.load(storage.AssetType.Project, id);
         promise.then(projectAsset => {
-            vm.loadProject(projectAsset.decodeText());
+            vm.loadProject(projectAsset.data.buffer);
         });
     }
 


src/virtual-machine.js
```diff
@@ -240,7 +240,7 @@ class VirtualMachine extends EventEmitter {
         const vm = this;
         const promise = storage.load(storage.AssetType.Project, id);
         promise.then(projectAsset => {
-            vm.loadProject(projectAsset.decodeText());
+            vm.loadProject(projectAsset.data.buffer);
```
kchadha: scratch-storage usually has `data` set to a json object for project assets and not an .sb2 / .sb3 buffer. The buffer stuff in loadProject is mainly for file uploads, but can handle the stuff coming from scratch-storage through the scratch 2 project server.

However, I was testing, and it looks like [scratch-storage has been updated recently](https://github.com/LLK/scratch-storage/commit/98d7bd9fced6316a1ada82ae23ef9b098a3bf4f8), and this change might have had an un-intended side effect of fetching the wrong thing from our servers...? The asset type returned from scratch-storage is now giving us buffers in the `data` property instead of JSON. @cwillisf, @rschamp, any thoughts on this?

gnarf:
@kchadha @rschamp - This change was needed to add newer projects to the scratch benchmark, it didn't break any of the old ones as far as I can tell.

src/virtual-machine.js
cwillisf: I would expect `projectAsset.decodeText()` to be the right thing here: `projectAsset.data` should always by a `Buffer` or `Uint8Array` regardless of the asset type, and in the case of `AssetType.Project` it should be the project JSON encoded in UTF-8 form (see `Project` in https://github.com/LLK/scratch-storage/blob/develop/src/AssetType.js for example).

Are you saying that LLK/scratch-storage#32 makes it so that `projectAsset.data` now sometimes contains a ZIP file instead of JSON? If so... how is the storage module ending up with an exported project stored as `AssetType.Project`?

gnarf:
You can test locally on https://github.com/LLK/scratch-vm/pull/1094 -- that project doesn't load from the playground suite without the PR.  I'm not sure what should be fixed to fix this, but that project breaks with `.decodeText()` -- The URL we fetch for it is a zip'ed sb2 - not a json.

kchadha (approved):
LG (pending travis build)! Thanks for catching/updating this.

PR #1096, opened 2018-04-30
Implement "all at once" block

towerofnix:
### Resolves

Towards #355 (implements "all at once" block). Should be merged alongside LLK/scratch-blocks#1483.

### Proposed Changes

Adds a spec mapping from `warpSpeed` to `control_all_at_once`.

Adds an implementation for `control_all_at_once`: When the block is called, run the inside blocks. It's functionally equivalent to "if 1 = 1".

### Reason for Changes

Compatibility with Scratch 2.0 projects. This is a hacked block; in total, [around 1800 projects use the "all at once" block](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752). (Its opcode is `warpSpeed` in 2.0.)

### Test Coverage

Unit test for the new block (it just makes sure `startBranch` is called). Existing tests pass. Tested manually with [this](https://scratch.mit.edu/projects/74719650/) project.

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -185,6 +185,12 @@ class Scratch3ControlBlocks {
     }
 
     allAtOnce (args, util) {
+        // Since the "all at once" block is implemented for compatiblity with
+        // Scratch 2.0 projects, it behaves the same way it did in 2.0, which
+        // is to simply run the contained script (like "if 1 = 1").
+        // (In early versions of Scratch 2.0, it would work the same way as
+        // "run without screen refresh" custom blocks do now, but this was
+        // removed before the release of 2.0.)
         util.startBranch(1, false);
     }
 }
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -35,7 +35,8 @@ class Scratch3ControlBlocks {
             control_delete_this_clone: this.deleteClone,
             control_get_counter: this.getCounter,
             control_incr_counter: this.incrCounter,
-            control_clear_counter: this.clearCounter
+            control_clear_counter: this.clearCounter,
+            control_all_at_once: this.allAtOnce
         };
     }
 
@@ -181,6 +182,10 @@ class Scratch3ControlBlocks {
 
     incrCounter () {
         this._counter++;
+    }
+
+    allAtOnce (args, util) {
+        util.startBranch(1, false);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_control.js
```diff
@@ -35,7 +35,8 @@ class Scratch3ControlBlocks {
             control_delete_this_clone: this.deleteClone,
             control_get_counter: this.getCounter,
             control_incr_counter: this.incrCounter,
-            control_clear_counter: this.clearCounter
+            control_clear_counter: this.clearCounter,
```
Kenny2github, line 38: nit: are we caring about alphabetization at this point yet?

towerofnix: I don't see exactly what you mean?

We're definitely sticking to `camelCase` for function/method names. I try to follow the same opcode-case that other opcodes in the same category use, so in this case, `snake_case`.

Kenny2github: I'm talking about the order of the names as they appear in the file - i.e. clearCounter should be before deleteClone, not after incrCounter.

towerofnix: Oh, got it. Arguably they should be in the order (get, clear, incr) because [that's the way they are in scratch-flash](https://github.com/LLK/scratch-flash/blob/35b6a84bfd508c17105d6bf68cd0962a5c8c94a0/src/Specs.as#L379-L381), or alphabetic, because consistency.. but the thing is, this PR adds "all at once", not the counter blocks (that was PR #1033). Maybe it's worth it to clean this up at some point, but the order definitely isn't causing any confusion right now.

src/blocks/scratch3_control.js
```diff
@@ -182,6 +183,10 @@ class Scratch3ControlBlocks {
     incrCounter () {
         this._counter++;
     }
+
+    allAtOnce (args, util) {
```
kchadha, line 187: @towerofnix, thanks for this PR! Do you mind adding the comment from scratch-blocks (or the part of it talking about 2.0 functionality) here as well?

kchadha (commented):
One minor change and then it's good to go

kchadha (commented):
Code changes look good, but do you mind copying over your comment from scratch-blocks here as well?

src/blocks/scratch3_control.js
towerofnix: Will do!

towerofnix:
@kchadha Thanks! The `allAtOnce` function has some documentation now. (I just used `//` comments here - not sure if it's convention here to use `/**`-style? I didn't see any block-implementation functions that had `/**`-style comments.)

kchadha (approved):

PR #1098, opened 2018-04-30
Use renderer bounds instead of stage bounds for positioning bubbles.

### LINKED ISSUE #1569: say bubble goes off of the stage with certain backgrounds
Expected:
If you use a say block with a sprite and it is near the edge of the screen the bubble should move so it stays on screen.

Actual:
The say bubble will stay in the upper right corner of sprite and move off the screen, above or below.

Reproduction steps:
NEEDS REPRO
I saw this in the screen capture from the Chelsea Clubhouse trip on 2/28/18 and pulled the screenshots from that.  Could not reproduce on 3/5/18.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/1569

### Proposed Changes

Instead of using the rendered bounds of the stage (which can be larger than the actual rendering area of the stage due to large costumes), use the `nativeSize` which is the size in scratch coords of the rendering area. Note that `nativeSize` is not the size of the canvas element or anything like that, it is the "logical size", i.e. the [width, height] in scratch coords.

### Test Coverage

Tested manually by using a large costume for the stage (like the basketball court), and testing the bubble bounds. 

bubble-positioning-scratch3-develop
bubble-positioning-scratch3-fixed

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -137,7 +137,13 @@ class Scratch3LooksBlocks {
                 bottom: target.y
             };
         }
-        const stageBounds = this.runtime.getTargetForStage().getBounds();
+        const stageSize = this.runtime.renderer.getNativeSize();
+        const stageBounds = {
+            left: -stageSize[0] / 2,
+            right: stageSize[0] / 2,
+            top: stageSize[1] / 2,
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
Ah! That makes sense.

PR #1102, opened 2018-05-01
"Implement" miscellaneous no-op obsolete blocks

towerofnix:
### Resolves

Towards #355 (implements "hide all sprites", "set stretch to", "change stretch by", and "user id"). Should be merged alongside LLK/scratch-blocks#1488.

### Proposed Changes

Adds spec mappings from `hideAll`, `setStretchTo`, `changeStretchBy`, and `getUserId` to their respective 3.0 opcodes.

Adds implementations for those blocks - none of them do anything.

### Reason for Changes

Compatibility with Scratch 2.0 projects. In total, [around 350 projects use the change/set stretch blocks](https://github.com/LLK/scratch-vm/issues/355#issuecomment-379418752). I don't have stats on how often "hide all sprites" and "user id" were used.

### Test Coverage

No new tests (since the blocks don't do anything), but tested manually with [this](https://scratch.mit.edu/projects/219797286/) project.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -255,6 +255,8 @@ class Scratch3LooksBlocks {
             looks_cleargraphiceffects: this.clearEffects,
             looks_changesizeby: this.changeSize,
             looks_setsizeto: this.setSize,
+            looks_changestretchby: this.doNothing,
+            looks_setstretchto: this.doNothing,
             looks_gotofrontback: this.goToFrontBack,
             looks_goforwardbackwardlayers: this.goForwardBackwardLayers,
             looks_size: this.getSize,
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -244,6 +244,7 @@ class Scratch3LooksBlocks {
             looks_thinkforsecs: this.thinkforsecs,
             looks_show: this.show,
             looks_hide: this.hide,
+            looks_hideallsprites: this.doNothing,
```
kchadha: Instead of defining a `doNothing` member in different categories, I would be okay with just replacing that with `() => {} // legacy no-op block`

kchadha (changes requested):
Minor change request in comments.

src/blocks/scratch3_looks.js
towerofnix: Cool, will update this and the scroll-blocks PR.

kchadha (approved):

PR #1105, opened 2018-05-01
Dynamically load music extension manifest

mzgoddard:
### Proposed Changes

Dynamically load the 1.6MB of music extension sample files dynamically.

### Reason for Changes

Split the music extension samples to be loaded separate from the main scripts to need less to be downloaded and for faster first execution of those scripts so Scratch starts earlier.

This change uses webpack's chunking to dynamically load the music extension files. An improvement would be to fetch the files as data through fetch/xhr base64 encoded json content or as a sound sprite.

Dynamically loaded separate from the main scripts, music samples could be preloaded with a script tag, in this format, or with an audio or link tag in the html body if it was encoded in a json file or sound sprite. This would load the file near the same time of the main scripts without blocking their execution.

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -6,17 +6,6 @@ const MathUtil = require('../../util/math-util');
 const Timer = require('../../util/timer');
 
 /**
- * The instrument and drum sounds, loaded as static assets.
- * @type {object}
- */
-let assetData = {};
-try {
-    assetData = require('./manifest');
-} catch (e) {
-    // Non-webpack environment, don't worry about assets.
-}
-
-/**
  * Icon svg to be displayed at the left edge of each extension block, encoded as a data URI.
  * @type {string}
  */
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_music/index.js
```diff
@@ -112,14 +112,19 @@ class Scratch3MusicBlocks {
     _storeSound (filePath, index, bufferArray) {
         const fullPath = `${filePath}.mp3`;
 
-        if (!assetData[fullPath]) return;
+        return this._fetchSounds()
+            .catch(() => ({}))
```
rschamp: This does the same thing as the previous empty catch above?  Would you comment why that's ok to do here?

src/extensions/scratch3_music/index.js
```diff
@@ -102,6 +91,17 @@ class Scratch3MusicBlocks {
         });
     }
 
+    _fetchSounds () {
+        if (!this._assetData) {
+            this._assetData = new Promise((resolve, reject) => {
+                require.ensure([], () => {
```
rschamp: Would you please add a comment here, since it's not used anywhere else?  And what will happen when not in a webpack environment?

rschamp (changes requested):
Thanks @mzgoddard, I think this looks good, just some comments would be helpful.

src/extensions/scratch3_music/index.js
mzgoddard: Yes. In case loading the sounds throws an error running the dynamic module or require.ensure is not available, this'll let the extension gracefully fail.

src/extensions/scratch3_music/index.js
mzgoddard: - Something without require.ensure: should throw an error trying to call an undefined function or derefencing a member off of an undefined variable. In those cases the wrapping promise constructor will reject the promise and the caller of `_fetchSounds` can handle the error with a catch handler on the promise.
- Webpack supports require.ensure for node. If you require the node target built script, this will load the sounds successfully.

rschamp (approved):

PR #1106, opened 2018-05-01
Use src/index as browser entry point

mzgoddard:
### Proposed Changes

- Use src/index.js as the entry by default in tools building scratch-vm into a larger package
- List src dependencies in the package `dependencies` category so they are installed for users of this package (scratch-gui for example)

### Reason for Changes

This saves file space by excluding extra webpack boilerplate, excluding extra copies of dependencies and may allow for deeper optimizations.

As is, scratch-* packages build in their dependencies. Three such packages build in `text-encoding`: scratch-vm, scratch-storage, and scratch-gui. These three copies are each ~500KB of source before minification and compression. Importantly they are also duplicate execution paths, that build up 3 copies of the same runtime memory structures and 3 copies of compiled functions. There are two copies of scratch-svg-renderer's 1.8MB source depended on by scratch-render and scratch-gui.

Removing these duplicate copies brings the scratch-gui lib.min.js file size from (webpack 4's) 9.2MB (3.6 gzipped) to 7.1MB (2.8 gzipped). (Additionally with https://github.com/LLK/scratch-vm/pull/1105 this further drops to 5.2MB (1.4 gzipped). The music sample js with [webpack-4](https://github.com/LLK/scratch-vm/pull/1104) branch's arraybuffer-loader is 2.0MB (1.4 gzipped))

With the webpack-4 branch this reduces runtime memory usage on a Samsung Chromebook Plus from 28MB (develop) to 21MB (webpack-4) to 18MB.

#### caveats ####

The main caveat for this change is that some building details would need repeating in some of the packages depending on other scratch packages like scratch-vm and scratch-gui. Packages with their module rules configurations are not automatically included in builds consuming that package. scratch-render for instance uses `ify-loader` on its linebreak and `grapheme-breaker` dependencies. scratch-vm and scratch-gui need to support handling this.

There are some ways that this could be resolved in the dependencies. Parts that need this extra handling could be prebuilt. Instead of building the entire package for use in later browser builds, the parts with webpack configuration details could be prebuilt and depended on by the package's source.

Some babel details can be handled by compiling to another build folder with modules individually compiled using the `babel` cli tool.

Probably the easiest option would be to inline the loader configuration details. Many binary files in the scratch packages have their loader stated in the `require` request. This could be extended to other dependencies like `linebreak` in scratch-vm to resolve this.

##### merging #####

A second minor caveat is this PR should be merged with its sibling PRs in each package or from top down, starting at scratch-gui. Directly related to build configuration packages using other scratch packages needed to handle cases building where the consumed packages have build details that are not already handled.


rschamp (approved):
Thanks @mzgoddard, I think we should do this! The only concern is what you listed in caveats, but luckily I think most of our tests for `scratch-render` are going to end up being integration tests on the built bundle, so this shouldn't be an issue — we'll just make the ify loader inline on those two requirements.

package.json
```diff
@@ -27,6 +27,21 @@
...
+    "arraybuffer-loader": "^1.0.3",
+    "decode-html": "2.0.0",
+    "escape-html": "1.0.3",
+    "format-message": "5.2.1",
+    "htmlparser2": "3.9.2",
+    "immutable": "3.8.1",
+    "jszip": "^3.1.5",
+    "minilog": "3.1.0",
+    "nets": "3.2.0",
+    "scratch-parser": "^3.0.0",
```
kchadha: Why is this in here? I think it's only a dependency of scratch-vm.

PR #1110, opened 2018-05-03
Support for Sprite Upload

kchadha:
### Resolves

Towards resolving LLK/scratch-gui#222.

### Proposed Changes

Adds a generic API to add a sprite from a .sprite2 or .sprite3 or (json versions of these) representation. Validates given input to determine which version of sprite this is, and uses addSprite2 or addSprite3 based on that. 

### Reason for Changes

Sprite upload -- available in 2.0.

### Test Coverage

Added integration test for .sprite2. Tests for sprite3 to come when .sprite3 exporting is implemented.

#### Note:

This PR contains a breaking change in that addSprite2 and addSprite3 have become intentionally private (as they no longer accept json strings). The new `addSprite` API should be used instead.


src/virtual-machine.js
```diff
@@ -212,7 +212,7 @@ class VirtualMachine extends EventEmitter {
         }
 
         const validationPromise = new Promise((resolve, reject) => {
-            validate(input, (error, res) => {
+            validate(input, false /* this is not a single sprite */, (error, res) => {
```
thisandagain: Could we move this to a line comment or a comment above?

src/virtual-machine.js
```diff
@@ -378,51 +378,81 @@ class VirtualMachine extends EventEmitter {
...
+            // or an ArrayBuffer view (this includes all typed arrays and DataViews)
+            // turn the object into a JSON string, because we suspect
+            // this is a project.json as an object
+            // validate expects a string or buffer as input
+            // TODO not sure if we need to check that it also isn't a data view
+            input = JSON.stringify(input);
+        }
+
+        const validationPromise = new Promise((resolve, reject) => {
+            validate(input, true /* single sprite */, (error, res) => {
```
thisandagain: Same here

src/virtual-machine.js
```diff
@@ -378,51 +378,81 @@ class VirtualMachine extends EventEmitter {
...
+                if (error) {
+                    reject(error);
+                }
+                if (res) {
```
thisandagain: With no error handled above is there ever an instance where `res` is false-y?

thisandagain (commented):
Looks good!

src/virtual-machine.js
kchadha: Probably not. =]

kchadha: The thing about removing it is that since reject isn't a return, if there's an error, it executes the reject and then executes the resolve. The actual error still bubbles up, but the fact that the resolve executes after the reject seems weird (and possibly error-prone) to me...

kchadha:
Just force pushed a rebase off of develop for a cleaner history because there were previously some merge commits from things that are already in develop that snuck in somehow.

kchadha:
@thisandagain, @paulkaplan, @rschamp, this PR is ready for re-review.

thisandagain (approved):

PR #1111, opened 2018-05-03
Use arraybuffer-loader in place of buffer-loader

mzgoddard:
### Resolves

Pull out the `arraybuffer-loader` part of https://github.com/LLK/scratch-vm/pull/1104.

### Proposed Changes

Replace buffer-loader with arraybuffer-loader.

### Reason for Changes

> buffer-loader inlines binary files in webpack builds by encoding them as numeric literals. Each byte encoded this way uses 2 to 4 bytes. arraybuffer-loader encodes the data as base64 and decodes it back into numbers at runtime using 1.33 bytes for each encoded bytes. arraybuffer-loader saves on bandwidth to download the resulting js and the amount of space in memory the script takes.

Quoted from https://github.com/LLK/scratch-vm/pull/1104

### Test Coverage

Builds and runs the existing music extension tests.


rschamp (approved):
Thank you!

PR #1113, opened 2018-05-04
Add micro:bit Scratch 3.0 extension

khanning:
### Proposed Changes

Adds the micro:bit extension to the scratch-vm.

### Reason for Changes

Centralize development of the micro:bit extension for Scratch 3.0


thisandagain:
Thanks @khanning!

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+
+    /**
+     * @return {number} - the latest value received for the motion gesture states.
+     */
+    get gestureState () {
+        return this._sensors.gestureState;
+    }
+
+    /**
+     * @return {number} - the current state of the 5x5 LED matrix.
```
ericrosenbaum: should be Uint8Array?

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+        }
+    }
+
+    /**
+     * @param {object} args - the block's arguments.
+     * @return {boolean} - the touch pin state.
+     * @private
+     */
+    whenPinConnected (args) {
+        const pin = parseInt(args.PIN, 10);
```
ericrosenbaum, line 655: You can use our Cast utility for this (see e.g. https://github.com/LLK/scratch-vm/blob/f3404f4275c659ca2901d5d13105f64c8f8f9e8b/src/extensions/scratch3_music/index.js#L590)

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+        if (isNaN(pin)) return;
+        return this._device._checkPinState(pin);
```
ericrosenbaum: The pin number should also be wrapClamped to the number of pins (see e.g. https://github.com/LLK/scratch-vm/blob/f3404f4275c659ca2901d5d13105f64c8f8f9e8b/src/extensions/scratch3_music/index.js#L593)

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+     * Map integer values between two ranges
+     * @param {number} val - the value to map.
+     * @param {number} iMin - the minimum value for the input range.
+     * @param {number} iMax - the maximum value for the input range.
+     * @param {number} oMin - the minimum value for the output range.
+     * @param {number} oMax - the maximum value for the ouput range.
+     * @return {number} - the scaled value
+     * @private
+     */
+    _mapSensorVal (val, iMin, iMax, oMin, oMax) {
```
ericrosenbaum: unused? or here for future use?

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+    displayText (args) {
+        window.postMessage({type: 'command', uuid: 'text', buffer: String(args.TEXT)}, '*');
+        return;
+    }
+
+    /**
+     * Display a predefined symbol on the 5x5 LED matrix.
+     * @param {object} args - the block's arguments.
+     */
+    displaySymbol (args) {
```
ericrosenbaum, line 545: This is awesome! I 0xAAC544 it.

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+                    }
+                },
+                {
+                    opcode: 'displayClear',
+                    text: 'set all lights off',
+                    blockType: BlockType.COMMAND
+                },
+                {
+                    opcode: 'whenTilted',
+                    text: 'when tilted [DIRECTION]?',
```
ericrosenbaum: No question mark on this block I think

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+                            defaultValue: '0'
+                        }
+                    }
+                }
+            ],
+            menus: {
+                buttons: ['A', 'B', 'any'],
+                rowcol: ['1', '2', '3', '4', '5'],
+                pinState: ['on', 'off'],
+                symbols: ['❤', '♫', '☓', '✓', '↑', '↓', '←', '→', '◯', '☀', '☺', '!', '?'],
```
ericrosenbaum: To avoid having this list in two places, you could generate this array from the keys of symbols2hex

src/extensions/scratch3_microbit/index.js
```diff
@@ -0,0 +1,676 @@
...
+     * @return {boolean} - true if the micro:bit is free falling.
+     */
+    whenJumped () {
+        return (this._device.gestureState >> 1) & 1;
+    }
+
+    /**
+     * Display text on the 5x5 LED matrix.
+     * @param {object} args - the block's arguments.
+     * Note the limit is 20 characters
```
ericrosenbaum, line 533: Should we trim the string down to 20 characters here, to avoid sending e.g. the entire lyrics of "All Star" by Smash Mouth over the socket?

ericrosenbaum (approved):
This looks great- just a few suggestions

src/extensions/scratch3_microbit/index.js
khanning: 👍

src/extensions/scratch3_microbit/index.js
khanning: @ericrosenbaum the Cast utility returns 0 if NaN, but 0 is a valid micro:bit pin. I would prefer toNumber to return -1 if it's passed a non-number so I could catch it in whenPinConnected.

src/extensions/scratch3_microbit/index.js
khanning: The mapping is now being done on the device side so I will drop this function until it's needed.

src/extensions/scratch3_microbit/index.js
khanning: 👍

src/extensions/scratch3_microbit/index.js
Kenny2github: **SOMEBODY ONCE TOLD ME** ok I'll stop

src/extensions/scratch3_microbit/index.js
khanning: wrapClamp probably isn't the write strategy here. If a number is given out of range (i.e not 0, 1, or 2) then the function should return false not try to clamp (or wrap clamp) the pin number.

src/extensions/scratch3_microbit/index.js
khanning: 👍

src/extensions/scratch3_microbit/index.js
khanning: This is done automatically because of the maximum bluetooth characteristic length, but it's good practice to catch this in the extension too.

PR #1114, opened 2018-05-04
Fix broken rotation center on vector sprites

kchadha:
### Resolves

Resolves issue with rotation center being broken on stage.

### Proposed Changes

Do not delete bitmap resolution on vector update.

### Reason for Changes

Fix issue found in smoke test.


paulkaplan (approved):
LG!

PR #1118, opened 2018-05-04
Add Thread.StackFrame class

mzgoddard:
### Proposed Changes

Use a private StackFrame class to help internally manage use of Stack Frame values and memory. Create params, reported, and executionContext on demand.

### Reason for Changes

params, reported and executionContext are created whether a stack frame will need them or not. Creating those members on demand in the stack frame saves time instantiating them and later garbage collecting them.

### DIFF
--- a/src/engine/thread.js
+++ b/src/engine/thread.js
@@ -1,4 +1,117 @@
 /**
+ * Recycle bin for empty stackFrame objects
+ * @type Array<_StackFrame>
+ */
+const _stackFrameFreeList = [];
+
+/**
+ * A frame used for each level of the stack. A general purpose
+ * place to store a bunch of execution context and parameters
+ * @param {boolean} warpMode Whether this level of the stack is warping
+ * @constructor
+ * @private
+ */
+class _StackFrame {
+    constructor (warpMode) {
+        /**
+         * Whether this level of the stack is a loop.
+         * @type {boolean}
+         */
+        this.isLoop = false;
+
+        /**
+         * Whether this level is in warp mode.  Is set by some legacy blocks and
+         * "turbo mode"
+         * @type {boolean}
+         */
+        this.warpMode = warpMode;
+
+        /**
+         * Reported value from just executed block.
+         * @type {Any}
+         */
+        this.justReported = null;
+
+        /**
+         * Persists reported inputs during async block.
+         * @type {Object}
+         */
+        this.reported = null;
+
+        /**
+         * Name of waiting reporter.
+         * @type {string}
+         */
+        this.waitingReporter = null;
+
+        /**
+         * Procedure parameters.
+         * @type {Object}
+         */
+        this.params = null;
+
+        /**
+         * A context passed to block implementations.
+         * @type {Object}
+         */
+        this.executionContext = null;
+    }
+
+    /**
+     * Reset all properties of the frame to pristine null and false states.
+     * Used to recycle.
+     * @return {_StackFrame} this
+     */
+    reset () {
+
+        this.isLoop = false;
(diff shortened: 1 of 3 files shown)


src/engine/thread.js
```diff
@@ -1,3 +1,54 @@
...
+        const stackFrame = _stackFrameFreeList.pop();
+        if (typeof stackFrame !== 'undefined') {
+            stackFrame.warpMode = warpMode;
+            return stackFrame;
+        }
+        return new StackFrame(warpMode);
+    }
+
+    static release (stackFrame) {
+        _stackFrameFreeList.push(stackFrame.reset());
```
gnarf: maybe double check that `stackFrame` exists, since the travis test seems to have failed here calling reset on undefined.

src/engine/thread.js
```diff
@@ -1,3 +1,116 @@
...
+        return new _StackFrame(warpMode);
+    }
+
+    /**
+     * Put a stack frame object into the recycle bin for reuse.
+     * @param {_StackFrame} stackFrame The frame to reset and recycle.
+     */
+    static release (stackFrame) {
+        if (typeof stackFrame !== 'undefined') {
+            _stackFrameFreeList.push(stackFrame.reset());
```
rschamp, line 109: How possible would it be for this list to get unhelpfully large?

rschamp (approved):
In my testing this works great and it makes sense to me. Just a question about the possibility of a memory leak — I'm not sure what the situation would be for frames to be continually added to the recycle bin without being removed, but seems like the kind of thing that could lead to things crashing if left running a long time?

src/engine/thread.js
gnarf: We could possibly cap the recycle bin out...

I setup a `setInterval(() => console.log('frames', _stackFrameFreeList.length), 1000);` when reviewing this -- I saw it get up to like ~20 frames in a few of my tests, but never much larger.

We could even just do something like `if (_stackFrameFreeList < MAX_STACK_FRAMES_FREE)` on this push line...  somewhere around 20?

mzgoddard: The list should be limited in growth to the largest stack height for the maximum thread count. With thread count being limited by the clone limit, this is currently limited by that as well.

Trying to be succinct, this grows when the number of active threads running deeply nested control blocks increases exits those control block sections. Otherwise the list would mostly only grow to the peak thread use.

This would never grow past what could actually have been used at a single moment.  It should be really hard to grow to unhelpfully large sizes unless you are growing the number of clones with lots of scripts and threads all running for those scripts and then enter deep frame regions at the same time. In which case you are also contending with the issues that arise from so many clones, threads, drawables, sound players and other resources.

I think the solution we would want if we add one is an interval to occasionally cut down to some count below its target max when the list is at or over the max. Like set the length to 512 if the length is 1024 or larger for example. Stopping it from growing when its at a large number would introduce a temporary performance issue as that means the runtime NEEDS more frames to exist, and has to return to creating and garbage collecting them instead of temporarily letting the available frame set grow and clean it up later. Occasional automated cleanup would introduce some of that cost back but then say for only 1 out of 60 frames if the cleanup was ever second as opposed to 60 out of 60 frames, if it reaches the max and starts creating large amounts of stack frames that are not already available.

-----

This was making me curious about tracking resource usage, probably with the profiler/benchmark.

rschamp: Thanks for the explanation @mzgoddard — given that the size of the list is basically tied directly to the size of the stacks, I think this is good to go, since it would be hard for this to run away, even with clones since those are capped. If we see issues come up we can deal with them then.

PR #1119, opened 2018-05-04
Recycle block args

mzgoddard:
### Proposed Changes

Create an object on the BlockExecuteCache for each block that is assigned the arguments that is passed on to a block function call.

### Reason for Changes

Reuse the args object built for each block function call. The args can be reused in relation to their BlockExecuteCache data. The shape of the args instance is defined by its fields and inputs, and every field and input is assigned each iteration of the same block so old values can <ins>not</ins> persist into a call of a block function with data from another block.

In a performance critical region like `execute` reducing object instantiation greatly improves performance.

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -118,6 +118,28 @@ const FieldKind = {
     DYNAMIC: 'DYNAMIC'
 };
 
+const BlockCached = function (cached) {
+    this._initialized = false;
+
+    this.opcode = cached.opcode;
+    this.fields = cached.fields;
+    this.inputs = cached.inputs;
+    this.mutation = cached.mutation;
+
+    this._isHat = false;
+    this._blockFunction = null;
+    this._definedBlockFunction = false;
+    this._isShadowBlock = false;
+    this._shadowValue = null;
+    this._fields = null;
+    this._fieldKind = FieldKind.NONE;
+    this._fieldVariable = null;
+    this._fieldList = null;
+    this._fieldBroadcastOption = null;
+    this._argValues = null;
+    this._inputs = null;
+};
+
 /**
  * Execute a block.
  * @param {!Sequencer} sequencer Which sequencer is executing.
@@ -132,12 +154,12 @@ const execute = function (sequencer, thread, recursiveCall) {
     const currentStackFrame = thread.peekStackFrame();
 
     let blockContainer = thread.blockContainer;
-    let block = blockContainer.getBlock(currentBlockId);
-    if (typeof block === 'undefined') {
+    let blockCached = BlocksExecuteCache.getCached(blockContainer, currentBlockId, BlockCached);;
+    if (blockCached === null) {
         blockContainer = runtime.flyoutBlocks;
-        block = blockContainer.getBlock(currentBlockId);
+        blockCached = BlocksExecuteCache.getCached(blockContainer, currentBlockId, BlockCached);
         // Stop if block or target no longer exists.
-        if (typeof block === 'undefined') {
+        if (blockCached === null) {
             // No block found: stop the thread; script no longer exists.
             sequencer.retireThread(thread);
             return;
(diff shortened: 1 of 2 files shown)


src/engine/execute.js
```diff
@@ -107,16 +107,161 @@ const handleReport = function (
...
+        // Assign opcode isHat and blockFunction data to avoid dynamic lookups.
+        this._isHat = runtime.getIsHat(opcode);
+        this._blockFunction = runtime.getOpcodeFunction(opcode);
+        this._definedBlockFunction = typeof this._blockFunction !== 'undefined';
+
+        const fieldKeys = Object.keys(fields);
+
+        // Store the current shadow value if there is a shadow value.
+        this._isShadowBlock = fieldKeys.length === 1 && Object.keys(inputs).length === 0;
+        this._shadowValue = this._isShadowBlock && fields[fieldKeys[0]].value;
```
kchadha: I think shadowBlock/shadowValue is not the best name for what this represents. It looks like this refers to any 'field' of a block, whereas shadows are specifically used for block fields that are 'pluggable'. Maybe rename this to _isfieldBlock and _fieldValue?

src/engine/execute.js
```diff
@@ -107,16 +107,161 @@ const handleReport = function (
...
+            } else if (fieldName === 'LIST') {
+                this._argValues.LIST = {
+                    id: fields.LIST.id,
+                    name: fields.LIST.value
+                };
+            } else if (fieldName === 'BROADCAST_OPTION') {
+                this._argValues.BROADCAST_OPTION = {
+                    id: fields.BROADCAST_OPTION.id,
+                    name: fields.BROADCAST_OPTION.value
+                };
```
kchadha: why not combine the above cases into one condition and use fieldName as a reference for the specific fields key?

src/engine/execute.js
```diff
@@ -107,16 +107,161 @@ const handleReport = function (
...
+            } else {
+                this._argValues[fieldName] = fields[fieldName].value;
+            }
+        }
+
+        // Remove custom_block. It is not part of block execution.
+        delete this._inputs.custom_block;
```
kchadha, line 236: I'm not sure I understand what's happening here.

src/engine/execute.js
```diff
@@ -107,16 +107,161 @@ const handleReport = function (
...
+
+            // We can go ahead and compute BROADCAST_INPUT if it is a shadow value.
+            const broadcastInput = this._inputs.BROADCAST_INPUT;
+            if (broadcastInput.block === broadcastInput.shadow) {
+                // Shadow dropdown menu is being used.
+                // Get the appropriate information out of it.
+                const shadow = blockContainer.getBlock(broadcastInput.shadow);
+                const broadcastField = shadow.fields.BROADCAST_OPTION;
+                this._argValues.BROADCAST_OPTION.id = broadcastField.id;
+                this._argValues.BROADCAST_OPTION.name = broadcastField.name;
```
kchadha: the RHS on this line should read `broadcastField.value`

kchadha (changes requested):
I had some minor comments/questions and found a bug.

src/engine/execute.js
mzgoddard: This is how shadow values are stored in blocks. They are are in blocks with no inputs and 1 field. These two members are specifically used in execute by a block of code that determines the block is a block containing a shadow value and reports the shadow value in the only field.

src/engine/execute.js
mzgoddard: `custom_block` is information stored in inputs of a block that determines non-execution details. Before BlocksExecuteCache the for loop on inputs in execute skipped any `inputName === 'custom_block'` inputs. Every input of every executed block made that test. With BlocksExecuteCache we can make a clone of `inputs` and remove the ignored field for a performance gain.

src/engine/execute.js
mzgoddard: Here's the comment from the relevant block in execute.

```js
// Hats and single-field shadows are implemented slightly differently
// from regular blocks.
// For hats: if they have an associated block function, it's treated as a
// predicate; if not, execution will proceed as a no-op. For single-field
// shadows: If the block has a single field, and no inputs, immediately
// return the value of the field.
```

mzgoddard: Instead of changing the names I made the `_isShadowBlock` containing an additional detail for whether its a shadow block or not. Now `_isShadowBlock` is true when the block can only refer to a block and false when it is not a shadow block.

mzgoddard:
Thank you for your review @kchadha.

rschamp (approved):
This looks great, thanks @mzgoddard. I will let @kchadha give the final sign-off.

kchadha (approved):

PR #1121, opened 2018-05-05
Initial version of the speech extension.  There is certainly still so…

picklesrus:
…me stuff to figure out but the core of it is here.

### Proposed Changes

Add the core changes for the speech extension.

### Reason for Changes

Because people like to talk to stuff.

### Test Coverage

Still TBD


picklesrus:
Number 3 below is slightly concerning and magnifies the "this code is fragile" comment I made earlier :)  I'm not sure why (or when) it changed but the opcodes for the extension blocks used to be extensionName.block and are now extensionName_block so I changed the test in the if to be speech_ instead. 🙃 Hopefully there is a better way to do that part of the code.

Did the things we chatted about offline:
1. extension manager now has speech in it.
2. moved over to using arraybuffer-loader
3. fixed the phrase list.

ericrosenbaum:
Let's initialize _currentUtterance to the empty string, so that it doesn't report "null" before anything has been recognized.

ericrosenbaum:
Looks like the comments for finalResponseTimeoutDurationMs and listenAndWaitBlockTimeoutMs need clarification to distinguish between them

ericrosenbaum:
I'm often getting:
`vm Not sending data because not in ready state. State: 2`
or
`vm Not sending data because not in ready state. State: 3`
Not sure why- things otherwise seem to be working as expected.

picklesrus:
Whoops. Fixed that comment and current_utterance initialization.

I'm curious how you're getting those errors - do you have a project/block set you can share? I'd love to be able to repro to figure out why it happens.  Also, there's a question marked in the code, but since you're now seeing this for real, would it make sense to call resetListening there?  I imagine the error throws continuously since we're probably still processing the audio data? We could also try to re-open the socket in addition/instead?

ericrosenbaum:
I saw this sequence of results, suggesting there are problems with the way fuzzy matching is being used (I think). I had a hat block with "go left" in it:

vm Got result: {"alternatives":[{"words":[],"transcript":"go left","confidence":0}],"isFinal":false,"stability":0.009999999776482582}

vm Keeing result: eft

picklesrus:
I think I solved the closed socket issue - the socket was getting closed but then more data came in through the processAudioCallback and we gave up.  I'm now removing the listener on the scriptNode at the same time that I call disconnect on it.  Then I put the listener back on when listening resumes.

picklesrus:
For the keeping "eft" bit:
I think this could be a result of this bit of code with the comment ~line 448 that starts with "Decide whether this is the right thing"  I removed it for now -  See commit https://github.com/LLK/scratch-vm/pull/1121/commits/3eec3170b2008b582636f52623c8186f4f8d114d

The reason I put it there initially is because of the case where the fuzzy match matches the thing in the when I hear block but isn't exactly what you said.  Example: put "card" in a when I hear hat, and say "car".  The when I hear hat triggers (for me at least) and the transcription result is "car".  It seemed odd to me that the value in the speech reporter didn't match the value in the "When I hear" hat, yet the hat triggered.
So, I could either leave it like this, or perhaps instead of putting the result from the fuzzy matcher, I should set it to the value from the when I hear hat.  Thoughts? I could also rip out fuzzy matching altogether, though I do remember thinking it made things a bunch better when I originally added it... Things change though :)

ericrosenbaum:
My intuition right now is that the the speech reporter's value should be the "raw" value, and not be affected by the hat block contents. That way the hat blocks with fuzzy matching will have a slightly more forgiving behavior, but the speech reporter will show you what the system is actually hearing.

picklesrus:
Okay. I'll leave it as is for now then.

picklesrus:
I think so far the code reflects everything we've talked about.  Let me know if there are more issues you've found!

ericrosenbaum (approved):
Hooray! Now we can talk to Scratch!

PR #1122, opened 2018-05-05
Handle random backdrop menu option

### LINKED ISSUE #1500: Switch to random backdrop
### Expected Behavior

It would be nice if `switch backdrop to %1` included `random backdrop` as one of the options.

### Actual Behavior

The options list contains the current backdrops along with 'next' and 'previous'

Kenny2github:
### Resolves

Resolves LLK/scratch-blocks#1500
Must be merged BEFORE LLK/scratch-gui#1989

### Proposed Changes

Changes the `_setCostumeOrBackdrop` function to set the backdrop to backdrop no. `Math.floor(Math.random() * target.getCostumes().length))` if the `requestedCostume === 'random backdrop'`.

### Reason for Changes

To resolve a help wanted issue.

### Test Coverage

Existing tests pass.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -346,11 +346,9 @@ class Scratch3LooksBlocks {
             } else if (requestedCostume === 'random backdrop') {
                 const numCostumes = target.getCostumes().length;
                 if (numCostumes > 1) {
-                    let index;
-                    do {
-                        index = Math.floor(Math.random() * numCostumes);
-                    } while (index === target.currentCostume);
-                    target.setCostume(index);
+                    let selectedIndex = Math.floor(Math.random() * (numCostumes - 1));
+                    if (selectedIndex === target.currentCostume) selectedIndex += 1;
+                    target.setCostume(selectedIndex);
                 }
             } else {
                 const forcedNumber = Number(requestedCostume);
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -343,6 +343,8 @@ class Scratch3LooksBlocks {
             } else if (requestedCostume === 'next costume' ||
                        requestedCostume === 'next backdrop') {
                 target.setCostume(target.currentCostume + 1);
+            } else if (requestedCostume === 'random backdrop') {
+                target.setCostume(Math.floor(Math.random() * target.getCostumes().length));
             } else {
                 const forcedNumber = Number(requestedCostume);
                 if (!isNaN(forcedNumber)) {
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -344,7 +344,8 @@ class Scratch3LooksBlocks {
                        requestedCostume === 'next backdrop') {
                 target.setCostume(target.currentCostume + 1);
             } else if (requestedCostume === 'random backdrop') {
-                if ((numCostumes = target.getCostumes().length) > 1) {
+                const numCostumes = target.getCostumes().length;
+                if (numCostumes > 1) {
                     let index;
                     do {
                         index = Math.floor(Math.random() * numCostumes);
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -344,7 +344,13 @@ class Scratch3LooksBlocks {
                        requestedCostume === 'next backdrop') {
                 target.setCostume(target.currentCostume + 1);
             } else if (requestedCostume === 'random backdrop') {
-                target.setCostume(Math.floor(Math.random() * target.getCostumes().length));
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -343,6 +343,8 @@ class Scratch3LooksBlocks {
             } else if (requestedCostume === 'next costume' ||
                        requestedCostume === 'next backdrop') {
                 target.setCostume(target.currentCostume + 1);
+            } else if (requestedCostume === 'random backdrop') {
+                target.setCostume(Math.floor(Math.random() * target.getCostumes().length));
```
kchadha: I think it would be nice to skip the currently selected backdrop when choosing a random one to switch to. That way, every time the user executes the switch backdrop to random block, it always switches backdrops (if there's more than one available).

You can do this by calculating a random number between 0 and one less than the number of backdrops. Then, if the selected random number is greater than or equal to the currently selected costume index (`target.currentCostume`), add 1 to it.

kchadha (commented):
@Kenny2github, I think the code looks good, but I have a small suggestion in the comments. Let me know if it makes sense or if you have any questions.

src/blocks/scratch3_looks.js
towerofnix: > Then, if the selected random number is greater than or equal to the currently selected costume index (target.currentCostume), add 1 to it.

Would this cause issues if the current costume is the last costume? So adding 1 to the costume number would make it switch to a nonexistant costume. You could make it so that it `%`'s the picked index plus one with the number of costumes, so that in this case it would select `1` instead of `(number of costumes)`. But then it would pick `1` more often than the other numbers, because there would be two cases that could result in `1`: randomly selecting `1`, or randomly selecting the current costume index which is the last costume.

A solution would be to keep picking random indexes while the picked index is equal to the current index:

```js
let index;
do {
  index = Math.floor(Math.random() * target.getCostumes().length);
} while (index === target.currentCostume);
target.setCostume(index);
```

You would also have to make sure that there *is* more than one costume to select, though; otherwise it would be a never-ending loop!

Kenny2github:
I went with @towerofnix's solution, with a slight tweak to account for the last problem with their method they mention. If the number of costumes is not greater than 1 (i.e. this is the only costume), do nothing; otherwise, keep choosing a random index until the index is not the current index, then switch to that index.

src/blocks/scratch3_looks.js
```diff
@@ -344,7 +344,13 @@ class Scratch3LooksBlocks {
                        requestedCostume === 'next backdrop') {
                 target.setCostume(target.currentCostume + 1);
             } else if (requestedCostume === 'random backdrop') {
-                target.setCostume(Math.floor(Math.random() * target.getCostumes().length));
+                if ((numCostumes = target.getCostumes().length) > 1) {
```
towerofnix: Watch out - this probably makes a global `numCostumes` variable, since assigning to a variable that does not exist creates a property on `window`. Probably better to do `const numCostumes = target.getCostumes().length` before.

Kenny2github: Huh? You're reviewing an outdated version - that's exactly what I did.

towerofnix: Whoops. I'm not sure why GitHub decided to notify me about a commit here, but then show me the changes of the oldest new commit instead of all the new changes.

src/blocks/scratch3_looks.js
kchadha: @towerofnix, sorry for the late response, but the proposed solution does not cause problems if the currentCostume is the last costume. We'd be calculating a random number capping at one less than the last index, so that adding 1 to it only gives us the last index. `setCostume` clamps inputs that are out of bounds anyways. see above for the `next backdrop` and `previous backdrop` cases which already rely on the `setCostume` clamp working.

kchadha:
@Kenny2github, @towerofnix, see my response to @towerofnix's original comment/proposal. If my comment makes sense to you, I think I prefer my originally proposed solution instead of continuously calculating a random number every time currently selected costume is picked (this would happen pretty frequently when the number of costumes is small), and is unnecessary.

my proposed solution is to do something like the following inside your `if (numCostumes  > 1)` block:
```
let selectedIndex = Math.floor(Math.random() * (numCostumes - 1));
if (selectedIndex === target.currentCostume) selectedIndex += 1;
target.setCostume(selectedIndex);
```

src/blocks/scratch3_looks.js
towerofnix: Ah, got it! Sounds good.

Kenny2github:
@kchadha done **Done**

kchadha (approved):
Looks good! Thanks for making the changes @Kenny2github!

PR #1126, opened 2018-05-08
Run quirks mode fixes on costumes loaded from sb2s

fsih:
This is part of the work to run quirks mode updates only on files loaded from SB2s. This should also ensure that the renderer, thumbnails, and paint editor are in sync in terms of what svg they're getting. Previously, the paint editor and renderer could get different fonts.

Depends on https://github.com/LLK/scratch-2.0-coping-strategies/pull/16


fsih:
@cwillisf in terms of next steps, would it make sense to take transformText and transformMeasurements out of SVG loading in the renderer?

src/import/load-costume.js
```diff
@@ -31,7 +33,24 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
     if (costumeAsset.assetType === AssetType.ImageVector) {
         // createSVGSkin does the right thing if rotationCenter isn't provided, so it's okay if it's
         // undefined here
```
kchadha: This comment should move down to where renderer.createSVGSkin is being called.

src/import/load-costume.js
```diff
@@ -31,7 +33,24 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
...
-        costume.skinId = renderer.createSVGSkin(costumeAsset.decodeText(), rotationCenter);
+        let svgString = costumeAsset.decodeText();
+        if (costume.version && costume.version === 2) {
+            // SVG Renderer load fixes "quirks" associated with Scratch 2 projects
```
kchadha: Can we move this comment above the if statement and add a note about costume.version here as well (that it should only ever exist for scratch 2 costumes?

src/import/load-costume.js
```diff
@@ -31,7 +33,24 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
...
+            const svgRenderer = new SvgRenderer();
+            svgRenderer.loadString(svgString);
+            svgString = svgRenderer.toString();
+            delete costume.version;
```
kchadha: As I understand it, costume.version is getting created for all sb2 costumes (vector and bitmap) but it's only getting deleted for the vector costumes. It should also get deleted for bitmap costumes (if it exists).

kchadha (commented):
Code changes look good, some minor change requests in the comments.

src/import/load-costume.js
fsih: Good catch! Let me know if this makes sense, or maybe we should pass it without messing around with the costume object.

kchadha: I think you might be right that it makes more sense to add an optional argument at the end of `loadCostume` that only ever gets passed in from sb2 costumes. I don't think any of the other arguments are optional, so I think we won't need to modify the other calls to loadCostume.
It's possible that this only works out because none of the costumes/backdrops from the library have text in them (is that true?)...

Is there any other way to identify, from the SVG data itself that it's a Scratch 2.0 SVG?

src/import/load-costume.js
```diff
@@ -29,9 +31,29 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime) {
...
+
+            // Put back into storage
+            const storage = runtime.storage;
+            costume.assetId = storage.builtinHelper.cache(
+                storage.AssetType.ImageVector,
+                storage.DataFormat.SVG,
+                (new TextEncoder()).encode(svgString)
+            );
```
cwillisf: Could you use `costumeAsset.encodeTextData(...)` here instead of calling the `TextEncoder` directly? If so you might be able to skip the `TextEncoder` dependency in this repository.

cwillisf (commented):
looks good -- one question, though.

src/import/load-costume.js
fsih: Cool! I tried to switch it to use encodeTextData

src/import/load-costume.js
fsih: I switched it to an argument.

I think SVGs from 2.0 can be identified by all of them containing a comment
`
`
We might want to add a different one for 3.0

cwillisf (approved):

cwillisf:
(pending `jimp` test fix of course)

kchadha (approved):
LGTM but I think @cwillisf should also take another look for a sanity check.

cwillisf (approved):
Looks great!

PR #1127, opened 2018-05-08
Update scratch-parser.

kchadha:
Update scratch-parser version to fix issue with uploading sprites (that don't have any sounds) from .sprite2 files (e.g. bananas sprite from scratch 2.0 library).

Pin scratch-parser to specific version so that greenkeeper PRs get created for any updates to scratch-parser.


paulkaplan (approved):
LG, might need to talk to @rschamp about this test failure, looks like it is related to the recent webpack update?

PR #1128, opened 2018-05-08
Import monitors from sb2 files.

### LINKED ISSUE #343: Import watchers/monitors in SB2 importer
Moving from TODO comment. Currently these are ignored immediately.

### LINKED ISSUE #683: Import Doesn't Display Variable Monitors
### Expected Behavior

A visible monitor in a saved Scratch 2 project should remain visible.

### Actual Behavior

It doesn't load.

### Steps to Reproduce

- Make any project with a visible variable monitor in 2.0
- Load said project into 3.0
- Note that the monitor is missing

Ubuntu 17.03, Chrome 61.0.3163.100

### Cause

The parseScratchObject function doesn't handle the stage monitor object when loading correctly. Specifically:
The stage monitor object looks like:
```
cmd:"getVar:"
color:15629590
isDiscrete:true
label:"test"
mode:1
param:"test"
sliderMax:100
sliderMin:0
target:"Stage"
visible:true
x:5
y:5
```
The parseScratchObject looks for a name, since there is none, it discards the object
Furthermore, it doesn't handle command events, labels, etc.

### Suggested Solution

1. Add a new function 'parseScratchCommand'
2. Before calling parseScratchObject on children, check if child is command, if it is, call parseScratchCommand instead of parseScratchObject.

### LINKED ISSUE #790: Monitors can cause undefined variables to appear
### Steps to Reproduce
1. Create a (global) variable on Sprite1
2. Drag it out to the workspace
3. Add a sprite Sprite2
4. Delete the variable from Sprite2.
Note: This deletes the monitor block that monitors the variable, even though the block was actually not deleted globally because it still exists in sprite 1.

5. Switch to sprite1.
6. Turn on the monitor for the variable in sprite 1
Note: This calls lookup or create variable when the monitor runs. Since no monitor block exists for this variable, it creates a variable. Note that look up or create variable creates a _local_ variable.
7. Switch to sprite 2.
Since the variable is local to the other sprite, this does the same thing on sprite 2 and creates a local variable. However it has no name and so an undefined variable appears.

Sometimes you have to switch between the sprites a couple of times here for the undefined variable to appear

### What should happen?
It's unclear what the state of the variable should be when it's deleted from one of the sprites. Should it actually be a local variable? In this case this bug will be fixed when we get around to implementing per-sprite variables and per-sprite variable monitors.

### LINKED ISSUE #1848: Monitors persist if you switch projects
### Expected Behavior

Monitors from the old project disappear

### Actual Behavior

Monitors stick around

### Steps to Reproduce
1. Go to https://llk.github.io/scratch-gui/develop/
2. Turn on a bunch of monitors. E.g. click the checkmark next to x position
3. Change the URL to navigate to a different project. For instance add #73399968 to the end

The monitors from the old project are still there, even though sprite 1 no longer exists

Windows 10 Firefox

paulkaplan:
Paired with @kchadha on all of this.

### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/343
Fixes https://github.com/LLK/scratch-vm/issues/790
Fixes https://github.com/LLK/scratch-vm/issues/683
Fixes https://github.com/LLK/scratch-gui/issues/1848

### Proposed Changes

Add a step, deferred until after targets are created, where the runtime processes monitors from sb2 files.

Also make the sprite-specific nature of certain monitors more explicit and consistent. 

Also support the fixed monitor ids for sprite specific monitors, etc.

### Reason for Changes

Monitors! We need them.

### Test Coverage

Added a kind of "kitchen-sink" sb2 import test with a fixture file that looks like this
image

And I added a lot of assertions that cover:
- monitor types set correctly (all 4)
- monitor visibility set correctly
- monitor threads not running if not visible
- monitor positioning imported
- monitor width/height for lists
- custom monitor slider range
- sprite specificity


paulkaplan:
@kchadha I updated this with fixes for the existing tests and a new fixture/integration test that covers a lot of the edge cases we found.

src/blocks/scratch3_sensing.js
```diff
@@ -74,10 +74,18 @@ class Scratch3SensingBlocks {
...
+                getId: () => 'answer'
+            },
+            sensing_loudness: {
+                getId: () => 'loudness'
+            },
+            sensing_timer: {
+                getId: () => 'timer'
+            },
+            sensing_current: {
+                getId: (_, param) => `current_${param}`
```
kchadha: We should probably add a note here about why this id is different from what's in the default toolbox xml. E.g. we want monitors to be imported as separately from sb2, but we want the block in the toolbox to only ever create one monitor.

I also just realized that this getId function is specific to sb2 now, (e.g. I won't be able to use it in sb3 deserialization) but we can potentially cross that bridge when we get there.

src/engine/blocks.js
```diff
@@ -441,22 +441,34 @@ class Blocks {
...
-                    // @todo(vm#564) this will collide if multiple sprites use same block
                     id: block.id,
                     targetId: block.targetId,
                     spriteName: block.targetId ? optRuntime.getTargetById(block.targetId).getName() : null,
                     opcode: block.opcode,
                     params: this._getBlockParams(block),
                     // @todo(vm#565) for numerical values with decimals, some countries use comma
-                    value: ''
+                    value: '',
+                    mode: block.opcode === 'data_listcontents' ? 'list' : 'default'
```
kchadha, line 471: yay for making mode a string instead of a number! =]

src/serialization/sb2.js
```diff
@@ -208,6 +209,104 @@ const globalBroadcastMsgStateGenerator = (function () {
     };
 }());
 
+/**
+ * Parse a single monitor object and create all its in-memory VM objects.
+ * @param {!object} object - From-JSON "Scratch object"
```
kchadha: We can provide more info here. The `object` param should represent a monitor and not a sprite (which also qualifies as a 'Scratch object').

src/serialization/sb2.js
```diff
@@ -208,6 +209,104 @@ const globalBroadcastMsgStateGenerator = (function () {
...
+                return variable.type === Variable.LIST_TYPE && variable.name === object.listName;
+            });
+            if (listVariables.length > 0) {
+                target = currTarget; // Keep this target for later use
+                object.target = currTarget.getName(); // Set target name to normalize with other monitors
+            }
+        }
+    }
+
+    // Create a block for the monitor blocks container
```
kchadha: This comment doesn't correspond with the lines below it.

src/serialization/sb2.js
```diff
@@ -208,6 +209,104 @@ const globalBroadcastMsgStateGenerator = (function () {
...
+        [object.cmd, object.param], // Scratch 2 monitor blocks only have one param.
+        null, // `addBroadcastMsg`, not needed for monitor blocks.
+        getVariableId,
+        extensions
+    );
+
+    let isSpriteLocalVariable;
+    if (object.cmd === 'getVar:' || object.cmd === 'contentsOfList:') {
+        // These monitors are sprite-specific if they are not targetting the stage.
+        isSpriteLocalVariable = !target.isStage;
```
kchadha: Just realized... at this point, don't we know that any monitor is sprite specific if it's `target` property is not `'Stage'`? We don't need to do a check for local variables/lists (we normalized lists above).

src/serialization/sb2.js
```diff
@@ -208,6 +209,104 @@ const globalBroadcastMsgStateGenerator = (function () {
...
+        // Variable getters have special block IDs for the toolbox that match the variable ID.
+        block.id = getVariableId(object.param);
+    }
+
+    block.id = runtime.monitorBlockInfo.hasOwnProperty(block.opcode) ?
```
kchadha: I think there should be a comment here which explains that `runtime.monitorBlockInfo` has all monitored block opcodes except `data_variable` and `data_listcontents`, so rewriting the id here doesn't rewrite the id we just set in the block above.

src/serialization/sb2.js
```diff
@@ -208,6 +209,104 @@ const globalBroadcastMsgStateGenerator = (function () {
...
+        runtime.monitorBlockInfo[block.opcode].getId(target.id, object.param) : block.id;
+
+    // Block needs a targetId if it is sprite specific or a local variable.
+    // Consult the monitorBlockInfo in the runtime for sprite-specificity.
+    const isSpriteSpecific = isSpriteLocalVariable ||
+        (runtime.monitorBlockInfo.hasOwnProperty(block.opcode) &&
+        runtime.monitorBlockInfo[block.opcode].isSpriteSpecific);
```
kchadha: we make the `runtime.monitorBlockInfo.hasOwnProperty(block.opcode)` at least twice now, it probably makes sense to make that check and pull out the monitorBlockInfo into a variable up above

test/integration/monitors.js
```diff
@@ -0,0 +1,95 @@
...
+const test = require('tap').test;
+const makeTestStorage = require('../fixtures/make-test-storage');
+const readFileToBuffer = require('../fixtures/readProjectFile').readFileToBuffer;
+const VirtualMachine = require('../../src/index');
+const Variable = require('../../src/engine/variable');
+
+const projectUri = path.resolve(__dirname, '../fixtures/monitors.sb2');
+const project = readFileToBuffer(projectUri);
+
+test('complex', t => {
```
kchadha: Should this test be called something more descriptive?

src/serialization/sb2.js
paulkaplan: Do you mean we could replace the "isSpriteSpecific" logic with just a `!target.isStage`?

paulkaplan:
@kchadha thanks for the review, updated to all the comments. Did some extra refactoring of the `isSpriteSpecific` area, should make it simpler to read.

kchadha (approved):
Changes look good to me!

PR #1133, opened 2018-05-09
Add implementations for hiding and showing variables and lists.

### LINKED ISSUE #308: Implement monitor manager, show/hide variable/list blocks
Needs to interact with the scratch-blocks flyout checkboxes and https://github.com/LLK/scratch-render/issues/19 to display monitors when appropriate. Also need to update the SB2 importer to know when to display monitors by default.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/308

### Proposed Changes

Implement the show/hide list and variable blocks by simply issuing the same change event to the runtime monitor blocks as scratch-blocks does when the monitor checkboxes are checked.

### Reason for Changes

The opcodes need to be implemented in the VM. 

Note: to get them back into the toolbox, we need to change that in blocks.

Tested using this example project: #221755484

It has keyboard controls for show/hide variable ("space", "left arrow") and show/hide list ("a", "b").

---

@kchadha currently the blocks container responds to these events by slicing the monitor record out of the monitor state, which like we talked about will lose some state we might want to keep track of. In the future, as an improvement, we could follow up with this by setting the visibility instead of removing the monitor.

### DIFF
--- a/src/blocks/scratch3_data.js
+++ b/src/blocks/scratch3_data.js
@@ -18,6 +18,8 @@ class Scratch3DataBlocks {
             data_variable: this.getVariable,
             data_setvariableto: this.setVariableTo,
             data_changevariableby: this.changeVariableBy,
+            data_hidevariable: this.hideVariable,
+            data_showvariable: this.showVariable,
             data_listcontents: this.getListContents,
             data_addtolist: this.addToList,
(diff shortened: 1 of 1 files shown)


kchadha (approved):
Looks good!

PR #1140, opened 2018-05-10
Add translate extension to the extension manager.

picklesrus:
### Proposed Changes

Add translate extension to extension manager so all changes to enable it are gui only.

### Reason for Changes

For now just makes my life easier... :)  But someday it will be necessary anyway.


ericrosenbaum (approved):
LGTM!

PR #1142, opened 2018-05-11
Update tx/config for Scratch 3.0

chrisgarrity:
Update transifex config to be able to use transifex command line to sync with Scratch 3.0 editor project on transifex..


thisandagain (approved):
🎉

PR #1144, opened 2018-05-11
Re-apply "Dynamically load music extension manifest"

rschamp:
Reverts LLK/scratch-vm#1143 (Re-applies #1105 )

This is a good optimization for file size and memory usage, but we need to resolve configuring scratch-gui before pulling this in.


thisandagain:
Closing until @rschamp and @ericrosenbaum have time to readdress.

PR #1147, opened 2018-05-15
Fix say bug with show/hide

### LINKED ISSUE #2078: Show block shows old speech/thought bubbles
### Expected Behavior

When you run a Show block it should not show speech or thought bubbles that you haven't told to appear.

### Actual Behavior

When you run a show block it will cause the speech or thought bubble that was previously run.

saybubbleshow

### Steps to Reproduce

Launch the editor
Drag a "Say hello for 2 seconds" block onto the work area
Drag a "Show" block onto the work area
Click the Show block
Note no speech bubbles appear
Click the Say Hello for 2 seconds block
Note the speech bubble appears and then disappears 2 seconds later
Click the Show block
Note that the say bubble reappears
Note that the say bubble never disappears (unless you stop the program)

This also works with the "Think" block
After getting the thought bubble to appear with the show block you can get it to go away by running the say for 2 seconds block again.

Chrome on Mac
Also on Firefox on Mac

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/2078

### Proposed Changes

Revert a small part of the change from https://github.com/LLK/scratch-vm/pull/1089. Make sure that the bubble state text is set to blank after a timed say/hide. 

### Reason for Changes

Previously it was not clearing the say text, but there was a second flag which considered whether to show/hide the bubble. We removed that other flag in https://github.com/LLK/scratch-vm/pull/1089, so now only the text of the bubble is used to decide if it is visible. The bubble drawable gets torn down when the target is hidden, but the text is left around in case the target is re-shown. Make sure to explicitly set the text to '' in timed say/hide to make sure it doesn't come back. Intervening say/thinks are handled by the `usageId` conditional. 

### Test Coverage

I repeated the steps for the bug: 
- say for 2 seconds
- hide
- show (bubble came back, no longer does)

as well as a few other tests for the bubbles, including 
- say
- hide
- show (bubble should come back)

and made sure that https://github.com/LLK/scratch-gui/issues/1817 did not come back.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -291,7 +291,7 @@ class Scratch3LooksBlocks {
                 this._bubbleTimeout = null;
                 // Clear say bubble if it hasn't been changed and proceed.
                 if (this._getBubbleState(target).usageId === usageId) {
-                    this._onTargetWillExit(target);
+                    this._updateBubble(target, 'say', '');
                 }
                 resolve();
             }, 1000 * args.SECS);
@@ -311,7 +311,7 @@ class Scratch3LooksBlocks {
                 this._bubbleTimeout = null;
                 // Clear think bubble if it hasn't been changed and proceed.
                 if (this._getBubbleState(target).usageId === usageId) {
-                    this._onTargetWillExit(target);
+                    this._updateBubble(target, 'think', '');
                 }
                 resolve();
             }, 1000 * args.SECS);


cwillisf (commented):
Do we need the other functionality from the `_onTargetWillExit` method that these cases were calling before? Specifically:
- destroying the render components (I assume we don't need this?)
- requesting a redraw (this might be important?)
- removing the `EVENT_TARGET_VISUAL_CHANGE` listener (seems like we definitely don't want this?)

cwillisf:
Oops, sorry about that: I meant to unassign myself and accidentally unassigned you, @paulkaplan. Rude! :/

paulkaplan:
@cwillisf onTargetWillExit still gets called, it gets called because the text is empty string https://github.com/paulkaplan/scratch-vm/blob/c18d5faafb89db9ab020900a5e3fe0cf1b286fc9/src/blocks/scratch3_looks.js#L182

cwillisf (approved):

PR #1148, opened 2018-05-16
Layer group ordering.

### LINKED ISSUE #1066: "go to [back]" and "go [backward] (1) layers" can cause sprites to go behind the pen layer
### Expected Behavior

- Sprites should never render behind the pen layer

### Actual Behavior

- The "go to [back]" block can cause sprites to go behind the pen layer

### Steps to Reproduce

https://scratch.mit.edu/projects/217041627 (Run in Turbo Mode)
https://llk.github.io/scratch-gui/develop/#217041627 (Run in Turbo Mode)

kchadha:
### Resolves

#1066

### Proposed Changes

Create layer groups. See LLK/scratch-render#285 for a more detailed description of how they work.
Note, I wasn't sure where to put the stage layering contents. I put it in src/engine for now. I wanted to put it in runtime, but then there's a circular dependency between runtime and video.

### Reason for Changes

#1066.

### Test Coverage

Existing tests pass.

#### Related PRs
LLK/scratch-render#285.
The tests for render depend on this PR so it should be merged first.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -15,6 +15,7 @@ const StringUtil = require('../util/string-util');
 const specMap = require('./sb2_specmap');
 const Variable = require('../engine/variable');
 const MonitorRecord = require('../engine/monitor-record');
+const StageLayering = require('../engine/stage-layering');
 
 const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
@@ -327,7 +328,7 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
     // Blocks container for this object.
     const blocks = new Blocks();
     // @todo: For now, load all Scratch objects (stage/sprites) as a Sprite.
-    const sprite = new Sprite(blocks, runtime, topLevel /* whether this sprite is a stge or not */);
+    const sprite = new Sprite(blocks, runtime);
     // Sprite/stage name from JSON.
     if (object.hasOwnProperty('objName')) {
         sprite.name = topLevel ? 'Stage' : object.objName;
@@ -399,7 +400,7 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
     }
 
     // Create the first clone, and load its run-state from JSON.
-    const target = sprite.createClone();
+    const target = sprite.createClone(topLevel ? StageLayering.BACKGROUND_LAYER : StageLayering.SPRITE_LAYER);
 
     const getVariableId = generateVariableIdGetter(target.id, topLevel);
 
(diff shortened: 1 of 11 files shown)


src/serialization/sb2.js
```diff
@@ -327,7 +327,7 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
     // Blocks container for this object.
     const blocks = new Blocks();
     // @todo: For now, load all Scratch objects (stage/sprites) as a Sprite.
-    const sprite = new Sprite(blocks, runtime);
+    const sprite = new Sprite(blocks, runtime, topLevel /* whether this sprite is a stge or not */);
```
cwillisf: `s/stge/stage/`

src/sprites/rendered-target.js
```diff
@@ -912,7 +920,7 @@ class RenderedTarget extends Target {
         newClone.effects = JSON.parse(JSON.stringify(this.effects));
         newClone.variables = JSON.parse(JSON.stringify(this.variables));
         newClone.lists = JSON.parse(JSON.stringify(this.lists));
-        newClone.initDrawable();
+        newClone.initDrawable(false); // this,sprite is not a stage if we're calling makeClone on it
```
cwillisf: `s/,/ /`?

src/serialization/sb3.js
```diff
@@ -684,7 +684,7 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
     const blocks = new Blocks();
 
     // @todo: For now, load all Scratch objects (stage/sprites) as a Sprite.
-    const sprite = new Sprite(blocks, runtime);
+    const sprite = new Sprite(blocks, runtime, object.isStage);
```
cwillisf: Instead of adding the `isStage` ctor arg and corresponding property on all `Sprite` objects, what do you think about adding a little bit of code here (and on the SB2 side) to switch the sprite's layer group if `object.isStage` is true?
I guess that might require an API change where calling `setDrawableOrder` without specifying a layer group keeps a sprite in its existing layer group, but that doesn't sound like a bad thing to me.

cwillisf (dismissed):
This looks great! I did notice a typo or two and have an idea about the API for you to consider, but I think this approach is fine if there's a reason to prefer it.

kchadha:
@cwillisf, this and the scratch-render PR are ready for re-review. Things have changed significantly (only for me to find out that the changes weren't actually necessary for performance), but I think this is  cleaner and only covers the use cases we have at hand.

cwillisf (approved):
This looks even better than the previous version :D
I'm especially glad you were able to remove the extra `Sprite` constructor arg, but a lot of the other stuff looks cleaner too!

PR #1153, opened 2018-05-17
Add option for Extensions to have different colour

### LINKED ISSUE #1151: Extensions default to the same colour
### Expected Behavior

Specifying primary, secondary and tertiary colours on an extension should allow the extension to use those colours instead of the default hardcoded colours

### Actual Behavior

All extensions have the same colour scheme, without the ability to change it 

### Steps to Reproduce

Add multiple extensions

Mac OS 10.13.3, Chrome Version 66.0.3359.139 (Official Build) (64-bit)

vincentbriglia:
### Resolves

#1151 

### Proposed Changes

This pull request allows for an extension category to identify blocks in a separate colour

### Reason for Changes

We need a different colour than pink, as we are treating each category as a separate extension

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,9 @@ npm-*
 /.nyc_output
 /coverage
 
-# IDEA
+# Editor
 /.idea
+/.vscode
 
 # Build
 /dist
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
Looks good -- thanks!

PR #1154, opened 2018-05-17
Allow override of icons per block

### LINKED ISSUE #1152: Extensions should also allow for icons per block
### Expected Behavior

Specifying a blockIconURI per block should show a different icon

### Actual Behavior

There is no option to add an icon per block, only currently on the category level

### Steps to Reproduce

Api is missing, this is default behavior at the moment, this is a feature request

Mac OS 10.13.3, Chrome Version 66.0.3359.139 (Official Build) (64-bit)

vincentbriglia:
### Resolves

#1152 

### Proposed Changes

current behavior of Extension icons: 

Use a menu icon if there is one. Otherwise, use the block icon. If there's no icon, the category menu will show its default colored circle.

this adds the capability to specify an icon per block if it needs to be overwritten, instead of using the specified blockIconURI from the extension specification itself.

### Reason for Changes

It would allow us to specify individual icons on the blocks, without having to add this in the block definition itself, to avoid mixing the location where to add icons

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,8 +10,9 @@ npm-*
 /.nyc_output
 /coverage
 
-# IDEA
+# Editor
 /.idea
+/.vscode
 
 # Build
 /dist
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
Cool - I like this approach. Thanks!

PR #1156, opened 2018-05-17
Publish specific files and folders to NPM

mzgoddard:
### Proposed Changes

- Publish only `src`, `dist`, `TRADEMARK`, `README.md`, `package.json`, and `LICENSE` to npm

### Reason for Changes

Other packages that depend on scratch-vm need to download it and locally install a copy to require into. As scratch-vm is, it's publishing `playground` and `test`. These folders are not used by scratch-vm dependents. These directories take up a relatively large amount of space in the download file and once extracted on disk. Importantly this extra space means that CI servers along with the developers working on another scratch-* package need to spend more time downloading and for CI, more time storing and retrieving the node_modules cache to perform CI work quicker.

Specifying the files published with this change, reduces the download from **27MB** to **15MB**, and space used once extracted from **64MB** to **35MB**. Most of the savings is from omitting `playground`.

-----

`README.md`, `package.json` and `LICENSE` are automatically included in the files published to npm whether they are stated in `package.json` or not. I can add those if the explicit nature of stating them that way would be preferred.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,20 @@
+# Development files
+.eslintrc.js
+/.editorconfig
+/.eslintignore
+/.gitattributes
+/.github
+/.travis.yml
+/.tx
+/test
+/webpack.config.js
+
+# Build created files
+/playground
+
+# Coverage created files
 /.nyc_output
 /coverage
+
+# Exclude already built packages from testing with npm pack
+/scratch-vm-*.{tar,tgz}


mzgoddard:
If you want to check locally what files will be published, one way to do so is to make the package npm will publish with the `npm pack` command. It'll make a gzip'd tar file with the content that would be published. You can then use a tool to see the contents (like `tar -tf scratch-vm-0.1.0.tgz` on a command line) or extract the contents and look at the extracted directory.

rschamp (changes requested):
As in https://github.com/LLK/scratch-blocks/pull/1537, IMO I think this would be less surprising if we did it via exclusion with `.npmignore`. (Even though this method is explicit so more foolproof.)

cwillisf (approved):
Maybe someday we'll be able to omit `dist` as well...?

cwillisf:
Oops, apparently my page was out of date & I didn't see @rschamp's comment before I submitted my review.

Maybe we should do both? If somebody checks in a file we don't want to distribute and forgets to include it in `.npmignore`, or if Travis starts generating an unexpected file, then the `package.json` `files` approach will exclude the file whereas the `.npmignore` approach won't.

mzgoddard:
Cool @rschamp. I'll redo them with `.npmignore`.

rschamp:
Thanks @mzgoddard!

@cwillisf I don't think we need to worry very much about unexpectedly publishing files since publishing everything follows the principle of least astonishment to the developer. For us working on public git repos, the reason to limit what we publish would mostly just be to optimize our npm package sizes.  Anything Travis generates should be due to running commands that we give it so should be reproducible locally and dealt with in `.gitignore` as well, so considered for `.npmignore` at the same time.

But like I said I think the `files` approach is nice since it's explicit, but I can see someone who's not well-versed on what happens during `npm publish` being confused after adding a file and being confused when it works locally but not after publishing.

rschamp (approved):

rschamp:
Thank you @mzgoddard!

PR #1159, opened 2018-05-21
Translation extension imports translated language names

ericrosenbaum:
- Use the new scratch-translate-extension-languages module to get a list of the available languages translated into each of those languages. 

- Get the viewer's language using their locale or navigator language.

- Language menu defaults to the viewer's language.

- Convert strings dropped onto the language menu to language codes where possible.


towerofnix:
The LLK/scratch-translate-extensions-languages repository (linked from [the NPM page](https://www.npmjs.com/package/scratch-translate-extension-languages)) is private to LLK members only - that should probably be made public?

src/extensions/scratch3_translate/index.js
```diff
@@ -166,13 +163,9 @@ class Scratch3TranslateBlocks {
...
-        // TODO: Set default to project language instead of english?
-        let urlBase = `${serverURL}translate?language=`;
-        let lang = 'en';
-        if (args.LANGUAGE) {
-            lang = args.LANGUAGE;
-        }
+        const lang = this.getLanguageCodeFromArg(args.LANGUAGE);
 
+        let urlBase = `${serverURL}translate?language=`;
         urlBase += lang;
```
picklesrus, line 169: since this is now user droppable, should we also encodeUriComponent the lang too?

ericrosenbaum: Interesting! as far as I can tell this is not necessary. getLanguageCodeFromArg takes the dropped argument, but always returns one of the two-letter language codes provided by scratch-translate-extension-languages (which are all english letters). so by the time we get to the URL, lang shouldn't need encoding.

ericrosenbaum:
@towerofnix right! I think that is a question for @thisandagain, but I believe we will do so once the extension has stabilized and been deployed.

src/extensions/scratch3_translate/index.js
```diff
@@ -20,17 +23,27 @@ const serverURL = 'https://translate-service.scratch.mit.edu/';
...
+         */
+        this._viewerLanguageCode = this.getViewerLanguageCode();
+
+        /**
+         * List of supported language name and language code pairs, for use in the block menu.
          * @type {Array.<object.<string, string>>}
          * @private
          */
-        this._supportedLanguages = null;
+        this._supportedLanguages = languageNames.menuMap[this._viewerLanguageCode].map(entry => {
```
rschamp, line 43: @picklesrus Would it make sense for us to export this format, so we don't need to map?

rschamp (approved):
LGTM!

thisandagain:
@towerofnix Yup! What @ericrosenbaum said. We have to stabilize things and get MIT approval before we can open source them which sometimes take a little while.

PR #1160, opened 2018-05-22
Round (x position) and (y position) to 8 places

### LINKED ISSUE #1150: (x position) is more precise than in 2.0
### Expected Behavior

For perfect compatibility, `(x position)` should round to 8 decimal places

### Actual Behavior

It looks like `(x position)`  is rounding to 16 decimal places

### Steps to Reproduce

https://llk.github.io/scratch-gui/develop/compatibility-testing.html#223591096

This project breaks in 3.0: https://llk.github.io/scratch-gui/develop/#117386152/

MacOS 10.12.4, Chrome 66.0.3359.139 (64-bit)

Kenny2github:
### Resolves

Fixes #1150 

### Proposed Changes

Adds a `round_coord` function as a quick and easy shortcut to rounding to N digits, then uses that in `getX` and `getY` to round the coordinates to 8 decimal places at most.

### Reason for Changes

To fix a help wanted issue

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -238,6 +238,17 @@ class RenderedTarget extends Target {
     }
 
     /**
+     * Round a number to n digits
+     * @param {number} value The number to be rounded
+     * @param {number} places The number of decimal places to round to
+     * @return {number} The rounded number
+     */
+    _roundCoord (value, places) {
+        const power = Math.pow(10, places);
+        return Math.round(value * power) / power;
+    }
+
+    /**
      * Set the X and Y coordinates.
      * @param {!number} x New X coordinate, in Scratch coordinates.
      * @param {!number} y New Y coordinate, in Scratch coordinates.
@@ -250,6 +261,8 @@ class RenderedTarget extends Target {
         const oldY = this.y;
         if (this.renderer) {
             const position = this.renderer.getFencedPositionOfDrawable(this.drawableID, [x, y]);
+            position[0] = this._roundCoord(position[0], 8);
+            position[1] = this._roundCoord(position[1], 8);
             this.x = position[0];
             this.y = position[1];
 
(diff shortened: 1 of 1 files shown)


src/blocks/scratch3_motion.js
```diff
@@ -258,12 +258,16 @@ class Scratch3MotionBlocks {
         util.target.setXY(util.target.x, y);
     }
 
+    roundCoord (value, places) {
+        return Math.round(value * Math.pow(10, places)) / Math.pow(10, places);
+    }
+
     getX (args, util) {
-        return util.target.x;
+        return this.roundCoord(util.target.x, 8);
```
mzgoddard: Applying this on get would fix projects that use setX. The precision would not be bound if you are using changeXBy command blocks. Should this rather be done in target's setXY so the stored value is the correct precision instead of the temporary value returned from getX?

mzgoddard: I think the concern of truncating is better placed in Target than in the command blocks.

mzgoddard: What I mean by changeXBy not being affected is you could have multiple changeXBy operations that change by more precision that getX returns.

I'm not fully sure my comments here are helpful.

src/blocks/scratch3_motion.js
```diff
@@ -258,12 +258,16 @@ class Scratch3MotionBlocks {
         util.target.setXY(util.target.x, y);
     }
 
+    roundCoord (value, places) {
+        return Math.round(value * Math.pow(10, places)) / Math.pow(10, places);
```
mzgoddard: Using a constant variable in place of the Math.pow calls I think would perform better and let the constant names better describe why the rounding is needed? Does that seem fair?

mzgoddard (commented):
Hope these help.

src/blocks/scratch3_motion.js
Kenny2github: I get what you mean, but I'm not sure where the code for `util.target` is 😛

Kenny2github: Ah, found it.

Kenny2github:
Note: rebased and force pushed to clean up stupid mistakes

mrjacobbloom:
Should this only apply in compatibility mode? it makes more sense to me that in 3.0 for everything should use the same 64-bit precision (and avoid a few math operations if it matters)

rschamp (approved):
Thanks @Kenny2github! And @mzgoddard for your review. Looks/works great.

rschamp:
@mrjacobbloom Unfortunately "compatibility mode" has lost some of its meaning, as we aren't planning to conditionally apply that flag any time soon.

PR #1166, opened 2018-05-24
Use default asset if asset cannot be loaded

paulkaplan:
### Resolves

Fixes an issue where scratch crashes if you try to load a project with an asset that isn't on the servers.

### Proposed Changes

Use the default asset which is in storage.

### Reason for Changes

Projects with unlinked assets should not crash.

### Test Coverage

None, there are no unit test around this code right now.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -16,6 +16,7 @@ const log = require('../util/log');
  * @returns {?Promise} - a promise which will resolve after skinId is set, or null on error.
  */
 const loadCostumeFromAsset = function (costume, costumeAsset, runtime, optVersion) {
+
     costume.assetId = costumeAsset.assetId;
     const renderer = runtime.renderer;
     if (!renderer) {
@@ -119,9 +120,18 @@ const loadCostume = function (md5ext, costume, runtime, optVersion) {
     const ext = idParts[1].toLowerCase();
     const assetType = (ext === 'svg') ? AssetType.ImageVector : AssetType.ImageBitmap;
 
+    const onLoad = asset => {
+        costume.dataFormat = asset.dataFormat;
+        return loadCostumeFromAsset(costume, asset, runtime, optVersion);
+    };
+
     return runtime.storage.load(assetType, md5, ext).then(costumeAsset => {
-        costume.dataFormat = ext;
-        return loadCostumeFromAsset(costume, costumeAsset, runtime, optVersion);
+        if (costumeAsset) {
+            return onLoad(costumeAsset);
+        }
+        const defaultAssetId = runtime.storage.getDefaultAssetId(assetType);
+        return runtime.storage.load(assetType, defaultAssetId).then(onLoad);
+
     });
 };
 
--- a/src/import/load-sound.js
+++ b/src/import/load-sound.js
@@ -46,14 +46,24 @@ const loadSound = function (sound, runtime) {
         log.error('No storage module present; cannot load sound asset: ', sound.md5);
         return Promise.resolve(sound);
     }
+    const assetType = runtime.storage.AssetType.Sound;
     const idParts = StringUtil.splitFirst(sound.md5, '.');
     const md5 = idParts[0];
     const ext = idParts[1].toLowerCase();
-    return runtime.storage.load(runtime.storage.AssetType.Sound, md5, ext)
-        .then(soundAsset => {
-            sound.dataFormat = ext;
-            return loadSoundFromAsset(sound, soundAsset, runtime);
-        });
+
+    const onLoad = asset => {
+        sound.dataFormat = asset.dataFormat;
+        return loadSoundFromAsset(sound, asset, runtime);
+    };
+
+    return runtime.storage.load(assetType, md5, ext).then(soundAsset => {
+        if (soundAsset) {
+            return onLoad(soundAsset);
+        }
+        const defaultAssetId = runtime.storage.getDefaultAssetId(assetType);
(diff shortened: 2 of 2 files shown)


src/import/load-costume.js
```diff
@@ -16,6 +16,7 @@ const log = require('../util/log');
  * @returns {?Promise} - a promise which will resolve after skinId is set, or null on error.
  */
 const loadCostumeFromAsset = function (costume, costumeAsset, runtime, optVersion) {
+
```
rschamp, line 19: Cruft?

rschamp (commented):
The thing that worries me about this kind of thing is if there is a temporary misconfiguration with the assets server which causes assets to start appearing as not found or otherwise erroring out.  In those cases, I wouldn't want default assets to overwrite the real ones, and then have that bad data saved, essentially destroying the project. I don't think this will protect against that situation, will it?  @cwillisf Do we have anything in Scratch 2 that protects against that?

cwillisf:
Thinking about @rschamp's comments:

There's no protection against this sort of thing in Scratch 2. If a project loads but a particular asset fails, then that asset is replaced by an "empty" asset (0-length sound, 0x0 costume, etc.) and life goes on. If the project gets saved that way and the asset failure was only temporary, then the user has effectively lost work. I definitely don't recommend this approach :/

I've been thinking about what _should_ happen if your project JSON loads successfully but at least one of your project's assets fails to load. If it's a permanent failure -- say, the project JSON was modified to refer to an asset which doesn't exist -- then there's not much we can do other than replace the asset with one which doesn't fail. I expect the "permanent failure" case is very rare, though, compared to temporary failures caused by network issues, proxy misconfiguration, etc.

So... what _should_ happen with a temporary failure? Some thoughts:
- As a project viewer: depending on the project, its runtime behavior might be severely impacted. For example, a game which fails to load the images it uses to define the level/map would be unplayable. On the other hand, a project might "just" be missing its background music -- though that's a pretty big deal for an AMV project. A pen-only project would probably run fine, I guess, but it seems to me that **the average project would likely be at least somewhat, and maybe severely, degraded by missing assets.** Maybe, in player mode specifically, projects with failed assets should just display a big "failed" notice or "try again" button instead of attempting to play?
- As a project author:
  - Saving a project with references to default assets would effectively mean losing work.
  - We could introduce a system to retain the original asset ID even when the "contents" of the asset correspond to a default asset, but that seems very risky: I imagine we'd end up with cases where the empty asset gets saved over real assets after a transient issue becomes resolved. 
  - We could leave the asset contents empty and use a default asset when sending a costume to the renderer, sound to the sound engine, etc., just for presentation purposes. This would make working on code work correctly and would reduce the risk of accidentally saving an empty asset over a "good" asset, but editing an empty/default asset would effectively reintroduce that risk.

With all that in mind, here's my proposal for the ideal solution:
- The project JSON (and equivalent runtime structures) should retain an asset ID even if it fails to load
- When it comes time to "use" an asset at runtime, such as sending a costume to the renderer, substitute a default asset if the actual asset failed to load. Conceptually, something like `renderer.setBitmap(sprite.costumes[0] || defaultBitmapCostume)`
- Don't allow the user to use the paint editor or sound editor to edit an asset which failed to load. Maybe fade them out or draw a red X over them, or something.
- Do allow the user to remove a failed asset from the costume, backdrop, or sound list.
- Optionally, each failed asset should display a "refresh"-style button in the costume, backdrop, or sound list which prompts the editor try again. (Maybe this is the "draw a red X over them, or something.")

Thoughts?

fsih:
For the project in the original bug https://github.com/LLK/scratch-gui/issues/2137, one of the assets was loading as a 0x0 costume. I then edited the project to try to isolate the issue, by deleting all of the other costumes. Then today, the empty costume isn't empty anymore! So scratch 2 seems to be doing something to keep from saving over the original failing asset with the empty one.

In a copy of the project, I did edit the empty costume, and that overrode the original asset.

fsih:
I think not being able to edit a failing asset in the paint or sound editor could be really confusing.

thisandagain:
Thanks for all the analysis @paulkaplan and @cwillisf. Just to attempt to recap, there are two common types of asset load failures that occur:

1) An asset that is referenced in the project JSON is not available (404) on our servers because it does not exist. Likely causes:
    - The async nature of saving projects and assets in 2.0 is somewhat susceptible to the project JSON and assets getting out of sync (though not likely)
    - The project JSON has been edited by the user ("hacked" projects)

2) An asset that is referenced in the project JSON is not available (503, 500, etc.) on our servers because the client cannot reach the server. Likely causes:
    - The user's network connection is unstable / unavailable
    - Our server is unavailable / down

Doing a quick analysis via ElasticSearch, I have to agree with @cwillisf and the first case appears to be fairly uncommon (perhaps 1 in every ~30,000 projects) but that still means many existing Scratch projects include references to asset IDs that do not exist on our servers (small percentages of big numbers are still big numbers). The Scratch 2.0 method of "default to an empty asset" works well to handle this case, but for the reasons @cwillisf illustrated above the implementation can be disastrous when applied to the second. 

I agree with @cwillisf's first two suggestions for how to handle this:
> - The project JSON (and equivalent runtime structures) should retain an asset ID even if it fails to load
> - When it comes time to "use" an asset at runtime, such as sending a costume to the renderer, substitute a default asset if the actual asset failed to load. Conceptually, something like renderer.setBitmap(sprite.costumes[0] || defaultBitmapCostume)

As @paulkaplan pointed out (offline) the last three suggestions could probably use more discussion and we should file a new issue to track.

thisandagain:
Closing until we have time to re-address.

Issue #1029, opened 2018-04-06
Revisions to "Translate" extension


thisandagain:
## Icon
The current icon depicts a Chinese character of "文" which means "text" or "article" and is part of the official Google Translate logo (see below). Currently the rendering of this glyph is a little rough / clipped which makes the glyph difficult to read. In addition, focusing the extension on this glyph may be a little alienating to other groups in our large community which spans every part of the globe. For these reasons we should design and develop some alternatives.

![image](https://user-images.githubusercontent.com/747641/38437664-32859a2a-39a6-11e8-9080-0802823ef4e6.png)

## Additional Blocks
We would like to add the following reporter block which will provide the current language being used by the current viewer of a given project:

`viewer language` (reporter)

/cc @carljbowman @ntlrsk

picklesrus:
For the logo, we'll probably need to circle back with the folks that Erik was talking to before.  I can dig some of the old conversations/slide decks if that's useful.

Adding some more of the TODO's I've been tracking to the list, though I can put these in a separate bug if you'd prefer.
- Allow the language menu to accept language names 
- Use the project language as the default instead of English (first step is to change to the browser language)
- Retry the supported languages request if it fails
- Set a default value in the language menu
- pick default translate text
- i18n for block text

fsih:
Probably irrelevant, but I think it's the Japanese http://jisho.org/search/%23kanji%20%E6%96%87 which differs from Chinese (and Korean) in that it has a vertical short line at the top. (Still, it's readable in 3 Asian languages, so that might have been why they chose that glyph.)

thisandagain:
@fsih Interesting! Thanks. It still feels a bit odd to focus on this specific for the same reason as mentioned above though.

ntlrsk:
I encourage using a different icon for translation for the reasons @thisandagain mentioned. 

(Hard to say what variation of the character the illustrator was going for https://en.wiktionary.org/wiki/%E6%96%87   )

Kenny2github:
I would suggest going with something similar to Wikipedia's globe:
![globe](https://en.wikipedia.org/static/images/project-logos/enwiki.png)
(of course without the wordmark)
Or maybe the Scratch Wiki's "ball of many countries" image:
![ball](https://en.scratch-wiki.info/w/images/Ball_of_many_countries.png)
I suspect that it would be easier to use the Wiki countries ball, since a) it seems to be original and b) it's public domain (I think the WP globe is copyrighted?)

[...]

PR #1178, opened 2018-05-28
Add language reporter block to translate extension

picklesrus:
### Resolves
Progress toward https://github.com/LLK/scratch-vm/issues/1029

Adding the reporter block for viewer language. The block reports the human readable language name since that is easier to understand and the translate block accepts language names dropped on top of the menu.

Also adds messages for the text in the blocks

Removes the icons from the blocks. These will be replaced in a subsequent PR.

### DIFF
--- a/src/extensions/scratch3_translate/index.js
+++ b/src/extensions/scratch3_translate/index.js
@@ -100,6 +100,12 @@ class Scratch3TranslateBlocks {
                             defaultValue: this._viewerLanguageCode
                         }
                     }
+                },
+                {
+                    opcode: 'getViewerLanguage',
+                    text: 'viewer language',
+                    blockType: BlockType.REPORTER,
+                    arguments: {}
                 }
             ],
             menus: {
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
I can't figure out how to push directly to this PR- here's what I was playing with. It updates the viewer language code, and converts it to the language name in that language:

```
    getViewerLanguage () {
        this._viewerLanguageCode = this.getViewerLanguageCode();
        const names = languageNames.menuMap[this._viewerLanguageCode];
        const langNameObj = names.find(obj => obj.code === this._viewerLanguageCode);
        let langName = this._viewerLanguageCode;
        if (langNameObj) {
            langName = langNameObj.name;
        }
        return langName;
    }
```

picklesrus:
@ericrosenbaum Added that bit of code and added messages in block text. I also went ahead and removed the icons since I was here.  Updated the PR description to reflect this.  Should be ready for another look.

ericrosenbaum (approved):
Looks great!

PR #1180, opened 2018-05-31
Import Comments from 2.0 Projects

kchadha:
### Proposed Changes

Allow importing comments from scratch 2.0 projects. Comments should layout reasonably. Block comments should be able to retain minimized/non-minimized state. Block comments should remain connected to the correct blocks.

#### Known Issues
- Changes made to the comments (resizing, changing the text, repositioning, minimized state) will not persist across workspace refreshes/swaps
- Workspace comments are TINY, look different from block comments, and cannot be minimized.

### Reason for Changes

Comments!

### Test Coverage

Existing tests pass, used project #223875256 to test nesting of block comments / layout.


src/engine/comment.js
```diff
@@ -0,0 +1,53 @@
...
+
+const uid = require('../util/uid');
+const cast = require('../util/cast');
+
+class Comment {
+    /**
+     * @param {string} id Id of the variable.
+     * @param {string} name Name of the variable.
+     * @param {string} type Type of the variable, one of '' or 'list'
+     * @param {boolean} isCloud Whether the variable is stored in the cloud.
```
paulkaplan: @kchadha Looks like this JSdoc does not apply here

src/engine/comment.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * @fileoverview
+ * Object representing a Scratch Comment (block or workspace).
+ */
+
+const uid = require('../util/uid');
+const cast = require('../util/cast');
```
paulkaplan: Elsewhere in the code this is imported using capital `Cast`, fwiw.

src/serialization/sb2.js
```diff
@@ -110,15 +111,26 @@ const flatten = function (blocks) {
...
     let previousBlock = null; // For setting next.
     for (let i = 0; i < blockList.length; i++) {
         const block = blockList[i];
         // eslint-disable-next-line no-use-before-define
-        const parsedBlock = parseBlock(block, addBroadcastMsg, getVariableId, extensions);
+        const parsedBlockAndComments = parseBlock(block, addBroadcastMsg, getVariableId,
+            extensions, comments, commentIndex);
+        const parsedBlock = parsedBlockAndComments[0];
+        // Update commentIndex
+        commentIndex = parsedBlockAndComments[1];
```
paulkaplan, line 138: Since this is not used again until it is returned, it may be good to assign a new variable instead to avoid confusion. It would also let you write line 128 as `const [parsedBlock, newCommentIndex] = parseBlock(...`

src/serialization/sb2.js
```diff
@@ -423,9 +444,54 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
...
+    const blockComments = {};
+    if (object.hasOwnProperty('scriptComments')) {
+        const comments = object.scriptComments.map(commentDesc => {
+            const isBlockComment = commentDesc[5] >= 0;
+            const newComment = new Comment(
+                null, // generate a new id for this comment
+                commentDesc[6], // text content of sb2 comment
+                // Only serialize x & y position of comment if it's a workspace comment
+                // If it's a block comment, we'll let scratch-blocks handle positioning
+                isBlockComment ? null : commentDesc[0] * 1.5, // x position of comment
```
paulkaplan: Where do these 1.5 and 2.2 numbers come from? Maybe they could use an explanatory variable name instead?

src/serialization/sb2.js
```diff
@@ -423,9 +444,54 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
...
+                isBlockComment ? null : commentDesc[1] * 2.2, // y position of comment
+                commentDesc[2] * 1.5, // width of comment
+                commentDesc[3] * 2.2, // height of comment
+                !commentDesc[4] // commentDesc[4] -- false means minimized, true means full screen
```
paulkaplan: `full screen`? Do you mean... expanded?

src/serialization/sb2.js
```diff
@@ -423,9 +444,54 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
         }
     }
 
+    // If included, parse any and all comments on the object (this includes top-level
+    // workspace comments as well as comments attached to specific blocks)
+    const blockComments = {};
+    if (object.hasOwnProperty('scriptComments')) {
+        const comments = object.scriptComments.map(commentDesc => {
+            const isBlockComment = commentDesc[5] >= 0;
```
paulkaplan: It might be nice to destructure the `commentDesc` here, and use that as place to condense the documentation of what each field is, e.g.

```
const [
    xPosition, 
    yPosition,
    width,
    height,
    expanded,
    flattenedBlockIndex,
    textContent
] = commentDesc;

File: src/serialization/sb2.js [hunk 7]
```diff
@@ -645,6 +717,21 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
         shadow: false, // No shadow blocks in an SB2 by default.
         children: [] // Store any generated children, flattened in `flatten`.
     };
+
+    // Attach any comments to this block..
+    const commentsForParsedBlock = (comments && typeof commentIndex === 'number' && !isNaN(commentIndex)) ?
+        comments[commentIndex] : null;
+    if (commentsForParsedBlock) {
+        // TODO currently only attaching the last comment to the block if there are multiple...
```
paulkaplan, line 737: Is this something that can happen through the scratch 2 editor, or is it a hacked project thing?

src/virtual-machine.js
```diff
@@ -1005,12 +1005,16 @@ class VirtualMachine extends EventEmitter {
         );
 
         const variables = Object.keys(variableMap).map(k => variableMap[k]);
+        const wkspcComments = Object.keys(this.editingTarget.comments)
```
paulkaplan: Can I buy a vowel?

;) I think we should stick with the full word as much as possible

test/unit/serialization_sb2.js
```diff
@@ -40,7 +39,6 @@ test('default', t => {
         t.type(targets[1].id, 'string');
         t.type(targets[1].blocks, 'object');
         t.type(targets[1].variables, 'object');
-        t.type(targets[1].lists, 'object');
```
paulkaplan, line 43: It would probably be good replace this expectation with one about `target.comments`, to stay consistent with these tests

paulkaplan (changes requested):
A few minor updates requested.

A bigger ask is to add an sb2 integration fixture with some block and workspace comments and make some useful assertions there.

src/serialization/sb2.js
kchadha: I found them as [existing magic numbers](https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb2.js#L153-L154) to translate block coordinates. =]

I'll make constants and use them in both places.

src/serialization/sb2.js
kchadha: yes, good catch

src/serialization/sb2.js
kchadha: It's something that can happen through the 2.0 editor, you can add multiple comments to the same block, but you can only ever see the top one you created, even though the others exist behind it. If you try to move the top comment around to see the other block comments below it, it gets detached from the block and becomes a workspace comment. But if you don't move it around, the block just has more than one comment, and all of those comments just have references to the same block. =/

src/serialization/sb2.js
paulkaplan: 🌮

src/serialization/sb2.js
paulkaplan: Ah, so the "use the last one in the list" behavior chooses the one that was visible when imported? That sounds reasonable.

src/serialization/sb2.js
kchadha: Unfortunately this doesn't work because I'm using the commentIndex outside of the for loop where the 'parseBlock' call is happening.

paulkaplan (approved):

PR #1188, opened 2018-06-01
fix: resolve scratch-*/src/index.js for expose-loader

mzgoddard:
require.resolve is a nodejs api that resolves as node resolves. An
enhanced-resolve instance, the library webpack uses, could resolve as
webpack will or we can use some slightly more manual definitions here
for exposing parts of scratch on the global object for the playground's
benchmark to use.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -119,15 +119,15 @@ module.exports = [
                     loader: 'expose-loader?Blockly'
                 },
                 {
-                    test: require.resolve('scratch-audio'),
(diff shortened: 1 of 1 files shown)


mzgoddard:
The scratch benchmark currently does not work without this change. Its unable to create needed AudioEngine and Renderer as they are not exposed to the playground script.

rschamp (approved):
Thanks @mzgoddard!

PR #1198, opened 2018-06-04
Update microbit extension for latest firmware

khanning:
### Proposed Changes

Update the micro:bit extension for the simplified TX/RX bluetooth firmware

### Reason for Changes

The current model of having many bluetooth characteristics isn't scalable with the memory constraints of the micro:bit

### Test Coverage

Load the latest scratch-firmware-combined.hex and run the extension

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -259,6 +259,17 @@ const symbols2hex = {
 };
 
 /**
+ * Enum for micro:bit BLE command protocol.
+ * @readonly
+ * @enum {number}
+ */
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_microbit/index.js
```diff
@@ -544,13 +560,14 @@ class Scratch3MicroBitBlocks {
      */
     displaySymbol (args) {
         const hex = symbols2hex[args.SYMBOL];
```
evhan55, line 562: Handle (by returning?) if `hex` ends up being `0` or `undefined` here

PR #1203, opened 2018-06-05
APIs for getting and setting variable values

### LINKED ISSUE #1145: Add top-level API for changing variable values
This is a stub issue to remember to make move the code in GUI that is setting variables from monitors into the VM.

From the review on https://github.com/LLK/scratch-gui/pull/2019

/cc @kchadha @rschamp

paulkaplan:
Adds a few top-level APIs for getting/setting variable values.

The match the "prototype" that was made in the GUI, which can now be migrated to use these methods.

Additionally, these tests uncovered a bug, which was fixed :)

Fixes https://github.com/LLK/scratch-vm/issues/1145

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -1071,6 +1071,42 @@ class VirtualMachine extends EventEmitter {
             this.editingTarget.postSpriteInfo(data);
         }
     }
+
+    /**
+     * Set a target variables value if it exists. Return whether it succeeded.
+     * @param {!string} targetId ID of the target which owns the variable.
+     * @param {!string} variableId ID of the variable to set.
+     * @param {!*} value The new value of that variable.
+     * @returns {boolean} whether the target and variable were found and updated.
+     */
+    setVariableValue (targetId, variableId, value) {
+        const target = this.runtime.getTargetById(targetId);
+        if (target) {
+            const variable = target.lookupVariableById(variableId);
+            if (variable) {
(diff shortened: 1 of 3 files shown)


src/virtual-machine.js
```diff
@@ -1071,6 +1071,42 @@ class VirtualMachine extends EventEmitter {
             this.editingTarget.postSpriteInfo(data);
         }
     }
+
+    /**
+     * Set a target variables value if it exists. Return whether it succeeded.
```
kchadha: 'Set a target's variable's value' ?

src/virtual-machine.js
```diff
@@ -1071,6 +1071,42 @@ class VirtualMachine extends EventEmitter {
...
+            if (variable) {
+                variable.value = value;
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Set a target variables value if it exists. Return whether it succeeded.
```
kchadha: Same as above if you end up fixing that one.

kchadha: Also it should be 'get' instead of 'set' here and below in the description for the variableId param.

kchadha (commented):
LG minus nit pick comments about comments.

src/virtual-machine.js
paulkaplan: ((target's variable)'s value)

kchadha (approved):

PR #1204, opened 2018-06-05
Add methods for reordering costumes and sounds

paulkaplan:
Part of implementing the asset reordering, this implements the top-level API that was outlined in the doc as well as rendered-target level methods to make testing better.

Adds unit tests for all methods covered.

Also fixes a bug that was uncovered by the tests, which was a `if (number) { ... ` falsey-ness bug where you could not add costumes/sounds at index 0.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -488,10 +488,10 @@ class RenderedTarget extends Target {
      * @param {?int} index Index at which to add costume
      */
     addCostume (costumeObject, index) {
-        if (index) {
-            this.sprite.addCostumeAt(costumeObject, index);
-        } else {
+        if (typeof index === 'undefined') {
             this.sprite.addCostumeAt(costumeObject, this.sprite.costumes.length);
+        } else {
+            this.sprite.addCostumeAt(costumeObject, index);
         }
     }
 
@@ -551,10 +551,10 @@ class RenderedTarget extends Target {
     addSound (soundObject, index) {
         const usedNames = this.sprite.sounds.map(sound => sound.name);
         soundObject.name = StringUtil.unusedName(soundObject.name, usedNames);
-        if (index) {
-            this.sprite.sounds.splice(index, 0, soundObject);
-        } else {
+        if (typeof index === 'undefined') {
             this.sprite.sounds.push(soundObject);
+        } else {
+            this.sprite.sounds.splice(index, 0, soundObject);
         }
     }
 
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -488,10 +488,10 @@ class RenderedTarget extends Target {
      * @param {?int} index Index at which to add costume
      */
     addCostume (costumeObject, index) {
-        if (typeof index === 'undefined') {
-            this.sprite.addCostumeAt(costumeObject, this.sprite.costumes.length);
-        } else {
+        if (typeof index === 'number' && !isNaN(index)) {
             this.sprite.addCostumeAt(costumeObject, index);
+        } else {
+            this.sprite.addCostumeAt(costumeObject, this.sprite.costumes.length);
         }
     }
 
@@ -551,10 +551,10 @@ class RenderedTarget extends Target {
     addSound (soundObject, index) {
         const usedNames = this.sprite.sounds.map(sound => sound.name);
         soundObject.name = StringUtil.unusedName(soundObject.name, usedNames);
-        if (typeof index === 'undefined') {
-            this.sprite.sounds.push(soundObject);
-        } else {
+        if (typeof index === 'number' && !isNaN(index)) {
             this.sprite.sounds.splice(index, 0, soundObject);
+        } else {
+            this.sprite.sounds.push(soundObject);
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/sprites/rendered-target.js
```diff
@@ -488,10 +488,10 @@ class RenderedTarget extends Target {
      * @param {?int} index Index at which to add costume
      */
     addCostume (costumeObject, index) {
-        if (index) {
-            this.sprite.addCostumeAt(costumeObject, index);
-        } else {
+        if (typeof index === 'undefined') {
```
kchadha: Any particular reason not to check `typeof index === 'number' && !isNaN(index)`. I suspect we wouldn't want to call `addCostumeAt` on the provided index if either of these checks is false.

src/sprites/rendered-target.js
```diff
@@ -551,10 +551,10 @@ class RenderedTarget extends Target {
     addSound (soundObject, index) {
         const usedNames = this.sprite.sounds.map(sound => sound.name);
         soundObject.name = StringUtil.unusedName(soundObject.name, usedNames);
-        if (index) {
-            this.sprite.sounds.splice(index, 0, soundObject);
-        } else {
+        if (typeof index === 'undefined') {
```
kchadha: Same as above.

src/sprites/rendered-target.js
```diff
@@ -640,6 +640,51 @@ class RenderedTarget extends Target {
...
 
+    /**
+     * Reorder costume list by moving costume at costumeIndex to newIndex.
+     * @param {!number} costumeIndex Index of the costume to move.
+     * @param {!number} newIndex New index for that costume.
+     * @returns {boolean} If a change occurred (i.e. if the indices do not match)
+     */
+    reorderCostume (costumeIndex, newIndex) {
+        const clampedNewIndex =
+            Math.max(0, Math.min(this.sprite.costumes.length - 1, newIndex));
```
kchadha: Isn't there a util function for clamping that can be used here?

src/virtual-machine.js
```diff
@@ -1071,6 +1071,38 @@ class VirtualMachine extends EventEmitter {
...
+     * @param {!string} targetId ID of the target which owns the costumes.
+     * @param {!number} costumeIndex index of the costume to move.
+     * @param {!number} newIndex index that the costume should be moved to.
+     * @returns {boolean} whether the target was found for reordering.
+     */
+    reorderCostume (targetId, costumeIndex, newIndex) {
+        const target = this.runtime.getTargetById(targetId);
+        if (target) {
+            target.reorderCostume(costumeIndex, newIndex);
+            return true;
```
kchadha: Should this return what `target.reorderCostume` returns? (If that is the case, the comment above should probably also be updated to say that this function returns 'whether a reorder occurred').

src/virtual-machine.js
```diff
@@ -1071,6 +1071,38 @@ class VirtualMachine extends EventEmitter {
...
+     * @param {!string} targetId ID of the target which owns the sounds.
+     * @param {!number} soundIndex index of the sound to move.
+     * @param {!number} newIndex index that the sound should be moved to.
+     * @returns {boolean} whether the target was found for reordering.
+     */
+    reorderSound (targetId, soundIndex, newIndex) {
+        const target = this.runtime.getTargetById(targetId);
+        if (target) {
+            target.reorderSound(soundIndex, newIndex);
+            return true;
```
kchadha: Same question here as above.

test/unit/sprites_rendered-target.js
```diff
@@ -427,3 +427,100 @@ test('#renameCostume does not duplicate names', t => {
...
+        a.setCostume(0);
+        s.costumes = [o1, o2, o3, o4, o5];
+    };
+    const costumeIds = () => a.sprite.costumes.map(c => c.id);
+
+    resetCostumes();
+    t.deepEquals(costumeIds(), [0, 1, 2, 3, 4]);
+    t.equals(a.currentCostume, 0);
+
+    // Returns false if the costumes are the same an no change occurred
```
kchadha: 'an' -> 'and'

test/unit/sprites_rendered-target.js
```diff
@@ -427,3 +427,100 @@ test('#renameCostume does not duplicate names', t => {
...
+    const resetSounds = () => {
+        s.sounds = [o1, o2, o3, o4, o5];
+    };
+    const soundIds = () => a.sprite.sounds.map(c => c.id);
+
+    resetSounds();
+    t.deepEquals(soundIds(), [0, 1, 2, 3, 4]);
+
+    // Return false if indices are the same and no change occurred.
+    t.equal(a.reorderSound(100000, 99999), false);
```
kchadha: This same check should exist for the `#reorderCostume` test above.

kchadha (dismissed):
Some questions in the comments and some minor changes requested.

kchadha:
@paulkaplan, Also, I just noticed that this branch has merge conflicts.

src/sprites/rendered-target.js
rschamp: I think that would be making this a bit too smart — if you provided the string `'one'` as the index, rather than `1`, it would put it at the end. I think I would rather it throw in that case.

paulkaplan:
@kchadha thanks for the comments, fixed them and then fixed the merge conflicts, no real conflicts, just the last PR added things in the exact same place.

kchadha (approved):
LG!

Issue #97, opened 2016-08-08
Pen Size Limit


thisandagain:
_Does this belong in scratch-vm?_

In Scratch 2.0, pen size was limited to 255 due to a technical limitation of Flash (see [here](https://github.com/LLK/scratch-flash/blob/5cff62b909856b7d7b3d116a5dcc2b4f03de8482/src/scratch/ScratchSprite.as#L264)). This limit was not imposed in Scratch 1.4. Has any consideration been made as to whether this limit will remain in Scratch 3? The only real reason it could be desired is for backwards-compatibility. Is that reason enough to keep the limit?

---

Moved from LLK/scratch-blocks#552
/cc @PullJosh @pulljosh

mrjacobbloom:
Another use case: I tried to re-create Agar.io in Scratch once, and for most characters I used one pen dot for each character, but for large characters I had to use a circle-drawing algorythm

ed-cooper:
In built support for pen size > 255 should definitely be done - I've never seen a circle filler algorithm even come close to working at the speed of a single pen down and pen up.

As for backwards compatibility, I think the real question is are there actually any projects that take advantage of the limit? Because I can't find any

CosmicWebServices:
Maybe the pen size cap should be at where it covers the full full screen with just one pen down if that make sense but that may differ for some browsers/computers.

thisandagain:
I don't see any reason why this limit shouldn't be enough to cover the entire size of the stage.

PR #1208, opened 2018-06-05
Increase maximum pen size to 1200

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/97

### Proposed Changes

Increase the maximum diameter of the pen in the pen extension to 1200px.

### Reason for Changes

The previous max of 255 was set by a limitation of flash. The new maximum was chosen to be twice the length of the diagonal of the stage, so that even if a sprite is off the stage, at the max pen size it can draw a single dot that completely fills it.

Some quick performance tests revealed that drawing with this very large pen size is very slightly slower, but we're still seeing thousands of pen operations per second, so it seems like an acceptable tradeoff for an edge case.

### DIFF
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -91,10 +91,12 @@ class Scratch3PenBlocks {
 
     /**
      * The minimum and maximum allowed pen size.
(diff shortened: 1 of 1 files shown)


chrisgarrity (approved):
Looks good to me :)

PR #1209, opened 2018-06-05
Check that bubble bounds do not extend past stage boundaries

### LINKED ISSUE #1135: Bubbles can have their near edge off the screen
### Expected Behavior

Bubbles in scratch 2 stay completely within the stage

### Actual Behavior

speech-off

### Steps to Reproduce

1. Open blank project
2. Set cat X position to -10000 (gets fenced)
3. Trigger "say" bubble
4. see the left edge is offscreen.

---

The fix for this is pretty straightforward, we should just "max" the bubble left with the stage left if the bubble is "facing right" and "min" the bubble right with the stage right if the bubble is facing left.

hyperobject:
### Resolves
Fixes #1135.

### Proposed Changes
This PR adds additional logic (and comments) to the `_positionBubble` function to check that a bubble's boundaries do not extend past the stage boundaries, even when the bubble cannot be flipped.

### How to test
In any sprite, set the x position to either a very large number or a very negative one. Use a say or think block. The resulting bubble should stay fully on the stage.
demonstration of bug fix

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -153,10 +153,17 @@ class Scratch3LooksBlocks {
             this.runtime.renderer.updateDrawableProperties(bubbleState.drawableId, {
                 position: [
                     bubbleState.onSpriteRight ? (
-                        Math.min(stageBounds.right - bubbleWidth, targetBounds.right)
+                        Math.max(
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
LG!

PR #1210, opened 2018-06-05
Add initial working version of the Amazon Polly extension

thisandagain:
Adds initial working version of the Amazon Polly (text-to-speech) extension. There are still many features that we would like to add, but this should get us started with testing.


ericrosenbaum (approved):
woohoo! 💬

PR #1211, opened 2018-06-06
yield a tick if broadcastandwait is waiting on threads waiting for a future tick

### LINKED ISSUE #1196: broadcast_and_wait busy waits on threads that have entered status_promise_wait
### Expected Behavior

broadcast_and_wait should next run when its likely that threads it created have completed.

### Actual Behavior

leaves a thread spinning until the vm performs a redraw or hits its per frame work limit.

### Steps to Reproduce

compare http://localhost:8073/playground/index.html#155128646,5000,5000 versus https://llk.github.io/scratch-vm/index.html#155128646,5000,5000

-----

I noticed this because the noted project uses broadcast_and_wait and sound blocks changed by #1184 that led the current vm version to locking to the frame rate and unable to make use of turbo mode. More relevant to this issue, this seems to be a battery performance issue as it leads the vm to waste cpu cycles in circumstances like this or if you had a broadcast_and_wait wait on a wait block.

mzgoddard:
### Resolves

Fix #1196

### Proposed Changes

Use STATUS_YIELD_TICK in event_broadcastandwait and looks_switchbackdroptoandwait if all threads are waiting for a some future time.

### Reason for Changes

Normal yielding lets other threads do work. Yielding a tick pauses the thread until the next tick of the runtime. This way if threads are looped 10 times, the block yielding a tick runs once instead of 10 times.

This is important for benchmarking as it makes these block deterministic in the number of times they run. If they are waiting on "waiting" threads and normally yield instead of yielding a tick they case the runtime to busy wait checking for a change that will not happen for up to 33 milliseconds.

### Test Coverage

As a performance issue this can be "tested" in the benchmark suite. Before this change event_broadcastandwait in http://llk.github.io/scratch-vm/index.html#155128646,5000,5000 might run hundreds of thousands of times. With the change event_broadcastandwait returns to its previous behaviour for this project running hundreds or thousands of times depending on the performance of the local hardware.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -409,7 +409,17 @@ class Scratch3LooksBlocks {
         const instance = this;
         const waiting = util.stackFrame.startedThreads.some(thread => instance.runtime.isActiveThread(thread));
         if (waiting) {
-            util.yield();
+            // If all threads are waiting for the next tick or later yield
+            // for a tick as well. Otherwise yield until the next loop of
+            // the threads.
+            if (
+                util.stackFrame.startedThreads
+                    .every(thread => instance.runtime.isWaitingThread(thread))
+            ) {
+                util.yieldTick();
+            } else {
+                util.yield();
+            }
         }
     }
 
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -100,7 +100,17 @@ class Scratch3EventBlocks {
             const instance = this;
             const waiting = util.stackFrame.startedThreads.some(thread => instance.runtime.isActiveThread(thread));
             if (waiting) {
-                util.yield();
+                // If all threads are waiting for the next tick or later yield
+                // for a tick as well. Otherwise yield until the next loop of
+                // the threads.
+                if (
+                    util.stackFrame.startedThreads
+                        .every(thread => instance.runtime.isWaitingThread(thread))
+                ) {
+                    util.yieldTick();
+                } else {
+                    util.yield();
+                }
             }
         }
     }
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -231,9 +231,7 @@ class Scratch3SoundBlocks {
         util.target.audioPlayer.setEffect(effect, soundState.effects[effect]);
 
         // Yield until the next tick.
-        return new Promise(resolve => {
-            resolve();
-        });
+        return Promise.resolve();
     }
 
     _syncEffectsForTarget (target) {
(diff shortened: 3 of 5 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -409,7 +409,17 @@ class Scratch3LooksBlocks {
         const instance = this;
         const waiting = util.stackFrame.startedThreads.some(thread => instance.runtime.isActiveThread(thread));
         if (waiting) {
-            util.yield();
+            // If all threads are waiting for the next tick or later yield
+            // for a tick as well. Otherwise yield until the next loop of
+            // the threads.
+            if (
+                util.stackFrame.startedThreads
+                    .some(thread => !instance.runtime.isWaitingThread(thread)) === false
```
rschamp: I can never fully understand double negatives no matter how many time I read them. Is this any different than

```js
               util.stackFrame.startedThreads
                   .every(thread => instance.runtime.isWaitingThread(thread))
```
? And if not can we change it to that so I can put my brain back right side in?

rschamp (changes requested):
LG other than the brain knot condition!

src/blocks/scratch3_looks.js
rschamp: Er..  I guess the opposite of what I said. Not every one is waiting?
```js
               !util.stackFrame.startedThreads
                   .every(thread => instance.runtime.isWaitingThread(thread))
```

mzgoddard: This is one way to prove your point. :wink:

Your first guess was right. Everyone is waiting.

rschamp: I promise I didn't get confused on purpose! Thanks for changing that.

rschamp (approved):

paulkaplan:
It looks like this caused this regression with broadcast and wait: https://github.com/LLK/scratch-gui/issues/2461

I made a test project to demonstrate, #235582577:
image

On develop, this is what the playground shows
broadcast-wait--develop


Reverting this PR, I see this
broadcast-wait--reverted

paulkaplan:
It looks like there is a problem with the `STATUS_YIELD_TICK` inside the `stepThread` function of sequencer. After calling `execute` inside `stepThread`, line 215 says:
```
            // If the thread has yielded or is waiting, yield to other threads.
            if (thread.status === Thread.STATUS_YIELD) { 
                 ... return
            } else if (thread.status === Thread.STATUS_PROMISE_WAIT) {
                 ... return 
            }
            // Switch to the next block or look on the stack for more blocks
```
I think the first condition is missing a check for `Thread.STATUS_YIELD_TICK`, which seems to be implied by the comment. Without that check, threads that are in that state after executing do not get set back to `RUNNING` and simply resolve right away and don't get executed again.

PR #1214, opened 2018-06-06
Add when touching hat

ericrosenbaum:
### Resolves

Part of https://github.com/LLK/scratch-gui/issues/607
Depends on https://github.com/LLK/scratch-blocks/pull/1559

### Proposed Changes

Add a hat block for when a sprite is touching the mouse pointer, edge of the stage, or another sprite.

This includes moving the touchingObject function into renderedTarget, since it is now used by both a sensing block (the boolean reporter) and an event block (the hat).

### Reason for Changes

This is a "low floor" feature, making it easier to make simple projects where you want a sprite to do something when it touches e.g. another sprite.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -725,11 +725,11 @@ class RenderedTarget extends Target {
     }
 
     /**
-     * Return whether touching the mouse, an edge, or a sprite.
+     * Return whether this target is touching the mouse, an edge, or a sprite.
      * @param {string} requestedObject an id for mouse or edge, or a sprite name.
      * @return {boolean} True if the sprite is touching the object.
      */
-    touchingObject (requestedObject) {
+    isTouchingObject (requestedObject) {
         if (requestedObject === '_mouse_') {
             if (!this.runtime.ioDevices.mouse) return false;
             const mouseX = this.runtime.ioDevices.mouse.getClientX();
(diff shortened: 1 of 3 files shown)


src/sprites/rendered-target.js
```diff
@@ -724,6 +724,23 @@ class RenderedTarget extends Target {
         return null;
     }
 
+    /**
+     * Return whether touching the mouse, an edge, or a sprite.
+     * @param {string} requestedObject an id for mouse or edge, or a sprite name.
+     * @return {boolean} True if the sprite is touching the object.
+     */
+    touchingObject (requestedObject) {
```
kchadha: Nit: Can we change the name of this function to be 'isTouchingObject' since that indicates that this returns a boolean? Otherwise, this function name can also be interpreted as 'return whatever object this target is touching, or null if none'

src/sprites/rendered-target.js
```diff
@@ -724,6 +724,23 @@ class RenderedTarget extends Target {
         return null;
     }
 
+    /**
+     * Return whether touching the mouse, an edge, or a sprite.
```
kchadha: 'whether touching' -> 'whether this target is touching' (for clarity)

kchadha (changes requested):
@ericrosenbaum most of this looks good! I had two very minor nits in the comments.

kchadha (approved):
LGTM!

Issue #1202, opened 2018-06-05
Fix Speech Extension in Safari


picklesrus:
For now, I'm only testing browsers on my mac laptop since those are easy
**Chrome Mac**: Okay so far
**Firefox Mac**: Okay so far

**Safari on a mac:** throws this error:
[Info] [react-ga] – "with trackers: undefined" (lib.min.js, line 46451)
[Log] vm (3) (lib.min.js, line 184265)
"warn"
"Primitive rejected promise: "
ReferenceError: Can't find variable: AudioContext
_initializeMicrophone — index.js:549
_initListening — index.js:539
_startListening — index.js:518
(anonymous function) — index.js:734
initializePromise
Promise
(anonymous function) — index.js:730
promiseReactionJob

picklesrus:
Update: works on chrome on Android and on an old Surface running Edge so I think safari is the main problem. 

For Safari, I think I just need to use webkitAudioContext instead of AudioContext so fix should be easy.

Since enough browsers work(!), I'll change the title of the bug and file anything else browser specific that comes up separately.

picklesrus:
Update: While #1215 fixed the js error, safari still doesn't work.  Audio data seems to be all arrays of 0s. Continuing to investigate.

picklesrus:
#1297 will fix the *first* listen and wait block we encounter to not have all 0s. 

For fixing subsequent ones, I'm still weeding through information from https://kaliatech.github.io/web-audio-recording-tests/dist/#/test1 to figure out how to re-initialize the pieces that Safari wants properly.

picklesrus:
After reading the code for the examples in https://kaliatech.github.io/web-audio-recording-tests/dist/#/test1 and the discussions on the bugs on that page, I figured out the main problem here is that I was reusing the mediaStream.  If I call getUserMedia for every listenAndWait block, it works fine in desktop Safari (don't currently have an ipad to test with).  Tab switching is still problematic at times, but that seems to be a larger problem.

Just a note for any future readers.  This Safari thing becomes really frustrating to debug because once you get it into a "stuck" state, you have to restart the browser to "unstick" it. Subsequent code changes go unnoticed until you remember :-/.  

PR coming soon.

ericrosenbaum:
Fixed by https://github.com/LLK/scratch-vm/pull/1467

PR #1215, opened 2018-06-06
Add webkitAudioContext if AudioContext does not exist. This might fix…

picklesrus:
… the speech extension in safari, but it is tricky to tell since I am running into problems with permissions and localhost.

### Resolves

Helps with #1202. Might fix it completely.

### Test Coverage

Chrome still works and Safari no longer throws an error about the AudioContext object.  There are still some permissions issues with running on localhost so I can't quite tell if this completely fixes the problem just yet.

### DIFF
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -546,7 +546,8 @@ class Scratch3SpeechBlocks {
      * @private
      */
     _initializeMicrophone () {
-        this._context = new AudioContext();
+        // Safari still needs a webkit prefix for audio context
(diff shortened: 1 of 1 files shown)


rschamp (approved):
LGTM! Looks familiar!

https://github.com/LLK/scratch-vm/blob/c2645f3da5a5ca5c974152b8e7491ace2671b087/src/extensions/scratch3_speak/index.js#L135

https://github.com/LLK/scratch-gui/blob/b593f11e04c8ebf4555c4491d773eff1c7f5b885/src/lib/audio/shared-audio-context.js#L6

PR #1217, opened 2018-06-07
Comment state fixes

kchadha:
### Proposed Changes

Changes to comment state-keeping.

The API for the CommentChange event emitted by scratch-blocks has been updated. Now, when a text change is emitted, the text change is represented as an object containing a 'text' property (to match the other kinds of changes a CommentChange event can encompass) instead of just a string. 

This change needs to be accommodated in the VM, addressed by this PR.

Previously, I had added code to update a block comment's position if the block it's attached to was moved. This causes bugs where the comment keeps moving on sprite switch even if the user did not indicate a move. This is because a block move event is emitted every time the workspace containing the block is updated. This PR reverts this change. 

There is now instead, a different bug where moving a block (with a comment attached to it) and switching sprites and back shows the comment in a different position relative to the block. This bug will be fixed with the fix for LLK/scratch-blocks#1566.

### Reason for Changes

See above.

### Test Coverage

Manual testing with scratch-gui.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -295,9 +295,8 @@ class Blocks {
                 oldInput: e.oldInputName,
                 newParent: e.newParentId,
                 newInput: e.newInputName,
-                newCoordinate: e.newCoordinate,
-                oldCoordinate: e.oldCoordinate
-            }, optRuntime);
+                newCoordinate: e.newCoordinate
+            });
             break;
         case 'dragOutside':
             if (optRuntime) {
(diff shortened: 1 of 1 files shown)


paulkaplan (changes requested):
To prevent future "is this here for a reason?", I think we should make it possible to change all the possible fields in the change event instead of only certain ones at a time.

kchadha:
@paulkaplan made that changes.

paulkaplan (approved):

BryceLTaylor:
I found several issues with comments generally and attached them to the comments issue: https://github.com/LLK/scratch-blocks/issues/1554

PR #1218, opened 2018-06-07
Fix importing hidden monitors from Scratch 2 projects

### LINKED ISSUE #1165: Variable monitors do not import correctly if they are initially hidden
### Expected Behavior

All variable monitors should import in their correct position and mode (normal/large/slider). The example project (after the "show 4-6" button is clicked) should look like this:

Screenshot of project in 2.0

### Actual Behavior

Variable monitors which are hidden when the 2.0 project is saved do not import with their position or mode. When shown (via the palette checkbox or a "show variable" block), they all line up at the top left, the default position for variable monitors, and are in normal-readout mode (with the min/max values for the slider the default 0-100). The example project (again, after "show 4-6" is pressed) looks like this:

Screenshot of project in 3.0

### Steps to Reproduce

Create a Scratch 2.0 project and place a variable in any position or mode, then hide the variable. Save the project and open it in Scratch 3.0, then show the variable. Note that the variable is in the top left of the stage and the default/normal mode instead of however it was positioned/set.

My example project: https://scratch.mit.edu/projects/225201357/ ([3.0](https://llk.github.io/scratch-gui/develop/#225201357))

Many projects which use variables as part of the user interface are affected by this issue. I stumbled across this issue [here (warning, this project has many assets)](https://scratch.mit.edu/projects/172355986/).

Issue exists on both Firefox 61.0b7 and Chromium 62.0.3202.89, running Debian (testing).

hyperobject:
### Resolves
This PR fixes #1165, and addresses some of llk/scratch-gui#2033.

### Proposed Changes
* Changes to the `changeBlock` function in blocks.js to add monitors with additional relevant parameters (x, y, mode, sliderMin, sliderMax) so shown monitors are created with those parameters
* Changes to the behavior of `requestRemoveMonitor` so it sets the `visible` parameter to false instead of deleting monitor state on remove
* Changes to how `requestAddMonitor` and `requestChangeMonitor` merge in a passes object
   * The `merge` function allows undefined objects to overwrite keys that already have a defined value, so I changed it to `mergeWith` and some extra logic to handle undefined or null values.

### Potential Gotchas
Monitor position is handled in scratch-gui, so there might be some weird interactions with the logic added here in edge cases. I wasn't able to find any examples of this, though.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -541,7 +541,7 @@ class Blocks {
                         params: this._getBlockParams(block),
                         // @todo(vm#565) for numerical values with decimals, some countries use comma
                         value: '',
-                        mode: block.opcode === 'data_listcontents' ? 'list' : block.mode
+                        mode: block.opcode === 'data_listcontents' ? 'list' : 'default'
                     }));
                 }
             }
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -529,22 +529,21 @@ class Blocks {
             block.targetId = isSpriteSpecific ? optRuntime.getEditingTarget().id : null;
 
             if (wasMonitored && !block.isMonitored) {
-                optRuntime.requestRemoveMonitor(block.id);
+                optRuntime.requestHideMonitor(block.id);
             } else if (!wasMonitored && block.isMonitored) {
-                optRuntime.requestAddMonitor(MonitorRecord({
-                    id: block.id,
-                    targetId: block.targetId,
-                    spriteName: block.targetId ? optRuntime.getTargetById(block.targetId).getName() : null,
-                    opcode: block.opcode,
-                    params: this._getBlockParams(block),
-                    // @todo(vm#565) for numerical values with decimals, some countries use comma
-                    value: '',
-                    x: block.x,
-                    y: block.y,
-                    mode: block.opcode === 'data_listcontents' ? 'list' : block.mode,
-                    sliderMin: block.sliderMin,
-                    sliderMax: block.sliderMax
-                }));
+                // Tries to show the monitor for specified block. If it doesn't exist, add the monitor.
+                if (!optRuntime.requestShowMonitor(block.id)) {
+                    optRuntime.requestAddMonitor(MonitorRecord({
+                        id: block.id,
+                        targetId: block.targetId,
+                        spriteName: block.targetId ? optRuntime.getTargetById(block.targetId).getName() : null,
+                        opcode: block.opcode,
+                        params: this._getBlockParams(block),
+                        // @todo(vm#565) for numerical values with decimals, some countries use comma
+                        value: '',
+                        mode: block.opcode === 'data_listcontents' ? 'list' : block.mode
+                    }));
+                }
             }
             break;
         }
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -539,6 +539,8 @@ class Blocks {
                     params: this._getBlockParams(block),
                     // @todo(vm#565) for numerical values with decimals, some countries use comma
                     value: '',
+                    x: block.x,
+                    y: block.y,
                     mode: block.opcode === 'data_listcontents' ? 'list' : block.mode,
                     sliderMin: block.sliderMin,
                     sliderMax: block.sliderMax
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -540,12 +540,8 @@ class Blocks {
                     // @todo(vm#565) for numerical values with decimals, some countries use comma
                     value: '',
                     mode: block.opcode === 'data_listcontents' ? 'list' : block.mode,
-                    x: block.x,
-                    y: block.y,
                     sliderMin: block.sliderMin,
-                    sliderMax: block.sliderMax,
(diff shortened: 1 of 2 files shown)


kchadha:
@paulkaplan also added you to this, for sanity checking =D

src/engine/runtime.js
```diff
@@ -1522,7 +1522,18 @@ class Runtime extends EventEmitter {
      * @param {!MonitorRecord} monitor Monitor to add.
      */
     requestAddMonitor (monitor) {
-        this._monitorState = this._monitorState.set(monitor.get('id'), monitor);
+        const id = monitor.get('id');
+        if (this._monitorState.has(id)) {
```
kchadha: Maybe you can just use `requestUpdateMonitor` here instead of defining the functionality twice. `requestUpdateMonitor` can return a boolean representing whether a monitor already existed in the state, and then if it didn't, you can set it here as in the else case below.

kchadha (approved):
LGTM!

kchadha (changes requested):
Blocks don't have some of the properties you're referencing in the 'checkbox' case for the changeBlock function. Maybe we should try to repro the case we talked about offline.

src/engine/blocks.js
```diff
@@ -529,18 +529,21 @@ class Blocks {
...
+                if (!optRuntime.requestShowMonitor(block.id)) {
+                    optRuntime.requestAddMonitor(MonitorRecord({
+                        id: block.id,
+                        targetId: block.targetId,
+                        spriteName: block.targetId ? optRuntime.getTargetById(block.targetId).getName() : null,
+                        opcode: block.opcode,
+                        params: this._getBlockParams(block),
+                        // @todo(vm#565) for numerical values with decimals, some countries use comma
+                        value: '',
+                        mode: block.opcode === 'data_listcontents' ? 'list' : block.mode
```
kchadha: `block.mode` here should go back to being `default` since `block.mode` doesn't exist.

src/engine/runtime.js
```diff
@@ -1522,24 +1522,37 @@ class Runtime extends EventEmitter {
      * @param {!MonitorRecord} monitor Monitor to add.
```
kchadha, line 1522: This comment should be updated to replace 'overwrites it.' with something like: 'updates those properties that are defined in the given monitor record.'

src/engine/runtime.js
```diff
@@ -1522,24 +1522,37 @@ class Runtime extends EventEmitter {
...
+                        return prev;
+                    }
+                    return next;
+                }, monitor));
+            return true;
         }
+        return false;
     }
-
+
```
kchadha: stray whitespace (indent) added here

src/engine/runtime.js
```diff
@@ -1549,6 +1562,32 @@ class Runtime extends EventEmitter {
         this._monitorState = this._monitorState.delete(monitorId);
     }
 
+    /**
+     * Hides a monitor. Does nothing if the monitor already does
```
kchadha: Replace 'Does nothing' with 'Returns false'

src/engine/runtime.js
```diff
@@ -1522,24 +1522,37 @@ class Runtime extends EventEmitter {
...
+        const id = monitor.get('id');
+        if (!this.requestUpdateMonitor(monitor)) { // update monitor if it exists in the state
+            // if the monitor did not exist in the state, add it
+            this._monitorState = this._monitorState.set(id, monitor);
+        }
     }
 
     /**
-     * Update a monitor in the state. Does nothing if the monitor does not already
+     * Update a monitor in the state. Does nothing and returns false if the monitor does not already
```
kchadha: Maybe replace this description with: 'Update a monitor in the state and report its success/failure.'

src/engine/runtime.js
kchadha: JK -- new suggestion (same as above):
> Maybe replace this description with: 'Update a monitor in the state and report its success/failure.'

src/engine/runtime.js
```diff
@@ -1549,6 +1562,32 @@ class Runtime extends EventEmitter {
         this._monitorState = this._monitorState.delete(monitorId);
     }
 
+    /**
+     * Hides a monitor. Does nothing if the monitor already does
+     * not exist in the state.
+     * @param {!string} monitorId ID of the monitor to remove.
```
kchadha: 'remove' -> 'hide'

src/engine/runtime.js
```diff
@@ -1549,6 +1562,32 @@ class Runtime extends EventEmitter {
...
+        return this.requestUpdateMonitor(new Map([
+            ['id', monitorId],
+            ['visible', false]
+        ]));
+    }
+
+    /**
+     * Shows a monitor. Does nothing if the monitor already does
+     * not exist in the state.
+     * @param {!string} monitorId ID of the monitor to remove.
```
kchadha: 'remove' -> 'show'

kchadha: Also same comment as above re: 'Does nothing...'

kchadha (changes requested):
Just had some last few minor comments!

kchadha (approved):
LGTM

PR #1222, opened 2018-06-08
Speech extension icons

picklesrus:
### Resolves
A couple TODOs in the speech extension.

### Proposed Changes
Replace the icon links with data uri versions downloaded from https://material.io/tools/icons/


ericrosenbaum (approved):
🎤looks good🎤

Issue #1223, opened 2018-06-09
Speech Extension: Pick default text for When I hear block


picklesrus:
This is currently set to "cat".  What should it be?

thisandagain:
From discussion: use "let's go" as the default text

kyleplo:
What about "Scratch on!"?

towerofnix:
@kyleplo "Scratch on" is usually something people say at the end of whatever they're writing, like a goodbye, as far as I've seen. But "let's go" (as @thisandagain mentioned) really means to start doing something, which I think makes sense for this.

kyleplo:
@towerofnix Good point.

picklesrus:
Fixed in #1224

PR #1224, opened 2018-06-09
Turn speech block text into i18n messages.

picklesrus:
### Resolves
i18n issues with speech extension.

### Proposed Changes

Add messages for the text in the speech extension blocks.

### DIFF
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -718,7 +718,7 @@ class Scratch3SpeechBlocks {
                             type: ArgumentType.STRING,
                             defaultValue: formatMessage({
                                 id: 'speech.defaultWhenIHearValue',
-                                default: 'cat',
+                                default: 'let\'s go',
                                 description: 'The default phrase/word that, when heard, triggers the event.'
                             })
                         }
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -689,7 +689,7 @@ class Scratch3SpeechBlocks {
             name: formatMessage({
                 id: 'speech.extensionName',
                 default: 'Google Speech',
-                description: 'Name of extension that add speech recognition blocks.'
+                description: 'Name of extension that adds speech recognition blocks.'
             }),
             menuIconURI: menuIconURI,
             blockIconURI: iconURI,
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -689,7 +689,7 @@ class Scratch3SpeechBlocks {
             name: formatMessage({
                 id: 'speech.extensionName',
                 default: 'Google Speech',
-                description: 'Name of extension that adds speech recognition blocks.'
+                description: 'Name of extension that adds speech recognition blocks. Do Not translate Google.'
             }),
             menuIconURI: menuIconURI,
             blockIconURI: iconURI,
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -1,6 +1,7 @@
 const ArgumentType = require('../../extension-support/argument-type');
 const Cast = require('../../util/cast');
 const BlockType = require('../../extension-support/block-type');
+const formatMessage = require('format-message');
 const log = require('../../util/log');
 const DiffMatchPatch = require('diff-match-patch');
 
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_speech/index.js
```diff
@@ -691,23 +692,35 @@ class Scratch3SpeechBlocks {
             blocks: [
                 {
                     opcode: 'listenAndWait',
-                    text: 'Listen and Wait',
+                    text: formatMessage({
+                        id: 'speech.ListenAndWait',
```
ericrosenbaum: please change to camel case (speech.listenAndWait)

src/extensions/scratch3_speech/index.js
```diff
@@ -691,23 +692,35 @@ class Scratch3SpeechBlocks {
...
+                        default: 'listen and wait',
+                        description: 'Start listening to the microphone and wait for data.'
```
ericrosenbaum: maybe change to 'Start listening to the microphone and wait for a result from the speech recognition system'

src/extensions/scratch3_speech/index.js
```diff
@@ -691,23 +692,35 @@ class Scratch3SpeechBlocks {
...
+                        id: 'speech.whenIHear',
+                        default: 'when I hear [PHRASE]',
+                        description: 'Text on hat block that triggers when phrase is heard.'
+                    }),
                     blockType: BlockType.HAT,
                     arguments: {
                         PHRASE: {
                             type: ArgumentType.STRING,
-                            defaultValue: 'cat'
+                            defaultValue: 'cat' // TODO: Choose something.
```
ericrosenbaum: I think this will need to be a formatMessage as well (we can leave 'cat' for now as the message)

src/extensions/scratch3_speech/index.js
```diff
@@ -691,23 +692,35 @@ class Scratch3SpeechBlocks {
...
                         }
                     }
                 },
                 {
                     opcode: 'getSpeech',
-                    text: 'speech',
+                    text: formatMessage({
+                        id: 'speech.speechReporter',
+                        default: 'speech',
+                        description: 'Text on speech reporter block.'
```
ericrosenbaum: These descriptions can be verbose, for use by translators, and this one-word block might be ambiguous (I guess the word "speech" might have many possible translations in different parts of speech etc). Let's try something like: 'Get the text of spoken words transcribed by the speech recognition system'

src/extensions/scratch3_speech/index.js
```diff
@@ -691,23 +692,35 @@ class Scratch3SpeechBlocks {
...
+                    }),
                     blockType: BlockType.COMMAND
                 },
                 {
                     opcode: 'whenIHearHat',
-                    text: 'When I hear [PHRASE]',
+                    text: formatMessage({
+                        id: 'speech.whenIHear',
+                        default: 'when I hear [PHRASE]',
+                        description: 'Text on hat block that triggers when phrase is heard.'
```
ericrosenbaum: How about "Event that triggers when the text entered on the block is recognized by the speech recognition system"

ericrosenbaum (changes requested):
I talked with @chrisgarrity about how we use the description fields, and we realized that we don't have clear standards yet for descriptions for hat blocks and reporter blocks. I made comments with suggested changes, where hats use "event that triggers when..." and reporters use "get the..." 

Also, we realized that we will need the extension name to be translated too! Please add that one as well.

picklesrus:
Thanks for the feedback. Messages have been updated!

src/extensions/scratch3_speech/index.js
picklesrus: done!

src/extensions/scratch3_speech/index.js
picklesrus: done!

src/extensions/scratch3_speech/index.js
picklesrus: done!

src/extensions/scratch3_speech/index.js
picklesrus: done!

src/extensions/scratch3_speech/index.js
picklesrus: done!

src/extensions/scratch3_speech/index.js
```diff
@@ -685,29 +686,51 @@ class Scratch3SpeechBlocks {
     getInfo () {
         return {
             id: 'speech',
-            name: 'Google Speech',
+            name: formatMessage({
+                id: 'speech.extensionName',
+                default: 'Google Speech',
+                description: 'Name of extension that adds speech recognition blocks.'
```
chrisgarrity: We may want to add 'Do not translate Google' to the description. I'm guessing that people usually wouldn't, but it doesn't hurt to be explicit.

picklesrus: Good point! Done! The conventions are all starting to come back to memory :)

picklesrus:
Replaced default 'cat' with 'let's go' as discussed.  This will also resolve https://github.com/LLK/scratch-vm/issues/1223

ericrosenbaum (approved):
LGTM!

PR #1229, opened 2018-06-12
Update position for comments imported from 2.0 projects after auto-positioning.

kchadha:
### Proposed Changes

Update the position of block comments imported from 2.0 when scratch-blocks emits a comment_create event.

### Reason for Changes

Block comments imported from 2.0 projects have their position coordinates initially set to `null` so that they can be auto positioned by scratch-blocks. However, because of this, the position does not get updated on the block when scratch-blocks emits a create events. Thus, every time the workspace is reloaded, the comments are auto-positioned again. This looks fine for the most part, except for when changing the minimized state of a block comment before moving it. See GIFs below for before and after.

Before:

block_comment_positioning_before

After: 

block_comment_positioning_after

### Test Coverage

Added unit tests for commentCreate function and added a unit test for testing this particular change.
This latter test was added to test/unit/virtual-machine, which might be a weird place for it because it's not necessarily testing functionality for that specific file/module. The reason I put it there was because other tests in virtual machine already have some of the boiler plate needed for this test as well (e.g access to the runtime and the target).

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -362,6 +362,18 @@ class Blocks {
                 const currTarget = optRuntime.getEditingTarget();
                 currTarget.createComment(e.commentId, e.blockId, e.text,
                     e.xy.x, e.xy.y, e.width, e.height, e.minimized);
+
+                if (currTarget.comments[e.commentId].x === null &&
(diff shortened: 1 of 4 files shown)


paulkaplan (approved):
LG, and great keeping up with adding more tests!

PR #1231, opened 2018-06-13
Add top-level `reorderTarget` API to VM

paulkaplan:
Allows reordering of the targets themselves. 

Add a unit test for this reordering, make it match the way the costume/sounds APIs work (i.e. consume an index to move and a new index, clamp the indices and return true/false based on whether a reorder occurs).

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -6,6 +6,7 @@ const Buffer = require('buffer').Buffer;
 const centralDispatch = require('./dispatch/central-dispatch');
 const ExtensionManager = require('./extension-support/extension-manager');
 const log = require('./util/log');
+const MathUtil = require('./util/math-util');
 const Runtime = require('./engine/runtime');
 const sb2 = require('./serialization/sb2');
 const sb3 = require('./serialization/sb3');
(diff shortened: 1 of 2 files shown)


src/virtual-machine.js
```diff
@@ -1033,6 +1034,25 @@ class VirtualMachine extends EventEmitter {
         return null;
     }
 
+    /**
+     * Reorder target by index. Return whether a change was made.
+     * @param {!string} targetIndex Index of the target.
+     * @param {!number} newIndex index that the target should be moved to.
+     * @returns {boolean} Whether a target was reordered.
+     */
+    reorderTarget (targetIndex, newIndex) {
```
kchadha, line 1043: I'm wondering if it would be more useful to have this function take a targetId instead of a targetIndex? I guess then it would be called 'setTargetOrder' or 'setTargetIndex' or something.

src/virtual-machine.js
```diff
@@ -1033,6 +1034,25 @@ class VirtualMachine extends EventEmitter {
...
+        targetIndex = MathUtil.clamp(targetIndex, 0, this.runtime.targets.length - 1);
+        newIndex = MathUtil.clamp(newIndex, 0, this.runtime.targets.length - 1);
+        if (targetIndex === newIndex) return false;
+        const target = this.runtime.targets[targetIndex];
+        let targets = this.runtime.targets;
```
kchadha: You reference `this.runtime.targets` a few times before this line. You can move this declaration up.

kchadha (changes requested):
Looks good pending an actual change request and an optional suggestion.

src/virtual-machine.js
paulkaplan: I initially had this, but the way the drag/drop code is working with respect to the GUI it is easiest  to just treat it similar to costume/sounds, just referring by index. It would just take more work to look up the targets id first.

PR #1232, opened 2018-06-13
Make the end listening sound only play one time even if multiple bloc…

picklesrus:
…ks stop listening at the same time.

### Resolves

Loud meows noticed during smoke test.

### Reason for Changes

If they all play at the same time, the end sound stack on top of each other and it can get quite loud.


ericrosenbaum (approved):
LGTM!

PR #1233, opened 2018-06-13
Add extension status button

ericrosenbaum:
### Proposed Changes

Add support for extensions to optionally show a status button, to be used by hardware extensions. Also add the relevant flag to the microbit extension for testing.

Depends on https://github.com/LLK/scratch-blocks/pull/1586

### Reason for Changes

Work in progress on supporting hardware extensions, where the user needs to be able to initiate a bluetooth connection workflow.


rschamp (approved):
Looks good!

PR #1236, opened 2018-06-14
Comment Save & Load

kchadha:
### Resolves

Saving and loading comments.

### Proposed Changes

Serialize and deserialize information about comments when saving and loading a project respectively.

### Reason for Changes

Comments should be saved and loaded along with other project data.

### Test Coverage

Manual testing


src/serialization/sb3.js
```diff
@@ -834,6 +860,24 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
...
+                comment.id,
+                comment.text,
+                comment.x,
+                comment.y,
+                comment.width,
+                comment.height,
+                comment.minimized
+            );
+            if (comment.blockId) {
+                newComment.blockId = comment.blockId;
```
paulkaplan, line 876: Is it worth checking `target.blocks` to see if that blockId exists?

paulkaplan (approved):
Looks good! 

This is fine how it is, considering how little test coverage there is for sb3's right now, but I would love if you could include an entry for comments in the `sb3-roundtrip` test, and/or this is a good time to at least add a little something to the `serialization_sb3` unit test?

PR #1239, opened 2018-06-15
Use new scratch-audio SoundBank in vm / sounds

gnarf:
## Depends on llk/scratch-audio#92

Still a little bit work in progress, but ready to start talking about.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -170,21 +170,30 @@ class RenderedTarget extends Target {
         }
     }
 
+    get audioPlayer () {
+        /* eslint-disable no-console */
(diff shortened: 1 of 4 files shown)


src/sprites/sprite.js
```diff
@@ -155,6 +161,12 @@ class Sprite {
 
         return Promise.all(assetPromises).then(() => newSprite);
     }
+
+    dispose () {
```
gnarf, line 165: i don't think this is getting called, @ericrosenbaum where would I call it?

src/sprites/rendered-target.js
```diff
@@ -170,21 +170,30 @@ class RenderedTarget extends Target {
         }
     }
 
+    get audioPlayer () {
```
gnarf, line 173: kill this

PR #1240, opened 2018-06-15
Add comment id to block when converting from XML to vm representation.

kchadha:
### Resolves

Resolves LLK/scratch-gui#2366

### Proposed Changes

Add comment id to block when converting block xml (containing a block comment) to VM representation of block.

### Reason for Changes

LLK/scratch-gui#2366

### Test Coverage

Updated tests for block xml adapter.


paulkaplan (approved):
LG

PR #1244, opened 2018-06-18
Add shareSound and shareCostume APIs with unit tests to the VM

paulkaplan:
Another step towards drag and drop sharing of assets between sprites. 

The inner part of the duplication  follows the pattern from the `duplicateSound` and `duplicateCostume` methods. 

It is worth noting that the method for sharing costumes, like duplicating costumes, also enforces the feature where the new costume becomes the active costume. This is true in 2.0 as well, and gives a nice bit of feedback that a transfer has occurred.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -936,6 +936,46 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kchadha (approved):
Looks good!

PR #1246, opened 2018-06-18
Localize category labels

### LINKED ISSUE #1245: Localize extension names for category labels.
### Expected Behavior

Labels for extension categories should be translatable, for extensions that are localized.

### Actual Behavior

The name attribute isn't localized, so category labels are all in English.

chrisgarrity:
Added formatMessage to the name attribute for:
* music
* pen
* translate
* video motion

Skipped speak extension - trademarked name
Skipped extensions that are not yet localized at all.

### Resolves

Fixes #1245 

### Proposed Changes

Adds `formatMessage` calls to the `name` attribute in `getInfo`

### Reason for Changes

### Testing
[Note: there are issues with extensions related to switching languages after loading an extension]

- [ ] Start with or switch to French, add the 'Pen' Extension - make sure the category label is translated
- [ ] Start or switch into a non-latin language (Japanese, Chinese, Thai), add all extensions

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -623,7 +623,11 @@ class Scratch3MusicBlocks {
     getInfo () {
         return {
             id: 'music',
-            name: 'Music',
+            name: formatMessage({
+                id: 'music.categoryName',
+                default: 'Music',
+                description: 'Label for the Music extension category'
+            }),
             menuIconURI: menuIconURI,
             blockIconURI: blockIconURI,
             blocks: [
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -282,7 +282,11 @@ class Scratch3PenBlocks {
     getInfo () {
         return {
             id: 'pen',
-            name: 'Pen',
+            name: formatMessage({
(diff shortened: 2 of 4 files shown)


cwillisf (approved):
Looks good! My one hesitation is the term `categoryName`: my first impulse is to refer to the name of the extension rather than the category, and if we want to allow those to differ then `categoryName` would be the optional one (defaulting to the extension's name).

That said, if it's no big deal to rename the field in the future then I would say let's put off that decision until/unless we find out that we need it, and leave this change as you have it.

chrisgarrity:
@cwillisf good point. I called it `categoryName` to be clear this is the text that will be used in the category labels. However, I can imagine that there will be extensions where the extension has a particular name, and the label will actually be something different (probably shorter). If that happens, there will need to be a new 'category' property - it will use the `<extension>.categoryName` id, and there should be a new id `<extension>.extensionName`. Since most of the extensions will use the same default text for both, Transifex should automatically fill in any previous translations of the category name for the extension name. 

/cc @ericrosenbaum @picklesrus

BryceLTaylor:
@chrisgarrity could you add test cases to this pr so I can test them?  I'm not sure what to test here.

PR #1254, opened 2018-06-19
Skip unknown opcode blocks while parsing

### LINKED ISSUE #2392: Preview Crashing on project 219715238
### Expected Behavior

Project should load with ~5000+ assets

### Actual Behavior

Immediately crashes

### Steps to Reproduce

Load https://preview.scratch.mit.edu/#219715238

macOS 10.13.14 Version 67.0.3396.87 (Official Build) (64-bit)

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/2392

### Proposed Changes

Be very slightly more careful when parsing blocks to make sure they exist before trying to use them.

### Reason for Changes

Basically we are treating `parseBlock` like it always returns an array, but it can return undefined if the specmap lookup fails. So for now, just skip those blocks in the processing. 

image

### Test Coverage

Added an sb2 test fixture derived from the project in https://github.com/LLK/scratch-gui/issues/2392 

The test explains what is in the fixture vs. what the expectations assert


colbygk:
You are the **best**

towerofnix:
Is there a warning for skipped blocks, so that if, for example, a project relies on a block that wasn't implemented in 3.0 (somehow), we would see that the block is missing in the console?

src/serialization/sb2.js
```diff
@@ -133,17 +133,20 @@ const parseBlockList = function (blockList, addBroadcastMsg, getVariableId, exte
...
             extensions, comments, commentIndex);
-        const parsedBlock = parsedBlockAndComments[0];
-        // Update commentIndex
-        commentIndex = parsedBlockAndComments[1];
 
-        if (typeof parsedBlock === 'undefined') continue;
-        if (previousBlock) {
-            parsedBlock.parent = previousBlock.id;
-            previousBlock.next = parsedBlock.id;
+        if (parsedBlockAndComments) { // Could fail due to unrecognized op-codes
```
kchadha: If a block that is undefined in Scratch 3.0 had a block comment on it in Scratch 2.0, it should probably get removed from the comments list. Right now, it gets attached to the next block in the stack (if there are more blocks in the stack) or doesn't display anywhere if there are no more blocks in the stack, but there's still a comment associated with it (that isn't done being parsed -- e.g. a blockId gets associated with it), so saving and loading that project in 3.0 then fails.

kchadha: So that actually means two things: 
1) `commentIndex` should still be updated in `parseBlock` even if it returns undefined (because that commentIndex refers to a position in a 2.0 stack).
2) Any block comment associated with an undefined block should get removed from the target's comments map (so that it doesn't get serialized later)

Testing this made me realized that this is also an actual issue for how I'm handling importing comments from 2.0 where multiple block comments are attached to the same block. Tracking in issue #1257

src/serialization/sb2.js
```diff
@@ -133,11 +133,12 @@ const parseBlockList = function (blockList, addBroadcastMsg, getVariableId, exte
         // eslint-disable-next-line no-use-before-define
         const parsedBlockAndComments = parseBlock(block, addBroadcastMsg, getVariableId,
             extensions, comments, commentIndex);
+
```
kchadha: Nit: stray newline

src/serialization/sb2.js
```diff
@@ -706,9 +707,19 @@ const specMapBlock = function (block) {
  * and second item is the updated comment index (after this block and its children are parsed)
  */
 const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extensions, comments, commentIndex) {
+    const commentsForParsedBlock = (comments && typeof commentIndex === 'number' && !isNaN(commentIndex)) ?
+        comments[commentIndex] : null;
     const blockMetadata = specMapBlock(sb2block);
     if (!blockMetadata) {
-        return;
+        // No block opcode found, exclude this block, increment the comment id and
+        // send all linked comments back to zero/zero to prevent serialization issues.
```
kchadha: You should probably also say that we're making them workspace comments

kchadha (commented):
Looks good apart from two nits.

paulkaplan:
@kchadha thanks, fixed those two issues. Struggling with build issues right now, but I assume good to merge after that is resolved?

src/serialization/sb2.js
```diff
@@ -711,8 +710,9 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
         comments[commentIndex] : null;
     const blockMetadata = specMapBlock(sb2block);
     if (!blockMetadata) {
-        // No block opcode found, exclude this block, increment the comment id and
-        // send all linked comments back to zero/zero to prevent serialization issues.
+        // No block opcode found, exclude this block, increment the comment id,
```
kchadha: Last nit, sorry! `comment id` should be `commentIndex` above.

PR #1255, opened 2018-06-19
Serialize extensions used in a project

hyperobject:
### Proposed Changes
This PR adds an additional section to target objects in an sb3 file to hold a list of extension IDs used by blocks in that target. This list of IDs is generated in the `serializeBlocks` function, which has been modified to include an opcode parsing step in the first `for (const blockID in blocks)` loop.

### Reason for Changes
This change is necessary so the project page can easily find the extensions used in a project and display them to a user.

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -466,11 +466,16 @@ const serializeTarget = function (target, extensions) {
 const serialize = function (runtime) {
     // Fetch targets
     const obj = Object.create(null);
+    // Create extension set to hold extension ids found while serializing targets
+    const extensions = new Set();
     const flattenedOriginalTargets = JSON.parse(JSON.stringify(
         runtime.targets.filter(target => target.isOriginal)));
-    obj.targets = flattenedOriginalTargets.map(t => serializeTarget(t, runtime));
+    obj.targets = flattenedOriginalTargets.map(t => serializeTarget(t, extensions));
 
     // TODO Serialize monitors
+
+    // Assemble extension list
+    obj.extensions = Array.from(extensions);
 
     // Assemble metadata
     const meta = Object.create(null);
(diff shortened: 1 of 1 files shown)


src/serialization/sb3.js
```diff
@@ -901,6 +909,9 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
     if (object.hasOwnProperty('isStage')) {
         target.isStage = object.isStage;
     }
+    if (object.hasOwnProperty('extensions')) {
+        target.extensions = object.extensions;
```
kchadha: Does this actually need to be deserialized?

hyperobject: No, not necessarily. I mainly added it for the symmetry of the thing but it doesn't really serve any purpose right now.

src/serialization/sb3.js
```diff
@@ -416,13 +422,15 @@ const serializeComments = function (comments) {
  */
 const serializeTarget = function (target) {
     const obj = Object.create(null);
+    const extensionIDs = new Set();
```
kchadha: I think I would rather keep the pattern of the rest of the file of each function initializing and returning a sub structure. E.g. I think serializeBlocks should be the one to initialize extensionIDs (and then return it) since it's the function that's populating it.

src/serialization/sb3.js
kchadha: I think it should be removed then. We can add it if/when we need it.

kchadha (approved):

rschamp (changes requested):
After speaking with @thisandagain, it would be better for the project index if the extensions were stored at the top level of the schema (next to `meta` and `targets`), rather than within each target.

rschamp (approved):

rschamp:
(except for the conflict)

PR #1258, opened 2018-06-20
Use new scratch-audio APIs in music extension

mzgoddard:
- [ ] Depends on https://github.com/LLK/scratch-audio/pull/96
- [ ] Depends on #1239

### Proposed Changes

Update music extension to not use AudioPlayer so AudioPlayer may be deprecated and removed.

### Reason for Changes

AudioPlayer is being deprecated in favor of SoundBank and SoundPlayer.

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -52,18 +52,25 @@ class Scratch3MusicBlocks {
         this._concurrencyCounter = 0;
 
         /**
-         * An array of audio buffers, one for each drum sound.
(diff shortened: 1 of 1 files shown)


mzgoddard:
At this moment, the code should not work. It needs to be paired with the [SoundBank vm pr](#1239). I have separated it out so it can be reviewed separately. https://github.com/mzgoddard/scratch-vm/tree/soundbank-music can be used to try the change.

gnarf:
Reading through this I'm wondering if it's intentional that changes to pan/pitch won't affect already playing music notes for a target.

PR #1269, opened 2018-06-25
delete clone should not stop sound

gnarf:
### Resolves
Deleting a clone **SHOULD NOT** stop it's sounds from playing

### Proposed Changes

Remove stop sounds hook from `dispose()` method, and update `onStopAll` to stop all sounds on the sprite (so deleted clone sounds still stop)


ericrosenbaum (approved):
👍

PR #1270, opened 2018-06-25
Enable extension status button via XML attribute

ericrosenbaum:
Depends on https://github.com/LLK/scratch-blocks/pull/1610

### Proposed Changes

For extensions that wish to display a status indicator button in the blocks menu, instead of adding a button directly to the XML, add an attribute to the XML: `showStatusButton="true"`. The blocks will then add a new header element containing both the category label and the status button.


cwillisf (approved):
👍

PR #1272, opened 2018-06-26
Update speech extension to use new audio engine api since the old ver…

picklesrus:
…sion no longer exists.

Update to use newer audio api.

### Reason for Changes

Audio api changed


ericrosenbaum (approved):
LGTM!

PR #1280, opened 2018-06-28
Fix a few say bubble issues

### LINKED ISSUE #1205: Say bubble should hide on stop even when sprite is hidden
### Expected Behavior

Pressing the stop button should always hide the say bubble.

### Actual Behavior

If the stop button is pressed while the sprite is hidden, and then you show the sprite, the say bubble comes back.

### Steps to Reproduce

- Run a "say" block
- Hide the sprite
- Press the stop button
- Show the sprite (bubble should be gone)

### LINKED ISSUE #1268: Say bubble defaults to the left now, should be to the right
### Actual Behavior

"Say" using the default sprite and message used to put the bubble on the right, as it does in scratch 2, but now for some reason it goes to the left?

### Steps to Reproduce

1. Open https://llk.github.io/scratch-gui/develop/
2. Click "Looks" and then click the "Say" bubble
3. It goes to the left of the sprite.

image

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/1205
Fixes https://github.com/LLK/scratch-vm/issues/1268

### Proposed Changes

Two separate fixes, but very nearby. Separated into two separate commits. One makes sure to reset the `text` in bubble state when resetting due to "stop" button, the other commit uses the bounds of the renderer instead of the bounds of the stage target to compare bubble size against.

Run the repros from the linked issues.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -107,6 +107,8 @@ class Scratch3LooksBlocks {
      */
     _onResetBubbles () {
         for (let n = 0; n < this.runtime.targets.length; n++) {
+            const bubbleState = this._getBubbleState(this.runtime.targets[n]);
+            bubbleState.text = '';
             this._onTargetWillExit(this.runtime.targets[n]);
         }
         clearTimeout(this._bubbleTimeout);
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
Looks good!

PR #1285, opened 2018-07-02
Device connection update for micro:bit and EV3 extensions

ericrosenbaum:
Add functions and events to the VM for managing connections to hardware extensions and communicating with the GUI's device connection modals (in https://github.com/LLK/scratch-gui/pull/2537). 

The changes include updates to `BLESession` and `BTSession`, changes to the micro:bit extension, an all-new EV3 extension, and removal of the peripheral chooser.

**Notes**

- We need to rename some things to make clear the distinction between extensions, peripherals and devices.
- EV3 extension needs re-factoring to clean up enums, static vars, and generation of command byte-codes. 
- EV3 extension is missing most of the motor blocks (rotate by degrees, set position, set power), and does not yet correctly handle braking and coasting.
- EV3 "all" motor options not added. 
- EV3 buttons need to be read individually by port.
- Microbit extension does not correctly handle timing of command blocks that update the LED display.

**Known issues**

- The EV3 extension randomly disconnects after some amount of time (from a few seconds to several minutes) with the error: `WebSocket connection to 'ws://localhost:20110/scratch/ble' failed: 
A server must not mask any frames that it sends to the client.`
- There are a number of other errors related to connection and disconnection that need further work to handle correctly.
- EV3 signal strength (RSSI) is always reported as 127.
- Stubbed-out need re-naming, and need to be filled in.


src/engine/runtime.js
```diff
@@ -254,6 +254,8 @@ class Runtime extends EventEmitter {
             video: new Video(this)
         };
 
+        this.extensionDevices = {};
```
cwillisf, line 257: I'd like to standardize on the term "peripheral" to match the terminology in Scratch Link. I don't know if that's the right term for user-facing text, but I think it's a good, unambiguous term to use in the code at least

src/extensions/scratch3_ev3/index.js
```diff
@@ -0,0 +1,921 @@
...
+        if (speed < 0) {
+            speed = -1 * speed;
+            n = -1 * n;
+        }
+
+        // If the input value is less than 0
+        const dir = (n < 0) ? 0x100 - speed : speed; // step negative or possitive
+        n = Math.abs(n);
+
+        // TODO ^^^^^ Is all of this shit really needed? ^^^^^ - ascii
```
cwillisf: It looks like this comment started out in the prototype code, but I'd like to keep the language in our official repositories clean :)

src/util/base64-util.js
```diff
@@ -28,6 +28,21 @@ class Base64Util {
...
+    */
+    static arrayBufferToBase64 (buffer) {
+        let binary = '';
+        const bytes = new Uint8Array(buffer);
+        const len = bytes.byteLength;
+        for (let i = 0; i < len; i++) {
+            binary += String.fromCharCode(bytes[ i ]);
+        }
+        return btoa(binary);
+    }
```
cwillisf, line 44: This looks to me like it might be very slow for large buffers. Is this only used for small buffers, say under 1k? If so, I think it's fine to leave it as is and come back if it becomes a problem later. If it does get used for large buffers then I'd suggest checking it with a profiler: it could turn out that it's fine, or maybe we need to come up with a faster approach.

src/util/jsonrpc-web-socket.js
```diff
@@ -27,6 +28,9 @@ class JSONRPCWebSocket extends JSONRPC {
 
     _onSocketMessage (e) {
         const json = JSON.parse(e.data);
+        if (json.method !== 'characteristicDidChange') {
+            // log.info(`received message: ${json}`);
+        }
```
cwillisf: Remove?

cwillisf (changes requested):
It's up to you how many of my suggestions you'd like to address now vs future work, but I at least think it's important to clean up that one comment...

src/engine/runtime.js
ericrosenbaum: yup, this will be part of a re-factor

src/extensions/scratch3_ev3/index.js
ericrosenbaum: good catch!

src/util/base64-util.js
ericrosenbaum: yup the buffers for EV3 communication are small

src/util/jsonrpc-web-socket.js
ericrosenbaum: yup

src/engine/runtime.js
evhan55: 👍

Issue #1278, opened 2018-06-28
Should not be able to create a global variable if one with the same name and type exists locally in any sprite.


kchadha:
### Expected Behavior

Should not be able to create a name conflict with existing variables in the scope of the variable being created.

### Actual Behavior

If a project has a sprite local variable 'foo' and you select a sprite that doesn't have that variable, and you create a new (global) variable named 'foo' there is no complaint about name collisions. Switching back to the sprite with the local variable named 'foo' causes a console error and the blocks on that sprite's workspace don't load.

### Steps to Reproduce

1. Create a project in Scratch 2.0 with 2 sprites.
2. Add a new sprite-local variable to only one of the sprites.
3. Open the 2.0 project in 3.0
4. Select the sprite that doesn't have the sprite-local variable
5. Create a variable using the toolbox with the same name as the sprite-local variable in the other sprite.
6. Observe the new variable blocks in the toolbox and no errors in the console.
7. Switch to the other sprite. Observe errors in the console.

PR #1301, opened 2018-07-04
Variable scope

kchadha:
### Resolves

Allows creating local variables.
Towards resolving #1278

### Proposed Changes

Allows creating local variables. Adds variable to appropriate target based on a new `isLocal` flag from a `var_create` event.

Global vars now check for conflict against all target local variables instead of just the currently selected one.

### Test Coverage

Manual testing.

#### Related PRs

This PR should be merged second.

This PR is related to the following PRs:

LLK/scratch-blocks#1621
LLK/scratch-gui#2548

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -330,19 +330,25 @@ class Blocks {
             this.deleteBlock(e.blockId);
             break;
         case 'var_create':
-            // New variables being created by the user are all global.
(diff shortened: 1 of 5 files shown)


paulkaplan:
Can you update this to include serializing scope to xml?

paulkaplan (approved):
LG. Are there any existing tests for the `var_create` etc. events? If so we should probably add to them, if not then ...

PR #783, opened 2017-11-14
Add an extension to support KOOV

enami:
### Resolves

Currently, scratch3 doesn't have an ability to control KOOV (an educational kit to build blocks with some sensors and actuators).  This PR adds an extension to add scratch3 blocks for sensors and actuators comes with KOOV.  Visit following page for more detail on KOOV (https://www.koov.io)

### Proposed Changes

Implements some scratch3 blocks to interact with sensors and actuators.

To test this extension on Mac, you will need to
- connect KOOV to Mac via USB.
- run this helper [server](https://github.com/sonyged/koov_scratch_server) on your Mac.  It will listen on 127.0.0.1:3030 for new connection.
- modify src/io/deviceManager.js to use http instead of https to access device manager.
```
diff --git a/src/io/deviceManager.js b/src/io/deviceManager.js
index d2d0046..bb32961 100644
--- a/src/io/deviceManager.js
+++ b/src/io/deviceManager.js
@@ -257,7 +257,7 @@ class DeviceManager {
      * @return {string} - The default Scratch Device Manager connection URL.
      */
     static get DEFAULT_SERVER_URL () {
-        return 'https://device-manager.scratch.mit.edu:3030';
+        return 'http://device-manager.scratch.mit.edu:3030';
     }
 
     constructor () {
```
- add following entry to the list of extensions in scratch-gui
```
diff --git a/src/lib/libraries/extensions/index.js b/src/lib/libraries/extensions/index.js
index 33465d9..49bdabc 100644
--- a/src/lib/libraries/extensions/index.js
+++ b/src/lib/libraries/extensions/index.js
@@ -23,5 +23,12 @@ export default [
         iconURL: wedoImage,
         description: 'Build with motors and sensors.',
         featured: true
+    },
+    {
+        name: 'KOOV',
+        extensionURL: 'koov',
+        iconURL: 'https://www.koov.io/favicon.ico',
+        description: 'Play with your KOOV.',
+        featured: true
     }
 ];
```
- run scratch-gui server and connect using Google Chrome.  Then, load this extension and play with it.

### Reason for Changes

So that people can program on scratch3 to control KOOV.

### Test Coverage

At least existing test passes.

PR #1302, opened 2018-07-04
Add an extension to support KOOV

enami:
This PR is an update of #783.

### Resolves

Currently, scratch3 doesn't have an ability to control KOOV (an educational kit to build blocks with some sensors and actuators).  This PR adds an extension to add scratch3 blocks for sensors and actuators comes with KOOV.  Visit following page for more detail on KOOV (https://www.koov.io)

### Proposed Changes

Implements some scratch3 blocks to interact with sensors and actuators.

To test this extension on Mac, you will need to
- connect KOOV to Mac via USB.
- run this helper [server](https://github.com/sonyged/koov_scratch_server) on your Mac.  It will listen on 127.0.0.1:3030 for new connection.
- to test without mit.edu cert, modify src/io/deviceManager.js to use http instead of https to access device manager.
```
diff --git a/src/io/deviceManager.js b/src/io/deviceManager.js
index d2d0046..bb32961 100644
--- a/src/io/deviceManager.js
+++ b/src/io/deviceManager.js
@@ -257,7 +257,7 @@ class DeviceManager {
      * @return {string} - The default Scratch Device Manager connection URL.
      */
     static get DEFAULT_SERVER_URL () {
-        return 'https://device-manager.scratch.mit.edu:3030';
+        return 'http://device-manager.scratch.mit.edu:3030';
     }
 
     constructor () {
```
- modify scratch-gui to include KOOV entry.  make sure to enable the extension to test.
https://github.com/LLK/scratch-gui/compare/develop...sonyged:add-initial-support-for-koov

- build scratch-gui with modified scratch-vm and run scratch-gui server and connect using Google Chrome.  Then, load this extension and play with it.

### Reason for Changes

So that people can program on scratch3 to control KOOV.

### Test Coverage

At least existing test passes.


thisandagain:
Thanks for the submission! This looks good overall, but we'll want you folks to use "Scratch Link" as the helper application and there are a few others details we should discuss via email. I'm going to close this for now so we can discuss. Thanks.

PR #1303, opened 2018-07-05
Add implementation for "When loudness >" block

### LINKED ISSUE #338: Finish when [sensor] > _ hat block
Moving this from a TODO comment. Currently, the "when [sensor] > _" hat block has the timer implemented, but it will need to be extended once we have support for video motion and the microphone input.

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/338

### Proposed Changes

adds an implementation for the `loudness` menu option for the `when > ` block

### Test Coverage

Added a unit test for this hat.

### DIFF
--- a/src/blocks/scratch3_event.js
+++ b/src/blocks/scratch3_event.js
@@ -69,9 +69,11 @@ class Scratch3EventBlocks {
     hatGreaterThanPredicate (args, util) {
         const option = Cast.toString(args.WHENGREATERTHANMENU).toLowerCase();
         const value = Cast.toNumber(args.VALUE);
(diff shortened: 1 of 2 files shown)


paulkaplan:
@rschamp  I could move the getLoudness stuff into an ioQuery if you want, the other place it is used is in the sensing block where there is some additional caching logic because it can be used a zillion times per frame there, but because this is a hat, the caching isn't needed. Let me know if you think that is reqf it should be filed as a follow-up, or what.

/cc @ericrosenbaum

kchadha (approved):
LGTM

PR #1315, opened 2018-07-10
Break out loadVector and loadBitmap from loadCostumeAsset, and handle bitmap resolution 1 assets in loadBitmap

fsih:
Work toward https://github.com/LLK/scratch-svg-renderer/issues/36


src/import/load-costume.js
```diff
@@ -1,69 +1,39 @@
...
-    costume.assetId = costumeAsset.assetId;
-    const renderer = runtime.renderer;
-    if (!renderer) {
-        log.error('No rendering module present; cannot load costume: ', costume.name);
-        return costume;
+const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, optVersion) {
+    let svgString = costumeAsset.decodeText();
+    // SVG Renderer load fixes "quirks" associated with Scratch 2 projects
+    if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
+        log.error('No V2 SVG adapter present; SVGs may not render correctly.');
```
kchadha, line 8: Should this be a warning instead?

src/import/load-costume.js
```diff
@@ -82,21 +52,85 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime, optVersio
...
+            log.error('No V2 bitmap adapter present; bitmaps may not render correctly.');
+        } else if (costume.bitmapResolution === 1) {
+            src = runtime.v2BitmapAdapter.convertResolution1Bitmap(src, dataURI => {
+                if (dataURI) {
+                    // Put back into storage
+                    const storage = runtime.storage;
+                    costume.assetId = storage.builtinHelper.cache(
+                        storage.AssetType.ImageBitmap,
+                        storage.DataFormat.PNG,
+                        runtime.v2BitmapAdapter.convertDataURIToBinary(dataURI)
```
kchadha, line 69: Any reason why convertResolution1Bitmap doesn't call convertDataURIToBinary itself?

src/import/load-costume.js
```diff
@@ -82,21 +52,85 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime, optVersio
...
+                    );
+                    costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
+                    if (rotationCenter) {
+                        rotationCenter[0] = rotationCenter[0] * 2;
+                        rotationCenter[1] = rotationCenter[1] * 2;
+                        costume.rotationCenterX = rotationCenter[0];
+                        costume.rotationCenterY = rotationCenter[1];
+                    }
+                    costume.bitmapResolution = 2;
```
kchadha: So if an error occurs during the conversion process, we don't end up updating the bitmapResolution, is that any cause for concern?

src/import/load-costume.js
```diff
@@ -82,21 +52,85 @@ const loadCostumeFromAsset = function (costume, costumeAsset, runtime, optVersio
...
+        } else {
+            imageElement.src = src;
+        }
     }).then(imageElement => {
         // createBitmapSkin does the right thing if costume.bitmapResolution or rotationCenter are undefined...
-        costume.skinId = renderer.createBitmapSkin(imageElement, costume.bitmapResolution, rotationCenter);
-        costume.size = renderer.getSkinSize(costume.skinId);
+        costume.skinId = runtime.renderer.createBitmapSkin(imageElement, costume.bitmapResolution, rotationCenter);
+        const renderSize = runtime.renderer.getSkinSize(costume.skinId);
+        costume.size = [renderSize[0] * 2, renderSize[1] * 2]; // Actual size, since all bitmaps are resolution 2
```
kchadha, line 93: 😂😅

kchadha (changes requested):
Some changes requested:
1) change callback for `convertResolution1Bitmap` function to error first style (see comment in scratch-svg-renderer PR), 
2) It looks like you have a merge conflict with the .npmignore file
and 3)some minor comments/questions.

In general, I think these changes look good, though as with all things involving bitmap resolution, they are a little scary. 

Thank you for making this refactor!! Also, thanks for breaking up the SVG and bitmap loading functionality! It helps to makes things clearer.

.npmignore
```diff
@@ -6,8 +6,6 @@
 /.github
 /.travis.yml
 /.tx
-/test
-/webpack.config.js
```
kchadha: It looks like this change partially already exists on develop now. I guess removing /tests wasn't necessary for this repo?

src/import/load-costume.js
fsih: It's similar to "No rendering module present", so I made it also an error. gnarf requested that it throw an error, since Z was hitting an issue he was having a very hard time fixing that was caused by this being missing.

src/import/load-costume.js
fsih: Here the VM needs both the dataURI (to set as the image src) and the Uint8Array (to put back in storage) so I think it makes sense to return the dataURI and let VM convert it to a Uint8Array when it wants.

src/import/load-costume.js
fsih: If there's an error, then we use the original data we are given, which has not been doubled in size, so the costume will be tiny.
Does it make sense to set it to bitmap resolution = 2 and accept that in the error case, the costume will be half the size that it should have been? Fwiw if it errors, it's likely the image is unrenderable anyway. So I think you're right that we should set the bitmap resolution and rotation center as if it's bitmap resolution 2, if it errors.

kchadha (approved):
LGTM

PR #1321, opened 2018-07-11
Issue/1299 switch language

### LINKED ISSUE #1299: Error in extension blocks after switching languages
### Expected Behavior

Extensions should work after switching language. It may just be the pen extension, but I suspect it hits all of them. Is it related to the fact that the VM has changed the block definitions internally, but the toolbox is still showing the previous language (due to recycling)?

### Actual Behavior

`vm Could not get implementation for opcode: pen_stamp`

### Steps to Reproduce

* load the editor in another language: https://preview.scratch.mit.edu/?enable=language&locale=ja#112439198
* Click the green flag, and start the game. Click arrows at the top to see tiles stamp when they reach the bottom. 
* switch to another language
* click arrows - game no longer works - and the message above is in the console.

Mac OS 10.13.5 Chrome

chrisgarrity:
### Resolves

Resolves #1299 

### Proposed Changes

Converts setLocale to return a promise when all the blocks have updated. Then GUI can refresh the toolbox and workspace with the extension blocks in a new language. Also calls `_prepareExtensionInfo` for each extension loaded to make sure all the block functions are correctly initialized.

### Reason for Changes

Allows extensions to keep working after switching language and renders the blocks in the new language.

### Testing

- [ ] Extension blocks (pen, music etc) in the toolbox render in a new language after switching language
- [ ] Extension blocks in the workspace render in a new language after switching language
- [ ] Extension blocks continue to work after switching language (e.g., pen `stamp` still stamps)

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -144,18 +144,21 @@ class ExtensionManager {
     }
 
     /**
-    * regenerate blockinfo for any loaded extensions
-    */
+     * Regenerate blockinfo for any loaded extensions
+     * @returns {Promise} resolved once all the extensions have been reinitialized
+     */
     refreshBlocks () {
-        this._loadedExtensions.forEach(serviceName => {
+        const allPromises = Array.from(this._loadedExtensions.values()).map(serviceName =>
             dispatch.call(serviceName, 'getInfo')
                 .then(info => {
+                    info = this._prepareExtensionInfo(serviceName, info);
                     dispatch.call('runtime', '_refreshExtensionPrimitives', info);
                 })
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
Looks good to me!

Sooner or later I think it would make sense to refactor `_prepareExtensionInfo` into something like `_getExtensionInfo` which would return a promise for the "prepared" extension info by doing the `dispatch.call` to `getInfo` and putting what `_prepareExtensionInfo` currently does into a `then` on the `getInfo` call. Something like that would make it more difficult to forget to "cook" the extension's info before trying to use it. I don't think that's a priority right now, though.

PR #1326, opened 2018-07-11
Fix variable conflicts that arise when sharing the love with the stage.

### LINKED ISSUE #1304: What should happen when a local variable is shared with the stage?
Users can share blocks between sprites or between sprites and the stage. If the set of blocks being shared from a sprite to a stage contain a reference to a local variable on the sprite, what should happen? Below are some possible options discussing the response to the following scenario.

Scenario:
1. Create a project with a sprite `Sprite1`
2. Create a local variable `foo` on `Sprite1`
3. Drag a `change [foo] by 1` block onto `Sprite1`'s workspace.
4. Create a new sprite `Sprite2`
5. Create a local variable `foo` on `Sprite2`
6. Drag a `change [foo] by 1` block onto `Sprite2`'s workspace.
7. Share the `change [foo] by 1` block from `Sprite1`'s workspace to the stage.
8. Share the `change [foo] by 1` block from `Sprite2`'s workspace to the stage.
9. Repeat steps 7 and 8.

#### Option # 1 - Local variable gets renamed on the stage  (creating a new global variable with a name that does not conflict with any local variables); local vars of the same name from the same sprites get the same name, local vars of the same name from a different sprite get a different name.

In this option, two global variables are created. One is created by step 7 and one is created by step 8. These global variables have new names: `Stage: foo` and `Stage: foo2` respectively. `Sprite1` and `Sprite2` still have their own local variables named `foo`.

Step 9 does not create any more new variables, instead we have one more `change [...] by 1` block for each of the new variables.

#### Option # 2 - Local variable gets renamed on the stage (creating a new global variable with a name that does not conflict with any local variables); all local variables of the same name (regardless of the sprite sharing them) get mapped to the same new global variable.

In this option, one new global variable is created: `Stage: foo`. After step 9 in the scenario, the stage has four `change [Stage: foo] by 1` blocks. `Sprite1` and `Sprite2` still have their own local variables named `foo`.

#### Option # 4 - Local variables are converted to global variables.

In this option, step 7 of the scenario converts the `foo` variable on both `Sprite1` and `Sprite2` to a global variable. The global variable's value is set to 0. No local variables named `foo` exist anymore and any blocks referencing them are converted to use the new global variable.

#### Additional Notes on the Options Above
- Option # 1 has already been implemented.
- Option # 2 would require a runtime map of original local variable name to new global variable name.

kchadha:
### Resolves

Resolves #1304 (with option # 1 discussed in that issue).

### Proposed Changes

Option # 1 in #1304.

### Reason for Changes

Resolving variable conflicts created by the 'share the love' feature and by adding code from the backpack.

### Test Coverage

Added unit tests for the new variable conflict resolution functionality.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -693,12 +693,15 @@ class Blocks {
     /**
      * Returns a map of all references to variables or lists from blocks
      * in this block container.
+     * @param {Array<object>} optBlocks Optional list of blocks to constrain the search to.
+     * This is useful for getting variable/list references for a stack of blocks instead
+     * of all blocks on the workspace
      * @return {object} A map of variable ID to a list of all variable references
      * for that ID. A variable reference contains the field referencing that variable
      * and also the type of the variable being referenced.
      */
-    getAllVariableAndListReferences () {
-        const blocks = this._blocks;
+    getAllVariableAndListReferences (optBlocks) {
+        const blocks = optBlocks ? optBlocks : this._blocks;
         const allReferences = Object.create(null);
         for (const blockId in blocks) {
             let varOrListField = null;
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -13,6 +13,7 @@ const Thread = require('./thread');
 const log = require('../util/log');
 const maybeFormatMessage = require('../util/maybe-format-message');
 const StageLayering = require('./stage-layering');
+const Variable = require('./variable');
 
 // Virtual I/O devices.
 const Clock = require('../io/clock');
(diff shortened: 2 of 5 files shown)


paulkaplan (changes requested):
Functionality-wise, i think this is working fine. I think we should do a bit of refactoring/moving things around though, for clarity.

First I think we should move this entire function into a target function (could also be called `shareBlocksToTarget`). That would at least eliminate the currTarget/target confusion (because `this` would be more clearly the `from target`.

Second maybe we could split up the blocks inside the if/else into two separate helper functions. That would simplify the function, and make it more clear what the two major situations are, and allow those to be documented.

Does that sound good?

paulkaplan (approved):

PR #1327, opened 2018-07-11
Add opcode implementation for `data_deletealloflist`

thisandagain:
### Blocked by
https://github.com/LLK/scratch-blocks/pull/1640

### Proposed Changes
Adds "delete all of [list]" block as per discussion.

### Reason for Changes
Maintaining and further exposing useful functionality from Scratch 2.0

### Reference
load


kchadha (approved):
LG!

griffpatch:
Will this auto translate from existing Scratch 2 projects that make use of
the existing <delete <all> of <list>> block?

thisandagain:
@griffpatch No. We decided to keep the spec map stable for existing SB2 projects (which map to the existing `delete (all) of [list]` block) and treat this as an addition.

PR #883, opened 2018-01-02
midiInstrument: support

apple502j:
### Resolves
#861 
Projects with old 1.4 "midiInstrument:" block fail to open
### Proposed Changes
Supporting midiInstrument: as Scratch 2.0 supporting.
src/serialization/sb2_specmap.js added 
```
'midiInstrument:':{
+        opcode: 'music.setInstrument',
+        argMap: [
+            {
+                type: 'input',
+                inputOp: 'math_number',
+                inputName: 'INSTRUMENT'
+            }
+        ]
+    },
```
### Reason for Changes
Old one cannot read 1.4 projects with this block.
### Test Coverage
Read 2.0 project uses it and was uploaded by 1.4 like
https://llk.github.io/scratch-gui/develop/#138256836

PR #1329, opened 2018-07-12
Add support for "midiInstrument:" block

### LINKED ISSUE #861: Projects with old 1.4 "midiInstrument:" block fail to open
### Expected Behavior

Project opens and editor loads. (Behavior for `midiInstrument:` is not 100% certain yet - see LLK/scratch-audio#13.)

### Actual Behavior

Editor crashes, with this error in the console:

```
vm warn Couldn't find SB2 block:  midiInstrument:
```

I recall projects with this block opening at some point in the past? That might be my bad memory, though.

### Steps to Reproduce

Open a project which uses the `midiInstrument:` 1.4 block, such as https://scratch.mit.edu/projects/138256836/ ([3.0](https://llk.github.io/scratch-gui/develop/#138256836)).

Firefox 58.0, Debian (Testing).

towerofnix:
### Resolves

Resolves #861. Closes #883 (since that PR also resolves the issue - but it doesn't use the MIDI mapping, so it's not a solution that makes the block's behavior compatible with 2.0).

### Proposed Changes

Adds support for the old Scratch block "midiInstrument:". This block behaves basically the same as the ordinary "set instrument to" block, but refers to a (hard-coded) mapping of MIDI instrument numbers (from Scratch 1.4 and earlier) to Scratch 2.0+ instrument numbers.

Support exists by way of adding a new block to the Music extension (and keeping it hidden from the block palette, since it's an obsolete/compatibility block). Since `setInstrument` and `midiSetInstrument` behave so similarly, I created a helper function `_setInstrument`. 2.0 just uses a single `primSetInstrument` function, and [directly checks the block's opcode](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/primitives/SoundPrims.as#L162-L165) to decide whether or not to use the MIDI mapping, but I didn't see any way to do something exactly like that in scratch-vm.

### Reason for Changes

Compatibility with old Scratch projects. The mapping is taken from [here, in scratch-flash](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/primitives/SoundPrims.as#L191-L257).

### Test Coverage

Manually tested with [this project](https://scratch.mit.edu/projects/138256836/).

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -778,7 +778,7 @@ class Scratch3MusicBlocks {
                     text: formatMessage({
                         id: 'music.midiSetInstrument',
                         default: 'set instrument to [INSTRUMENT]',
-                        description: 'set the instrument for notes played according to a mapping of MIDI codes to Scratch instrument numbers'
+                        description: 'set the instrument for notes played according to a mapping of MIDI codes'
                     }),
                     arguments: {
                         INSTRUMENT: {
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -785,7 +785,8 @@ class Scratch3MusicBlocks {
                             type: ArgumentType.NUMBER,
                             defaultValue: 1
                         }
-                    }
+                    },
+                    hideFromPalette: true
                 },
                 {
                     opcode: 'setTempo',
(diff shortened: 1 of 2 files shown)


ericrosenbaum (approved):
This looks and works great! Thanks for your help @towerofnix- Sorry it it took us a while to get to this one.

PR #1331, opened 2018-07-12
Microbit display commands yield for a tick

### LINKED ISSUE #1290: MicroBit "write" blocks need to resolve on a timeout?
Right now they resolve on a Promise, and the execution model seems unclear.
Need to correctly handle timing of command blocks that update the LED display.

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1290

### Proposed Changes

Each opcode that updates the microbit display returns a promise that immediately resolves. This has the effect of yielding the stack for a tick.

### Reason for Changes

This rate-limits commands sent to the microbit to 30 per second.

You can now make a stack of animation frames with the new matrix field block, and it will play at 30fps.


paulkaplan (commented):
LGTM

PR #1332, opened 2018-07-12
Microbit write bails out if not connected

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1310

### Proposed Changes

If no device is connected, bail out before attempting to send data to the microbit.

### Reason for Changes

This prevents errors that otherwise occur if you run any of the microbit display blocks while a microbit is not connected.


paulkaplan (approved):
LG

PR #1333, opened 2018-07-12
Change extension name to micro:bit

ericrosenbaum:
This matches the official micro:bit branding.


paulkaplan:
@ericrosenbaum This needs to be updated in the GUI extension library too right?

PR #1335, opened 2018-07-12
fix asynchronous block execution

### LINKED ISSUE #1328: Translate block in a say block shows "undefined"
### Expected Behavior

The translate block should not result in "undefined"

### Actual Behavior

If you put the translate block inside a say block, if the translation result is not cached, the result arrives after the say block runs, so the say bubble says "undefined":

translate-bug

Also it's pretty weird that after the block runs, the bubble comes out of the block with the correct translation!

mzgoddard:
### Resolves

#1328

### Proposed Changes

Freeze promise waiting blocks into an array of reported values. Thaw those reported values into parent values of matching blocks by id. If the reporting block was waiting on a promise and reported a value assign it to its parent value at that future time.

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -102,7 +102,7 @@ const handleReport = function (resolvedValue, sequencer, thread, blockCached) {
     }
 };
 
-const handlePromise = (primitiveReportedValue, sequencer, thread, blockCached) => {
+const handlePromise = (primitiveReportedValue, sequencer, thread, blockCached, lastOperation) => {
     if (thread.status === Thread.STATUS_RUNNING) {
         // Primitive returned a promise; automatically yield thread.
         thread.status = Thread.STATUS_PROMISE_WAIT;
@@ -110,7 +110,8 @@ const handlePromise = (primitiveReportedValue, sequencer, thread, blockCached) =
     // Promise handlers
     primitiveReportedValue.then(resolvedValue => {
         handleReport(resolvedValue, sequencer, thread, blockCached);
-        if (typeof resolvedValue === 'undefined') {
+        // If its a command block.
+        if (lastOperation && typeof resolvedValue === 'undefined') {
             let stackFrame;
             let nextBlockId;
             do {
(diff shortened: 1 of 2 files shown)


src/engine/execute.js
```diff
@@ -407,7 +406,44 @@ const execute = function (sequencer, thread) {
...
+            const lastExisting = reported.reverse().find(report => ops.find(op => op.id === report.opCached.id));
+            if (lastExisting) {
+                i = ops.findIndex(opCached => opCached.id === lastExisting.opCached.id) + 1;
+            } else {
+                i = 0;
+            }
+        }
+
+        // The reporting block must exist and must be the next one in the sequence of operations.
+        if (thread.justReported !== null && ops[i] && ops[i].id === currentStackFrame.reporting.id) {
```
mzgoddard: Handling was missing to assign the `justReported` value.

src/engine/execute.js
```diff
@@ -131,8 +132,6 @@ const handlePromise = (primitiveReportedValue, sequencer, thread, blockCached) =
             } while (stackFrame !== null && !stackFrame.isLoop);
 
             thread.pushStack(nextBlockId);
-        } else {
-            thread.popStack();
```
mzgoddard, line 135: No longer need to pop reporter blocks.

src/engine/execute.js
mzgoddard: Changed `.reporting` to be the id itself.

paulkaplan (commented):
This fixes the issue where the block was executing with `undefined`, but there is still the issue that clicking a command block with a promise reporter in it will report the value visually, which it shouldn't. This seems to be due to line 83 in `handleReport`, which seems to assume that if there is a reported value and the thread is at the top of the stack, a visual report should be issued. Those assumptions do not seem to be correct for promise reporters inside a single command block that is stack clicked.

~@thisandagain because this fixes the "fatal" issue of the block not functioning correctly, i feel like it is better to move forward with this instead of backward. What do you think. Or even better, @mzgoddard it may be a simple issue to fix the visual report, what do you think?~

Discussed offline, we can move ahead with this without waiting on the reported value issue. I'll file a follow-up for that.

PR #1341, opened 2018-07-13
fix unnecessary stackClick report from nested promise reporters

mzgoddard:
### Resolves

https://github.com/LLK/scratch-vm/pull/1335#pullrequestreview-137015171

### Proposed Changes

Check that a reporting block is the last operation as part of some
parent block. The last operation is the block itself which may report
values for stack clicks and monitor updates if it meets other
requirements.

### Reason for Changes

Do not report values from promise operations that are not the top most block of a clicked stack.


paulkaplan (approved):
LG

PR #1342, opened 2018-07-13
EV3 port and sensor polling updates

evhan55:
### Proposed Changes

Adds more features to the EV3 extension:
- Can unplug and replug sensors and motors at any time
- Stop button stops all motors
- Added `beep (note) for (time) secs` block
- "this way" and "that way" for motor turn blocks
- Removes `motor rotate by` and `motor set position` blocks until motor coasting is resolved

### Test Coverage

Will be preliminarily tested at a Smoke Test on July 13, 2018


ericrosenbaum (approved):
LGTM!!

PR #1346, opened 2018-07-15
Use 2x png for microbit icon

ericrosenbaum:
There was a performance problem caused by the large SVG file we were loading as a dataURI in the microbit extension (reported by @seotts), where scrolling the blocks workspace became laggy with large stacks of microbit blocks.

This PR resolves that issue by:
- ~~Changing the webpack config to use file-loader for .png files~~
- ~~Adding a PNG file at 2x resolution (i.e. 80x80px, for the 40x40px block icon) for the microbit extension block icon and menu icon~~
- Replacing the SVG data uri with a 2x resolution PNG data uri


ericrosenbaum:
@rschamp can you please have a look and help understand why this is failing on travis?

thisandagain:
@ericrosenbaum Looks like this is failing on a syntax error while trying to load the PNG. A smaller / less invasive fix may be to simply encode the PNG as a datauri (as we have now with the SVG) and omit the use of `file-loader`.

thisandagain (approved):

PR #1347, opened 2018-07-15
Fix local var rename/delete

### LINKED ISSUE #1343: Local variables don't stay renamed or deleted
### Expected Behavior

If a variable is renamed or deleted, that change should persist across all user interactions.

### Actual Behavior

Renaming/deletion of a *local* variable is undone after switching sprites.

kchadha:
### Resolves

Resolves #1343

### Proposed Changes

Fixes the issue where local variable renaming or deletion was getting undone after switching sprites. Also fixes the issue where the monitor name wasn't getting updated after renaming a local variable.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -354,7 +354,7 @@ class Blocks {
             }
             break;
         case 'var_rename':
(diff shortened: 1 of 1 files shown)


paulkaplan (approved):
LG!

PR #1355, opened 2018-07-17
Ignore display command with blank string

khanning:
### Resolves

If you run the `display [Hello!]` block with a blank string the micro:bit will sometimes lock up requiring a hard reset. 

To replicate, run the following stack a couple of times and the micro:bit will become unresponsive.

screen shot 2018-07-17 at 11 11 10 am

### Proposed Changes

Only send the command to the micro:bit if the input string length is greater than 0.

### Reason for Changes

Fixes micro:bit device lockup.


ericrosenbaum (approved):
LGTM!

PR #1356, opened 2018-07-17
Fix microbit display matrix clear function

### LINKED ISSUE #1344: micro:bit display matrix block doesn't work when empty
### Expected Behavior

The display matrix block should clear the micro:bit LED matrix.

screen shot 2018-07-13 at 2 44 15 pm

### Actual Behavior

Nothing happens.

### Steps to Reproduce

1. Connect to micro:bit
2. Drag out `display [♡]` and clear the matrix.
3. Run the block, nothing happens.

all

khanning:
### Resolves

Resolves #1344 

### Proposed Changes

Only send matrix data to micro:bit if input does not equal `null`.

### Reason for Changes

`if (!hex)` returns if the input data is '0' which is valid input data. Passing '0' to the matrix field should clear the screen.

This also fixes the clear button not working.

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -509,13 +509,14 @@ class Scratch3MicroBitBlocks {
             return value;
         };
         const hex = symbol.split('').reduce(reducer, 0);
-        if (!hex) return;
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM!

PR #1361, opened 2018-07-17
Fix sprite upload variable conflict

kchadha:
### Resolves

Fix variable conflict that arises during sprite upload where the sprite may reference a global variable that conflicts with the name of a local variable on a different sprite in the project.

### Proposed Changes

If uploading a new sprite `Sprite2` that has a block which references a variable `foo` that does not exist in `Sprite2` (e.g. `foo` was a variable that was a global variable in the project `Sprite2` was exported from), check for variable conflicts with local variables on other sprites in the project before creating a new global variable `foo`. If the project we're working in had a `Sprite1` with a local variable `foo`, then uploading `Sprite2` should create a new global variable `foo2`. 

### Reason for Changes

Variable conflicts are bad and lead to crashing/block disappearing errors.

### Test Coverage

Added a unit test specific to this case.

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -535,6 +535,7 @@ class Target extends EventEmitter {
      * then the local variable is renamed to distinguish itself from the pre-existing variable.
      * All blocks that reference the local variable will be updated to use the new name.
      */
+    // TODO (#1360) This function is too long, add some helpers for the different chunks and cases...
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):
Thanks for including the todo issue :) LGTM

PR #1367, opened 2018-07-18
Add function to return the current VM/blocks locale

chrisgarrity:
### Resolves

Needed for https://github.com/LLK/scratch-gui/issues/2530

### Proposed Changes

Adds a function that returns the current VM locale

### Reason for Changes

GUI shouldn't update the toolbox locale when the blocks aren't visible. If the locale switches while the blocks are hidden (e.g. in sound editor), the gui needs a way to know to call setLocale when the visibility changes. This function is so that GUI can tell when the VM locale doesn't match the current locale, and call vm.setLocale to update the blocks.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -928,6 +928,14 @@ class VirtualMachine extends EventEmitter {
     }
 
     /**
+     * get the current locale for the VM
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LG!

PR #1368, opened 2018-07-18
Change 'video motion' to 'video sensing'. See LLK/scratch-gui#2621

thisandagain:
### Resolves
Follow-up for LLK/scratch-gui#2621

### Proposed Changes
Change "Video Motion" to "Video Sensing" in the extension definition


kchadha (approved):
LG

PR #1377, opened 2018-07-18
Update the category info name when the extension is refreshed.

paulkaplan:
Fixes an issue where changing languages did not update the name in the category menu.

Fixes https://github.com/LLK/scratch-gui/issues/2600


chrisgarrity (approved):
Thanks for looking at this! Looks good.

PR #1378, opened 2018-07-18
Use the newer date format for vm versions, to get greenkeeper working

paulkaplan:
Fixes an issue where the last version in the npm version list was not the most recent version. Hopefully this fixes greenkeeper. It also brings the version tag more in line with the audio/rest of the repos


kchadha (approved):
LG

PR #1381, opened 2018-07-19
Do not send coast if another motor command has been sent

paulkaplan:
### Resolves

Fixes an issue where sending another second motor command while another is running causes the second command to get interrupted by the first commands "coast" command.

Diagrammatically:
- 0s: Motor command (1) is sent to rotate motor for 5 seconds
- 2s: Motor command (2) is sent to rotate motor for 10 seconds
- 5s: Motor command (1) finishes, and the "coast" command is sent, interrupting motor command 2
- 12s: Motor command (2) should still be running, but it isn't. Sends a coast command anyway.

This changes that sequence to this:
- 0s: Motor command (1) is sent to rotate motor for 5 seconds, sets motor command id to X
- 2s: Motor command (2) is sent to rotate motor for 10 seconds, sets motor command id to Y
- 5s: Motor command (1) finishes, checks to see if motor command id is X. It isn't (it is now Y), so it does not send a coast command.
- 12s: Motor command (2) finishes running, checks if the current motor command is Y (it is), so it sends coast. 

### Proposed Changes

Store a map of command IDs, and refactor the coast timeout logic into a `coastAfter(port, time)` function that sets the command ID and creates the timeout for checking if the coast command should be sent. 

### Reason for Changes

Sending rotate for time commands in moderately quick succession leads to the motors moving for short and seemingly random amounts of time, because the "coast"  from out-of-date commands starts interrupting the current commands.

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -342,7 +342,7 @@ class EV3 {
         });
     }
 
-    coastAfter(port, time) {
+    coastAfter (port, time) {
         // Set the motor command id to check before starting coast
         const commandId = uid();
         this._motors.commandId[port] = commandId;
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
Works great! LG pending lint fix

Issue #1366, opened 2018-07-18
Turning micro bit power off doesn't disconnect microbit


BryceLTaylor:
### Expected Behavior

If you remove power from the micro bit, Scratch should disconnect from it.

### Actual Behavior

It does not disconnect from the microbit and still shows that they are connected.  When you disconnect from the microbit (that is unplugged) in Scratch it throws a console error:

`vm BLESession error: "ws onclose"`

### Steps to Reproduce

Add the micro bit extension to a project
Plug in and connect a micro bit
unplug the microbit
Note: there is no disconnection error of any kind and the microbit looks like it is still connected
Click the connection status dot
Note: it still says the micro bit is connected
Disconnect the microbit
Note: An error appears in the console

### Operating System and Browser

Chrome on Windows

BryceLTaylor:
It seems you get the `vm BLESession error: "ws onclose"` error every time you disconnect the microbit, not only when you turn it off first.

ericrosenbaum:
Fixed by https://github.com/LLK/scratch-vm/pull/1383

Issue #1363, opened 2018-07-17
EV3 motor block causes distance to return 0


BryceLTaylor:
### Expected Behavior

The distance block should not be related to the motor block.

### Actual Behavior

When you run the rotate motor block the distance briefly becomes 0.  This will trigger the `distance < ` block and affects the reporter.

### Steps to Reproduce

Add the EV3 Extension and connect an EV3
Connect a motor and a distance sensor to the EV3 hub
Build a stack that says Forever Say Distance
Run the stack
Click the rotate motor block
Note: the distance drops to 0 briefly

### Operating System and Browser

Edge on Windows and Chrome on Mac

thisandagain:
This causes a lot of confusion if you have the "when distance < [n]" hat block and a motor block in the same project.

evhan55:
Could no longer reproduce, so closing because it might be resolved.

PR #1383, opened 2018-07-20
Feature/critical extension fixes

evhan55:
### Resolves

[Turning micro bit power off doesn't disconnect microbit #1366](https://github.com/LLK/scratch-vm/issues/1366)
[EV3 motor block causes distance to return 0 #1363](https://github.com/LLK/scratch-vm/issues/1363)

### Notes

#1366: This is fixed with a 4.5 second delay.  I tried to make the delay shorter, but then it auto-disconnects too early sometimes because the microbit sends the first 2-3 packets in a delay before they are sent quickly.  If the 4.5 second delay feels like too long, I can try to add a special case to not start the timeout after the first few packets are sent back?

#1363: This is fixed with a minor hack based on packet length which seems to be the only way to differentiate EV3 responses.


ericrosenbaum (approved):
LGTM!

PR #1398, opened 2018-07-31
Publish jsdoc to gh-pages

paulkaplan:
Using the jsdoc config copied from the renderer, I made it so the vm publishes the jsdoc to `playground/docs`

I needed to fix several jsdoc issues that were (suspiciously) not caught by the linter. I don't like this at all, and makes me concerned. There is also really limited documentation about jsdoc syntax, and some things that were in the codebase simply do not exist (like documenting the types of a tuple, like `[Function, Function]` @kchadha) and the fact that type declarations cannot take description strings.

@rschamp I worry a lot about the in-editor linters not correctly catching all jsdoc syntax. It could be an issue of some versions not being correct. 

Take a look at all the changes in d66167e to get a sense of what the linter allows that kills the jsdoc builder


kchadha (approved):
@paulkaplan, these changes look good to me. Where should we look for JSDoc style/syntax ground truth?

kchadha (dismissed):
Actually, I guess there's a merge conflict w/the package-lock.json... not sure how we go about resolving those..

rschamp:
For resolving conflicts in `package-lock.json`: https://docs.npmjs.com/files/package-locks#resolving-lockfile-conflicts

rschamp:
Re: `jsdoc` ground truth, the syntax is described on http://usejsdoc.org/ and the types it uses are generally explained on http://usejsdoc.org/tags-type.html, but note it also accepts types defined in https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#type-expressions

rschamp:
@paulkaplan Did you look at if https://github.com/gajus/eslint-plugin-jsdoc works any differently for us than eslint's built-in `valid-jsdoc`?

paulkaplan:
@rschamp I checked out that plugin, and if we only enable the rule `jsdoc/valid-types`, that will catch the errors that I fixed manually in this PR. Should I just file an issue on eslint-config-scratch to add that rule?

rschamp (approved):
@paulkaplan Sounds good to me re: using the jsdoc plugin for validating types only.

PR #1399, opened 2018-07-31
Add an extra param for adding sounds and costumes to specific targets

paulkaplan:
Add an extra param for `addSound` and `addCostume` so that they can be used on specific targets, not just the editing target. This is required to implement adding costumes/sounds from the backpack to sprites/stage.


kchadha (approved):
LG -- maybe also add tests where the functions are used w/the new param?

PR #1409, opened 2018-07-31
Change variable toXML function to a static function that acts on vari…

### LINKED ISSUE #895: Duplicating/Cloning a sprite w/sprite-specific variables throws (non-fatal) type error.
### Expected Behavior

Duplicating or cloning a sprite should copy over variables correctly.

### Actual Behavior

Variables aren't serialized correctly when cloning or duplicating a sprite so their toXML function cannot be accessed from the emitWorkspaceUpdate, causing the following TypeError:

```
raven.js:420 Uncaught TypeError: v.toXML is not a function
    at scratch-vm.js:17116
    at Array.map (<anonymous>)
    at VirtualMachine.emitWorkspaceUpdate (scratch-vm.js:17115)
    at VirtualMachine.setEditingTarget (scratch-vm.js:17031)
    at VirtualMachine.stopDrag (scratch-vm.js:17166)
    at Stage.onStopDrag (stage.jsx:250)
    at Stage.wrapper [as onStopDrag] (index.js:550)
    at Stage.onMouseUp (stage.jsx:182)
    at HTMLDocument.wrapper (index.js:550)
```
#### Additional Notes
`RenderedTarget.duplicate()` and `RenderedTarget.makeClone` both call `JSON.parse(JSON.stringify(this.variables))` which only copies the properties of the object and not the functions. A new Variable instance should be created for each variable in this.variables.

### Steps to Reproduce

Create a new Scratch 2.0 project.
Click on Sprite1, and create a new sprite-specific variable 'spriteVar'.
Load this project in 3.0. 
Duplicate Sprite1 and observe the error. There will be a new error every time you switch to the duplicated sprite.

kchadha:
### Resolves

Resolves #895

### Proposed Changes

Replace `Variable.prototype.toXML` with a static function that can act on both variable instances as well as flattened objects (via `JSON.parse(JSON.stringify...)`). These flattened objects were creating issues with serializing the blocks workspace to XML whenever a sprite with local variables was duplicated or cloned. These workspace issues can eventually lead to a workspace crash.

### DIFF
--- a/src/engine/variable.js
+++ b/src/engine/variable.js
@@ -33,9 +33,15 @@ class Variable {
         }
     }
 
-    toXML (isLocal) {
+    /**
+     * Serialize a variable object to XML.
+     * @param {Variable | object} variable The Variable or Variable-like object to serialize.
+     * @param {boolean=} isLocal Whether the variable being serialized is local or not.
+     * @return {!string} The serialized variable in XML format.
+     */
+    static toXML (variable, isLocal) {
         isLocal = (isLocal === true);
-        return `<variable type="${this.type}" id="${this.id}" islocal="${isLocal}">${this.name}</variable>`;
+        return `<variable type="${variable.type}" id="${variable.id}" islocal="${isLocal}">${variable.name}</variable>`;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


paulkaplan:
I'm kinda concerned about how there are things that look like Variable objects but aren't, that seems bound to bite us again in the future. Is that an issue that should be addressed?

paulkaplan:
For posterity, @kchadha and I discussed offline, decided the right solution is to change the way we are cloning variables so that we don't get into a situation where objects are losing their prototypes unexpectedly.

kchadha:
The revisions discussed above are addressed in #1434.

PR #1415, opened 2018-08-02
Import current block menu as field not input

### LINKED ISSUE #1414: Current block menu imports from sb2 as "year" no matter what
image

This is the project I'm importing

[current-menu.sb2.zip](https://github.com/LLK/scratch-vm/files/2254562/current-menu.sb2.zip)

Saw this being discussed here https://scratch.mit.edu/discuss/topic/299791/?page=270#post-3200539

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/1414

### Proposed Changes

Import current block menu as field not an input.

### Reason for Changes

Because the current block has a field, not an input

### Test Coverage

On develop 

image

With this PR

image

### DIFF
--- a/src/serialization/sb2_specmap.js
+++ b/src/serialization/sb2_specmap.js
@@ -1110,9 +1110,8 @@ const specMap = {
         opcode: 'sensing_current',
         argMap: [
             {
-                type: 'input',
-                inputOp: 'sensing_currentmenu',
-                inputName: 'CURRENTMENU'
+                type: 'field',
+                fieldName: 'CURRENTMENU'
             }
         ]
     },


src/serialization/sb2_specmap.js
```diff
@@ -1110,9 +1110,8 @@ const specMap = {
         opcode: 'sensing_current',
         argMap: [
             {
-                type: 'input',
-                inputOp: 'sensing_currentmenu',
```
kchadha, line 1114: So I could just be confused about this, but it seems weird that we're not making use of the `sensing_currentmenu` anymore. Is this just a piece of code where we originally thought `sensing_currentmenu` was going to be a dynamic menu but it isn't?

paulkaplan: I think so yeah. I didn't find any sensing_currentmenu in scratch-blocks.

kchadha (approved):
LG

PR #1416, opened 2018-08-02
Make click hats trigger on mouse down instead of up

### LINKED ISSUE #1249: “when this sprite clicked” does not trigger until after the mouse button has been released
### Expected Behavior
The "when this sprite clicked" hat block should trigger on mouse down (as per Scratch 2.0)

### Actual Behavior
The "when this sprite clicked" hat block triggers on mouse up

### Steps to Reproduce
image

### Credit
HT to multiple users in the forums (including Mr-Dave, RaefWorks_Animations, and videogames102) for noticing this and providing repro cases.

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/1249

### Proposed Changes

Make the click hats activate on mousedown instead of mouse up

### Reason for Changes

We already had all the needed info, we just were explicitly triggering on mouse up. Understandable, since that is the technical way the word "click" is used, but it turns out that click in scratch 2 specifically means "on mouse down", and projects depended on that behavior. 

I do not think this impacts any dragging behavior, which is handled totally separately. But I want to be cautious because I don't fully understand the comment from @thisandagain in the linked issue.

### Test Coverage

I added a pretty long unit test to make sure:
1. hats are triggered on the "leading edge" of the mouse up -> mousedown transition.
2. mouse ups do not trigger the hats.
3. hats are triggered only for clicks within the canvas window, which was implemented previously but was not tested until this.

### DIFF
--- a/src/io/mouse.js
+++ b/src/io/mouse.js
@@ -14,40 +14,42 @@ class Mouse {
     }
 
     /**
-     * Activate "event_whenthisspriteclicked" hats if needed.
-     * @param  {number} x X position to be sent to the renderer.
-     * @param  {number} y Y position to be sent to the renderer.
-     * @param  {?bool} wasDragged Whether the click event was the result of
-     * a drag end.
+     * Activate "event_whenthisspriteclicked" hats.
+     * @param  {Target} target to trigger hats on.
      * @private
      */
-    _activateClickHats (x, y) {
+    _activateClickHats (target) {
+        // Activate both "this sprite clicked" and "stage clicked"
+        // They were separated into two opcodes for labeling,
+        // but should act the same way.
+        // Intentionally not checking isStage to make it work when sharing blocks.
+        // @todo the blocks should be converted from one to another when shared
+        this.runtime.startHats('event_whenthisspriteclicked',
+            null, target);
+        this.runtime.startHats('event_whenstageclicked',
+            null, target);
+    }
+
+    /**
+     * Find a target by XY location
+     * @param  {number} x X position to be sent to the renderer.
+     * @param  {number} y Y position to be sent to the renderer.
+     * @return {Target} the target at that location
+     * @private
+     */
+    _pickTarget (x, y) {
         if (this.runtime.renderer) {
             const drawableID = this.runtime.renderer.pick(x, y);
             for (let i = 0; i < this.runtime.targets.length; i++) {
                 const target = this.runtime.targets[i];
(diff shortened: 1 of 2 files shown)


kchadha:
This introduces an incompatibility regression w/Scratch 2.0 behavior. `When this sprite clicked` hats should not trigger if the sprite is being dragged. This is a good test project (dragging the cat around should not change the color of the cat):

image

kchadha:
Marking this as - needs work - because after discussions on Friday, we realized that Scratch 2.0 behavior around this is complicated and the functionality of `when this sprite clicked` is triggered is different depending on 1) the drag mode of the sprite and 2) editor mode vs. player/fullscreen mode.

kchadha:
Closing this for now. I will revisit this PR when I work on the corresponding issue #1249.

paulkaplan:
@kchadha after some investigation, we realized that "when sprite clicked" acts differently for draggable vs. non-draggable sprites. Draggable sprites get "clicked" on a non-drag mouse-up event. Non-draggable sprites get "clicked" on mouse down, hence the "when clicked, repeat until not mouse down" construction, which works for non-draggable sprites, but does nothing for draggable sprites.

This project is a good test to compare 2/3
[Mouse down test.sb2.zip](https://github.com/LLK/scratch-vm/files/2516371/Mouse.down.test.sb2.zip)

Another way to think about this:

A non-draggable sprite can trigger "when clicked" early, because it knows that the mouse down will never "turn into" a drag. A draggable sprite has to wait until mouse up to know if the gesture was a drag vs. a click.

paulkaplan:
@cwillisf could you take a look at this? @kchadha and I paired on figuring out the 2.0 behavior and implementing it, it would be good to get some new eyes on it.

fsih:
@paulkaplan cwillisf is on vacation until next Tuesday

fsih:
There seems to be a difference in behavior within the editor in scratch 2 vs this PR. In Scratch 2, it's possible to drag the cat to reposition it, while the cat runs away from you in this PR. Was this a conscious decision?
https://scratch.mit.edu/projects/227085207/
mousedown1
mousedown2

paulkaplan:
@fsih i should have mentioned, the goal here is only to make the behavior match in player/fullscreen mode, for project compatibility. In editor mode, there is some very unusual interaction between draggability and editor dragging in both 2.0 and 3.0.

fsih (approved):
LG!

PR #1431, opened 2018-08-07
Allow localizing micro:bit blocks

### LINKED ISSUE #1421: micro:bit commands should be able to be translated
### Expected Behavior

micro:bit commands should be available in other languages when the user switches language. The getInfo function for the extension has hard-coded strings for the command instead of formatMessage functions calls to define the string for translation and localize it at runtime. Look at the getInfo function for the [music](https://github.com/LLK/scratch-vm/blob/f38b96d70a539d932eef85ff85e5107be75070e2/src/extensions/scratch3_music/index.js#L621), pen or translate extensions as examples.

### Actual Behavior

micro:bit blocks are only available in English, and the strings are not available in Transifex for translators.

hyperobject:
### Resolves
This PR resolves #1421.

### Proposed Changes
This PR adds calls to `formatMessage` in the `getInfo` function to allow blocks to be translated, as well as moving micro:bit menus to separate getters and enums to allow them to be translated.

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -573,12 +573,7 @@ class Scratch3MicroBitBlocks {
                     text: formatMessage({
                         id: 'microbit.displayText',
                         default: 'display [TEXT]',
-                        description: `default text to display.
-                        IMPORTANT - the micro:bit only supports letters a-z, A-Z.
-                        Please substitute a default word in your language
-                        that can be written with those characters,
-                        substitute non-accented characters or leave it as "Hello!".
-                        Check the micro:bit site documentation for details`
+                        description: 'display text on the micro:bit display'
                     }),
                     blockType: BlockType.COMMAND,
                     arguments: {
@@ -587,7 +582,12 @@ class Scratch3MicroBitBlocks {
                             defaultValue: formatMessage({
                                 id: 'microbit.defaultTextToDisplay',
                                 default: 'Hello!',
-                                description: 'Hello!: the default text to display on the micro:bit'
+                                description: `default text to display.
+                                IMPORTANT - the micro:bit only supports letters a-z, A-Z.
+                                Please substitute a default word in your language
+                                that can be written with those characters,
+                                substitute non-accented characters or leave it as "Hello!".
+                                Check the micro:bit site documentation for details`
                             })
                         }
                     }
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_microbit/index.js
```diff
@@ -308,6 +341,151 @@ class Scratch3MicroBitBlocks {
         return 15;
     }
 
+    /**
+     * @return {array} - text and values for each buttons menu element
+     */
+    get BUTTONS_MENU () {
+        return [
+            {
+                text: formatMessage({
```
chrisgarrity: Just confirmed with Kreg, the buttons are physically labelled 'A' and 'B', so these shouldn't get translated.

src/extensions/scratch3_microbit/index.js
```diff
@@ -384,24 +578,40 @@ class Scratch3MicroBitBlocks {
...
                     blockType: BlockType.COMMAND,
                     arguments: {
                         TEXT: {
                             type: ArgumentType.STRING,
-                            defaultValue: 'Hello!'
+                            defaultValue: formatMessage({
+                                id: 'microbit.defaultTextToDisplay',
+                                default: 'Hello!',
+                                description: 'Hello!: the default text to display on the micro:bit'
+                            })
```
chrisgarrity: The micro:bit has a limited set of characters - we need to warn translators to only use numbers and 'a-zA-z'. How about:
`description: 'default text to display. IMPORTANT - the micro:bit only supports letters a-z, A-Z. Please substitute a default word in your language that can be written with those characters, substitute non-accented characters or leave it as "Hello!"'`

src/extensions/scratch3_microbit/index.js
```diff
@@ -438,7 +656,11 @@ class Scratch3MicroBitBlocks {
                 '---',
                 {
                     opcode: 'whenPinConnected',
-                    text: 'when pin [PIN] connected',
+                    text: formatMessage({
+                        id: 'microbit.whenPinConnected',
+                        default: 'when pin [PIN] connected',
+                        description: 'when a pin is connected to the micro:bit'
```
chrisgarrity: This is a difficult one to describe. I think we'll get better translations with a description like:
when the pin detects a connection to Earth/Ground

chrisgarrity (changes requested):
Overall it looks great. Just a couple of things that I think will help translators.

src/extensions/scratch3_microbit/index.js
hyperobject: Does that include punctuation like the exclamation point?

chrisgarrity (approved):
Thanks @technoboy10

PR #1432, opened 2018-08-07
ensure getTargetXY('_mouse_') is always whole numbers

mzgoddard:
### Resolves

I'm not sure this is a bug or derivation from how Scratch 2 works but I think it may be. 

https://llk.github.io/scratch-gui/develop/#238096787 horizontal and vertical lines "fade" in and out because the x, y values the lines are drawing from are not whole numbers. They have fractional values that causes the line to be drawn inbetween pixels causing a fading line for some mouse positions.

### Proposed Changes

Round mouse position x and y.

### Reason for Changes

Mouse position might not be whole numbers.

### Test Coverage

https://llk.github.io/scratch-gui/develop/#238096787

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -79,8 +79,8 @@ class Scratch3MotionBlocks {
         let targetX = 0;
         let targetY = 0;
         if (targetName === '_mouse_') {
-            targetX = util.ioQuery('mouse', 'getScratchX');
-            targetY = util.ioQuery('mouse', 'getScratchY');
+            targetX = Math.round(util.ioQuery('mouse', 'getScratchX'));
+            targetY = Math.round(util.ioQuery('mouse', 'getScratchY'));
         } else if (targetName === '_random_') {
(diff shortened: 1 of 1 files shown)


mzgoddard:
Talking with @rschamp, this is on to something, but I don't think this is the answer. It seems to me to follow Scratch 2 behaviour that changing the Sprite's position should do the rounding. Here is the [relevant code in Scratch 2](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/scratch/ScratchSprite.as#L180-L186). However currently in Scratch 3 we're only [rounding to the 8th decimal](https://github.com/LLK/scratch-vm/blob/develop/src/sprites/rendered-target.js#L274-L276). This rounding to the 8th decimal at this point is an optimization instead of needing to do it on every x, y, or direction request which Scratch 2 uses [snapToInteger](https://github.com/LLK/scratch-flash/blob/9fbac92ef3d09ceca0c0782f8a08deaa79e4df69/src/primitives/MotionAndPenPrims.as#L225) for.

PR #1434, opened 2018-08-07
Fix variable xml

### LINKED ISSUE #895: Duplicating/Cloning a sprite w/sprite-specific variables throws (non-fatal) type error.

kchadha:
### Resolves

Resolves #895 and is a replacement for #1409.

### Proposed Changes

Fix variable duplication during making a clone of a sprite or duplicating a sprite.
There are now functions for duplicating variables (and actually creating real variable objects from them so that they get serialized to XML correctly) in one of two ways:

1) The duplicated variable gets a new ID and all block fields that reference that variable in the provided blocks container get updated to use the new variable ID. This is necessary for duplicating a sprite, in which we want local variables of different sprites to have different IDs (which is necessary for those variables to have their own monitors).

2) The duplicated variable keeps the original variable's ID, and no block field update is necessary. This is required for making a clone of a sprite because sprite clones use the same blocks container as their parent sprite. Sprite clone variables cannot be monitored separately and thus do not need to have an ID separate from the original variable.

### Reason for Changes

Fixing issues where duplicating a sprite with local variables causes errors that can eventually cause the editor to crash.

### Test Coverage

Added unit tests for the new duplicate functions.

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -269,7 +269,7 @@ class Target extends EventEmitter {
 
     /**
      * Renames the variable with the given id to newName.
-     * @param {string} id Id of renamed variable.
+     * @param {string} id Id of variable to rename.
      * @param {string} newName New name for the variable.
      */
     renameVariable (id, newName) {
@@ -301,7 +301,7 @@ class Target extends EventEmitter {
 
     /**
      * Removes the variable with the given id from the dictionary of variables.
-     * @param {string} id Id of renamed variable.
+     * @param {string} id Id of variable to delete.
      */
     deleteVariable (id) {
         if (this.variables.hasOwnProperty(id)) {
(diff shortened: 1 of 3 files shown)


src/engine/target.js
```diff
@@ -313,6 +313,56 @@ class Target extends EventEmitter {
...
+    duplicateVariable (id, optKeepOriginalId) {
+        if (this.variables.hasOwnProperty(id)) {
+            const originalVariable = this.variables[id];
+            const newVariable = new Variable(
+                optKeepOriginalId ? id : null, // conditionally keep original id or generate a new one
+                originalVariable.name,
+                originalVariable.type,
+                originalVariable.isCloud
+            );
+            newVariable.value = originalVariable.value;
```
rschamp, line 334: Really, `value` isn't part of the constructor o.O?

rschamp (approved):

Issue #1424, opened 2018-08-04
Projects with old 1.4 drum block are mangled


as-com:
### Expected Behavior

This project (and any other project using the old 1.4 drum block) opens without ~~crashing~~ being mangled: https://scratch.mit.edu/projects/115797375/

### Actual Behavior

Scratch ~~crashes~~ makes the blocks disappear:
<img width="540" alt="image" src="https://user-images.githubusercontent.com/3819625/46045277-3c850c00-c0eb-11e8-94af-6a64fca3177e.png">
<img width="653" alt="image" src="https://user-images.githubusercontent.com/3819625/46045284-41e25680-c0eb-11e8-8151-dcf567b5e3c1.png">

Project in picture: https://scratch.mit.edu/projects/76004096/

### Operating System and Browser

~~macOS Mojave 10.14 Beta (18A347e), Safari 12.0 (14606.1.30)~~
macOS High Sierra 10.13.6 (17G2307), Firefox Nightly 64.0a1 (2018-09-25) (64-bit)

Related: #861

towerofnix:
I don't have much experience with that block - IIRC it's exactly equivalent to the other "play drum" block, but has an ordinary number input instead of one with a dropdown? If so, fixing this is as easy as adding an entry to [sb2_specmap.js](https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb2_specmap.js).

as-com:
They actually make different sounds. There’s another hard coded array of parameters buried somewhere within the old ActionScript codebase for this old block.

as-com:
Since https://github.com/LLK/scratch-vm/pull/1329 was merged, the project now opens without crashing, but the drum block disappears. See the updated issue description...

towerofnix:
Looks like the relevant code is [here](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/sound/SoundBank.as#L103-L118). It doesn't look too complicated to implement in 3.0. (Edit - specifically, the algorithm is "Is it the play MIDI (1.4) drum block? If so, subtract 35 from the index.") (I'm working on this atm.)

Double edit: oops, yes, [there's a mapping for drums too.](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/sound/SoundBank.as#L339-L388)

towerofnix:
[...]

PR #1435, opened 2018-08-08
Fix errors in music extension: Piano note 108 sample is 109 instead

DevanWolf:
Fix for the piano note 108 being C# 109 instead. Other parameters have been changed.


towerofnix:
I already created a PR for the `midiInstrument:` block which is compilant with Scratch 2.0 (it includes the mapping from 1.4 instruments to 2.0/3.0): #1329 - Whereas this PR, as well as the similar PR #883, do not include that mapping.

According to @as-com in https://github.com/LLK/scratch-vm/issues/1424#issuecomment-410541554, the `drum:duration:elapsed:from:` block will take similar treatment, which I don't *think* (but I may be wrong) this PR addresses.

thisandagain:
@ericrosenbaum Bump

ericrosenbaum:
I'm going to close this. The PR title suggests that the piano sample for note 108, a C, is actually a C#. I looked at the original sample set and confirmed that the sample I used is labeled as a C8. I also did a pitch analysis, which confirms that the sample is quite sharp. My best guess is that the piano that was recorded for the sample was tuned using "octave stretch" (see https://en.wikipedia.org/wiki/Stretched_tuning).

PR #1437, opened 2018-08-08
Allow localizing EV3 blocks

### LINKED ISSUE #1422: EV3 commands should be able to be translated
### Expected Behavior

EV3 commands should be available in other languages when the user switches language. The getInfo function for the extension has hard-coded strings for the command instead of formatMessage functions calls to define the string for translation and localize it at runtime. Look at the getInfo function for the [music](https://github.com/LLK/scratch-vm/blob/f38b96d70a539d932eef85ff85e5107be75070e2/src/extensions/scratch3_music/index.js#L621), pen or translate extensions as examples.

Keep in mind that the description field in the definition is context for **translators**, not help for the block.

### Actual Behavior

EV3 blocks are only available in English, and the strings are not available in Transifex for translators.

hyperobject:
### Resolves
This PR resolves #1422.

### Proposed Changes
This PR changes the EV3 extension's `getInfo` function to make it possible to translate block text. Note that none of the menus are translated, since they refer to labeled ports on an EV3.

### Test Coverage
Tested by making sure all blocks display as expected before and after moving block text into `formatMessage` calls. (macOS Firefox 63, Chrome 68)

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -1,6 +1,7 @@
 const ArgumentType = require('../../extension-support/argument-type');
 const BlockType = require('../../extension-support/block-type');
 const Cast = require('../../util/cast');
+const formatMessage = require('format-message');
 const uid = require('../../util/uid');
 // const log = require('../../util/log');
 const Base64Util = require('../../util/base64-util');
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_ev3/index.js
```diff
@@ -943,7 +968,11 @@ class Scratch3Ev3Blocks {
                 },
                 {
                     opcode: 'whenBrightnessLessThan',
-                    text: 'when brightness < [DISTANCE]',
+                    text: formatMessage({
+                        id: 'ev3.whenBrightnessLessThan',
+                        default: 'when brightness < [DISTANCE]',
```
chrisgarrity, line 973: @ericrosenbaum is this the correct block command? with the placeholder `[DISTANCE]` when the block is brightness? (Localization part is fine 👍 )

chrisgarrity (approved):
All the localization is great. Just wondering about the actual definition of the `when brightness...` block.

chrisgarrity (approved):
thanks @technoboy10 LGTM!

PR #1441, opened 2018-08-09
WeDo2 Extension: First Pass

evhan55:
### Resolves

Adds an implementation of the WeDo extension and its input and output blocks, using Scratch Link.

### TODO

- check all the opcodes and their input ranges / etc.
- set LED back to blue on connect
- debug rendering loop
- parse incoming messages by using table of port connections instead of packet length
- what to do if two distance or tilt sensors plugged in?
- make set motor direction block work while motors are running
- distance should report 0 - 100 in steps of ten
- check for battery level?
- set volume of piezo?
- rename some functions to simplify them

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -5,12 +5,37 @@ const log = require('../../util/log');
 const BLESession = require('../../io/bleSession');
 const Base64Util = require('../../util/base64-util');
 
+// TODO:
+// 1. keep track of who is connected to channels 1 and 2, set motor indices appropriately
+// 2. zero out sensor values when disconnected
+// 3. check that all blocks do something and are ready for Eric's spec confirmations
+// 4. refactor where you can before renaming things like 'Peripheral' throughout
+// 5. ???
+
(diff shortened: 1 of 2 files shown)


src/extensions/scratch3_wedo2/index.js
```diff
@@ -278,63 +350,223 @@ class WeDo2 {
...
+            }
+
+            // Distance Sensor
+            if (type === WeDo2Types.DISTANCE) {
+                const cmd = new Uint8Array(11);
+                cmd[0] = 1; // sensor format
+                cmd[1] = 2; // command type: write
+                cmd[2] = connectID; // connect id
+                cmd[3] = WeDo2Types.DISTANCE; // type
+                cmd[4] = 0; // mode: detect
```
evhan55, line 540: I might want to put the modes up in an enum

evhan55: ... but then I'd probably put the units and other things up there too and it'd be lots of enums, so I am still debating it

ericrosenbaum (approved):
LG!

PR #1442, opened 2018-08-10
Let wait, sayforsecs, thinkforsecs - handle big numbers

### LINKED ISSUE #1010: Can't wait longer than 2147483.647 seconds
### Expected Behavior

Any number of seconds entered into e.g. a "say" or "wait" block should cause the block to wait that number of seconds.

### Actual Behavior

Numbers of seconds over a certain large threshold cause the wait time to be zero. We often see kids typing in very large numbers like this.

With help from @rschamp, we found the exact threshold, which is 2147483.647 (1/1000 of max int, presumably because it is converted to milliseconds). So, waiting for 2147483.647 seconds seems to work (though we didn't actually wait for 24.8 days to find out), but waiting for 2147483.648 seconds waits for 0 instead:

big-wait-bug 

### Steps to Reproduce

See gif

joker314:
### Resolves

This PR resolves #1010

### Proposed Changes

For each block definition which uses `setTimeout`, ensure that the second argument passed to that function is no greater than the largest signed 32-bit integer.

As a result, the `wait` block in the Control category and the `thinkforsecs` and `sayforsecs` blocks in the Looks category will now do their action for a very long time when passed an unreasonably large argument -- rather than instantly completing the relevant action.

### Reason for Changes

This maintains compatibility with the way Scratch 2.0 behaved.

@ericrosenbaum stated that:
> We often see kids typing in very large numbers like this.

### Test Coverage

Manually tested through dragging out the aforementioned trio of blocks and verifying that they have a ring around them for more than a few seconds. Additionally, it was verified that when small numbers (like 5) were passed, the correct amount of time was taken.

On Windows 10,
- Chrome v68
- Microsoft Edge

`npm test` was run, only warnings appeared -- and these were regarding todo comments that existed pior to this pull request's filing.

`npm run coverage` was terminated half-way through, it was not enjoying that the sprite library was missing. This was not identified as an issue that this pull request created.

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -1,4 +1,5 @@
 const Cast = require('../util/cast');
+const MathUtil = require('../util/math-util.js');
 
 class Scratch3ControlBlocks {
     constructor (runtime) {
@@ -108,11 +109,11 @@ class Scratch3ControlBlocks {
     }
 
     wait (args) {
-        const duration = Math.max(0, 1000 * Cast.toNumber(args.DURATION));
+        const duration = 1000 * Cast.toNumber(args.DURATION);
         return new Promise(resolve => {
             setTimeout(() => {
                 resolve();
-            }, duration);
+            }, MathUtil.clamp(duration, 0, 2**31 - 1)); // setTimeout only supports up to 2**31 - 1
         });
     }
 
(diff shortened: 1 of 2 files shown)


ericrosenbaum:
Thanks! We decided to go a different strategy, which I will implement (see https://github.com/LLK/scratch-vm/issues/1010#issuecomment-421478816)

PR #1453, opened 2018-08-13
round x and y like Scratch 2 does

mzgoddard:
### Proposed Changes

Use Math.round instead of this._roundCoord.

### Reason for Changes

https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchSprite.as#L180-L186

Scratch 2 rounds the X and Y to the nearest whole number.

### Test Coverage

Projects like https://llk.github.io/scratch-gui/develop/#238746871 with fine lines and fractional positions needs rounded x and y values so that their lines are not offset from whole numbers. Offset lines from whole numbers in Scratch 3 render "inbetween pixels" and come out faded out or with striations.

https://scratch.mit.edu/projects/238746871/
https://llk.github.io/scratch-gui/develop/#238746871

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -272,8 +272,8 @@ class RenderedTarget extends Target {
         const oldY = this.y;
         if (this.renderer) {
             const position = this.renderer.getFencedPositionOfDrawable(this.drawableID, [x, y]);
-            position[0] = this._roundCoord(position[0], 8);
-            position[1] = this._roundCoord(position[1], 8);
+            position[0] = Math.round(position[0]);
+            position[1] = Math.round(position[1]);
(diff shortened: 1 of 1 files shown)


towerofnix:
This is the only place in the code which uses the (private) function `_roundCoord`, so does it make sense to get rid of that function now?

cwillisf (approved):

mrjacobbloom:
From some testing it looks like in 2.0, the x position reporter is precise to 8 digits and the x position monitor is precise to 6 digits. That specific rounding was added to fix project [117386152](https://llk.github.io/scratch-gui/develop/#117386152/) (see #1150) -- does rounding to integer happen to the stored value or does it only round while rendering?

PR #1467, opened 2018-08-13
Fix speech extension in Safari (#1202) by calling getUserMedia for ev…

picklesrus:
…ery listen and wait block rather than trying to reuse it.  Also get rid of some code that was not actually doing anything.

### Resolves

### Proposed Changes

### Reason for Changes

### Test Coverage


ericrosenbaum (approved):
LGTM!

PR #1468, opened 2018-08-14
export deserializeBlocks method

gnarf:
### Resolves

Export a helper method from serialization/sb3 to take a serialized `blocks` and produce something a bit more "programmatically parseable"

### Proposed Changes

Takes some inner functionality to the larger `deserialize` method and creates `deserializeBlocks` which does that and exports it for external utilities to be able to depend on.

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -742,6 +742,34 @@ const deserializeFields = function (fields) {
 };
 
 /**
+ * Covnert serialized INPUT and FIELD primitives back to hydrated block templates.
+ *
+ * @param {object} blocks Serialized SB3 "blocks" property of a target. Will be mutated.
+ * @return {object} input is modified and returned
+ */
+const deserializeBlocks = function (blocks) {
+    for (const blockId in blocks) {
+        if (!blocks.hasOwnProperty(blockId)) {
+            continue;
+        }
+        const block = blocks[blockId];
+        if (Array.isArray(block)) {
+            // this is one of the primitives
+            // delete the old entry in object.blocks and replace it w/the
+            // deserialized object
+            delete block[blockId];
+            deserializeInputDesc(block, null, false, blocks);
+            continue;
+        }
+        block.id = blockId; // add id back to block since it wasn't serialized
+        block.inputs = deserializeInputs(block.inputs, blockId, blocks);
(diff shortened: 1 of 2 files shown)


src/serialization/sb3.js
```diff
@@ -690,6 +690,7 @@ const deserializeInputs = function (inputs, parentId, blocks) {
     for (const inputName in inputs) {
         if (!inputs.hasOwnProperty(inputName)) continue;
         const inputDescArr = inputs[inputName];
+        if (!Array.isArray(inputDescArr)) continue;
```
kchadha: I think I'm still a little hesitant about this change. 

At the very least, this change (and the similar one below) needs a comment to explain that this line is solely for accommodating deserializing deserialized blocks (which is not the normal use case of this function).

In general, I think it's good for functions like this to be idempotent, but one thing that's weird about making deserialize work on already deserialized things is that there isn't really a similar analogy for `serialize` to work that way.

The closest (albeit somewhat different) thing to this is JSON.parse and JSON.stringify... 
`JSON.stringify(JSON.stringify(x))` and `JSON.stringify(x)` are not the same thing, and `JSON.parse(JSON.parse(x))` doesn't work.

gnarf: basically - there is no way to install targets from json without getting passed through this `deserialize`, I think it makes sense to make `deserialize(alreadyDeserialized)` result in no change (as opposed to weird undefined variables and broken code as it does with the current setup)

kchadha: Do you mind adding comments above both of those lines since they are not required for the normal/common usage flow of deserialize?

gnarf:
@kchadha i added those comments and also added some test cases for serializeBlocks / deserializeBlocks

I noticed while working on the testing that a lot of stuff was using `someObj.hasOwnProperty(prop)` and I replaced some of those with just using the `Object.prototype.hasOwnProperty.call(someObj, prop)` rather than assuming that the objects will be connected to the object prototype.

kchadha (approved):
LG

gnarf:
Thanks for working with me on this one!

This had me thinking that it might be useful to write a `scratch-sb3` and `scratch-sb2` serialize/deserialize package on npm for third parties.

PR #1471, opened 2018-08-14
Regulate micro:bit send timing

ericrosenbaum:
In order to prevent sending data too rapidly to the micro:bit over BLE:

- Make all display blocks yield for 100ms
- Add a "busy" flag that is set when we start sending data, and cleared when we finish
- Add a timeout to clear the busy flag if it has been true for too long (which may have occurred if a write response never came back, e.g. because the device was powered off)

Also, update BLESession with an optional `withResponse` param for Scratch Link that we set to true for micro:bit. See https://github.com/LLK/scratch-link/pull/71

- ~~Add a cache for data to send, which gets updated if a display block runs while we are busy sending. The cache gets overwritten with new data, so if it is updated multiple times, we will send only the latest data.~~
- ~~If we finish sending and the cache has been updated while we were busy, immediately send again. This prevents the need for additional timers.~~

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -27,6 +27,10 @@ const BLECommand = {
 
 const BLETimeout = 4500; // TODO: might need tweaking based on how long the device takes to start sending data
 
+/**
+ * A time interval to wait (in milliseconds) while a block that sends a BLE message is running.
+ * @type {number}
+ */
 const BLESendInterval = 100;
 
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
Looks good!
I'm a little worried about the case we discussed yesterday -- where sending two micro:bit animation frames quickly then doing nothing could result in showing the first frame instead of the second -- but it sounds like that might be OK. If it turns out to be a problem we can always change it later.

PR #1475, opened 2018-08-14
Use round() instead of floor() in repeat

### LINKED ISSUE #1083: Repeat block treats its argument differently than in Scratch 2.0
### Expected Behavior

The repeat block should follow the Scratch 2.0 repeat block behavior of rounding its argument.

### Actual Behavior

The repeat block appears to do a floor (int?) function on its argument.

### Steps to Reproduce

repeat issue

Windows 7 (64 bit), Chrome Version 66.0.3359.117 (Official Build) (64-bit)

The Scratch 3.0 repeat block will break projects like https://scratch.mit.edu/projects/42202752/, which, however wrongly, depend on the rounding of the repeat argument.

---
Moved from LLK/scratch-blocks#1470
/cc @Mr-Dave2

towerofnix:
### Resolves

Fixes #1083.

### Proposed Changes

Make the "repeat" control block round the given number of times according to `round()` (down if decimal is less than 0.5, otherwise up) instead of `floor()` (always down).

### Reason for Changes

Compatibility with Scratch 2.0.

### Test Coverage

A new unit test is added which verifies that the "repeat" block is behaving the right way. It passes, and did not pass before the `Math.floor` to `Math.round` change.

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -49,7 +49,7 @@ class Scratch3ControlBlocks {
     }
 
     repeat (args, util) {
-        const times = Math.floor(Cast.toNumber(args.TIMES));
+        const times = Math.round(Cast.toNumber(args.TIMES));
         // Initialize loop
         if (typeof util.stackFrame.loopCounter === 'undefined') {
             util.stackFrame.loopCounter = times;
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LG, thank you for adding the test as well!

PR #1476, opened 2018-08-14
Fix #1395 by resetting the utterance we check for in the edge trigger…

### LINKED ISSUE #1395: Speech recognition hats trigger again when you press green flag
### Expected Behavior

Pressing the green flag should not cause a "when I hear" block to run.

### Actual Behavior

If the most recent speech recognition result matches what is in a "when I hear" hat, when you press the green flag, it triggers the hat.

when-i-hear-bug

picklesrus:
… hat on green flag

### Resolves

#1395 

### Proposed Changes

Reset edgeTriggerUtterance phrase.

### Test Coverage

Manually tested.

### DIFF
--- a/src/extensions/scratch3_speech/index.js
+++ b/src/extensions/scratch3_speech/index.js
@@ -82,7 +82,8 @@ class Scratch3SpeechBlocks {
         this._currentUtterance = '';
 
         /**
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM!

Issue #1461, opened 2018-08-13
WeDo 2.0 distance sensor should report 0 - 100 in steps of ten


ericrosenbaum:

evhan55:
Resolved by #1481

Issue #1460, opened 2018-08-13
WeDo 2.0 "set motor direction" block should work while a motor is running


ericrosenbaum:

evhan55:
Resolved by #1481

Issue #1459, opened 2018-08-13
WeDo 2.0 extension should handle two of the same sensor


ericrosenbaum:
If two of the same sensor are plugged in (i.e. two distance sensors, or two tilt sensors), we should do something consistent (probably just always report the value of the sensor that was most recently plugged in).

evhan55:
Resolved by #1481

Issue #1457, opened 2018-08-13
WeDo 2.0 extension should parse incoming messages using port table


ericrosenbaum:
(instead of message length)

evhan55:
Resolved by #1481

PR #1481, opened 2018-08-15
Various WeDo2 fixes and enhancements

### LINKED ISSUE #1454: WeDo 2.0 extension should cast and clamp all input args

evhan55:
### Resolves

#1454: WeDo 2.0 extension should cast and clamp all input args
- 'motor on for' duration limited to **15 seconds**
- 'set motor power' clamped to **0 - 100**
- 'set light color' clamped to **0 - 100**
- 'play note' note clamped to **25 - 125** and duration limited to **3 seconds**
- invalid droppable input values ignored

#1455: WeDo 2.0 extension should set LED to blue on connecting
#1457: WeDo 2.0 extension should parse incoming messages using port table
#1459: WeDo 2.0 extension should handle two of the same sensor
#1460: WeDo 2.0 "set motor direction" block should work while a motor is running
#1461: WeDo 2.0 distance sensor should report 0 - 100 in steps of ten

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1,9 +1,11 @@
 const ArgumentType = require('../../extension-support/argument-type');
 const BlockType = require('../../extension-support/block-type');
+const Cast = require('../../util/cast');
 const color = require('../../util/color');
 const log = require('../../util/log');
 const BLESession = require('../../io/bleSession');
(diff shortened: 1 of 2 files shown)


ericrosenbaum:
Looks like setting a motor direction while the motor is running works great for "turn for seconds", but is not implemented yet for the case where the motor was started with "motor on" or "set motor power".

ericrosenbaum (approved):
LGTM!

PR #1482, opened 2018-08-15
WeDo command translation enabled.

### LINKED ISSUE #1423: WeDo commands should be able to be translated
### Expected Behavior

WeDo commands should be available in other languages when the user switches language. The getInfo function for the extension has hard-coded strings for the command instead of formatMessage functions calls to define the string for translation and localize it at runtime. Look at the getInfo function for the [music](https://github.com/LLK/scratch-vm/blob/f38b96d70a539d932eef85ff85e5107be75070e2/src/extensions/scratch3_music/index.js#L621), pen or translate extensions as examples.

Keep in mind that the description field in the definition is context for **translators**, not help for the block.

### Actual Behavior

WeDo blocks are only available in English, and the strings are not available in Transifex for translators.

evhan55:
Fixes #1423: WeDo commands should be able to be translated.

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1,5 +1,6 @@
 const ArgumentType = require('../../extension-support/argument-type');
 const BlockType = require('../../extension-support/block-type');
+const formatMessage = require('format-message');
 const color = require('../../util/color');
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
👍

PR #1487, opened 2018-08-16
Fix importing WeDo 2.0 blocks from SB2

ericrosenbaum:
Update the SB2 specmap to correctly import WeDo 2.0 blocks. This requires the menu name, arg name, and inputOP to all match.

Also, hide the WeDo's "play note" block from the palette, because the sound from the WeDo's tiny buzzer is often very quiet or inaudible.


evhan55:
Looks good to me!  Tested with project #205351780.

evhan55 (approved):

PR #1494, opened 2018-08-17
Some small EV3 fixes

### LINKED ISSUE #1353: Stop button should stop EV3 "beep" sound

### LINKED ISSUE #1358: EV3 and Microbit: shouldn't send requests if not connected?
The EV3 and Microbit extensions shouldn't try to send data requests to the devices if they aren't connected.  I thought this was correctly implemented in the Microbit but it seems to be gone now.

Might be good enhancements for both EV3 and Microbit so the console doesn't fill up with vm error statements.

### LINKED ISSUE #1373: EV3 motor does not brake at the end of a motor command
### Expected Behavior

After you run a `motor turn this way for 1 seconds` block, the EV3 motor should brake, so it stops, and then after a briefly delay switch to coast mode.

### Actual Behavior

As far as I can tell, then motor does not brake, and immediately goes to coast mode.

evhan55:
Resolves #1353: Stop button should stop EV3 "beep" sound
Resolves #1358: EV3 and Microbit: shouldn't send requests if not connected?
Resolves #1373: EV3 motor does not brake at the end of a motor command
- coasts after a 1 second delay

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -359,6 +359,7 @@ class EV3 {
     }
 
     motorCoast (port) {
+        if (!this.getPeripheralIsConnected()) return;
 
         const cmd = [];
         cmd[0] = 9; // length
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM!

PR #1503, opened 2018-08-20
Restore sound

kchadha:
### Resolves

Towards restoring a deleted sound.

### Proposed Changes

Deleting a sound returns a function to restore the deleted sound.

### Reason for Changes

Restore sound feature.

### Test Coverage

Added new unit test for vm.deleteSound and updated a unit test for rendered-target.deleteSound.

### Related PRs
LLK/scratch-gui#2935

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -589,7 +589,7 @@ class RenderedTarget extends Target {
      */
     deleteSound (index) {
         // Make sure the sound index is not out of bounds
-        if (index < 0 || index >= this.sounds.length) {
+        if (index < 0 || index >= this.sprite.sounds.length) {
             return null;
         }
         // Delete the sound at the given index
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -585,12 +585,17 @@ class RenderedTarget extends Target {
     /**
      * Delete a sound by index.
      * @param {number} index Sound index to be deleted
+     * @return {object} The deleted sound object, or null if no sound was deleted.
      */
     deleteSound (index) {
-        this.sprite.sounds = this.sprite.sounds
-            .slice(0, index)
-            .concat(this.sprite.sounds.slice(index + 1));
+        // Make sure the sound index is not out of bounds
+        if (index < 0 || index >= this.sounds.length) {
+            return null;
+        }
+        // Delete the sound at the given index
+        const deletedSound = this.sprite.sounds.splice(index, 1)[0];
         this.runtime.requestTargetsUpdate(this);
+        return deletedSound;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/virtual-machine.js
```diff
@@ -675,9 +675,20 @@ class VirtualMachine extends EventEmitter {
     /**
      * Delete a sound from the current editing target.
      * @param {int} soundIndex - the index of the sound to be removed.
+     * @return {?Function} A function to restore the sound that was deleted,
```
kchadha, line 678: Should the `F` in function be capital or not? I've seen it both ways in this same file... O.o

src/sprites/rendered-target.js
```diff
@@ -585,12 +585,17 @@ class RenderedTarget extends Target {
...
     deleteSound (index) {
-        this.sprite.sounds = this.sprite.sounds
-            .slice(0, index)
-            .concat(this.sprite.sounds.slice(index + 1));
+        // Make sure the sound index is not out of bounds
+        if (index < 0 || index >= this.sprite.sounds.length) {
+            return null;
+        }
+        // Delete the sound at the given index
+        const deletedSound = this.sprite.sounds.splice(index, 1)[0];
```
kchadha, line 596: Note the switch to splice here. I wanted to make sure this is okay, or check if there was any particular reason why we weren't using splice before...

test/unit/sprites_rendered-target.js
```diff
@@ -211,8 +211,10 @@ test('deleteSound', t => {
     const renderer = new FakeRenderer();
     a.renderer = renderer;
 
-    a.deleteSound(0);
+    const firstDeleted = a.deleteSound(0);
     t.deepEqual(a.sprite.sounds, [o2, o3]);
+    t.type(firstDeleted, 'object');
+    t.equals(firstDeleted.id, 1);
```
rschamp: Can we not do a similar `t.deepEqual` for the returned sound?

src/sprites/rendered-target.js
rschamp: I don't see anything about using `slice` over `splice`, so might just be an improvement!

src/virtual-machine.js
rschamp: Looking at https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler#function-type it looks like lowercase...

rschamp (approved):
Looks good!

src/virtual-machine.js
kchadha: According to this thread I found earlier, `function` with a lowercase `f` doesn't exist...?
https://github.com/Microsoft/TypeScript/issues/10946#issuecomment-248084184

scratch-blocks seems to also use both lowercase and uppercase, so I'm just going to stick with uppercase.

test/unit/sprites_rendered-target.js
kchadha: Done!

PR #1509, opened 2018-08-21
WeDo2 rate limiting

### LINKED ISSUE #1456: WeDo2 extension should rate-limit message sends

evhan55:
Resolves #1456: WeDo2 extension should rate-limit message sends

This is an attempt to rate limit the messages sent to the WeDo2 and executed by blocks in order to prevent crashing Scratch or the Scratch Link connection.

Two main strategies are used:
- don't send a message to the WeDo2 if a message is currently being sent and a reply/confirmation has not yet been received by the WeDo2 (i.e. drop messages to be sent if we are currently busy sending one already)
- yield on block op codes for some duration regardless of how long they take to execute, to prevent a stack of multiple blocks that send messages from being dropped/ignored if the first block takes a while to send

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -25,6 +25,12 @@ const UUID = {
 };
 
 /**
+ * A time interval to wait (in milliseconds) while a block that sends a BLE message is running.
+ * @type {number}
+ */
+ const BLESendInterval = 100;
+
+/**
  * Enum for WeDo2 sensor and output types.
  * @readonly
  * @enum {number}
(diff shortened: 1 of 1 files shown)


evhan55:
@ericrosenbaum This PR is beginning with what we worked on together, but is open to changes based on the various scenarios you were envisioning, for example:

- one forever block changing the LED
- another stack that tries to turn on and off the motor?

ericrosenbaum:
This looks good! I tested that specific scenario that I thought might be an issue, where set light color blocks are in a loop, and I press the stop button to stop the motor, but it seems to work fine.

ericrosenbaum (approved):
LGTM!

PR #1517, opened 2018-08-22
Make "switch costume" and "switch backdrop" blocks compatible with 2.0

### LINKED ISSUE #1490: Switch costume/backdrop to boolean value switches to costume nr. 0 or 1
### Expected Behavior

Using a switch costume or backdrop to block with a parameter that is a boolean block in 2.0 will switch to a backdrop or costume named 'true' or 'false'. It should do the same in 3.0.

### Actual Behavior

In 3.0, it switches to the costume/backdrop number 0 (last costume/backdrop) or 1.

### Steps to Reproduce
Found in 86981455
See project 239254767 for a simple example (thanks @BryceLTaylor)
See project 239255071 for a few more examples

Mac OS, Chrome

joker314:
### Resolves

Resolves #1490 

### Proposed Changes

- `switch backdrop to` block is (EDIT: [almost](https://github.com/LLK/scratch-vm/pull/1517#discussion_r223779175)) compatible with 2.0 behaviour
- `switch costume to` block is compatible with 2.0 behaviour
- unit tests guard against future incompatibility
- make "random backdrop" have lower priority than a costume with that name

See [2.0 behaviour description](https://github.com/LLK/scratch-vm/issues/1490#issuecomment-413679793).

### Reason for Changes

Making sure that 2.0 projects still work in 3.0.

I placed the "random backdrop" feature as _less_ important than a costume name. This means that if a 2.0 project already had a backdrop with the name "random backdrop", references to it by name will still hold. Was this a good idea?

This is different to how "previous backdrop" and "next backdrop" behave -- they can't be overridden (per 2.0); however this is the same behaviour as "previous costume" and "next costume", which _can._

### Test Coverage

- 2 automatic unit tests were added. They all pass after my changes. Expand below to see which failed before (more than was initially documented in the issue).
 
<details><summary>Assertions which failed prior to changes</summary>
<h4>Costumes</h4>
<ul>
    <li> <code>switch costume to NaN</code> (e.g. <code>0/0</code>) should flawlessly switch to the first costume. </li>
    <li> <code>switch costume to Infinity</code> (e.g. <code>1/0</code>) should flawlessly switch to the first costume. </li>
    <li> <code>switch costume to -Infinity</code> (e.g. <code>-1/0</code>) should flawlessly switch to the first costume. </li>
    <li> <code>switch <strong>costume</strong> to previous <strong>backdrop</strong></code> </li> should do nothing
    <li> <code>switch <strong>costume</strong> to next <strong>backdrop</strong></code> should do nothing</li>
    <li> Strings should only be interpreted as indices if they contain at least one numeric digit.</li>
    <li> <code>switch costume to &lt;1 = 0></code> should switch to the costume named "false" if it exists #1490 </li></ul>
  <h4>Backdrops</h4>
    <ul><li> <s><code>switch backdrop to [previous backdrop]</code> decrements the backdrop <em>even if there's a backdrop with that name</em> </s></li>
    <li> <s><code>switch backdrop to [next backdrop]</code> increments the backdrop <em>even if there's a backdrop with that name</em></s></li>
    <li> <code>switch backdrop to NaN</code> (e.g. <code>0/0</code>) should flawlessly switch to the first backdrop.</li>
    <li> <code>switch backdrop to Infinity</code> (e.g. <code>1/0</code>) should flawlessly switch to the first backdrop.</li>
    <li> <code>switch backdrop to -Infinity</code> (e.g. <code>-1/0</code>) should flawlessly switch to the first backdrop.</li></ul>
</details>
<hr>

- Manually tested with 239254767
- Manually tested with 239255071
- Unable to manually test with 86981455 as it wasn't a minimal example
- Manually confirmed that switching costume by name still worked

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -344,65 +344,86 @@ class Scratch3LooksBlocks {
     }
 
     /**
-     * Utility function to set the costume or backdrop of a target.
+     * Utility function to set the costume of a target.
      * Matches the behavior of Scratch 2.0 for different types of arguments.
-     * @param {!Target} target Target to set costume/backdrop to.
+     * @param {!Target} target Target to set costume to.
      * @param {Any} requestedCostume Costume requested, e.g., 0, 'name', etc.
      * @param {boolean=} optZeroIndex Set to zero-index the requestedCostume.
      * @return {Array.<!Thread>} Any threads started by this switch.
      */
-    _setCostumeOrBackdrop (target,
-        requestedCostume, optZeroIndex) {
+    _setCostume (target, requestedCostume, optZeroIndex) {
         if (typeof requestedCostume === 'number') {
-            target.setCostume(optZeroIndex ?
-                requestedCostume : requestedCostume - 1);
+            target.setCostume(optZeroIndex ? requestedCostume : requestedCostume - 1);
         } else {
-            const costumeIndex = target.getCostumeIndexByName(requestedCostume);
-            if (costumeIndex > -1) {
+            const costumeIndex = target.getCostumeIndexByName(requestedCostume.toString());
+
+            if (costumeIndex !== -1) {
                 target.setCostume(costumeIndex);
-            } else if (requestedCostume === 'previous costume' ||
-                       requestedCostume === 'previous backdrop') {
-                target.setCostume(target.currentCostume - 1);
-            } else if (requestedCostume === 'next costume' ||
-                       requestedCostume === 'next backdrop') {
+            } else if (requestedCostume === 'next costume') {
                 target.setCostume(target.currentCostume + 1);
-            } else if (requestedCostume === 'random backdrop') {
-                const numCostumes = target.getCostumes().length;
-                if (numCostumes > 1) {
-                    let selectedIndex = Math.floor(Math.random() * (numCostumes - 1));
-                    if (selectedIndex === target.currentCostume) selectedIndex += 1;
-                    target.setCostume(selectedIndex);
-                }
-            } else {
-                const forcedNumber = Number(requestedCostume);
-                if (!isNaN(forcedNumber)) {
-                    target.setCostume(optZeroIndex ?
-                        forcedNumber : forcedNumber - 1);
-                }
+            } else if (requestedCostume === 'previous costume') {
+                target.setCostume(target.currentCostume - 1);
+            } else if (!isNaN(requestedCostume) &&
+                        (typeof requestedCostume !== 'string' || /\d/g.test(requestedCostume))) {
+                target.setCostume(optZeroIndex ? Number(requestedCostume) : Number(requestedCostume) - 1);
             }
         }
-        if (target === this.runtime.getTargetForStage()) {
-            // Target is the stage - start hats.
-            const newName = target.getCostumes()[target.currentCostume].name;
-            return this.runtime.startHats('event_whenbackdropswitchesto', {
-                BACKDROP: newName
-            });
-        }
+
+        // Per 2.0, 'switch costume' can't start threads even in the Stage.
         return [];
     }
 
+    /**
+     * Utility function to set the backdrop of a target.
+     * Matches the behavior of Scratch 2.0 for different types of arguments.
+     * @param {!Target} stage Target to set backdrop to.
+     * @param {Any} requestedBackdrop Backdrop requested, e.g., 0, 'name', etc.
+     * @param {boolean=} optZeroIndex Set to zero-index the requestedBackdrop.
+     * @return {Array.<!Thread>} Any threads started by this switch.
+     */
+    _setBackdrop (stage, requestedBackdrop, optZeroIndex) {
+        if (typeof requestedBackdrop === 'number') {
+            stage.setCostume(optZeroIndex ? requestedBackdrop : requestedBackdrop - 1);
+        } else if (requestedBackdrop === 'next backdrop') {
+            stage.setCostume(stage.currentCostume + 1);
+        } else if (requestedBackdrop === 'previous backdrop') {
+            stage.setCostume(stage.currentCostume - 1);
+        } else {
+            const costumeIndex = stage.getCostumeIndexByName(requestedBackdrop.toString());
+
+            if (costumeIndex !== -1) {
+                stage.setCostume(costumeIndex);
+            } else if (requestedBackdrop === 'random backdrop') {
+                const numCostumes = stage.getCostumes().length;
+                if (numCostumes > 1) {
+                    let selectedIndex = Math.floor(Math.random() * (numCostumes - 1));
+                    if (selectedIndex === stage.currentCostume) selectedIndex += 1;
+                    stage.setCostume(selectedIndex);
+                }
+            } else if (!isNaN(requestedBackdrop) &&
+                        (typeof requestedBackdrop !== 'string' || /\d/g.test(requestedBackdrop))) {
+                stage.setCostume(optZeroIndex ? Number(requestedBackdrop) : Number(requestedBackdrop) - 1);
+            }
+        }
+
+        const newName = stage.getCostumes()[stage.currentCostume].name;
+        return this.runtime.startHats('event_whenbackdropswitchesto', {
+            BACKDROP: newName
+        });
+    }
+
     switchCostume (args, util) {
-        this._setCostumeOrBackdrop(util.target, args.COSTUME);
+        this._setCostume(util.target, args.COSTUME);
     }
 
     nextCostume (args, util) {
-        this._setCostumeOrBackdrop(
+        this._setCostume(
             util.target, util.target.currentCostume + 1, true
         );
     }
 
     switchBackdrop (args) {
-        this._setCostumeOrBackdrop(this.runtime.getTargetForStage(), args.BACKDROP);
+        this._setBackdrop(this.runtime.getTargetForStage(), args.BACKDROP);
     }
 
     switchBackdropAndWait (args, util) {
@@ -410,7 +431,7 @@ class Scratch3LooksBlocks {
         if (!util.stackFrame.startedThreads) {
             // No - switch the backdrop.
             util.stackFrame.startedThreads = (
-                this._setCostumeOrBackdrop(
+                this._setBackdrop(
                     this.runtime.getTargetForStage(),
                     args.BACKDROP
                 )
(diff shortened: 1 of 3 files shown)


joker314:
Regarding the unit tests added, would it be a good idea to split them into a separate file -- and, since the backdrop and costume tests are quite similar, use a loop/function for the common bits?

src/blocks/scratch3_looks.js
```diff
@@ -344,73 +344,94 @@ class Scratch3LooksBlocks {
...
-                target.setCostume(target.currentCostume - 1);
-            } else if (requestedCostume === 'next costume' ||
-                       requestedCostume === 'next backdrop') {
+            } else if (requestedCostume === 'next costume') {
                 target.setCostume(target.currentCostume + 1);
-            } else if (requestedCostume === 'random backdrop') {
-                const numCostumes = target.getCostumes().length;
+            } else if (requestedCostume === 'previous costume') {
+                target.setCostume(target.currentCostume - 1);
+            } else if (!isNaN(requestedCostume) &&
```
kchadha: The first part of this if will always be true because `NaN` is of type `'number'` which you already checked for in the parent if.

src/blocks/scratch3_looks.js
```diff
@@ -344,73 +344,94 @@ class Scratch3LooksBlocks {
...
+                        (typeof requestedCostume !== 'string' || /\d/g.test(requestedCostume))) {
```
kchadha: I'm not sure why you need the regex test here.

If you're tying to see if it's a string representation of a number e.g. "12", you could try casting requestedCostume to a number and see if it results in `NaN` or not.

src/blocks/scratch3_looks.js
```diff
@@ -344,73 +344,94 @@ class Scratch3LooksBlocks {
...
+     * @param {boolean=} optZeroIndex Set to zero-index the requestedBackdrop.
+     * @return {Array.<!Thread>} Any threads started by this switch.
+     */
+    _setBackdrop (stage, requestedBackdrop, optZeroIndex) {
+        if (typeof requestedBackdrop === 'number') {
+            stage.setCostume(optZeroIndex ? requestedBackdrop : requestedBackdrop - 1);
+        } else if (requestedBackdrop === 'next backdrop') {
+            stage.setCostume(stage.currentCostume + 1);
+        } else if (requestedBackdrop === 'previous backdrop') {
+            stage.setCostume(stage.currentCostume - 1);
```
kchadha: This behavior is consistent with Scratch 2.0, as you noted here: 

https://github.com/LLK/scratch-vm/issues/1490#issuecomment-413714669

But from a usability/consistency perspective this definitely seems like a bug.

My thought is to make backdrops more consistent with costumes in this regard (e.g. if you have a costume named 'next costume' you can switch to it).
@thisandagain do you have any thoughts on this?

At the very least, the `random backdrop` option should be handled the same way we decide to handle `previous backdrop` and `next backdrop`

src/blocks/scratch3_looks.js
kchadha: e.g. in the [old code here](https://github.com/LLK/scratch-vm/pull/1517/files?utf8=%E2%9C%93&diff=split#diff-8d62e377c35d800fc8cfe63d4e98232fL376)

src/blocks/scratch3_looks.js
thisandagain: Agreed @kchadha. I think this is a place I'd be ok with a breaking change. We should document it here: https://github.com/LLK/scratch-vm/wiki/Compatibility

thisandagain:
Thanks a ton @joker314. In addition to the PR, I really appreciate the documentation and research in https://github.com/LLK/scratch-vm/issues/1490. 😄

src/blocks/scratch3_looks.js
joker314: Scratch thinks that empty space is not a number (whereas JavaScript thinks it's 0). Looking for a digit was [the algorithm used by 2.0](https://github.com/LLK/scratch-flash/blob/develop/src/interpreter/Interpreter.as#L399-L413) -- but it didn't use a regular expression in its implementation.

I think this regular expression should be moved to `util/cast.js`

test/unit/blocks_looks.js
```diff
@@ -23,6 +25,146 @@ const fakeRuntime = {
...
+
+        looks.switchCostume({COSTUME: arg}, {target});
+
+        return target.currentCostume + 1; // Convert to 1-indexed.
+    };
+
+    // Non-existant costumes do nothing
+    t.strictEqual(testCostume(['a', 'b', 'c', 'd'], 'e', 3), 3);
+
+    // Difference between string and numeric arguments
```
kchadha: These tests are awesome and very thorough!

Could you clarify this comment a little bit e.g. "Number arguments are always treated as the costume index. String arguments are treated as the name of the costume, otherwise converted to the costume index if possible"  and add one more test case to illustrate all the possible cases here
`t.strictEqual(testCostume(['a', 'b', 'c'], '2'), 2);`

test/unit/blocks_looks.js
```diff
@@ -23,6 +25,146 @@ const fakeRuntime = {
...
+    // False is 0 (the last costume)
+    t.strictEqual(testBackdrop(['a', 'b', 'c', 'd'], false), 4);
+
+    // Booleans are backdrop names where possible.
+    t.strictEqual(testBackdrop(['a', 'true', 'false', 'd'], false), 3);
+    t.strictEqual(testBackdrop(['a', 'true', 'false', 'd'], true), 2);
+
+    // Backdrop indices should wrap around.
+    t.strictEqual(testBackdrop(['a', 'b', 'c', 'd'], -1), 3);
+    t.strictEqual(testBackdrop(['a', 'b', 'c', 'd'], -4), 4);
```
kchadha, line 160: Yay for fixing the negative number wrap-around bug for backdrops!

test/unit/blocks_looks.js
```diff
@@ -23,6 +25,146 @@ const fakeRuntime = {
...
+     * Test which costume index the `switch costume`
+     * block will jump to given an argument and array
+     * of costume names.
+     *
+     * @param {string[]} costumes List of costume names as strings
+     * @param {string|number|boolean} arg The argument to provide to the block.
+     * @param {number} [currentCostume=1] The 1-indexed default costume for the sprite to start at.
+     * @return {number} The 1-indexed costume index on which the sprite lands.
+     */
+    const testCostume = (costumes, arg, currentCostume = 1) => {
```
kchadha: `testCostume` and `testBackdrop` can be helpers outside of this test definition. You can also add one more optional argument at the end of testCostume to swap the relevant costume/backdrop bits so that the two functions can share code.

E.g. 
```
const testCostume = (costumes, arg, currentCostume = 1, costumeOrBackdrop = 'costume') => { ... };

const testBackdrop = (backdrop, arg, currentBackdrop = 1) => testCostume(backdrop, arg, currentBackdrop, 'backdrop');
```

I also think this code is okay as is though.

src/blocks/scratch3_looks.js
joker314: `isNaN` casts its argument to a number before checking if it is NaN. Hence, `isNaN("meow") -> true`. Many complain that the function is named improperly and hence its behaviour is unintuitive.

`Number.isNaN` does not do this casting. Hence, this PR could do the casting explicitly, and then use `Number.isNaN`.

Note: `Number.isNaN` is not supported in IE11. Is this okay?

kchadha (changes requested):
@joker314 Thanks for making this pull request and for detailing the Scratch 2.0 behavior in the comment on #1490. I apologize for the delay on this review.

I am requesting some minor changes/code cleanup, and would also like to request adding more comments in the code to illustrate what kinds of cases you're trying to address in which parts of the code.

Thanks for adding the thorough unit tests, they're awesome!

I added a comment about maybe making the `switch backdrop` behavior similar to `switch costume` in how it handles backdrops named 'previous backdrop' and 'next backdrop' but I think it's also okay to file that as a follow-up separate issue since what you have now is consistent with Scratch 2.0.

src/blocks/scratch3_looks.js
kchadha: Oh I didn't know about `isNaN` implicitly casting! Thanks. =] In that case, do you mind adding a comment in the places where you use it like this?

src/blocks/scratch3_looks.js
kchadha: Hmmm it seems like maybe we could use a combination of `parseInt` and `isNaN`. Though, unfortunately `parseInt` does not work the way we want for boolean inputs, so we would have to deal with those explicitly.

I'm also okay with leaving this code as is now that I know the reason.

In either case, could you add a comment here explaining the digit check (e.g. the fact that whitespace is treated as 0 when casting to a number, and we don't want that)

joker314: This is a good point. I think I'll use the `Cast` utility for this -- `Cast.isWhiteSpace`.

joker314:
Thank you very much for the review @kchadha. I've addressed your thoughtful feedback, and have introduced the breaking change which makes backdrop special arguments be handled in a similar way to costumes.

This pull request is ready for re-review.

Here's a base for what one might want to include in the [Compatibility wiki page](https://github.com/LLK/scratch-vm/wiki/Compatibility), if it's useful:

> ### Backdrops named 'previous backdrop' and 'next backdrop' can be switched to via blocks
>
> In Scratch 2.0, `switch backdrop to [next backdrop]` would switch to the backdrop sequentially following the current backdrop, even if there was a backdrop literally named "next backdrop". The same behaviour is seen in "previous backdrop".
>
> This differed to the behaviour of the `switch costume` block, which would not perform the block input's special function if there existed a costume with the corresponding name.
>
> To make the blocks more intuitive, the `switch backdrop` block now uses the `switch costume` behaviour. This is also likely to be what a user would expect, since the workaround for incrementing a backdrop is trivial and stable, whereas the workaround for switching to a backdrop by name would require trying every costume inefficiently until it is found (unless the order in which the backdrops have been arranged is hard-coded)

src/blocks/scratch3_looks.js
joker314: Since I'm now using the very readable utility function, I decided to remove the explanation that JS casts whitespace to 0.

kchadha (approved):
@joker314 Thank you for making the requested changes! This PR looks good to go.

Issue #1450, opened 2018-08-13
Some 2.0 Projects won't open in 3.0


BryceLTaylor:
Some 2.0 projects fail to load in 3.0 giving this error:
![screen shot 2018-08-13 at 1 24 33 pm](https://user-images.githubusercontent.com/1088291/44047607-d5fc74c4-9efc-11e8-8e62-6788232d5d84.png)

219549113
`Error: Failed to load project from server [id=#219549113]: TypeError: Cannot read property 'id' of undefined`
`gui Unhandled Error: Error: Failed to load project from server [id=#219549113]: TypeError: Cannot read property 'id' of undefined`

10118230
`Error: Failed to load project from server [id=#10118230]: Connection checks failed. Input "CONDITION" connection on "control_if_else" block (id="Ems8l;|SY:7JqPGQ,DVu") expected Boolean, found String`
`gui Unhandled Error: Error: Failed to load project from server [id=#10118230]: Connection checks failed. Input "CONDITION" connection on "control_if_else" block (id="Ems8l;|SY:7JqPGQ,DVu") expected Boolean, found String`

15200040
`Error: Failed to load project from server [id=#15200040]: Error: Serialized variable type with id '5_ow%Z#Tkawli]^Gs1j/-_Level' had type list, and does not match variable field that references it: <field xmlns="http://www.w3.org/1999/xhtml" name="VARIABLE" id="5_ow%Z#Tkawli]^Gs1j/-_Level" variabletype="">_Level</field>.`
`raven.js:55 Error: Tried to serialize a variable field with no variable.
    at Object.Blockly.Xml.fieldToDomVariable_ (lib.min.js:167)
    at Object.Blockly.Xml.fieldToDom_ (lib.min.js:167)
    at Object.Blockly.Xml.allFieldsToDom_ (lib.min.js:167)
    at Object.Blockly.Xml.blockToDom (lib.min.js:167)
    at new Blockly.Events.Delete (lib.min.js:167)
    at Blockly.BlockSvg.Blockly.Block.dispose (lib.min.js:167)
    at Blockly.BlockSvg.dispose (lib.min.js:167)
    at Blockly.WorkspaceSvg.Blockly.Workspace.clear (lib.min.js:167)
    at Blockly.WorkspaceSvg.clear (lib.min.js:167)
    at Blockly.WorkspaceSvg.Blockly.Workspace.dispose (lib.min.js:167)
[...]

PR #1524, opened 2018-08-23
Resolve project load error with undefined opcodes inside c/e blocks.

kchadha:
### Resolves

Project loading error for the first project mentioned in #1450.

### Proposed Changes

Fix issue where a c-block that contained only blocks with unknown opcodes, was causing an import error.

### Reason for Changes

#1254 skips unknown opcodes when importing blocks from sb2 projects. However, that PR missed the case where all of the statements inside a c-block are undefined. This PR fixes projects that were failing to load because of that issue.

### Test Coverage

Manually tested project 219549113 (mentioned in #1450). Tested some of the projects in the top 1000 projects list. Added fixture and integration test for example failing project.

cc/ @mzgoddard -- Just wanted to give you a heads up on this patch.

### How to test this
1. In Scratch 2.0, drag out a c-block (e.g. `forever`)
2. Plug in one or more blocks with unknown opcodes inside the c-block
(Examples are: `Lego WeDo 1.0` blocks, a custom procedure caller from a different project or sprite that came out of the backpack without its definition).
3. Don't plug in any blocks with known opcodes inside of the c-block (those already import correctly), but you may optionally plug these outside of the c-block if you wish.
4. Import the project into 3.0
The project *should* be able to load and there should be nothing inside the c-block (blocks with unknown opcodes get skipped when importing into 3.0). There *should* be warnings about skipped opcodes in the console.


kchadha:
Also cc/ @BryceLTaylor

rschamp (approved):

spectranaut (commented):
oh no I accidentally hit review in the wrong window! please ignore this.

PR #1528, opened 2018-08-24
Fix Comment import/save/load issue

kchadha:
### Resolves

Resolves issue found in smoke test on 08/24 where some comments weren't being imported correctly when testing the fix in #1522.

### Proposed Changes

Fix issue where block comments converted to workspace comments import correctly when there are no scripts on the workspace.

### Test Coverage

Added new integration test and fixture.

### How to Test this PR
1. Create a new 2.0 project.
2. Drag out a block, e.g. `move 10 steps`
3. Use the block context menu to add a block comment to the block.
4. Ensure that the project is saved.
5. Move the new block comment away from the block to detach it from the block.
6. Ensure that the project is saved.
7. Delete the original block, e.g. `move 10 steps`
8. Ensure that the project is saved.
9. Load this project in 3.0, the comment should exist as a workspace comment.
10. Download this project to your computer
11. Upload the project from your computer.
The project *should not* fail to load. The workspace comment *should* still be there.


rschamp (approved):
Thank you for the test!

BryceLTaylor:
Looks good.  It seems there is a bug in 2.0 where if you try to move a comment from the stage to the workspace and save, when you reload the project it is still attached to the block.  3.0 also imports the comment attached to the block, so this is consistent.

If there are no blocks, it appears as a workspace comment, but if there are any it picks one and attaches itself, similar to how it does in 2.0.

PR #1540, opened 2018-08-28
Rate-limiting with a token bucket for the WeDo extension

### LINKED ISSUE #1523: WeDo2 motor blocks do not work correctly with two motors plugged in
regression

### LINKED ISSUE #1531: Wedo2 sometimes drops commands
### Expected Behavior

All blocks that send commands to the WeDo 2.0 should take effect, including turning motors on and off, setting their direction and power, and setting the light color.

### Actual Behavior

While a forever loop is running that sends a steady stream of commands (such as forever set light color to random), clicking on a motor block to turn it on will sometimes not work.

We believe this is due to the command being dropped, because of our current rate-limiting strategy.

ericrosenbaum:
Resolves https://github.com/LLK/scratch-vm/issues/1531
Resolves https://github.com/LLK/scratch-vm/issues/1523

Previously, the WeDo extension used a simple rate limiting strategy that would drop any messages while it was currently sending. This approach caused the bugs in the issues above. 

This PR adds a rateLimiter utility that uses the [token bucket](https://en.wikipedia.org/wiki/Token_bucket) strategy. It limits sends to 30 per second. A counter accumulates tokens at the rate of 30 per second, and each send costs a token. If no tokens remain, the message is dropped.

The WeDo blocks that send messages all yield for 100ms, so a forever loop with a "set light color" in it for example is sending about 10 messages per second. With three or more such loops running in parallel, we will start to drop messages. Because of the limitations of the WeDo, this seems like enough bandwidth. 

It's possible to create lots of clones that all have forever loops with sending blocks, resulting in thousands of sends per second:

screen shot 2018-08-28 at 6 13 37 pm

This change protects against problems caused by this clone-sending-apocalypse situation (such as messages queuing up and running on the WeDo for a tens of seconds after the Scratch code has stopped). 

This PR also fixes an issue that was revealed by these changes: When you run fast-sending code with clones as shown above, and also turn on a motor, pressing the stop button would result in the code stopping, but the motor never getting turned off. Pressing the stop button additional times would still not stop it. This is because the initial stop button press, during the fast-sends, tried to send a motor off command that was dropped, but it nonetheless updated the internal motor state to off; later stop button presses would check the motor state, and not send the motor off command... So I just removed the check for motor.isOn, since we can safely send a motor off command to a motor that is off.

**Testing**
- [x] Mac Chrome
- [x] Mac Firefox
- [x] Mac Safari
- [x] Windows Chrome
- [x] Windows Edge
- [x] Windows Firefox

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -364,7 +364,7 @@ class WeDo2 {
          * @type {RateLimiter}
          * @private
          */
-        this._rateLimiter = new RateLimiter(30);
+        this._rateLimiter = new RateLimiter(20);
     }
 
     /**
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -358,6 +358,12 @@ class WeDo2 {
         this._onConnect = this._onConnect.bind(this);
         this._onMessage = this._onMessage.bind(this);
 
+        /**
+         * A rate limiter utility, to help limit the rate at which we send BLE messages
+         * over the socket to Scratch Link to a maximum number of sends per second.
+         * @type {RateLimiter}
+         * @private
+         */
         this._rateLimiter = new RateLimiter(30);
     }
 
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -32,6 +32,12 @@ const UUID = {
 const BLESendInterval = 100;
 
 /**
+ * A maximum number of BLE message sends per second, to be enforced by the rate limiter.
+ * @type {number}
+ */
+const BLESendRateMax = 20;
+
+/**
  * Enum for WeDo2 sensor and output types.
  * @readonly
  * @enum {number}
@@ -364,7 +370,7 @@ class WeDo2 {
          * @type {RateLimiter}
          * @private
          */
-        this._rateLimiter = new RateLimiter(20);
+        this._rateLimiter = new RateLimiter(BLESendRateMax);
     }
 
     /**
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -396,7 +396,7 @@ class WeDo2 {
      */
     stopAllMotors () {
         this._motors.forEach(motor => {
-            if (motor && motor.isOn) {
+            if (motor) {
                 motor.setMotorOff();
             }
         });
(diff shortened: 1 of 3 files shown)


src/extensions/scratch3_wedo2/index.js
```diff
@@ -369,6 +357,8 @@ class WeDo2 {
 
         this._onConnect = this._onConnect.bind(this);
         this._onMessage = this._onMessage.bind(this);
+
+        this._rateLimiter = new RateLimiter(30);
```
evhan55: Maybe a comment to describe this?

src/extensions/scratch3_wedo2/index.js
```diff
@@ -549,20 +539,11 @@ class WeDo2 {
      * @private
      */
     _send (uuid, message) {
-        if (!this.getPeripheralIsConnected()) return;
-        if (this._sending) return;
+        if (!this.getPeripheralIsConnected()) return Promise.resolve();
```
evhan55: oooh, what is Promise.resolve() doing here (and in line 544?)

src/util/rateLimiter.js
```diff
@@ -0,0 +1,74 @@
+const Timer = require('../util/timer');
+
+class RateLimiter {
```
evhan55, line 3: Looks really great!

evhan55 (approved):
It looks great to me!  Added a couple of comments.

evhan55:
With a single `forever>setLED` stack, `motor on`, `motor off` and stop button all behave well
Same with two `forever>setLED` stacks
and mostly same with three `forever>setLED` stacks
with four `forever>setLED` stacks, `motor on`, `motor off` and stop button only work intermittently, about one every 5 or 6 times that I try them
seems great! :+1:
also, both motors now work again with the `'motors'` and `'all motors'` options

src/extensions/scratch3_wedo2/index.js
ericrosenbaum: Good question! The JSDOC indicates that the function should return a promise, but we were just returning null from it when we needed to bail out, previously. The rate limiter change revealed a bug where another function (_stopAll) used a then() function on another function that returned the promise from _send(). In the case where the message is dropped because of rate limiting, the _send function was not returning a promise, and the attempt to use then() threw an error. 

I suppose this means we should reconsider whether to use promise chaining at all here!

src/extensions/scratch3_wedo2/index.js
ericrosenbaum: good catch!

src/util/rateLimiter.js
ericrosenbaum: thanks!

cwillisf (approved):
I know we talked about a few tweaks to this, but it looks good to me so far!

cwillisf (approved):
👍

PR #1550, opened 2018-08-30
Use rate limiter util in microbit extension

ericrosenbaum:
Switch the microbit extension to use the new rate limiter util. It uses a token bucket approach to limit the rate of BLE sends to 20 per second. 

This doesn't fix any known bugs, but it simplifies the logic here in the extension a bit, and makes it more similar to the WeDo 2.0 extension.


evhan55:
I am going to wait on this pending PR to see if we decide to use https://github.com/LLK/scratch-vm/pull/1587 instead.

evhan55:
Keep busy flag implementation for now

Issue #1535, opened 2018-08-28
text2speech extension should trim input to 128 chars


ericrosenbaum:
The server currently ignores input that is longer than this limit, so we should trim anything longer than that before sending.

PR #1551, opened 2018-08-31
Trim input to 128 chars

CosmicWebServices:
### Resolves

https://github.com/LLK/scratch-vm/issues/1535

### Proposed Changes

Trims text in text2speech blocks to be able to successfully be sent to the server

### Reason for Changes

It was an issue created by @ericrosenbaum 

### Test Coverage

Could not find any tests for text2speech blocks

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -286,7 +286,7 @@ class Scratch3SpeakBlocks {
      */
     speakAndWait (args, util) {
         // Cast input to string
-        let words = Cast.toString(args.WORDS);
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
Thanks @CosmicWebServices! This works, but I noticed an issue: the kitten and puppy voices modify the input (replacing each word with e.g. "meow"). This often makes the string a bit longer, which can then cause the problem where the server rejects it because it's longer than 128 chars. Would you mind putting the substring command lower down, just before the request is sent (i.e. before line 311)?

ericrosenbaum:
I'm going to close this one, and go with a new PR: https://github.com/LLK/scratch-vm/pull/1574

Issue #1536, opened 2018-08-28
text2speech voice should be louder


ericrosenbaum:
We should make the sounds produced by the text2speech extension somewhat louder, to match the typical level of Scratch sound library sounds.

Issue #1405, opened 2018-07-31
Text2Speech extension sounds should be stopped by stop button


ericrosenbaum:
### Expected Behavior

The stop button should stop all speech sounds created by the the Text-to-Speech extension.

### Actual Behavior

It doesn't yet!

PR #1554, opened 2018-09-05
Make text2speech louder, and enable stop all

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1536
Resolves https://github.com/LLK/scratch-vm/issues/1405

### Proposed Changes

- Create an effects chain for each speech sound, and set the volume to 250% (which brings it up to a reasonable loudness)
- Use a map to store currently playing speech sounds, and stop all speech sounds when the stop button is pressed

### Reason for Changes

The speech sounds were too quiet, and they should stop when you press stop!

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -61,12 +61,16 @@ class Scratch3SpeakBlocks {
          */
         this.runtime = runtime;
 
-        // @todo stop all speech sounds currently playing
-        // https://github.com/LLK/scratch-vm/issues/1405
-        // this._stopAllSpeech = this._stopAllSpeech.bind(this);
-        // if (this.runtime) {
-        //      this.runtime.on('PROJECT_STOP_ALL', this._stopAllSpeech);
-        // }
+        /**
+         * Map of soundPlayers by sound id.
+         * @type {Map<string, SoundPlayer>}
+         */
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_text2speech/index.js
```diff
@@ -337,10 +350,21 @@ class Scratch3SpeakBlocks {
...
                 this.runtime.audioEngine.decodeSoundPlayer(sound).then(soundPlayer => {
-                    soundPlayer.connect(this.runtime.audioEngine);
+                    this._soundPlayers.set(soundPlayer.id, soundPlayer);
+
                     soundPlayer.setPlaybackRate(playbackRate);
+
+                    // Increase the volume
+                    const engine = this.runtime.audioEngine;
+                    const chain = engine.createEffectChain();
+                    chain.set('volume', 250);
```
thisandagain: Can we put this magic number in a constant?

thisandagain (changes requested):
Nice! Looks great. One nit regarding magic numbers.

src/extensions/scratch3_text2speech/index.js
ericrosenbaum: thanks-all set!

thisandagain (approved):

PR #1562, opened 2018-09-06
ensure getParam operations are boolean when needed

mzgoddard:
Check that a getParam is a boolean in its proccode. If it is make sure
argument_reporter_boolean is used.

### Resolves

Part of #1450.

### Proposed Changes

Store procedure information in the sb2 deserializer so getParam can check the argument type from the proccode. If the argument type is boolean ensure a boolean 

### Reason for Changes

scratch-blocks errors if a non-boolean reporter is in a CONDITION input (like for a control_if). Scratch 2.0 allows boolean arguments to be serialized as `["getParam", "name", "r"]` which turns into a argument_reporter_string_number block, a non-boolean reporter.

### Test Coverage

Loading http://localhost:8601/#10118230 works with this change.

### Additional Information

This is an incomplete solution. Its possible for a getParam operation to be a script without the procDef initial block. That means the needed argument information will not be available for comparison. A different parseState check is needed for that so getParam can see if it is a CONDITION input.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -43,6 +43,9 @@ const CORE_EXTENSIONS = [
 const WORKSPACE_X_SCALE = 1.5;
 const WORKSPACE_Y_SCALE = 2.2;
 
+// Split on %n, %b, %s creating an array of those members of the original string.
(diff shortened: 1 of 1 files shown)


mzgoddard:
Closing in favor of #1563.

PR #1563, opened 2018-09-06
Get param condition input

mzgoddard:
### Resolves

Part of #1450.

### Proposed Changes

Pass parsing state information of the parent block's expectedArg. For getParam blocks ensure argument_reporter_boolean is used if the reporter is in a boolean input like CONDITION or a procedure call boolean input.

### Reason for Changes

scratch-blocks errors when a non-boolean reporter is in a boolean input. Like https://github.com/LLK/scratch-vm/pull/1562, this transforms "r" reporter blocks that Scratch 2 allows into "b" reporter blocks.

This even works for `getParam` blocks that are on scripts missing the procedure definition hat.

### Test Coverage

Loading http://localhost:8601/#10118230 works with this change.


src/serialization/sb2.js
```diff
@@ -1070,8 +1079,22 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             argumentids: JSON.stringify(parseProcedureArgIds(sb2block[1]))
         };
     } else if (oldOpcode === 'getParam') {
+        let returnCode = sb2block[2];
+
+        // Ensure the returnCode is "b" if used in a boolean input.
+        if (parentExpectedArg && (
+            // Used as a CONDITION input like for control_if.
+            parentExpectedArg.inputName === 'CONDITION' ||
```
kchadha: this check seems a little brittle, and it doesn't cover all our boolean cases (for example the not operator input is named OPERAND). The name doesn't actually prescribe the type information. Maybe we should add an additional property in the [sb2_specmap](https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb2_specmap.js) for all the boolean inputs we know about? That seems like an approach that is more consistent with everything else we're doing for sb2 import.

kchadha: hmmm or actually....

I wonder if you can just get away with the following check `parentExpectedArg.type === 'input' && !parentExpectedArg.inputOp`

kchadha (changes requested):
Information about expected boolean inputs should be encoded in the sb2_specmap as discussed in comments and discussions offline.

mzgoddard:
@kchadha Added explicit `boolean` and `substack` inputOp values to the input arguments missing a inputOp member. So all input arguments now have a defined inputOp value.

`boolean` and `substack` are specifically considered not to be shadow values, so the shadow block creation work skips them. And any input argument that may be later added without an inputOp value will log a warning when it is used.

kchadha (approved):
LGTM - Thanks for adding the warning for a missing `inputOp`!

PR #1564, opened 2018-09-07
Rename Stage object and Stage monitors together

### LINKED ISSUE #1322: Project won't load: Error: Cannot create monitor for target...
http://llk.github.io/scratch-vm/index.html#187694931

### Expected Behavior

Project loads and runs

### Actual Behavior

http://llk.github.io/scratch-vm/index.html#187694931 which is used in the vm benchmark suite will not longer load / execute due to this monitor error:

```
sb2.js:262 Uncaught (in promise) Error: Cannot create monitor for target that cannot be found by name
    at parseMonitorObject (sb2.js:262)
    at sb2.js:648
```

### Steps to Reproduce

Load the URL, look at console

Mac / Chrome - but probably everywhere.

mzgoddard:
### Resolves

Fix #1322 
Part of #1450.

### Proposed Changes

When naming the sprite, if the object is renamed to Stage, retarget monitors with the same old name to the new `'Stage'` name.

### Reason for Changes

Monitors referring to the old Stage name, leave monitors that cannot find corresponding sprites which later throws an Error.

### Test Coverage

http://localhost:8601/#187694931 loads with this change.


kchadha (approved):
LG - Thanks for fixing this!

PR #1565, opened 2018-09-07
Extension monitors

### LINKED ISSUE #693: Monitors for extension reporters
### Expected Behavior

It should be possible to monitor the value of an extension reporter.

### Actual Behavior

It's not possible to monitor the value of an extension reporter.

### Steps to Reproduce

1. Load an extension with at least one reporter.
2. Try to activate a monitor for a reporter provided by that extension.
3. can't

cwillisf:
### Resolves

Along with LLK/scratch-gui#3099, this resolves #693 

### Proposed Changes

1. Extension reporters which have no block arguments will get a checkbox in the flyout so the user may enable monitoring of that reporter, unless the extension sets the `disableMonitor` flag for the block.
2. The runtime now has a helper method to determine a user-friendly label for any block registered through the extension system. This is similar to the `OpcodeLabels` function in the `scratch-gui`.

### Reason for Changes

The `OpcodeLabels` function in the GUI currently hard-codes all non-trivial cases. We can't do that for extensions, and the GUI ~doesn't~ shouldn't have direct access to extension block metadata anyway, so instead it delegates calculation of the label to the VM's runtime.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -745,6 +745,9 @@ class Runtime extends EventEmitter {
         case BlockType.REPORTER:
             blockJSON.output = 'String'; // TODO: distinguish number & string here?
             blockJSON.outputShape = ScratchBlocksConstants.OUTPUT_SHAPE_ROUND;
+            if (!blockInfo.disableMonitor) {
+                blockJSON.checkboxInFlyout = true;
+            }
             break;
         case BlockType.BOOLEAN:
             blockJSON.output = 'Boolean';
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1828,6 +1828,31 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Get the label or label function for an opcode
+     * @param {string} extendedOpcode - the opcode you want a label for
+     * @return {object} - object with label and category
+     * @property {string} category - the category for this opcode
+     * @property {Function} [labelFn] - function to generate the label for this opcode
+     * @property {string} [label] - the label for this opcode if `labelFn` is absent
+     */
+    getLabelForOpcode (extendedOpcode) {
+        const [category, opcode] = extendedOpcode.split('_');
+        if (!(category && opcode)) return;
+
+        const categoryInfo = this._blockInfo.find(ci => ci.id === category);
+        if (!categoryInfo) return;
+
+        const block = categoryInfo.blocks.find(b => b.info.opcode === opcode);
+        if (!block) return;
+
+        // TODO: should this use some other category? Also, we may want to format the label in a locale-specific way.
+        return {
+            category: 'data',
+            label: `${categoryInfo.name}: ${block.info.text}`
+        };
+    }
+
+    /**
      * Create a new global variable avoiding conflicts with other variable names.
      * @param {string} variableName The desired variable name for the new global variable.
      * This can be turned into a fresh name as necessary.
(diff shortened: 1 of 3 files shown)


src/engine/runtime.js
```diff
@@ -1824,6 +1828,31 @@ class Runtime extends EventEmitter {
...
+    /**
+     * Get the label or label function for an opcode
+     * @param {string} extendedOpcode - the opcode you want a label for
+     * @return {object} - object with label and category
+     * @property {string} category - the category for this opcode
+     * @property {Function} [labelFn] - function to generate the label for this opcode
+     * @property {string} [label] - the label for this opcode if `labelFn` is absent
+     */
+    getLabelForOpcode (extendedOpcode) {
+        const [category, opcode] = extendedOpcode.split('_');
```
fsih: is the category_opcode format of the string well enforced? Like can the extension define opcode to have underscores?
Maybe category should be added to MonitorRecord?

cwillisf: Hmm... `category_opcode` is indeed enforced (see [here](https://github.com/LLK/scratch-vm/pull/1565/files/8602d5798a4a95f482aad570f7d5965595a9c2a8#diff-38d875e1d50b3158dc96404b65275a54R692)) but it's also possible for the extension author to put `_` in the extension ID (aka "category") or the opcode.

IMO it would make sense to forbid that and enforce it during load by replacing `_` with something else or even removing it, since that's only meant as the machine-readable ID anyway. That said, the conversion between "extended opcode" and category+opcode pair is effectively part of the project JSON spec, so it's not safe to change it unnecessarily.

Another approach would be to build a flat lookup table indexed by extended_opcode, which would mean that we wouldn't need to `split` here and extra `_` characters wouldn't cause any trouble.

Do you think adding the category to MonitorRecord would be a better approach? I'm not very familiar with the monitor framework...

rschamp (changes requested):
This looks good to me, but I think since `getLabelForOpcode` isn't actually used in this repo, it's especially important that we have unit tests so it's clear how it should work.

src/engine/runtime.js
kchadha: I think it's okay to split on `_` here like this, since we rely on that same method for determining the extension in both the sb2 and sb3 deserializers. I *do* think that `.split` here should be replaced with the `splitFirst` function defined in `src/util/StringUtil.js` so that the opcode portion is resilient against extra `_`s.

src/engine/runtime.js
```diff
@@ -1824,6 +1828,31 @@ class Runtime extends EventEmitter {
...
+
+        const categoryInfo = this._blockInfo.find(ci => ci.id === category);
+        if (!categoryInfo) return;
+
+        const block = categoryInfo.blocks.find(b => b.info.opcode === opcode);
+        if (!block) return;
+
+        // TODO: should this use some other category? Also, we may want to format the label in a locale-specific way.
+        return {
+            category: 'data',
```
kchadha, line 1851: why _is_ this using 'data' as the category?

kchadha (changes requested):
The call to `.split` should be replaced with the `splitFirst` util function.

src/engine/runtime.js
fsih: MonitorRecord is just a javascript object with a description of what keys are expected to be in the object. You could add 'category' if you want to have it conveniently available separate from the opcode, or get it from the opcode, whatever's easier.

src/engine/runtime.js
cwillisf: The monitors code in `scratch-gui` uses `data` as the default, and there doesn't seem to be a "correct" category for extensions currently.

test/unit/engine_runtime.js
```diff
@@ -50,7 +50,7 @@ test('monitorStateDoesNotEqual', t => {
     });
     r.requestAddMonitor(prevMonitorState);
     r.requestUpdateMonitor(newMonitorDelta);
-    
+
```
kchadha, line 53: stray newline

cwillisf: Nope: this was a blank line with 4 spaces on it, and now it's a blank line with no spaces on it :)

PR #1569, opened 2018-09-11
Remove branding from translate extension

ericrosenbaum:
- Remove block icon
- Remove the word "Google" from the extension name


src/extensions/scratch3_translate/index.js
```diff
@@ -18,7 +18,7 @@ const serverURL = 'https://translate-service.scratch.mit.edu/';
...
+const iconURI = '';
```
thisandagain: Could we just remove this (plus the `eslint` line above)?

thisandagain (changes requested):

thisandagain (approved):

Kenny2github:
May I ask why this change was necessary?

PR #1579, opened 2018-09-13
Limit answer length

apple502j:
### Resolves
Resolves LLK/scratch-gui#3112

### Proposed Changes
Clamp askAndWait answers to 400 letters

### Reason for Changes
Long string can make Scratch buggy

### Test Coverage
Tested manually on Windows 7 HomePremium 64bit SP1, Firefox 62 via
```
ask [foo] and wait
(answer "hello"*210 = 1050 letters)
say (length of (answer))
```
This returns 400 while current one returns 1050

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -45,6 +45,14 @@ class Scratch3SensingBlocks {
     }
 
     /**
+     * The limit of askAndWait answer.
+     * @const {number}
+     */
+    static get ANSWER_LIMIT () {
+        return 400;
+    }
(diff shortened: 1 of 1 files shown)


chexbox:
That might ruin some existing projects that have long strings like save codes.

towerofnix:
@chexbox agreed!

The issue is in very long speech bubbles, not in the "ask and wait" block, so the limit should be applied in the "say" block's code (well, the underlying speech bubble function).

apple502j:
Okay, I'll close this and change the behavior.

PR #1580, opened 2018-09-13
Send the language code of the project viewer to the server

picklesrus:
The server uses en by default if there is nothing sent. This isn't 100% right yet because it doesn't reconcile when the language of the project and the language of the viewer are different, but it is better than using english all the time...

### DIFF
--- a/src/extensions/scratch3_speech2text/index.js
+++ b/src/extensions/scratch3_speech2text/index.js
@@ -200,6 +200,14 @@ class Scratch3Speech2TextBlocks {
     }
 
     /**
+     * Get the viewer's language code.
+     * @return {string} the language code.
+     */
+    _getViewerLanguageCode () {
+        return formatMessage.setup().locale || navigator.language || navigator.userLanguage || 'en-US';
+    }
+
+    /**
      * Resets all things related to listening. Called on Red Stop sign button.
      *   - suspends audio processing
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_speech/index.js
```diff
@@ -284,6 +284,14 @@ class Scratch3SpeechBlocks {
         return words;
     }
 
+    /**
+     * Get the viewer's language code.
+     * @return {string} the language code.
+     */
+    _getViewerLanguageCode () {
+        return formatMessage.setup().locale || navigator.language || navigator.userLanguage || 'en-US';
```
chrisgarrity: Just wondering if the extension can get the current locale from the VM?
https://github.com/LLK/scratch-vm/blob/e2c6c7667b84e71d790f4a375af83849d50b9116/src/virtual-machine.js#L1015

Not a big deal either way.

chrisgarrity (approved):
Just a question about figuring out the locale. Everything looks good.

src/extensions/scratch3_speech/index.js
picklesrus: I have access to the Runtime, but not the VM which, iiuc, is intentional.

picklesrus:
Fixed merge conflicts (the file had moved), will merge once all the tests run again.

PR #1585, opened 2018-09-14
Bubble string length limit

apple502j:
### Resolves
Fixes LLK/scratch-gui#3112

### Proposed Changes
Strings passed to say or think blocks have length limit now.

### Reason for Changes
Not to show long strings and make the screen ugly.

In Scratch 2, normal stage mode, strings with more than 330 charactors will always overflow, so I choose 330. However, you can make it lower.


thisandagain (approved):
That works! Thanks @apple502j

Issue #1379, opened 2018-07-18
Resolve Global/Local Variable Conflicts from Importing 2.0 Projects


kchadha:
In Scratch 2.0, projects can sometimes have a global and local variable have the same name and type. This breaks monitors and and variable behavior in 3.0.

Resolve these conflicts when importing 2.0 projects:

**Solution:**
- If the stage and a sprite have a variable of the same name and type, rename the sprite variables to avoid conflicts with the stage variable. This maintains 2.0 behavior where the sprites that had the conflict could only access their local variable anyways. Only sprites without a name conflict could access the global variable.

**Repro**
1. Create a local variable in scratch 2 sprite
2. Share that reporter to the stage by dragging
3. Click on that variable block in the stage
4. Open that project in 3.0

PR #1586, opened 2018-09-14
conflict workaround for variables and lists of the same name

mzgoddard:
### Resolves

Successfully loads and runs projects with variables and lists that share a name in the same scope.

### Proposed Changes

Append the variable type (scalar or list) as a suffix in the id and scoped name lookup map.

### Reason for Changes

Currently the id for a scalar variable and a list variable are the same if they share a name. Loading the lists after the variables overwrites the scalar variable stored in the Target `variables` mapping with the list variable. This causes bugs in running variable data blocks for that name and issues rendering the blocks in scratch-blocks, when scratch-blocks tries to look up the variable in the passed variables list. It throws an error when there is a variable data block that shares a variable id with only the list type variable.

Adding the variable type to the variable id distinguishes the scalar variable and list values from each other so that they do not overwrite each other in the Target `variables` mapping.

### Test Coverage

Added two vm execution integration tests with conflicting local and global variables and lists.


kchadha (changes requested):
The code looks good, but I think the test fixtures should be updated so that they only have instances of a variable and list in the same scope in conflict. Currently, both files have exhibit a different issue as well which is that a global and local variable have a name conflict, and a global and local list have a name conflict. That is a different issue, being tracked in #1379 and isn't resolved by this PR.

mzgoddard:
@kchadha replaced the tests with tests that test variable/list name conflict either globally scoped and locally scoped.

kchadha (changes requested):
@mzgoddard, Thanks for splitting out the tests! I think two of the tests (the `-global` ones) are missing a `say end` block, so the tests are failing. After that, I think this PR is good to go.

mzgoddard:
@kchadha I fixed up the new tests. Added the missing end say blocks and fixed the plan say blocks (those were out of date too).

kchadha (approved):

kchadha:
LGTM!

Issue #1547, opened 2018-08-29
Use token bucket style rate limiting in the micro:bit extension


ericrosenbaum:
We should try replacing the current simple rate limiting approach in the micro:bit extension (while sending, drop other messages) with one based on the token bucket strategy, as we did for the WeDo [here](https://github.com/LLK/scratch-vm/pull/1540).

I expect this to improve the behavior of the LED display blocks, especially when rapidly sending them.

thisandagain:
@ericrosenbaum @evhan55 When did you folks want to give this a go?

evhan55:
@thisandagain : @ericrosenbaum and I did some minor triaging today and decided to go for this in November: https://github.com/LLK/scratch-vm/issues/1508.  Let us know if you prefer us to try earlier!

evhan55:
This is duplicated here: https://github.com/LLK/scratch-vm/issues/1508

PR #1587, opened 2018-09-15
Add a rate-limited task queue utility

cwillisf:
### Resolves

This utility code should make it possible to solve #1547 without dropping any messages -- in particular, without dropping animation frames.

### Proposed Changes

This `TaskQueue` class provides a way to rate-limit a series of tasks. The tasks run in the order they are submitted with `do` method and will be rate-limited according to the [token bucket](https://en.wikipedia.org/wiki/Token_bucket) algorithm. The `TaskQueue`'s `do` method returns a promise for the return value of the task.

### Reason for Changes

This version of rate limiting is built around each task resolving a `Promise` for each task as it completes. This offers a way to provide "back pressure" to the VM, preventing tasks from buffering too deeply when the rate limit takes effect. Combined with the `cancelAll` method, this also means that the editor and the stop button should always remain responsive even when running many hardware extension blocks quickly.

### Test Coverage

The new `test/unit/util_task-queue.js` contains tests for the task queue.

### DIFF
--- a/src/util/task-queue.js
+++ b/src/util/task-queue.js
@@ -3,13 +3,17 @@ const Timer = require('../util/timer');
 /**
  * This class uses the token bucket algorithm to control a queue of tasks.
  */
-class TokenBucket {
+class TaskQueue {
     /**
-     * Creates an instance of TokenBucket.
+     * Creates an instance of TaskQueue.
+     * To allow bursts, set `maxTokens` to several times the average task cost.
+     * To prevent bursts, set `maxTokens` to the cost of the largest tasks.
+     * Note that tasks with a cost greater than `maxTokens` will be rejected.
+     *
      * @param {number} maxTokens - the maximum number of tokens in the bucket (burst size).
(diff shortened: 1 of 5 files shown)


evhan55:
I have tested the Task Queue with Max=1000 Rate=1000/sec and with this program:

screen shot 2018-09-18 at 10 55 59 am

and it seems to work well:
- all three shapes display in order and flash by pretty quickly but distinctively
- pressing stop button stops all the display immediately, and LED output shows latest shape sent

This was the code I used:

`this._queue.do(this._ble.write(BLEUUID.service, BLEUUID.txChar, data, 'base64', true));`

[Edit January 2019]:
`TaskQueue.do` requires a function, so the line of code to use the TaskQueue should be:
`this._queue.do(() => this._ble.write(BLEUUID.service, BLEUUID.txChar, data, 'base64', true));`

evhan55 (approved):
Seems good to me!

PR #1588, opened 2018-09-16
Decode at-mark and backslash ( #1307 )

### LINKED ISSUE #1307: Custom blocks import "@" in the label as "\@"
### Expected Behavior

Imported custom blocks should have "@" characters exactly as they are.

### Actual Behavior

"@" characters are replaced with "\\@".

I can edit the custom block and replace the "\\@" with "@", so I don't think this is an issue in scratch-blocks (i.e. scratch-blocks can display "@" characters fine, the problem is they're not being imported from 2.0 projects correctly).

### Steps to Reproduce

Create a 2.0 project with a custom block that has "@" in the label.

I noticed the issue in this project: https://scratch.mit.edu/projects/150986523/ (See blocks like "3d position @ x y z focal").

Firefox 62.0b6, Debian (testing).

apple502j:
### Resolves
Fixes #1307 

### Proposed Changes
Decode `\\@` and `\\\\` using regex and replace. The issue didn't contain the backslash one, but I also found that, so it contains fix for that.

### Reason for Changes
Not to change custom block name.

### Test Coverage
https://scratch.mit.edu/projects/246543955/
This project has 2 custom blocks named `@atmark@` and `\backslash\`

Tested using above with Windows 7, Firefox 62 and Chrome 68

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -1061,7 +1061,8 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             children: [],
             mutation: {
                 tagName: 'mutation',
-                proccode: procData[0].replace(/\\@/g, '@').replace(/\\\\/g, '\\'), // e.g., "abc %n %b %s"
+                proccode: procData[0].replace(/\\@/g, '@').replace(/\\\\/g, '\\')
+                    .replace(/\\%/g, '％'), // e.g., "abc %n %b %s" @todo: It uses fullwidth percent!
                 argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
                 argumentids: JSON.stringify(parseProcedureArgIds(procData[0])),
                 argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -1061,7 +1061,7 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             children: [],
             mutation: {
                 tagName: 'mutation',
-                proccode: procData[0], // e.g., "abc %n %b %s"
+                proccode: procData[0].replace(/\\@/g, '@').replace(/\\\\/g, '\\'), // e.g., "abc %n %b %s"
                 argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
                 argumentids: JSON.stringify(parseProcedureArgIds(procData[0])),
                 argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -1061,8 +1061,8 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             children: [],
             mutation: {
                 tagName: 'mutation',
-                proccode: procData[0].replace(/\\@/g, '@').replace(/\\\\/g, '\\')
-                    .replace(/\\%/g, '％'), // e.g., "abc %n %b %s" @todo: It uses fullwidth percent!
+                proccode: procData[0].replace(/\\%/g, '％').replace(/\\(.)/g, '$1'),
+                // e.g., "abc %n %b %s" @todo: It uses fullwidth percent!
                 argumentnames: JSON.stringify(procData[1]), // e.g. ['arg1', 'arg2']
                 argumentids: JSON.stringify(parseProcedureArgIds(procData[0])),
                 argumentdefaults: JSON.stringify(procData[2]), // e.g., [1, 'abc']


joker314:
This is a nice pull request, but it doesn't account for `%` being encoded.

Scratch 3.0 with a custom block that has a percent symbol
*Take note of the backslash before the final percent sign*
Scratch 2.0 with a custom block that has a percent symbol
*Take note of the absence of a backslash before the final percent sign*

**I haven't tested your PR locally yet, I'm just assuming `%` won't work from looking through the code**. I may be mistaken however.

Note that, per my [issuecomment](https://github.com/LLK/scratch-vm/issues/1307#issuecomment-411915833), only backslash, at, and percent are likely to be affected.

apple502j:
OK, I will decode percent too. Thanks @joker314

joker314:
In Scratch 2.0, the algorithm for escaping the data is not the mirror image of the algorithm for unescaping that data.

This pull request, however, takes the mirror image of the escape function.

This is absolutely fine; and the only time this will not work is if there's 'hacked' JSON.

Somebody may modify their string like so:

```diff
- ask user how they are doing
+ \a\s\k\ \u\s\e\r\ \h\o\w\ \t\h\e\y\ \a\r\e\ \d\o\i\n\g
```

Upon performing such a modification, their project would behave _identically_. There's no reason to make such a change as a JSON hacker (**and so there's no reason to worry about it in this PR, maybe**), however, if such a modification has been made, this pull request would leave the backslashes in. 

I'd say that this is absolutely fine and we shouldn't worry about it.

<hr>

Regarding the chaining of `.replace()`s: 

```js
.replace(/\\([%@\\])/g, '$1')
``` 
might be shorter. I'm not sure if this is more readable? I also don't know how this may impact efficiency -- it's a style choice that you may wish to consider.

This could be modified to resolve the above issue about hacked JSONs like so:

```js
.replace(/\\(.)/g, '$1')
``` 

Though, as stated earlier, the hacked JSON issue seems like a non-issue.

apple502j:
I added percent decoding, but it uses full-width percent to avoid LLK/scratch-blocks#1368

src/serialization/sb2.js
```diff
@@ -1061,8 +1061,8 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             children: [],
             mutation: {
                 tagName: 'mutation',
-                proccode: procData[0].replace(/\\@/g, '@').replace(/\\\\/g, '\\')
-                    .replace(/\\%/g, '％'), // e.g., "abc %n %b %s" @todo: It uses fullwidth percent!
+                proccode: procData[0].replace(/\\%/g, '％').replace(/\\(.)/g, '$1'),
```
joker314: You can remove `.replace(/\\%/g, '％')` because `.` will already match it 🙂

joker314:
Hmm. Maybe I was wrong about us needing to decode `\%` as `%`. If Scratch 3.0 still considers it special, leave it as-is. Note that in my picture, only one of the percents was keeping the backslash: the one which didn't have a character after it. I'm investigating when Scratch 2.0 escapes it and when it doesn't.

<hr>

**EDIT**: Scratch 2.0 will _always_ encode percent symbols with a preceding backslash. Hence, Scratch 3.0 is assigning some percent signs special meanings -- but others not. Likely, if there's a character after the percent sign, it'll become a special token.

<hr>

**EDIT**: I can only seem to get the first percent sign to be treated as something that can be escaped.

joker314:
Okay, some things to note:
- It looks like the first escaped percent sign to appear in the string, will be stripped of its backslash. The rest won't. This leads me to believe that there may be a forgotten `g` flag on a regular expression, elsewhere, and would be out of the scope of this pull request. Perhaps not, though.
- This isn't strictly related to this PR, but it is related to the stuff it's addressing: 
https://github.com/LLK/scratch-vm/blob/64a1d3e02b6aaaac9d6ec5e7299fc13e8199e136/src/serialization/sb2.js#L60
This looks like it would not account for an escaped backslash before a percent sign (i.e., `\\%`), because it'd detect there was a `\%` present and incorrectly ignore it.
<br>Should an issue be filed?

apple502j:
@joker314 It's already help-wanted, LLK/scratch-blocks#1368

joker314:
@apple502j When you made the change to the full-width percent sign, did you actually manage to get a working example of a 2.0 project that would cause LLK/scratch-blocks#1368 to happen?

apple502j:
@joker314 Oops, 1368 was not related at all.

joker314 (commented):
It's probably okay to remove the replacement to the full-width percent sign, then?

kchadha:
@apple502j, Thanks for making this pull request! We are a little backed up on code reviews at the moment, but we'll take a look at these changes soon.

src/serialization/sb2.js
```diff
@@ -1061,7 +1061,8 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
             children: [],
             mutation: {
                 tagName: 'mutation',
-                proccode: procData[0], // e.g., "abc %n %b %s"
+                proccode: procData[0].replace(/\\%/g, '％').replace(/\\(.)/g, '$1'),
```
kchadha, line 1064: Please create a new function with a comment explaining that these symbols are escaped in scratch 2 project jsons and need to be unescaped when importing into 3.0. 

I think it would be best to use and credit [the `unescape` function in the original scratch-flash code](https://github.com/LLK/scratch-flash/blob/74ae0cbc0427d6acc0508e41a73a324f84ebc8c4/src/util/ReadStream.as#L116-L128) as the basis for this change.

kchadha (changes requested):
@apple502j, Thank you for the pull request! I have left a comment in the code changes.

src/serialization/sb2.js
adroitwhiz, line 1064: Note that this is a *fullwidth* percent sign-- "replace this character with another very similar looking character" strikes me as somewhat janky

apple502j:
Hmm, maybe I should make better one

Issue #1338, opened 2018-07-13
micro:bit extension identical display blocks


khanning:
### Expected Behavior

You should be able to tell the display text and display matrix blocks apart when a reporter is dropped on them.

<img width="303" alt="screen shot 2018-07-13 at 9 22 16 am" src="https://user-images.githubusercontent.com/1169507/42693807-49f00bda-867e-11e8-8953-5ae3d6e06fd1.png">

### Actual Behavior

1. Store '10101' in `my variable`
2. Drag out a `display [Hello!]` and `display [♡]` block.
3. Drop `my variable` on both blocks
4. The display text block scrolls the text '10101' on the micro:bit display
5. The display matrix block lights up every other LED in the first row of the micro:bit display

### Steps to Reproduce

1. Load the micro:bit extension
2. Drag out a `display [Hello!]` and `display [♡]` block.
3. Drop any reporter on both blocks
4. They now look identical

### Operating System and Browser

all

kyleplo:
Maybe rename to "display text []" and "display image [❤️]"

ericrosenbaum:
After discussion we've decided to keep the first one as is, and change the second one to "display text"

PR #1597, opened 2018-09-19
Rename block "display [Hello!]" to "display text [Hello!]"

khanning:
### Resolves

Issue #1338, micro:bit extension identical display blocks

### Proposed Changes

Rename `display [Hello!]` block to `display text [Hello!]`

image

### Reason for Changes

Avoids potential confusion when a block is dropped onto the input field of the `display [♡]` and `display [Hello!]` blocks.

image


ericrosenbaum (approved):
LGTM!

Issue #1339, opened 2018-07-13
Trim whitespace in micro:bit display matrix block


khanning:
### Expected Behavior

When dropping a list on the micro:bit display matrix block, the whitespace added between each item should be automatically trimmed. This way you could maintain a list of 5 items (1 for each row of LEDs) and send it to the micro:bit display by dropping the list on the `display [♡]` block.

<img width="108" alt="screen shot 2018-07-13 at 9 26 20 am" src="https://user-images.githubusercontent.com/1169507/42694079-26958100-867f-11e8-8308-68a55597d4b5.png">
<img width="143" alt="screen shot 2018-07-13 at 9 26 31 am" src="https://user-images.githubusercontent.com/1169507/42694088-2c596980-867f-11e8-8256-b984a549b21c.png">

Should the whitespace be automatically trimmed? Or should it be interpreted as a 0 (i.e. LED off)?

### Actual Behavior

If a list item contains more than one character, a blank space is automatically added to the reporter return value. The display matrix block interprets this blank space as a character and lights up the LED accordingly.

You can maintain a list of 25 items (1 for each LED) and as long as each item only contains 1 character (0 or 1) it will work as expected with the display matrix block.

### Steps to Reproduce

1. Load the micro:bit extension
2. Create a list with five items that each contain a combination of five 1's or 0's.
3. Drop the list reporter on the `display [♡]` micro:bit block and run.
4. The micro:bit interprets the blank spaces as characters shifting everything over.

### Operating System and Browser

all

paulkaplan:
Additionally, i think the line `if (!hex) return;` is not right, you should be able to write `0`

khanning:
Closed with #1598

PR #1598, opened 2018-09-19
Trim whitespace in "display [♡]" input field

khanning:
### Resolves

Issue #1339, Trim whitespace in micro:bit display matrix block

### Proposed Changes

Use `.replace(/\s/g, '')` to remove all whitespace from the `display [♡]` input field.

### Reason for Changes

This change makes it easy to use a variable list for maintaining the state of the LED matrix. 

#### List of 5 items, 1 for each row:
image
image

#### List of 25 items, 1 for each LED
image
image


ericrosenbaum (approved):
LGTM!

Issue #1340, opened 2018-07-13
micro:bit display text block wait until finished


khanning:
### Expected Behavior

The micro:bit extension `display [Hello!]` block should wait until the text is done printing on the micro:bit display.

The MicroBitDisplay driver has an [event notification](https://github.com/lancaster-university/microbit-dal/blob/master/source/drivers/MicroBitDisplay.cpp#L316) for when the print command has finished so we could easily make the block wait until it's done printing.

On the other hand, it is nice to be able to interrupt the display text block and you can implement your own wait after it.

<img width="140" alt="screen shot 2018-07-13 at 9 42 15 am" src="https://user-images.githubusercontent.com/1169507/42694621-0c7c9734-8681-11e8-9843-469c8d4cbf5b.png">

### Actual Behavior

`display [Hello!]` will wait for 1 second regardless of the input length.

### Steps to Reproduce

1. Load the micro:bit extension
2. Drag out a forever loop and drop in a `display [Hello!]` block
3. Run the loop
4. The display text block executes every seconds, restarting the micro:bit print function.

### Operating System and Browser

all

ericrosenbaum:
@khanning will proceed with this one. It will require a new microbit firmware version, so to maintain backward compatibility, we may need to introduce a new opcode for a new version of this block with the new behavior? Or! Maybe! the VM can actually predict the exact timing of the text scrolling and just yield for that amount of time.

khanning:
Closed with #1599

PR #1599, opened 2018-09-19
Make "display [Hello!] block yield until printing is done

khanning:
### Resolves

Issue #1340, micro:bit display text block wait until finished

### Proposed Changes

Make the `display [Hello!]` block yield until the scrolling text has finished printing.

The delay time is calculated in the scratch-vm by multiplying the number of horizontal pixels by the default micro:bit scroll delay time (120ms). This eliminates the need for making a firmware level change.

Each char has a 5px width + a 1px spacer. The micro:bit DAL also adds a 1px spacer before printing the first character.

Delay time can be calculated with the following formula:
`DELAY_TIME = (120 * ((6 * TEXT_LENGTH) + 1)))`

If a second `display [TEXT]` block runs while another is printing, the new text will interrupt and begin printing immediately. The first block will still only yield when it originally expected to finish printing.

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -812,16 +812,18 @@ class Scratch3MicroBitBlocks {
      * Note the limit is 19 characters
      * The print time is calculated by multiplying the number of horizontal pixels
      * by the default scroll delay of 120ms.
-     * # of horizontal pixels = 5 + 1 for each char with an additional 1px spacer.
+     * The number of horizontal pixels = 6px for each character in the string,
+     * 1px before the string, and 2px after the string.
      */
     displayText (args) {
         const text = String(args.TEXT).substring(0, 19);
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
Nice! It looks like the delay is just slightly too short. From what I can tell, the yield finishes at the moment the final character has reached its centered position on the display. It still has three or four 120ms frames remaining in its animation to make it off the edge. One side effect of this is that if you stack two "display text" blocks, the first one will not quite complete- the last character will disappear instead of scrolling off, and then the second one starts.

khanning:
Nice catch @ericrosenbaum! I had to add 5 additional 120ms frames to allow the largest characters (e.g. W) to clear the display. 

In this example, the first block now fully clears the display before the second one begins.
image

There is a very slight gap between the two display commands, but it's much better than cutting the first string short.

ericrosenbaum (approved):
LGTM!

Issue #1559, opened 2018-09-06
Fix name of micro:bit block "when pin connected"


ericrosenbaum:
### Expected Behavior

The block text should be "when pin 0 connected"

### Actual Behavior

The block currently says "when pin 0 connected test"

khanning:
Closed with #1600

PR #1600, opened 2018-09-19
Remove the word "test" from 'when pin connected' block

khanning:
### Resolves

Issue #1559, Fix name of micro:bit block "when pin connected"

### Proposed Changes

Delete the word "test" from the `when pin connected` block.

### Reason for Changes

It shouldn't say test 😜


ericrosenbaum (approved):
LGTM!

PR #1604, opened 2018-09-20
Add event to show and hide a mic indicator for Speech to Text

ericrosenbaum:
### Proposed Changes

- Add an event emitted by the runtime to indicate when an extension starts or stops streaming audio from the microphone to a server, so that the GUI can show or hide a microphone indicator icon.
- Use the event in the Speech to Text extension.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -458,6 +458,15 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Event name for requesting an update of the indicator that shows that the
+     * microphone is being used to stream audio.
+     * @const {string}
+     */
+    static get UPDATE_MIC_INDICATOR () {
+        return 'UPDATE_MIC_INDICATOR';
+    }
+
+    /**
      * Event name for reporting that blocksInfo was updated.
      * @const {string}
      */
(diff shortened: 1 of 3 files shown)


src/engine/runtime.js
```diff
@@ -457,6 +457,15 @@ class Runtime extends EventEmitter {
         return 'PERIPHERAL_SCAN_TIMEOUT';
     }
 
+    /**
+     * Event name for requesting an update of the indicator that shows that the
+     * microphone is being used to stream audio.
+     * @const {string}
+     */
+    static get UPDATE_MIC_INDICATOR () {
+        return 'UPDATE_MIC_INDICATOR';
```
rschamp: `UPDATE_MIC_INDICATOR: true/false` is a little ambiguous in what it should indicate (I think this name might also be too closely tied to the UI). I would suggest something like `MIC_LISTENING: true/false` or two events, `MIC_LISTENING_ON`/`MIC_LISTENING_OFF`.

rschamp (changes requested):

rschamp (approved):

rschamp (approved):

PR #1607, opened 2018-09-22
Show customized alerts on hardware extension peripheral errors.

evhan55:
### Resolves

This is progress towards: https://github.com/LLK/scratch-gui/issues/2956.

### Proposed Changes

Shows customized alerts for peripheral errors/disconnections including extension icon and name.

screen shot 2018-09-22 at 5 40 22 pm

### Reason for Changes

To warn the user that a peripheral has disconnected and blocks may no longer work.

### Notes
- Requires this PR to land: https://github.com/LLK/scratch-gui/pull/3209
- Sometimes it will show two errors for a single hardware malfunction, but that's to fix in a future PR.

### DIFF
--- a/src/io/ble.js
+++ b/src/io/ble.js
@@ -174,8 +174,9 @@ class BLE extends JSONRPCWebSocket {
     _sendError (/* e */) {
         this.disconnect();
         // log.error(`BLE error: ${JSON.stringify(e)}`);
+        console.log('extension id sending: ' + this._extensionId);
         this._runtime.emit(this._runtime.constructor.PERIPHERAL_ERROR, {
-            message: `Scratch lost connection to ${this._peripheralType}.`,
+            message: `Scratch lost connection to`,
             extensionId: this._extensionId
         });
     }
--- a/src/io/ble.js
+++ b/src/io/ble.js
@@ -8,10 +8,11 @@ class BLE extends JSONRPCWebSocket {
      * A BLE peripheral socket object.  It handles connecting, over web sockets, to
      * BLE peripherals, and reading and writing data to them.
      * @param {Runtime} runtime - the Runtime for sending/receiving GUI update events.
+     * @param {string} peripheralType - the type of peripheral.
      * @param {object} peripheralOptions - the list of options for peripheral discovery.
      * @param {object} connectCallback - a callback for connection.
      */
-    constructor (runtime, peripheralOptions, connectCallback) {
+    constructor (runtime, peripheralType, peripheralOptions, connectCallback) {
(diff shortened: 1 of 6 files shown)


src/io/bt.js
```diff
@@ -115,7 +117,10 @@ class BT extends JSONRPCWebSocket {
     _sendError (/* e */) {
         this.disconnect();
         // log.error(`BT error: ${JSON.stringify(e)}`);
-        this._runtime.emit(this._runtime.constructor.PERIPHERAL_ERROR);
+        this._runtime.emit(this._runtime.constructor.PERIPHERAL_ERROR, {
+            message: `Scratch lost connection to`,
+            extensionId: this._extensionId
```
rschamp, line 122: I think we might want to define this message in the GUI and only pass `extensionId` as the data of this. We might want the message to be different in different contexts.

src/io/ble.js
```diff
@@ -172,7 +174,10 @@ class BLE extends JSONRPCWebSocket {
     _sendError (/* e */) {
         this.disconnect();
         // log.error(`BLE error: ${JSON.stringify(e)}`);
-        this._runtime.emit(this._runtime.constructor.PERIPHERAL_ERROR);
+        this._runtime.emit(this._runtime.constructor.PERIPHERAL_ERROR, {
```
rschamp, line 177: Maybe we should just have an API for this, so we can just use `this._runtime.emitPeripheralError(this._extensionId)`?

rschamp (changes requested):

src/io/bt.js
evhan55: Yes, this suggestion feels better to me, I wasn't sure what that might look like on the GUI side.

src/io/ble.js
evhan55: An API like this sounds good, maybe for another pass that includes localization, like you said?

rschamp (approved):
Like we discussed off-thread, let's address these in a follow-up.

evhan55:
Great!  Will wait for https://github.com/LLK/scratch-gui/pull/3209 to land to merge this.  Thanks.

rschamp:
I think you want to merge in the other order, don't you? After you land this one, wait for the new `scratch-vm` version to be published, then bump to that version of `scratch-vm` in  LLK/scratch-gui#3209 and merge that, so `scratch-gui` builds with the version of `scratch-vm` it needs?

evhan55:
Oh, I see! I had not done that process yet myself, but I will try it now for these PRs, thanks.

PR #1610, opened 2018-09-24
Add “set language” block for the Text to Speech extension

ericrosenbaum:
As we progress toward making all three language extensions (text to speech, speech to text, translate) work well with each other, we're trying out different ways to make sure they work together across different languages. This PR adds a block to set the language for speech synthesis in the text to speech extension to one of the languages available from the service we are currently using. 

If we decide to go with this approach, there will be some additional chunks of work, including:

- Decide how to provide the list of available languages, instead of the way it is done here (e.g. possibly generate it with a module)
- Localize the language names that appear in the menu (using something like our existing scratch-translate-extension-languages module)
- Do the conversion between language codes on the server instead of in the extension (i.e. between two letter codes used by Scratch like 'en' and the four letter codes like 'en-US')
- Set the language on load using the Scratch locale, if it matches an available speech language (note that this may require a more sophisticated strategy, such as saving the project creator's language in the project, to handle a case where the project creator does not use a "set language" block, and it is opened by a user viewing Scratch in another language)

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -63,6 +63,12 @@ class Scratch3Text2SpeechBlocks {
         this.runtime = runtime;
 
         /**
+         * The current language code to use for speech synthesis.
+         * @type {string}
+         */
+        this.currentLanguage = 'en-US';
+
+        /**
          * Map of soundPlayers by sound id.
          * @type {Map<string, SoundPlayer>}
          */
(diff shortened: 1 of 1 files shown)


thisandagain:
- It seems like this list should come from the server (in the same way we get the list of languages for the translate extension)
- This is also handled in the translate extension by the server which is nice. :-)
- Happy to help with this if you file an issue (if you haven't already) 👍 
- 👍

thisandagain (approved):

PR #1612, opened 2018-09-25
Reset the supported language list in getInfo. This works because refr…

### LINKED ISSUE #1365: Languages dropdown menu in translate extension doesn't update when language is switched
### Expected Behavior

If the Google Translate extension is loaded when the user switches to another locale in gui, the list of languages should update to show the list of languages in the new locale

### Actual Behavior

If you load the extension while in English all the languages in the menu are listed in English. When you switch locales in the GUI, the list of languages stays in English.

### Steps to Reproduce

Load the translate extension, then switch to another language for the GUI.

picklesrus:
…eshBlocks is called

when the language gets changed. Fixes #1365.

### Resolves

#1365 

### Proposed Changes

Refresh the language list whenever getInfo is called.
I know this isn't the preferred place to fix this, but I put together the PR mostly to talk about how we'd want to handle these types of things in extensions. Right now setLocale eventually eventually calls getInfo so it works :)

### DIFF
--- a/src/extensions/scratch3_translate/index.js
+++ b/src/extensions/scratch3_translate/index.js
@@ -31,7 +31,7 @@ class Scratch3TranslateBlocks {
          * @private
          */
         this._viewerLanguageCode = this.getViewerLanguageCode();
-        
+
         /**
          * List of supported language name and language code pairs, for use in the block menu.
          * Filled in by getInfo so it is updated when the interface language changes.
@@ -116,8 +116,6 @@ class Scratch3TranslateBlocks {
                             type: ArgumentType.STRING,
                             menu: 'languages',
                             defaultValue: this._randomLanguageCode
-
-
                         }
                     }
                 },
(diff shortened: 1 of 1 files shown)


picklesrus:
Aww. I seem to be running a different lint configuration than Travis. Will sort that out.

picklesrus:
I still haven't sorted out the lint config/versioning issues, but I updated my branch so at least my line numbers match up well enough to use travis logs to fix the errors.

chrisgarrity (approved):
Looks good, but...
I think we do need to discuss when we expect service extensions to 're-initialize'. Obviously, we re-run getInfo on a language change, so an extension can rely on that. Are there other use-cases where a service might need to reset something?

cwillisf (approved):
👍
Someday we might be able to handle much of this by providing a function to implement the `languages` menu dynamically, but I think this approach makes sense right now.

PR #1613, opened 2018-09-26
Old MIDI drum block support, featuring approximations via pitch shift

### LINKED ISSUE #1424: Projects with old 1.4 drum block are mangled

towerofnix:
### Resolves

Fixes #1424. Sort of a follow up to #1329 and #355.

This also (not) accidentally allows the "set (audio) effect" block's "pitch" to have an effect on drums. I figure this is okay since the pan left/right effect already worked; and it's necessary to have the MIDI drum approximations.

### Proposed Changes

Adds support for the obsolete `drum:duration:elapsed:from:` block, including very similar MIDI drum approximations to the ones originally created [for Scratch 2.0](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/sound/SoundBank.as#L339-L388).

### Reason for Changes

Compatibility with old Scratch projects (and new ones which use that block).

### Test Coverage

Tested manually by confirming that:

* The examples in [this project](https://scratch.mit.edu/projects/248475643/) are all correct.
* @as-com's ["Canyon.mid"](https://scratch.mit.edu/projects/115797375/) project runs correctly. (It's *beautiful!*)

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -610,7 +610,7 @@ class Scratch3MusicBlocks {
 
     /**
      * An array that is a mapping from MIDI drum numbers in range (35..81) to Scratch drum numbers.
-     * @type {Array[Array]} an array of information about the drums, in the format [drumNum, pitch, decay].
+     * @type {Array[]} an array of information about the drums, in the format [drumNum, pitch, decay].
      */
     get MIDI_DRUMS () {
         return [
@@ -938,6 +938,8 @@ class Scratch3MusicBlocks {
      * the MIDI to Scratch drum mapping.
      * @param {number} drumNum - the drum number.
      * @param {beats} beats - the duration in beats to pause after playing the sound.
+     * @param {object} util - utility object provided by the runtime.
+     * @param {boolean} mapMidi - whether or not drumNum is a MIDI drum number.
      */
     _playDrumForBeats (drumNum, beats, util, mapMidi) {
         if (this._stackTimerNeedsInit(util)) {
@@ -1002,7 +1004,7 @@ class Scratch3MusicBlocks {
 
         // Dirty hack to make the pitch effect work - otherwise the effect isn't applied, for some reason.
         // (Other audio effects work fine without this, though.)
-        const pitchEffect = chain._effects.find(effect => effect.name === 'pitch')
+        const pitchEffect = chain._effects.find(effect => effect.name === 'pitch');
         if (pitchEffect) {
             pitchEffect.updatePlayer(player);
         }
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -799,7 +799,8 @@ class Scratch3MusicBlocks {
                             type: ArgumentType.NUMBER,
                             defaultValue: 0.25
                         }
-                    }
+                    },
+                    hideFromPalette: true
                 },
                 {
                     opcode: 'restForBeats',
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -1006,6 +1006,10 @@ class Scratch3MusicBlocks {
         });
 
         player.play();
+        // Connect the player to the gain node.
+        player.connect({getInputNode () {
+            return volumeGain;
+        }});
     }
 
     /**
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -613,6 +613,7 @@ class Scratch3MusicBlocks {
 
     /**
      * An array that is a mapping from MIDI drum numbers in range (35..81) to Scratch drum numbers.
+     * The pitch and decay properties are not currently being used.
      * @type {Array[]} an array of information about the drums, in the format [drumNum, pitch, decay].
      */
     get MIDI_DRUMS () {
@@ -940,7 +941,7 @@ class Scratch3MusicBlocks {
         if (midiDescription) {
             drumNum = midiDescription[0];
         } else {
-            drumNum = 2;
+            drumNum = 2; // Default instrument used in Scratch 2.0
         }
         drumNum += 1; // drumNum input to _playDrumForBeats is one-indexed
         this._playDrumForBeats(drumNum, args.BEATS, util);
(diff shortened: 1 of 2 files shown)


src/extensions/scratch3_music/index.js
```diff
@@ -608,6 +608,62 @@ class Scratch3MusicBlocks {
         ];
     }
 
+    /**
+     * An array that is a mapping from MIDI drum numbers in range (35..81) to Scratch drum numbers.
+     * @type {Array[Array]} an array of information about the drums, in the format [drumNum, pitch, decay].
```
towerofnix: Decay data is included, but there isn't currently a decay effect in scratch-audio, so I can't do anything with it.

src/extensions/scratch3_music/index.js
```diff
@@ -722,6 +778,26 @@ class Scratch3MusicBlocks {
                         }
                     }
                 },
+                {
+                    opcode: 'midiPlayDrumForBeats',
+                    blockType: BlockType.COMMAND,
+                    text: formatMessage({
+                        id: 'music.midiPlayDrumForBeats',
+                        default: 'play drum [DRUM] for [BEATS] beats',
```
towerofnix, line 791: Just a thought that came to mind - should there be some more obvious way to distinguish between the two "play drum" blocks? Presently the two blocks look totally identical if their DRUM slots are filled (because the only way to tell the difference is to see if it's a dropdown or a number input). Not really a big deal since this is an obsolete block, though.

src/extensions/scratch3_music/index.js
```diff
@@ -843,14 +919,46 @@ class Scratch3MusicBlocks {
...
+            drumNum = Cast.toNumber(drumNum);
+            drumNum = Math.round(drumNum);
+            let pitchShift = 0;
+            if (mapMidi) {
+                const midiDescription = this.MIDI_DRUMS[drumNum - 35];
+                if (midiDescription) {
+                    drumNum = midiDescription[0];
+                    pitchShift = midiDescription[1] * 10; // 10 = one semitone
+                } else {
+                    drumNum = 2;
```
towerofnix: This is the default instrument which plays when no instrument is found in 2.0.

src/extensions/scratch3_music/index.js
```diff
@@ -884,9 +993,20 @@ class Scratch3MusicBlocks {
...
+        }
+        chain.setEffectsFromTarget(soundEffects);
         player.connect(chain);
 
+        // Dirty hack to make the pitch effect work - otherwise the effect isn't applied, for some reason.
+        // (Other audio effects work fine without this, though.)
+        const pitchEffect = chain._effects.find(effect => effect.name === 'pitch')
+        if (pitchEffect) {
+            pitchEffect.updatePlayer(player);
+        }
```
towerofnix: This is a hack, of course, because it directly searches for `pitchEffect` by accessing `chain._effects`. I couldn't figure out any other way to make the effect apply correctly, though.

ericrosenbaum:
Thanks for this @towerofnix ! I'm excited to give this a look- due to other work it may have to wait a week or more.

ericrosenbaum:
Okay this is working well, but I'd like to make some structural changes, and maybe remove the use of the pitch effect (for simplicity). If it's okay with you I will push some changes directly to the PR branch (assuming I can figure out how!).

update: I seem to have got into some git trouble. Looking into it..

towerofnix:
@ericrosenbaum Okay! That's fine with me. Good luck fixing the git issues.. always a pain, of course :P

(In the uber-worst case scenario that this PR is utterly irredeemable, I have the commits separately on my computer - so it's easy to re-push/PR that if necessary.)

ericrosenbaum:
@towerofnix Okay, @paulkaplan helped me fix the branch. Phew! But now that it has been force-pushed from my side, you won't be able to push correctly to it from your existing local branch. I can take over from here (just moving a few things around). If you do want to push any changes, I think you would just have to rename your local branch, then pull this one down, make the changes there, and push from it. Sorry again about the confusion (I learned something!).

towerofnix:
@ericrosenbaum Alright, sounds good! And no problem! (:stars: \~The more you know~ :stars:)

Good to know - I've fixed up my own branch so it'll push/pull fine. I'm cool with you taking over from here. Just one thought:

> I'd like to [..] maybe remove the use of the pitch effect (for simplicity).

I'm okay with this, but I think it is a pretty important part of simulating the sounds - several sounds just don't sound the same when their pitches aren't right. It's a lot of constant numbers which we can't be 100% sure are all exactly correct, but I think some effort will make them sound a lot better than none. There's not a lot of pressure to get them perfect either, since 1.4 MIDI-drum projects are relatively rare. Thoughts?

Aside - there are a few "decay" values which can be removed from that constants array, since I don't have any code for the feature and it would probably take a significant amount of effort to implement.

as-com:
> Aside - there are a few "decay" values which can be removed from that constants array, since I don't have any code for the feature and it would probably take a significant amount of effort to implement.

I think decay is very important; if it's not a ridiculous amount of effort to implement, it should be done (since some of my projects just don't sound right without the shorter decays of some instruments).

ericrosenbaum:
Okay, I've now made the changes. The main thing is to keep the legacy Scratch 1.x-related code more separate from the rest of the code- so I just moved the drum number mapping into midiPlayDrumForBeats. 

I recently fixed a couple of other bugs in the music extension by removing the use of the effects chain (https://github.com/LLK/scratch-vm/pull/1735). This prevents us from using the pitch effect, so we will not be able to get the sound variations between the different 1.4 drums. I appreciate the clever work on this, but going forward we probably won't be able to use it.

Thanks for your help @towerofnix !

ericrosenbaum:
Hm. In another hilarious github-related plot-twist, the diffs are not showing up correctly (apparently this is sometimes an issue on branches that have been force-pushed). I can't figure out a way to fix this PR, so I'm just going to create a new one.

ericrosenbaum:
Replaced by https://github.com/LLK/scratch-vm/pull/1790

Issue #1247, opened 2018-06-18
Execution order of original sprite and clones is backwards


thisandagain:
### Expected Behavior
The order of execution for sprites and their clones should match Scratch 2.0

### Actual Behavior
The order of execution for sprites and their clones is the opposite of Scratch 2.0

### Reference (HT [Mr-Dave](https://scratch.mit.edu/users/Mr-Dave/))

A clone will execute events before the original sprite in Scratch 3.0 but the opposite is true for Scratch 2. [Mr-Dave](https://scratch.mit.edu/discuss/topic/299791/?page=181#post-3143035) noticed the problem in these projects:

*Tower Defense*
https://scratch.mit.edu/projects/71816138/
https://preview.scratch.mit.edu/#71816138

*Minimal Repro Case*
https://scratch.mit.edu/projects/229249805/
https://preview.scratch.mit.edu/#229249805

### Operating System and Browser
Windows 7, Firefox 60.0, Flash 30.0 (release 0)

thisandagain:
Note: I did very little investigation here aside from repro via the projects listed. This will need a little digging.

towerofnix:
Hat blocks are started by [`startHats`](https://github.com/LLK/scratch-vm/blob/6e2af4f3cc336d20173ab7725efc89a71cd113f4/src/engine/runtime.js#L1117-L1208), which calls [`allScriptsDo`](https://github.com/LLK/scratch-vm/blob/6e2af4f3cc336d20173ab7725efc89a71cd113f4/src/engine/runtime.js#L1094-L1115) and [pushes threads](https://github.com/LLK/scratch-vm/blob/6e2af4f3cc336d20173ab7725efc89a71cd113f4/src/engine/runtime.js#L1204-L1205) according to the order that `allScriptsDo` works.

[...]

PR #1617, opened 2018-09-26
Sprite clone and initial draw order execution order

mzgoddard:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1247

### Proposed Changes

- Add execution order execute integration tests
- Add a Runtime.executableTargets list

`executableTargets` lists the targets in the same order as the renderer's draw list. In other words it is order in the reverse that the targets are executed.

### Reason for Changes

In Scratch 2 the order sprites are drawn is the reverse of the order the sprite's scripts are executed in.

The targets list is also the order the sprites are listed in the gui and serialized to the sb3 format. As in the order sprites are added to the project. The VM needs to maintain two orders for the sprites, the order the project created the sprites and the order the sprites are executed in. Since the VM cannot track both of those orders in one list I've added a new list to track the order.

### Test Coverage

3 new integration tests. 2 that test that the draw order and execution order match based of how the sprites are loaded from the sb2 and sb3 formats. 1 test that also creates clones of its sprites and ensures they are ordered correctly.


mzgoddard:
I'm happy to rename executableTargets, its mostly a placeholder I decided to use so I could implement a version of this.

kchadha (commented):
@mzgoddard, could you investigate the test failures on travis?

mzgoddard:
@kchadha last travis issue is fixed by https://github.com/LLK/scratch-vm/pull/1618.

src/sprites/rendered-target.js
```diff
@@ -943,6 +943,9 @@ class RenderedTarget extends Target {
                 other.drawableID, 0, StageLayering.SPRITE_LAYER, true);
             this.renderer.setDrawableOrder(this.drawableID, otherLayer, StageLayering.SPRITE_LAYER);
         }
+
+        const executionPosition = this.runtime.moveExecutable(other, 0);
```
kchadha: why not:
`const executionPosition = this.runtime.executableTargets.indexOf(other)` ?

You can even add a call to the math util clamp function if you want to ensure that the returned index is within the bounds of the list...

mzgoddard: My first thought was to have a similar API to `RenderWebGL`'s setDrawableOrder without the boolean trap. I think your suggestion here would be <del>fine</del><ins>good</ins> since we're not crossing a package boundary.

mzgoddard:
The tests are passing. I don't know why it still thinks the tests are pending.

src/sprites/rendered-target.js
```diff
@@ -1142,6 +1145,7 @@ class RenderedTarget extends Target {
     dispose () {
         this.runtime.changeCloneCounter(-1);
         this.runtime.stopForTarget(this);
+        this.runtime.removeExecutable(this);
```
kchadha, line 1148: My only other question was.. is it possible for both calls to `this.runtime.removeExecutable(this)` to get called (e.g. this one and the one in `engine/target`), and if so, is that a bad thing or a good thing...

I haven't gotten a chance to investigate this yet, but just wanted to write it down for you in case you have any immediate thoughts on it. I'll try to look into this a bit tomorrow.

mzgoddard: It shouldn't be possible for them to both be called without `super.dispose();` being added here in `RenderedTarget`. If that was added and they were both called, `removeExecutable` is already written in a way to safely perform the removal if the target was already removed. (It'll check for existence, if it doesn't exist in the list, it won't try to alter the list.)

kchadha: Ah, yep that makes sense! Thanks for talking that through.

kchadha (approved):
LGTM!

Issue #1053, opened 2018-04-13
Execution Order of Sprites does not change with ordering


griffpatch:
### Expected Behavior

As encountered in #457 the execution order of scripts across multiple sprites should be in z-index order of the sprites from front to back. Changing the order of the sprites and then re-running the project, or changing them mid execution should cause the execution order to change (note: only for scripts that are not already running - these should maintain the original execution order - it will be worth keeping this in mind when implementing a fix - So if you have a number of forever loops already in progress, the order should not be affected by re-ordering them, but on stopping and re-starting the order will then be updated).

### Actual Behavior

In Scratch 3, the order of execution is correct at the time of loading as long as the order is not changed either by the user, or during execution. However, any change in order does not get reflected at runtime and the execution order stays as it was on project load / creation.

### Steps to Reproduce

This project randomly shuffles three sprites during runtime and 'says' the order of execution (1 being first, then 2 & 3).

https://llk.github.io/scratch-gui/develop/#216001494

In Scratch 2 the numbers continually shuffle around, but in Scratch 3 they stay constant.

### Operating System and Browser

Windows 10 PC - Chrome

PR #1619, opened 2018-09-27
Execution order 2

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-vm/issues/1053

### Proposed Changes

Builds on top of #1617.

Update execution order from look instructions. 

### Reason for Changes

Draw order is the reverse of execution order. Update the execution order so it matches the draw order and execute new scripts in the reverse direction.

### Test Coverage

New execute integration tests for changing order and if the scripts were already executing or not.


kchadha (commented):
These changes don't seem to resolve the issue in the project mentioned in #1053. The sprites in the project (216001494) should be constantly shuffling which numbers they're saying, but they stay constant e.g. 

2.0 Behavior:

execution_order_2 0

vs. 

3.0 Behavior

execution_order_3 0

mzgoddard:
@kchadha Ah. That project is encountering an extra detail. <del>`_restartThreads` doesn't honor the new sprite order and maintains thread position.</del>

mzgoddard:
Or rather the threads are not allowed to die before restarting.

kchadha (approved):
Looks good! Thank you for all the tests and the `broadcast and wait` fix.

PR #1630, opened 2018-10-02
Hardware extensions: fixes and menu localization

### LINKED ISSUE #1485: WeDo2/micro:bit block menus need localization

### LINKED ISSUE #1505: WeDo2 motor power between 0-30ish doesn't power the motor

### LINKED ISSUE #1592: WeDo2 and EV3 should not send motor command when motor power is 0
### Expected Behavior

In the WeDo2 and EV3 extensions, when the power level for a motor is set to 0, motor commands for that motor should not be sent. The opcode function should bail out (but in cases where the block waits for some amount of time, it should still wait).

### Actual Behavior

We send motor commands with zero power, which can cause the motors to make a whining sound.

### LINKED ISSUE #1615: Prevent multiple hardware disconnect alerts
### Expected Behavior

One alert is shown when hardware peripheral disconnects.

### Actual Behavior

In some cases, two, redundant alerts display.

### Cases to Consider:

- Preventing the VM from requesting the alert twice- it should only be when the socket closes
- Preventing the alert from getting sent if we were not already connected
- What about if you connect, disconnect, connect, disconnect. Okay to see two alerts? Lets ask Carl and Kathy

evhan55:
### Resolves

Resolves #1592: WeDo2 and EV3 should not send motor command when motor power is 0.

Resolves #1505: WeDo2 motor power between 0-30ish doesn't power the motor.
- Sometimes the WeDo2 still stalls at low numbers (no repro available), might be a separate issue?

Resolves #1485: WeDo2/micro:bit block menus need localization
- micro:bit menus were already localized, but I refactored the menu names to include the extension name to match other hardware extensions

Resolves #1615: Prevent multiple hardware disconnect alerts
- multiple alerts no longer show, but #1638 still needs to be addressed

These are all meant to work with Scratch Link version no. `1.18.827.1749`

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1219,7 +1219,7 @@ class Scratch3WeDo2Blocks {
                         text: formatMessage({
                             id: 'wedo2.tiltDirection.any',
                             default: 'any',
-                            description: 'label for any element in tilt direction any menu for LEGO WeDo 2 extension'
+                            description: 'label for any element in tilt direction menu for LEGO WeDo 2 extension'
                         }),
                         value: WeDo2TiltDirection.ANY
                     }
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1093,28 +1093,140 @@ class Scratch3WeDo2Blocks {
             ],
             menus: {
                 MOTOR_ID: [
-                    WeDo2MotorLabel.DEFAULT,
-                    WeDo2MotorLabel.A,
-                    WeDo2MotorLabel.B,
-                    WeDo2MotorLabel.ALL
+                    {
+                        text: formatMessage({
+                            id: 'wedo2.motorId.default',
+                            default: 'motor',
+                            description: 'label for motor element in motor menu for LEGO WeDo 2 extension'
+                        }),
+                        value: WeDo2MotorLabel.DEFAULT
+                    },
+                    {
+                        text: formatMessage({
+                            id: 'wedo2.motorId.a',
+                            default: 'motor A',
+                            description: 'label for motor A element in motor menu for LEGO WeDo 2 extension'
+                        }),
+                        value: WeDo2MotorLabel.A
+                    },
+                    {
+                        text: formatMessage({
+                            id: 'wedo2.motorId.b',
+                            default: 'motor B',
+                            description: 'label for motor B element in motor menu for LEGO WeDo 2 extension'
+                        }),
+                        value: WeDo2MotorLabel.B
+                    },
+                    {
+                        text: formatMessage({
+                            id: 'wedo2.motorId.all',
+                            default: 'all motors',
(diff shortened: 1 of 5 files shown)


ericrosenbaum (approved):
LGTM!

src/extensions/scratch3_wedo2/index.js
```diff
@@ -1077,28 +1093,140 @@ class Scratch3WeDo2Blocks {
...
+                        value: WeDo2MotorDirection.REVERSE
+                    }
                 ],
                 TILT_DIRECTION: [
-                    WeDo2TiltDirection.UP,
-                    WeDo2TiltDirection.DOWN,
-                    WeDo2TiltDirection.LEFT,
-                    WeDo2TiltDirection.RIGHT
+                    {
+                        text: formatMessage({
```
chrisgarrity, line 1157: You don't need two separate message ids for 'up', 'down' etc unless they are being used in very different contexts. For example, if the word 'right' is being used as the direction and also correctness.

However, there are a couple of things to note with respect to how the strings are automatically extracted. Only the first instance in the file will get 'saved', so make sure that has the description for translators. Any following messages only need the `id` and `default`.

src/extensions/scratch3_wedo2/index.js
```diff
@@ -1077,28 +1093,140 @@ class Scratch3WeDo2Blocks {
...
-                    WeDo2TiltDirection.DOWN,
-                    WeDo2TiltDirection.LEFT,
-                    WeDo2TiltDirection.RIGHT,
-                    WeDo2TiltDirection.ANY
+                    {
+                        text: formatMessage({
+                            id: 'wedo2.tiltDirectionAny.up',
+                            default: 'up',
+                            description: 'label for up element in tilt direction any menu for LEGO WeDo 2 extension'
+                        }),
```
chrisgarrity, line 1194: This could reuse the `wedo2.tiltDirection.up` message:
```
formatMessage({
    id: 'wedo2.tiltDirection.up',
    default: 'up'
}),
```
It doesn't need the description because that is above.

chrisgarrity (commented):
I needed to refresh my memory of the VM tools for format-message as they're a little different from the react ones used in gui. You probably don't want to define things like 'up', 'down' etc twice, so you can reuse the message id and default string everywhere after fully defining it on the first use. I used the 'up' messages as examples.

Note, due to a limitation of the automatic extraction tool only calls to formatMessage with data are extracted: 
```
formatMessage({id: 'myId', default: 'this', description: 'an example'})
```
will extract the data from the call. But
```
aMsg = {id: 'myId', default: 'this', description: 'an example'};
formatMessage(aMsg);
```
does not extract the data.

src/extensions/scratch3_wedo2/index.js
evhan55: thanks!

src/extensions/scratch3_wedo2/index.js
evhan55: Ok, makes sense, thanks!

chrisgarrity (approved):
Looks good!

Issue #1410, opened 2018-08-01
Stopping scripts in sprite/clone doesn't stop sounds that are playing "until done"


TheLogFather:
### Expected Behavior

In Scratch 2.0, stopping the scripts of a sprite/clone will stop a sound that is currently being played by a scripts that's still running a "play sound until done" block.

Note this is also the case for a clone that gets deleted and is currently running a "play sound until done" block (since, in Scratch 2, its scripts get stopped when the clone is deleted).

### Actual Behavior

In Scratch 3.0 the sounds continue playing even though the script running a "play sound until done" block has been terminated (either by "stop other scripts in sprite" or by deleting the clone).

### Steps to Reproduce

Click on a "play sound until done" block (for a sound that lasts several seconds).
While the sound is playing, click on a "stop other scripts in sprite" block.

(Alternatively, have two "when I start as a clone" scripts. One of them starts the long sound playing 'until done', the other waits for a second or so and then deletes the clone. Then create a clone by clicking on a "create clone of myself" block...)

In Scratch 2 the sound will stop playing.
In Scratch 3 it won't stop playing.

### Operating System and Browser

N/A

TheLogFather:
Forgot to reference https://github.com/LLK/scratch-vm/issues/949, which is related (since its resolution led to this happening).

BryceLTaylor:
This can be seen in [this project:](https://llk.github.io/scratch-gui/develop/compatibility-testing.html#238938875)
As well as this one: 228613336

Robomandude:
It should be noted that this issue also affects note blocks from the music extension.

PR #1631, opened 2018-10-02
Stop other asks

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-vm/issues/1076

### Proposed Changes

Add STOP_FOR_TARGET event and listen to it in the sensing blocks. On STOP_FOR_TARGET clear any questions that target is queued to ask. If the current question is from that target move to the next question or dismiss the question.

### Reason for Changes

`stop other scripts in sprite` was correctly stopping the threads with ask blocks but the ask blocks were only cleaning up after themselves for `stop all`.

### Test Coverage

- Add unit test to test that `stopAll` dismisses any question.
- Add unit tests to test `stopForTarget` responses.

PR #1632, opened 2018-10-02
Stop other sounds

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-vm/issues/1410

### Proposed Changes

Depends on https://github.com/LLK/scratch-vm/pull/1631
Move all sounds stopping to be controled by scratch3_sounds, RenderedTarget should not instruct the sounds to stop in `stopAll`.
Stop sounds for single stopping targets in scratch3_sounds.

### Reason for Changes

In scratch 2 the `stop other scripts in sprite` block also stops sounds playing from that target.

It seemed to make sense to have this per-target stopping and all targets stopping control in scratch3_sounds.

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -2,6 +2,12 @@ const MathUtil = require('../util/math-util');
 const Cast = require('../util/cast');
 const Clone = require('../util/clone');
 
+/**
+ * Occluded boolean value to make its use more understandable.
+ * @const {boolean}
+ */
+const STORE_WAITING = true;
+
 class Scratch3SoundBlocks {
     constructor (runtime) {
         /**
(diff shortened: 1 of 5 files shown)


kchadha (commented):
I noticed a slight difference between the changes in this PR and Scratch 2.0 behavior.

In Scratch 2.0, the `stop other scripts in sprite` block does not stop a sound started using the `play sound` block instead of the `play sound until done` block (because technically that script is already done running).

towerofnix:
BTW, a relevant test case for something similar to this: https://scratch.mit.edu/projects/96146764/

Probably worth considering alongside @kchadha's comment above.

mzgoddard:
@kchadha I added a change that only stops the `until done` sounds when `STOP_FOR_TARGET` is emitted.

src/blocks/scratch3_sound.js
```diff
@@ -10,14 +16,16 @@ class Scratch3SoundBlocks {
          */
         this.runtime = runtime;
 
+        this.waitingSounds = {};
+
         // Clear sound effects on green flag and stop button events.
         this.stopAllSounds = this.stopAllSounds.bind(this);
-        this._stopAllSoundsForTarget = this._stopAllSoundsForTarget.bind(this);
```
kchadha: Isn't this binding still necessary?

kchadha:
@mzgoddard, this looks good other than the one question I had about the binding for the `_stopAllSoundsForTarget` function.

src/blocks/scratch3_sound.js
kchadha: Ah, never mind, I guess this has been removed because [this line](https://github.com/LLK/scratch-vm/pull/1632/files/4d56dd38724da9853c684afb9523a9e9b7b868d8..e854d8d5639df71e243862e4169bde35cd35f596#diff-e297f21d789f60044405cdadcf095a43L20) has also been removed, right?

kchadha (approved):
LGTM

Issue #1433, opened 2018-08-07
Loading a project from 2.0 that ever checked the video motion monitor turns on video motion extension


BryceLTaylor:
### Expected Behavior

If a project doesn't use video motion blocks it should not turn on the video motion extension.

### Actual Behavior

If at any point ever you have checked the video motion sensing block in 2.0 and saved it, then even if you turn off the monitor it will be saved in the project json.  Opening the project in 3.0 then imports the extension and turns on the webcam.

Can be seen in this project: https://scratch.mit.edu/projects/238455324/#editor
As well as these:
50182430
196398611
71789076

### Steps to Reproduce

Create a project in 2.0
Choose Sensing in the category menu
Check the checkbox next to Video Motion for this sprite
Make sure the project is saved
Uncheck the video motion checkbox
Open the project in 3.0
Note: the video motion extension is added and the webcam turns on

### Operating System and Browser

Chrome on Mac

LiFaytheGoblin:
Another one: 216897461

ktbee:
Fixed in #1643

fsih:
Reopening since #1643 was reverted

kchadha:
Fixed with #1688.

PR #1634, opened 2018-10-03
Only parse visible monitors

ktbee:
### Resolves

Initializing the video sensing extension for monitors that aren't visible: https://github.com/LLK/scratch-vm/issues/1433

### Proposed Changes

I've added a check for a monitor's visibility before passing it to `parseMonitorObject` in sb2.js, which is where a sb2 project could have `senseVideoMotion` added to its extensions even though there are no video motion blocks in the project.

### Reason for Changes

Scratch 2 adds a video motion monitor to project.json when you check the video motion sensing block, but doesn't remove it once the box is unchecked. If we prevent monitor blocks with `visible: false` from being parsed, we won't initialize extensions for monitor blocks that don't appear to be in use. 

@mzgoddard and I were also thinking we should add checks to ensure the video motion extension also honors the project's initial property values for `videoOn` and `videoAlpha`, but we can do that in a separate PR.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -682,7 +682,9 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
                 }
             }
             for (let n = 0; n < deferredMonitors.length; n++) {
-                parseMonitorObject(deferredMonitors[n], runtime, targets, extensions);
+                if (deferredMonitors[n].visible) {
+                    parseMonitorObject(deferredMonitors[n], runtime, targets, extensions);
+                }
             }
             return targets;
         })


ktbee:
Hmm, looks like this fails in test/integration/monitors.js because the test expects hidden monitors. Sounds like we'll need more of a discussion of this change, so I'll hold off requesting a review for the time being.

paulkaplan:
I believe the issue here is that invisible monitors still need to be parsed to have their positions imported, so they can be in the right place when they are shown. There are a number of projects where monitors (like final score) are hidden when the project starts, but need to be in the right place when they do get shown.

ktbee:
@paulkaplan that makes sense, thanks for the context!

PR #1640, opened 2018-10-05
Set procedure param missing

mzgoddard:
### Resolves

Parameters outside of their procedure definition always return 0.

### Proposed Changes

Depends on https://github.com/LLK/scratch-vm/pull/1642
- Test that parameters outside of their procedure return 0.
- argumentReporterStringNumber and argumentReporterBoolean should return 0 when the parameter cannot be found
- If the first params object found on the stack does not have the parameter that is being looked up return null

### Reason for Changes

Scratch 2 behaviour for parameters outside their procedure (or when the procedure is stack clicked) always returns 0.

### Test Coverage

Added 6 integration tests to test parameters outside their procedure under a green flag event and under a different procedure.


thisandagain:
Thanks @mzgoddard. It looks like this is failing on the following unit test:
```
./test/unit/blocks_procedures.js ...................... 1/2 733ms
  calling a custom block with no definition does not throw
```

src/engine/thread.js
```diff
@@ -349,6 +349,7 @@ class Thread {
             if (frame.params.hasOwnProperty(paramName)) {
                 return frame.params[paramName];
             }
+            return null;
```
kchadha, line 352: Could you clarify this change some more? What was failing before this change?

kchadha (commented):
I have a minor question in the code comments.

src/engine/thread.js
kchadha: This was answered in a discussion offline:

> getParams searches for a params object in the stack. When it encounters a params object, it can check it for the parameter being searched for. After the defaults branch being merged, if its not there that implies the param is not from the current running procedure.
Since the parameter is not in the current procedure we can shortcut out of any more searching.

kchadha:
@mzgoddard, I think this PR is good to go since you answered my question offline. I copied your answer in response to my comment in the code changes. I'm going to merge this in so that it can get smoke tested this Friday.

PR #1642, opened 2018-10-05
Set procedure param defaults

### LINKED ISSUE #1534: Recursive Boolean Parameter Bug
This is a VM bug in custom blocks - and took some tracking down! ;)

### Expected Behavior

When calling a custom procedure (custom block) recursively, the boolean parameter values should be correctly passed in, even if the value is an empty value (falsey). The values should be set to false, not left as true.

### Actual Behavior

"Empty" boolean parameters evaluate to the true state when called recursively if the previous value of the parameter was true - Possibly the value is not being reset when called with an empty value? This is breaking projects like griffpatch's "snail" (project id 91723110), and no doubt other projects that make use of recursion and boolean reporters.

### Steps to Reproduce

Here is a test project: https://llk.github.io/scratch-gui/develop/#241681003
It should output into the list "true, false, false, true", but instead outputs "true, false, true, true".

PC win 10 chrome

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-vm/issues/1534

### Proposed Changes

Parse procedure parameter defaults in Blocks and return along with parameter names and ids. If a parameter does not appear in args for the procedure call, use the default.

### Reason for Changes

Recursive calls should not fall through the stack until they find a value for a parameter. They should be able to find the value in the closest params object in the stack. So it must have a passed value or a default available. Otherwise you can end up with a earlier in the stack's call value instead of the default for booleans. An earlier call could be passed true and a later call passed nothing and should return a default false value instead of true.

### Test Coverage

Add recursive procedure parameter execute integrations.


towerofnix:
Out of curiosity's sake, does take into account "hacked" custom blocks which have their own default values? I.e. in the SB2 representation `["procDef", "jump %n", "distance", 25, false]`, the default value for the "distance" argument is 25.

(Edit - probably not, sounds like something for a different PR/issue?)

mzgoddard:
@towerofnix https://github.com/LLK/scratch-vm/pull/1642/files#diff-c2e52c759bccc44bfc3789bbfe5d9b2eR279 should handle that. The code immediately before that handles the `"todo"` case created by scratch-blocks currently. ... Actually that brings up an important issue, if there is a project with hacked defaults that actually wants `"todo"` to be the default that code would overwrite that. Otherwise hacked defaults would be honored. So I probably have to remove that `"todo"` stuff because of that and projects made with 3.0 will have weird procedure defaults until procedure defaults are fixed in scratch-blocks.

mzgoddard:
I've removed the todo -> normal defaults transformation that was in an earlier version of this PR.

kchadha (commented):
The code changes lgtm.

I was looking at the tests you added and I'm a little confused about `procedures-recursive-default-string.sb2` and `procedures-recursive-default-number.sb2`. The boolean param `boolean1` should not exist in either of those projects. I assume that's a remixing error? Also, should the `say` blocks containing the `join` actually read `'fail'` instead of `'pass'`?

kchadha (approved):
Looks good! Thanks for making the test changes as well.

Issue #1639, opened 2018-10-04
Need an event to alert extensions when targets are done being installed


ktbee:
### Expected Behavior

Extensions can use an events listener to wait for when targets are installed, and then execute target dependent code accordingly.

### Actual Behavior

No `runtime` events indicate when all of the targets have finished installing and are ready to be used.

### Steps to Reproduce

In taking a look at https://github.com/LLK/scratch-vm/issues/1635, I found that there is a race condition where [we check for the stage target](https://github.com/LLK/scratch-vm/blob/develop/src/extensions/scratch3_video_sensing/index.js#L164) in the extension, but the stage target is always undefined because the extension [is loaded before the targets are installed](https://github.com/LLK/scratch-vm/blob/develop/src/virtual-machine.js#L432). If we had something like a "targetsInstalled" event, we could check again for the stage target once all the targets are fully loaded.

Issue #1635, opened 2018-10-03
Load videoOn and videoAlpha state from sb2 project's initial values


ktbee:
### Expected Behavior

A sb2 project's saved video state (video on or off, transparency level) should be used when loading the project.

### Actual Behavior

Transparency and the `videoOn` setting aren't always applied to the project. Transparency issue first reported here for 3.0: https://github.com/LLK/scratch-vm/issues/1062

### Steps to Reproduce

Steps to reproduce transparency issue: https://github.com/LLK/scratch-vm/issues/1062

Steps to repro `videoOn` issue:
1. Create new 2.0 project
2. Add video on/off block in sensing
3. Turn on video and save
4. turn off video and save
5. Observe that video is off in 2.0, but not in 3.0

example project: https://scratch.mit.edu/projects/250128622/

### Operating System and Browser

Mac OS 10.13.4,  Chrome 69.0.3497.100

Issue #1062, opened 2018-04-13
Load video transparency state from stage property when a project gets loaded


kchadha:
### Expected Behavior

A project that has a modified video transparency state should get loaded with the same video transparency state that was saved in the project.

### Actual Behavior

The video transparency state does not load from the saved project state.

### Steps to Reproduce

1. Create a new 3.0 project
2. Add the Video Motion extension
3. Change the video transparency state using the `set video transparency to (NUMBER)` block.
4. Download the project to your computer
5. Load the project and observe that the video transparency setting is not what you saved.

PR #1643, opened 2018-10-05
Confirm extension in use in sb2 serialization

ktbee:
### Resolves

A few related issues: 
- https://github.com/LLK/scratch-vm/issues/1433 
- https://github.com/LLK/scratch-vm/issues/1635 
- https://github.com/LLK/scratch-vm/issues/1062

### Proposed Changes

This PR checks if there is a discrepancy between the stage target's `videoOn` and `videoTransparency` properties and what the video currently is displaying, and updates those properties accordingly. 

### Reason for Changes

Currently there is a race condition when extensions are loaded and use `this.runtime.getTargetForStage`. Targets are installed after the extensions are loaded, so the stage target's properties aren't available to load until after the targets are updated. This PR checks for discrepancies whenever targets are updated, and updates the project's video properties if they are different. I made an issue to handle this a little more neatly by adding an event for when targets are installed (https://github.com/LLK/scratch-vm/issues/1639), but checking when targets are updated could be a solution in the meantime.


fsih (changes requested):
A couple issues
1. We would strongly prefer if the Video Sensing category didn't load into the project at all when no blocks from it are present. (It appears at the bottom of the list of block categories.) It's a particularly sketchy extension to have appear when you didn't expect it 
image

2. When the video sensing extension is added by the user, the video should show up on stage, but after this change it does not.
To repro click here
image
then here
image

@kchadha suggests that it might be a good solution to ignore loading a monitor from an extension category that isn't already loaded.

ktbee:
Thanks for taking a look, @fsih. That feedback all makes sense, I'll make those updates.

ktbee:
By the way @fsih, do you have any example projects where you're seeing the video sensing extension is added to the bottom of the list of categories or where you can't add it to the stage? I haven't been able to repro that yet, for example Bryce's project demoing the original issue isn't adding the video extension for me at the bottom of the block categories (238455324).

fsih:
@ktbee that's odd, I got it on all the projects listed in https://github.com/LLK/scratch-vm/issues/1433 . I'm on Mac Chrome 69 in case that's relevant. Here's what I see
extensionload

fsih:
@ktbee Here is what I see for the video loading (will pull vs develop)
videoload2

ktbee:
@fsih thanks for the clarification and the screencasts, very helpful! I've simplified this PR after some offline suggestions from @kchadha as well. Currently it just addresses #1433 by checking that only extensions used in blocks or visible monitors are added to the set of `extensionIDs` for a serialized sb2 project.  

In this version I confirmed that the video shows on the stage when you add the video extension. Also the video extensions isn't added the project when you go the project with a full page reload, as you showed in your screenshot.

One issue I'm seeing that is also in `develop` is that the video extension will stay added to a project and the video will stay on when you navigate to another non-video sensing project without a full page refresh. Steps to repro:

1) Open project that uses the video sensing extension
2) Change the project ID in the URL manually to a project that doesn't use the video extension at all. Hit enter.
3) Observe the video is still on and the extension is still added

I'll make another issue for this since I'm not seeing an existing one (let me know if I missed it though). However, I wanted to mention it in case it comes up as you're taking a look at this round of changes as well.

fsih:
@ktbee we have a bug for video staying on between projects here: https://github.com/LLK/scratch-gui/issues/1849

src/serialization/sb2.js
```diff
@@ -334,6 +334,58 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
...
 
+const confirmTargetExtensions = function (targets, extensions) {
+    // Ensure only extensions used by blocks or visible monitors are enabled
+    const extensionsInUse = new Set();
+
+    for (const ext of extensions.extensionIDs) {
+        let extensionConfirmed = false;
+
+        for (const target of targets) {
+            if (extensionConfirmed) {
```
fsih: Nit, I don't think you need this if block

fsih (approved):
Looks great, thank you!

ktbee:
@fsih good catch! I removed that if-statement you mentioned.

ktbee:
Also @fsih , let me know if I should merge this or if we're waiting to add the "good to merge" label for another reason

fsih:
@ktbee yes, when it's good to merge the writer of the change can merge the change. There was a Travis integration test failing, so I went into Travis and restarted it and it's succeeding now (unfortunately tests time out somewhat often :/) but now it's all passing so feel free to merge!

fsih:
One thing to note is that once you merge the pull request, Travis tests kick off again, and if they don't pass, then the change doesn't actually get deployed to NPM. So if a test flakes again you might have to check back in and restart the Travis build :s

BryceLTaylor:
This has caused the pen extension to not load: https://github.com/LLK/scratch-vm/issues/1676

Issue #1478, opened 2018-08-15
Some When I receive show/hide stacks don't trigger in 3.0


BryceLTaylor:
In this project the word Breathe appears and does not disappear.
https://llk.github.io/scratch-gui/develop/compatibility-testing.html#169401431
<img width="995" alt="screen shot 2018-08-14 at 11 28 28 pm" src="https://user-images.githubusercontent.com/1088291/44130200-52ba5f26-a01a-11e8-9103-20cfaa2acdf2.png">

The project uses broadcasting for each part and for some reason that sprite doesn't trigger the `When I receive 'part 14'` stack on the sprite "Part 13 Rocks"

In this project:
https://llk.github.io/scratch-gui/develop/compatibility-testing.html#12714608
On Level 10 the red enemy sprite never appears.
<img width="984" alt="screen shot 2018-08-14 at 11 44 20 pm" src="https://user-images.githubusercontent.com/1088291/44130522-0486fd30-a01c-11e8-8bbb-782ad446cdfd.png">

It similarly has a stack with a`When I receive 'go'` hat that never triggers.

And in this project:
https://llk.github.io/scratch-gui/develop/compatibility-testing.html#13701368
When you win (beat the pac man level) the ghosts and the monitors don't disappear on the end screen:
<img width="993" alt="screen shot 2018-08-14 at 11 53 49 pm" src="https://user-images.githubusercontent.com/1088291/44130843-75c76358-a01d-11e8-90a8-399ab4b69eea.png">

The ghosts and Pac Man have a `When I receive 'hide'` that doesn't trigger

BryceLTaylor:
Related to this issue: https://github.com/LLK/scratch-vm/issues/1053

Also, note this project:
https://llk.github.io/scratch-gui/develop/compatibility-testing.html#239873312

<img width="978" alt="screen shot 2018-08-23 at 1 50 40 pm" src="https://user-images.githubusercontent.com/1088291/44542690-8ee1a000-a6db-11e8-9a96-911ec2207941.png">

mzgoddard:
The projects in this issue have the same symptom due to 4 different bugs. In all of their cases the `when I receive` are kicking off threads.

[...]

PR #1648, opened 2018-10-09
Stop all next tick

mzgoddard:
### Resolves

Part of https://github.com/LLK/scratch-vm/issues/1478

### Proposed Changes

Match Scratch 2.0 `Interpreter.stopAllThreads()` behaviour.

### Reason for Changes

In Scratch 2.0, `stop all` only immediately stops the active thread and removes the remaining threads from executing next tick. The remaining threads are able to execute one last time. It is also possible because of this that the remaining threads can create new threads that will in fact execute next tick.

### Test Coverage

Add a execute integration test that makes sure threads after stop all get one more execution and that they can actually create new threads that will execute in a later tick.

The behaviour is demonstrated in Scratch 2.0 project 13701368.


src/engine/runtime.js
```diff
@@ -1494,9 +1494,14 @@ class Runtime extends EventEmitter {
                 newTargets.push(this.targets[i]);
             }
         }
+
```
kchadha: stray newline

BryceLTaylor:
in Scratch 2.0 if in the second thread you call another broadcast and the second broadcast has a forever in it, stop all doesn't stop that other forever:
https://scratch.mit.edu/projects/260463874/
endthreadtest
The above changes don't cause that to be true.
The cat does the full Meow in this example.

Here that situation is in 3.0 after this PR:
endthreadtestin3

The cat meows fully but does not enter the change costume loop.

Also, in 2.0, if you use a custom procedure to create the other thread and the custom procedure has a broadcast, the when I receive does not run.
https://scratch.mit.edu/projects/260471599/#editor
endthreadtest2

In that gif the cat meows briefly and then is stopped.

Here's that situation in 3.0 after this PR:
endthreadtest2in3

This behaves the same in 3.0 as 2.0 with half a meow.

kchadha:
@mzgoddard, marking this as needs work because of the issue that @BryceLTaylor mentioned with the   custom procedures. I would be open to filing that as a separate issue and landing this one if the fix is particularly complicated.

ktbee:
I think the issues @BryceLTaylor noted from when @mzgoddard first submitted this PR have been fixed! When I download this branch and merge in the latest changes from `develop`, I find that the example projects that @BryceLTaylor provided have compatible behavior with Scratch 2, ie the hat block stack now fires in #260463874 despite the "stop all" block. 

I believe [our recent fix to start hat threads in `startHats`](https://github.com/LLK/scratch-vm/pull/1943) has made this PR's behavior more compatible with Scratch 2. Now those threads are started before we have a chance to remove them from the runtime's threads array in `stopAll`.

@BryceLTaylor are these examples now resolved? Or am I missing some incompatible behavior? 

/cc @kchadha

mzgoddard:
- Fix conflict (It was an added comment.)

BryceLTaylor:
@mzgoddard  @kchadha I think this looks pretty good!

kchadha (approved):
LGTM! Thanks @mzgoddard!

PR #1649, opened 2018-10-09
Get attribute stage alt name

mzgoddard:
### Resolves

Part of #1053.

### Proposed Changes

Support 2.0 projects that have a `getAttribute:Of:` opcode with a OBJECT input set to `Stage` instead of the anticipated `_stage_`.

### Reason for Changes

Some projects have the OBJECT input set to `Stage` instead of `_stage_`. However the VM only understands _stage_ to mean the Stage target.

### Test Coverage

Add a integration test testing that getAttributeOf works with Scratch 2.0 projects that have the OBJECT input set to `Stage` instead of `_stage_`.

The Scratch 2.0 behaviour can be seen in project 12714608.


kchadha (changes requested):
It turns out this introduces a new "bug" where importing this block with another block plugged into it (obscured shadow) 

e.g. image

 and then unplugging the input block (e.g. the join block above) in 3.0 results in the underlying shadow default to the opcode of the plugged in block...

image

Before these changes (e.g. on beta) doing the above would have resulted in this:

image

which is less bad, but still not great. I think we should probably default to 'Stage' when the shadow is obscured e.g. add the following to the new else if block you added:

```
if (shadowObscured) {
    fieldValue = '__stage__';
}
```

mzgoddard:
Change applied.

kchadha (approved):
LGTM

PR #1652, opened 2018-10-10
Add icons for Video Sensing, Translate and Text to Speech extensions

ericrosenbaum:
Adding extension menu and block icons for three extensions, with artwork from @carljbowman 

They're SVG data URIs, except for the translate icons, which were larger SVG files, so I put them in as 2x PNGs.  

screen shot 2018-10-10 at 2 11 02 pm


thisandagain (approved):
Looks good! Thanks @carljbowman and @ericrosenbaum!

PR #1655, opened 2018-10-11
Text to Speech voices update

ericrosenbaum:
### Proposed Changes

- Change the first two voice names from "quinn" and "max" to "alto" and "tenor"
This makes them a bit more neutral (they are more like descriptions and less like names, which is hopefully less confusing).

- Adjust the pitch of the squeak voice to be a bit lower
This makes it easier to understand what the voice is saying.

- Add support for dropping a number onto the "set voice" block input
Now you can set the voice using a variable, where a one-indexed value maps to a voice. This makes it easier to make projects that interactively set the voice, and of course now you can set a random voice using the "pick random" block:

screen shot 2018-10-11 at 11 56 34 am


thisandagain (approved):
This is great. Thanks @ericrosenbaum!

PR #1669, opened 2018-10-17
Add runtime event for when the project is loaded

### LINKED ISSUE #1639: Need an event to alert extensions when targets are done being installed

ktbee:
### Resolves

This PR resolves #1639

### Proposed Changes

Add an event to a project's runtime to alert any listeners that the project's targets have been installed in the virtual machine.

### Reason for Changes

As mentioned in the issue, extensions don't currently have a way to know if targets are installed. This can lead to race conditions where extensions check for a target (like the stage) when they are being loaded, but the targets are only installed after the extensions load.

### Test Coverage

Add a test in `test/unit/engine_runtime.js` to check whether the event is fired by the runtime instance whenever targets are installed.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -401,6 +401,14 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Event name for targets installed report.
+     * @const {string}
+     */
+    static get TARGETS_INSTALLED () {
+        return 'TARGETS_INSTALLED';
+    }
+
+    /**
      * Event name for targets update report.
      * @const {string}
      */
@@ -1900,6 +1908,14 @@ class Runtime extends EventEmitter {
      */
     clonesAvailable () {
         return this._cloneCounter < Runtime.MAX_CLONES;
+    }
+
+    /**
+     * Report that targets have finished being installed in the Virtual Machine.
+     * @param {Target} installedTargets - the newly installedTargets.
+     */
+    emitTargetsInstalled (installedTargets) {
+        this.emit(Runtime.TARGETS_INSTALLED, installedTargets);
     }
 
     /**
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -401,7 +401,7 @@ class Runtime extends EventEmitter {
     }
 
     /**
-     * Event name for targets installed report.
+     * Event name for project loaded report.
      * @const {string}
      */
     static get PROJECT_LOADED () {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -404,8 +404,8 @@ class Runtime extends EventEmitter {
      * Event name for targets installed report.
      * @const {string}
      */
-    static get TARGETS_INSTALLED () {
-        return 'TARGETS_INSTALLED';
+    static get PROJECT_LOADED () {
+        return 'PROJECT_LOADED';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/engine/runtime.js
```diff
@@ -1902,6 +1910,14 @@ class Runtime extends EventEmitter {
         return this._cloneCounter < Runtime.MAX_CLONES;
     }
 
+    /**
+     * Report that targets have finished being installed in the Virtual Machine.
+     * @param {Target} installedTargets - the newly installed targets.
+     */
+    emitTargetsInstalled (installedTargets) {
```
kchadha: Is there a reason that this event lives in the runtime instead of the VM?

src/virtual-machine.js
```diff
@@ -460,6 +460,7 @@ class VirtualMachine extends EventEmitter {
             this.emitTargetsUpdate();
             this.emitWorkspaceUpdate();
             this.runtime.setEditingTarget(this.editingTarget);
+            this.runtime.emitTargetsInstalled(targets);
```
kchadha: Perhaps a note for the future... `installTargets` has the potential to get called multiple times for a single project.

`installTargets` gets called when a project gets loaded, but also whenever a sprite gets added or a deleted sprite gets restored. Is that okay?

I could see this maybe creating an issue down the line where we set extension state when all the targets are installed, and then accidentally reset it when adding a sprite... 
I'm not sure if it will actually be an issue, but just wanted to make sure you were aware of all the use cases of `installTargets`.

src/engine/runtime.js
kchadha: (This might be the right place for it to live, I just wanted to know what the reasoning for it living here is)

kchadha (commented):
I have some questions in the code comments.

src/engine/runtime.js
ktbee: Good Q! I wanted this event to be available to the extensions (ie listenable in `scratch3_video_sensing/index.js`), and they are already able to listen to runtime events. From what I can tell, currently they're not able to listen to VM events, but let me know if that's not the case.

src/virtual-machine.js
ktbee: Good call, I noticed that `installTargets` was being called each time a sprite was added as you mentioned, and I'm also not sure whether this is desirable not. I could imagine a use case where someone would want to know whenever a target is installed, though I can imagine more where it gets unintentionally called. 

In thinking about it again, is there a VM event that runs just once when a project has finished its initial load? Perhaps we can propagate that to the runtime? I've taken a look at `virtual-machine.js` to find an event that might work, and the closest seems to be `workspaceUpdate`, though that also runs whenever `installTargets` ones.

kchadha: > In thinking about it again, is there a VM event that runs just once when a project has finished its initial load? Perhaps we can propagate that to the runtime? I've taken a look at `virtual-machine.js` to find an event that might work, and the closest seems to be `workspaceUpdate`, though that also runs whenever `installTargets` ones.

There isn't already such an event that exists, but you should be able to achieve what you described by moving the event emitter line from the end of `installTargets` to the end of `loadProject` instead. You can chain another `.then` clause after this one that waits for the `deserializeProject` call to finish:
`.then(validatedInput => this.deserializeProject(validatedInput[0], validatedInput[1]))`

ktbee:
@kchadha thanks for the suggestions! I've switched this PR to add a runtime event for when a project is loaded, and that event should only run once per project load.

kchadha (approved):
LGTM!

PR #1670, opened 2018-10-17
Save and load the language setting for Text to Speech

ericrosenbaum:
### Proposed Changes

- Save and load a property on the stage for the language code used by the Text to Speech extension.
- Move the conversion to Polly's language code format into an adapter function, to be removed once we have this conversion on the proxy server.
- On extension load, set the language for the extension using the editor language, if it is one of the supported languages for speech synthesis.
- Only localize the default input to the `speak` block if the editor is in a supported language. i.e. if the editor is in an unsupported language, we will be using English language speech synthesis, so the block should use the default English input ("hello").

Some remaining improvements I think we need (for future work):
- Handling changes to the editor language after the extension has been loaded. It's not clear to me what should happen in this case (maybe nothing).
- The `set language` block should accept a dropped language name (and this should work in any language). E.g. You should be able to use the “language” reporter from translate to set the text2speech language.
 
### Reason for Changes

This is progress toward making the Text to Speech extension work well across languages. We need to save and load the language so that when a project is shared, it will remember the language of the creator, and correctly speak they text they have entered, if the viewer of the project is using a different language (i.e. even if the project does not contain a `set language` block).

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -158,21 +158,51 @@ class Scratch3Text2SpeechBlocks {
      */
     get LANGUAGE_INFO () {
         return {
-            'Danish': 'da-DK',
-            'Dutch': 'nl-NL',
-            'English': 'en-US',
-            'French': 'fr-FR',
-            'German': 'de-DE',
-            'Icelandic': 'is-IS',
-            'Italian': 'it-IT',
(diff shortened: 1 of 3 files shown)


thisandagain (approved):
This LGTM but I'd love @picklesrus to take a look!

picklesrus:
LGTM too!

PR #1683, opened 2018-10-19
Sequencer thread cleanup

### LINKED ISSUE #3344: Performance Regression with Geometry Dash
The geometry dash project (https://llk.github.io/scratch-gui/develop/#105500895) is running slow again despite performance issues being fixed recently.

The `Player: FPS` monitor reads out `15` on llk.github.io/scratch-gui/develop, but runs at normal speed (30 FPS) on beta. This means that this change was introduced sometime in the last week (Thursday Oct 4th - Thursday Oct 11th).

mzgoddard:
### Resolves

Fix https://github.com/LLK/scratch-gui/issues/3344

### Proposed Changes

Track if a thread has finished with a boolean called `stoppedThread`. If it is true after stepping all threads, iterate the threads a second time, filtering out threads from the list that have finished.

### Reason for Changes

https://github.com/LLK/scratch-vm/pull/1619 fixed a compatibility issue so that targets would execute in the opposite order that they are drawn instead of the fixed order stated by the project json. This introduced a side effect in the broadcast "tick" and wait block used to "tick" the project. The broadcast and wait block must wait for the threads to leave the list of executing threads. Scratch 2 removes finished threads from the running list multiple times a tick instead of once at the end of the tick. This side effect caused Geometry Dash to run at half the possible framerate because the broadcast and wait had to wait for those tick threads to be removed at the end of the tick instead of in the middle of the tick like Scratch 2 does.

Respecting the Scratch 2 behaviour Scratch 3 similarly needs to check and remove finished threads multiple times a tick.

### Test Coverage

Add a integration test that uses the order of threads, broadcast and wait, and repeat blocks to ensure that broadcast and wait may complete in the same tick it creates threads in and that those threads are cleaned up as soon as possible while matching Scratch 2 behaviour.


kchadha (changes requested):
@mzgoddard, these changes seem to fix the issue in the geometry dash project and also in the project that @towerofnix mentioned in #760. 

It looks like some tests are failing though...

mzgoddard:
I updated the tests that check runtime.threads. Most of them now check a new runtime member `_lastStepDoneThreads` that stores the doneThreads from the last `runtime._step` call. The tests now check runtime.threads and _lastStepDoneThreads to make the right number of threads are in each.

kchadha (approved):
@mzgoddard,

LGTM! I'm going to merge so that this can get tested this week, but I think we should possibly try to document the changes here somewhere (especially the ones in the test).

PR #1686, opened 2018-10-21
Monitor save & load

### LINKED ISSUE #1501: Monitor Save/Load
### Expected Behavior

Monitor state should be saved and loaded.
E.g.
- position on stage,
- mode (big, slider, slider min/max)

### Actual Behavior

It isn't yet.

### LINKED ISSUE #1513: Timer monitor does not import correctly.
### Expected Behavior

Timer monitor should get imported from 2.0 project.

### Actual Behavior

Timer monitor does not get imported correctly from 2.0 project.

### Steps to Reproduce

1. Open a 2.0 project
2. Click the checkbox in the toolbox for the timer monitor.
3. Save the project
4. Open the project in 3.0, and observe that there is a monitor that says "timer" on the stage, but it is empty, and the timer reporter in the toolbox does not have its monitor checkbox checked.

### LINKED ISSUE #1628: Backdrop name monitor is not imported correctly
### Expected Behavior

Monitor for 'backdrop name' reporter should get imported from 2.0

### Actual Behavior

Monitor label gets imported correctly, but has a blank value where the name should be.

### Steps to Reproduce

Try out this project:
https://beta.scratch.mit.edu/#249559537

#### Other Notes
Potentially related to #1513

kchadha:
### Resolves

Resolves #1501, monitor save/load, also mentioned in LLK/scratch-gui#3086

### Proposed Changes

- Serialize monitor information when saving a project.
- Deserialize monitor information when loading an sb3 (this adds a new monitor record to the runtime monitor state and either
    - finds and augments the corresponding monitor block in the monitor block container to set the `isMonitored` and `targetId` properties), or
    - creates a new monitor block for the monitor
- Changes monitor color for extension monitors from the `data` category color to the extension color.

#### Known Issues
- Imported variable monitors and monitors with parameters in the reporter don't display the correct checkbox state in the toolbox when the project is loaded. For variable monitors, the toolbox state becomes in-sync again after switching to a different sprite or tab causing the workspace and toolbox to reload. 
- For the monitors where the reporter has a parameter, switching sprites causes the checkbox in the toolbox to become unchecked (this is a remnant issue from #1742).
- Monitor information is not currently serialized when exporting a sprite. This was left out of this PR because there is currently still a problem with variables retaining their IDs when serialized and thus importing a sprite3 with local variables (and monitors) back into the project it was exported from causes the local variable IDs to no longer be unique. This creates confusion with the local variable monitors because the monitor block is identified by the variable ID (thus two variables share the same monitor and cannot be displayed at the same time).
- Monitors that are placed close to the right hand side of the stage, sometimes wrap when the project is saved and loaded.

### Reason for Changes

Monitors should get saved and loaded in sb3s.

### Test Coverage

Manual testing along with LLK/scratch-gui#3713.

Testing branch: https://llk.github.io/scratch-gui/monitors/

Added integration test monitor save/load, below is a screenshot of the stage in the fixture project added for this integration test:

integration-test-screenshot

Ensure the following:
- [ ] Can still import monitors from 2.0 projects
- [ ] Stage monitors like `loudness` and `timer` load correctly, (are visible, not empty, and update)
- [ ] Imported monitors retain position and mode, list monitors retain height and width
- [ ] Can save and load monitors in .sb3 projects. Monitors should load correctly and should update with changes to their values.
- [ ] Test that non-visible monitors are saved/loaded correctly. E.g. check a monitor, move it somewhere and change its mode, hide the monitor, and then save the project. Check that the monitor is still hidden when that project is uploaded, and that when you show the monitor, it retained its new position and mode.

### Related PRs

- This PR is related to LLK/scratch-gui#3713. This PR should be merged before the GUI PR.

cc/ @picklesrus


paulkaplan (changes requested):
Looks like we should add an integration test for loading sb3 files saved with monitors, to parallel the one testing loading sb2 files with monitors

src/serialization/sb2.js
```diff
@@ -295,12 +295,17 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
...
+    const existingMonitorBlock = runtime.monitorBlocks._blocks[block.id];
+    if (existingMonitorBlock) {
+        // A monitor block already exists if the toolbox has been loaded and
+        // the monitor block is not target specific (because the block gets recycled).
+        // Update the existing block with the relevant monitor information.
+        existingMonitorBlock.isMonitored = object.visible;
+        existingMonitorBlock.targetId = block.targetId;
     }
+    // In the case where a monitor block does not already exist, one will get created
+    // when the target is installed because the toolbox will reload.
```
paulkaplan: Make sure to check this works in player mode where there is no toolbox

kchadha:
As discussed offline, it looks like this approach of using the monitor block from the runtime monitorBlocks container is not going to work after all since we won't always have monitorBlock information when we need it. In player mode (thus also on the preview page), the toolbox doesn't get loaded when the project does, so the monitorBlocks container does not get populated and monitors appear empty when a project is loaded.

#### Possible Solutions for The Future™️

- Extension-ify just the monitor blocks (is this possible)? Specifically they have a handy json serialization function and a minimal specification
- SB3 -> SB3 specmap (for just the monitor blocks?). Extension monitors would still use their json serialization function
- Serialize the actual monitor blocks as well as the monitor record in the sb3. We would need to check how much this ends up bloating the file size.

kchadha:
Closing this until I can come back to it in the future.

src/serialization/sb3.js
```diff
@@ -1015,6 +1038,96 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
...
+    // If the runtime already has a monitor block for this monitor's id,
+    // update the existing block with the relevant monitor information.
+    const existingMonitorBlock = runtime.monitorBlocks._blocks[monitorData.id];
+    if (existingMonitorBlock) {
+        // A monitor block already exists if the toolbox has been loaded and
+        // the monitor block is not target specific (because the block gets recycled).
+        existingMonitorBlock.isMonitored = monitorData.visible;
+        existingMonitorBlock.targetId = monitorData.targetId;
+    } else {
+        // If a monitor block doens't already exist for this monitor,
```
paulkaplan: nit: typo "doens't"

src/serialization/sb3.js
```diff
@@ -1015,6 +1038,96 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
...
+
+        runtime.monitorBlocks.createBlock(monitorBlock);
+
+        // If the block is from an extension, record it.
+        const extensionID = getExtensionIdForOpcode(monitorBlock.opcode);
+        if (extensionID) {
+            extensions.extensionIDs.add(extensionID);
+        }
+    }
+    // Otherwise, the monitor block will get created when the toolbox updates
```
paulkaplan: Does this comment still apply?

paulkaplan (approved):
Apart from the nit and question, looks good.

src/serialization/sb3.js
kchadha: No it doesn't =]. Removed.

src/serialization/sb3.js
kchadha: Fixed

Issue #1676, opened 2018-10-18
Pen Extension doesn't load when importing projects, No blocks load in Sprite


BryceLTaylor:
### Expected Behavior

If a project uses pen blocks the pen extension should load and the blocks should appear in the sprites that use the pen blocks

### Actual Behavior

When a 2.0 project loads that uses the pen blocks the Pen extension doesn't load.  Any sprite that uses pen blocks shows no blocks in the workspace.

### Steps to Reproduce

Load dolphins 3d https://llk.github.io/scratch-gui/develop/#115870836
Look at the Pen sprite
Note: It has no blocks and the Pen Extension is not loaded (strangely it does have a workspace comment)

3d Solar System v1.6.1 https://llk.github.io/scratch-gui/develop/#89811578
Look at the OrbitDraw sprite
Note: It has no blocks and the Pen Extension is not loaded

Load Bouncy Heroes https://llk.github.io/scratch-gui/develop/#173918262
Look a the Loop sprite
Note: It has no blocks and the Pen Extension is not loaded

### Operating System and Browser

Mac Chrome
Safari Chrome

BryceLTaylor:
The PR #1643 is being reverted for now in this pr https://github.com/LLK/scratch-vm/pull/1675.

ktbee:
The original issue that broke the pen extension is fixed in #1693 and we now have tests for catching when the pen extension breaks as well in #1700

PR #1688, opened 2018-10-22
Prevent invisible monitors from loading extensions

### LINKED ISSUE #1433: Loading a project from 2.0 that ever checked the video motion monitor turns on video motion extension

ktbee:
### Resolves

#1433 and tangentially #1676 since the broken pen extension was caused by the last solution (#1643) and is avoided in this one. 

### Proposed Changes

This PR adds a check in the sb2 serialization process for monitors that are not visible and that have a `cmd` for an extension that is supported in Scratch 3.

### Reason for Changes

Invisible monitors for some extensions (like the video sensing example in #1433) add their extension to the serialized extension set. There isn't a use case (that I know of) for invisible monitors that use extensions, so I've added a check for these sorts of monitors, and prevent serializing them if found. In addition to the video sensing example in #1433, I was able to create two other examples with the music and WeDo 2.0 extension:

Music: https://llk.github.io/scratch-gui/develop/#255230688
Steps to repro in Scratch 2: 
- Check "tempo" under Sounds, save
- Uncheck reporter and save
- Note the extension is added to the bottom of the Scratch 3 project

WeDo2: https://llk.github.io/scratch-gui/develop/#255231406 
Steps to repro in Scratch 2: 
- Add WeDo 2.0 extension to project
- Check "distance", save
- Uncheck "distance" and save
- Note the extension is added to the bottom of the Scratch 3 project

### Test Coverage

I've added a check to confirm invisible extension monitors aren't added to the runtime in `test/unit/serialization_sb2.js`


kchadha (changes requested):
@ktbee, I think the important piece of information here that might have been missing is that the `extensions` argument passed into `parseMonitorObject` is already the minimal set of extensions that exists because actual blocks using those extensions have been found in the project. At the point where we get to the for loop that calls parseMonitorObject, we have already parsed all the targets and their blocks. I think these changes could be simplified by early exiting in `parseMonitorObject` if the opcode the monitor refers to is not in the passed in `extensions` list. 

Thus, if the monitor refers to a block that is part of an extension that's not in the list, parsing that monitor can be skipped.

I tried out the following code snippet and I think it fixes the original problem in #1433 

```
 const parseMonitorObject = (object, runtime, targets, extensions) => {
+    const opcode = specMap[object.cmd].opcode;
+
+    const index = opcode.indexOf('_');
+    const prefix = opcode.substring(0, index);
+    if (CORE_EXTENSIONS.indexOf(prefix) === -1) {
+        if (prefix !== '') return;
+    }
```

BTW, I think this PR is missing the changes to the video state (defaulting the video motion extension to the `OFF` state and then explicitly requesting it to turn on when necessary). I think we still want to keep those changes. =]

ktbee:
Thank @kchadha for the feedback and explanation! I believe I've addressed all of your requested changes, but let me know if this PR can still be improved.

ktbee:
@kchadha as we talked about offline, I've removed the changes to set default video state to "off", since these aren't testable at the moment. 

The plan now is to reintroduce these changes once we have a runtime event for when targets are installed in the runtime. This will allow us to check the video state again once we know the stage target is installed and enable/disable video accordingly.

kchadha (approved):
@ktbee LGTM! Thanks for making all the requested changes!

Issue #1577, opened 2018-09-12
Refactor asset instances to be held by VM, not Storage


rschamp:
To prepare for incremental saving, stop storing lists of assets in scratch-storage, and instead hold these instances as a part of the costume and sound lists.

When a costume or sound is updated, replace the instance in the list rather than updating — this will help ensure atomic saves.

PR #1691, opened 2018-10-23
Store asset objects on costumes and sounds

rschamp:
Depends on https://github.com/LLK/scratch-storage/pull/50 (merged!)

### Resolves

Towards https://github.com/LLK/scratch-vm/issues/1577.

### Proposed Changes

Stop using storage for in-memory asset storage, and keep these on the costume/sound objects. Then start retrieving the assets from the costumes/sounds rather than trying to load them from storage's cache. 

### Reason for Changes

See https://github.com/LLK/scratch-storage/pull/50.

### Test Coverage

Tests have been updated to account for these changes.


src/import/load-costume.js
```diff
@@ -11,12 +11,8 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
         // Put back into storage
         const storage = runtime.storage;
-        costumeAsset.encodeTextData(svgString, storage.DataFormat.SVG);
-        costume.assetId = storage.builtinHelper.cache(
-            storage.AssetType.ImageVector,
-            storage.DataFormat.SVG,
-            costumeAsset.data
-        );
+        costume.asset.encodeTextData(svgString, storage.DataFormat.SVG, true);
+        costume.assetId = costume.asset.assetId;
```
rschamp, line 15: The `assetId` on `costume.asset` is now regenerated when `encodeTextData` regenerates the asset's data.

src/import/load-costume.js
```diff
@@ -163,9 +162,12 @@ const loadCostume = function (md5ext, costume, runtime, optVersion) {
...
     const ext = idParts[1].toLowerCase();
     const assetType = (ext === 'svg') ? AssetType.ImageVector : AssetType.ImageBitmap;
-
-    return runtime.storage.load(assetType, md5, ext).then(costumeAsset => {
-        costume.dataFormat = ext;
+    costume.dataFormat = ext;
+    return (
+        (costume.asset && Promise.resolve(costume.asset)) ||
+        runtime.storage.load(assetType, md5, ext)
+    ).then(costumeAsset => {
```
rschamp, line 169: This was the clearest way I found to do this, open to suggestions. We only want to try to load the asset from storage if the asset is `null`.

src/import/load-sound.js
```diff
@@ -55,11 +55,14 @@ const loadSound = function (sound, runtime, sprite) {
...
-    return runtime.storage.load(runtime.storage.AssetType.Sound, md5, ext)
-        .then(soundAsset => {
-            sound.dataFormat = ext;
-            return loadSoundFromAsset(sound, soundAsset, runtime, sprite);
-        });
+    sound.dataFormat = ext;
+    return (
+        (sound.asset && Promise.resolve(sound.asset)) ||
+        runtime.storage.load(runtime.storage.AssetType.Sound, md5, ext)
+    ).then(soundAsset => {
```
rschamp, line 62: Same goes here.

src/serialization/deserialize-assets.js
```diff
@@ -22,37 +22,29 @@ const deserializeSound = function (sound, runtime, zip, assetFileName) {
...
     }
 
-    const assetId = sound.assetId;
-
-    // TODO Is there a faster way to check that this asset
-    // has already been initialized?
-    if (storage.get(assetId)) {
-        // This sound has already been cached.
-        return Promise.resolve(null);
-    }
```
rschamp, line 32: We'll have to see if this affects project load performance very much — previously we would only need to get a specific asset once. But now we will be relying on the browser cache to return this quickly.

src/serialization/deserialize-assets.js
```diff
@@ -22,37 +22,29 @@ const deserializeSound = function (sound, runtime, zip, assetFileName) {
...
-        storage.builtinHelper.cache(
-            storage.AssetType.Sound,
-            dataFormat,
-            data,
-            assetId
-        );
-    });
+    const dataFormat = sound.dataFormat.toLowerCase() === 'mp3' ?
+        storage.DataFormat.MP3 : storage.DataFormat.WAV;
+    return soundFile.async('uint8array').then(data => storage.createAsset(
```
rschamp, line 42: So one of the key points of this refactor is here — previously we'd ask `storage` to remember the contents of this asset. Now `storage` just creates it and we return it. The `storage` instance doesn't know about it after creating it.

src/serialization/sb2.js
```diff
@@ -452,7 +452,11 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
             // the file name of the costume should be the baseLayerID followed by the file ext
             const assetFileName = `${costumeSource.baseLayerID}.${ext}`;
             costumePromises.push(deserializeCostume(costume, runtime, zip, assetFileName)
-                .then(() => loadCostume(costume.md5, costume, runtime, 2 /* optVersion */)));
+                .then(asset => {
+                    costume.asset = asset;
```
rschamp, line 407: Here's the second part on that point — after we get the asset back from storage, cache it on the costume object. Previously it was cached in storage.

src/serialization/sb3.js
```diff
@@ -844,7 +844,10 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
         // any translation that needs to happen will happen in the process
         // of building up the costume object into an sb3 format
         return deserializeCostume(costume, runtime, zip)
-            .then(() => loadCostume(costumeMd5Ext, costume, runtime));
+            .then(asset => {
+                costume.asset = asset;
```
rschamp, line 848: Same deal but for sb3's

src/serialization/serialize-assets.js
```diff
@@ -16,12 +16,10 @@ const serializeAssets = function (runtime, assetType, optTargetId) {
         const currAssets = currTarget.sprite[assetType];
         for (let j = 0; j < currAssets.length; j++) {
             const currAsset = currAssets[j];
-            const assetId = currAsset.assetId;
-            const storage = runtime.storage;
-            const storedAsset = storage.get(assetId);
+            const asset = currAsset.asset;
```
rschamp, line 19: While serializing, pull the assets off of the costume/sound objects. Don't try to get them out of storage.

rschamp:
~~Tests will fail until LLK/scratch-storage#50 is merged, published and we update this PR to that version of storage.~~

New version published and updated, so they are passing now :)

src/import/load-costume.js
kchadha: Has the asset's data actually changed here though?

src/serialization/deserialize-assets.js
kchadha: We should probably test this on projects with lots of assets and on lots of devices...

src/serialization/deserialize-assets.js
```diff
@@ -22,37 +22,29 @@ const deserializeSound = function (sound, runtime, zip, assetFileName) {
...
+        storage.AssetType.Sound,
+        dataFormat,
+        data,
+        sound.assetId
```
kchadha, line 46: A note for the future... should we be checking that the asset ids actually match? Maybe we can do that on the storage side when creating the asset...

Or is that something that happens on the server side...?

src/import/load-costume.js
rschamp: I tried to make the changes for this move as little logic as possible and just update the API, to reduce variables so I didn't introduce regressions with the refactor. So in this case, the `encodeTextData` has the capability to set data, and generate a new assetId. But there may be cases where when it does this, it doesn't actually change the data that was already there. But that should have been true previously as well, but we were reassigning the assetId then too (the md5 was recalculated by the `cache` method instead).

I think after this is in and working, it would be a good idea to review the loading process and see if all of the information we're passing around is still necessary, or if we could just derive it from the asset when we needed to. I think we could also take that time to see if all the pre-existing logic is sound, or are we doing more work than we need to because we've done changes piecemeal and haven't looked at the process holistically in a while.

src/serialization/serialize-assets.js
```diff
@@ -16,12 +16,10 @@ const serializeAssets = function (runtime, assetType, optTargetId) {
...
             assetDescs.push({
-                fileName: `${assetId}.${storedAsset.dataFormat}`,
-                fileContent: storedAsset.data});
+                fileName: `${asset.assetId}.${asset.dataFormat}`,
```
kchadha, line 21: It seems like there's a lot of duplicate info flying around now...

e.g. both the costume/sound object and the assets themselves have assetId (and dataFormat?) now...

We should probably clean these at some point, but that's definitely a change for the future as I'm sure there's a lot of references to the properties on the top-level objects right now.

rschamp: Agreed — I sort of touched on that in the previous comment. There should be a second phase where we re-adjust everything to be better aware of the fact that there's an asset object sitting there that might have everything you need, so you don't need to have md5, assetId, etc. sitting on the top-level object at the same time.

src/virtual-machine.js
```diff
@@ -731,16 +734,16 @@ class VirtualMachine extends EventEmitter {
      *     a dataURI if it's a PNG or JPG, or null if it couldn't be found or decoded.
      */
     getCostume (costumeIndex) {
-        const id = this.editingTarget.getCostumes()[costumeIndex].assetId;
-        if (!id || !this.runtime || !this.runtime.storage) return null;
-        const format = this.runtime.storage.get(id).dataFormat;
+        const asset = this.editingTarget.getCostumes()[costumeIndex].asset;
+        if (!asset || !this.runtime || !this.runtime.storage) return null;
```
kchadha, line 738: Just to make sure, is there ever a case where we could have a sound/costume object in a functioning project and the `asset` property of the project is falsey?

test/integration/offline-custom-assets.js
```diff
@@ -52,22 +51,16 @@ test('offline-custom-assets', t => {
             t.equals(costumes.length, 1);
             const customCostume = costumes[0];
             t.equals(customCostume.name, 'A_Test_Costume');
-            const costumeMd5Ext = customCostume.md5;
-            const costumeIdParts = StringUtil.splitFirst(costumeMd5Ext, '.');
-            const costumeMd5 = costumeIdParts[0];
 
-            const storedCostume = vm.runtime.storage.get(costumeMd5);
+            const storedCostume = customCostume.asset;
```
kchadha, line 55: It's great that we can now replace all those lines of setup of trying to get the data out of storage with just this one line lookup. =D

kchadha (commented):
In general all the changes look good to me and also make sense! I made some questions/comments.

kchadha:
It looks like these changes are somewhat inter-dependent with the changes to the gui in LLK/scratch-gui#3491...

At least these changes shouldn't be pulled into the GUI without the changes in LLK/scratch-gui#3491 also being pulled in.

src/import/load-costume.js
kchadha: That sounds like a really good idea. =]

src/serialization/deserialize-assets.js
rschamp: The server will reject the asset upload if the contents don't match the md5 you send.

src/virtual-machine.js
rschamp: Maybe if there was some issue with creating assets while the project was loading? Not during normal functioning of the editor, as far as I'm aware.

PR #1696, opened 2018-10-24
Close Scratch Link web socket on every new peripheral scan attempt

### LINKED ISSUE #1671: Close web socket before making a new one for hardware extensions
From a test:

"I was connected to the EV3, then shut it down, restarted it, went to reconnect, and the scan never found the EV3 at all. Scratch Link showed 4 active sessions (seems to be the number of times I clicked “refresh” on the scanning step). Restarting bluetooth on the EV3 did not help. Restarting Chrome did not help. I was able to detect the same EV3 from a mac with Link 1.1.2. Restarting Scratch Link did not help."

evhan55:
### Resolves

- Resolves #1671: Close web socket before making a new one for hardware extensions
- Also fixes an error when LEGO WeDo2 modal step 1 was cancelled without attempting a scan

### Test Coverage

Should be tested on Windows with EV3

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -556,13 +556,12 @@ class EV3 {
      */
     scan () {
         if (this._bt) {
-            this._bt.requestPeripheral();
-        } else {
-            this._bt = new BT(this._runtime, this._extensionId, {
-                majorDeviceClass: 8,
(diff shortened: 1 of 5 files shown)


ericrosenbaum (approved):
LGTM! tested on windows with Scratch Link 1.1.9.0, and was able to disconnect and reconnect

PR #1697, opened 2018-10-24
Resolves #1687.  Clear peripheral discover timeout when needed.

### LINKED ISSUE #1687: Peripheral scan timeout is not correctly reset
### Expected Behavior

When you are scanning for devices, the scan should run for the same amount of time each time.

### Actual Behavior

Sometimes the scan times out early, when you scan multiple times in a row. This may be because the timeout is not being cleared correctly.

### Steps to Reproduce

- Add the microbit extension
- Press the "refresh" button several times
- Wait for the scan to timeout
- Press refresh again. The scan will timeout immediately.

scan-timeout

evhan55:
### Resolves

Resolves #1687: Peripheral scan timeout is not correctly reset

### Note

This needs PR https://github.com/LLK/scratch-vm/pull/1696 to land first in order to work correctly, since it is dependent on the socket now being closed on every scan attempt.

### DIFF
--- a/src/io/ble.js
+++ b/src/io/ble.js
@@ -38,6 +38,9 @@ class BLE extends JSONRPCWebSocket {
     requestPeripheral () {
         if (this._ws.readyState === 1) { // is this needed since it's only called on ws.onopen?
             this._availablePeripherals = {};
+            if (this._discoverTimeoutID) {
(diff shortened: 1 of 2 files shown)


ericrosenbaum (approved):
LGTM! Tested with Scratch Link 1.1.9 on both mac and windows

Issue #1689, opened 2018-10-22
Need tests to ensure current extensions load


ktbee:
### Expected Behavior

If a code change prevents an extension from loading, a failed test should indicate there is an issue. 

### Actual Behavior

No test necessarily covers the general case of ensuring all current extensions load. 

### Steps to Reproduce

An example case where a PR passes automated testing, but breaks the Pen extension in Scratch 3: https://github.com/LLK/scratch-vm/issues/1676

ktbee:
Closed in #1700

PR #1700, opened 2018-10-25
Tests to check whether extensions load

ktbee:
### Resolves

This PR is a work in progress for adding tests that check whether an extension has been successfully loaded in a project. This is meant to catch cases like #1676 and the issue for these tests is #1689 

### Proposed Changes

This adds a test that loads sample projects and then checks whether a particular extension was successfully loaded using `vm.extensionManager.isExtensionLoaded()`.

### Context and Questions

When I checked out SHA 899ce562 for the version of `develop` that had broken extensions for some projects (like Dolphins 3D, 115870836), I found my test worked as expected if my sample project was broken. In other words, when I created a simple sample project with just one extension-specific block (saved locally as `load-pen.sb2`), the test passed because the project wasn't broken in this branch. However, when I ran the test with Dolphins 3D, it was indeed failing and I confirmed in my localhost that that project was broken.

My questions:
- Should I use more complicated or real projects to test extension loading?
- If I should use real projects to test extensions, are there any you recommend I use? Perhaps the most popular ones that use a certain extension?
- Does the way I'm saving these test projects make sense? There are a lot of them, so I figured we'd want a separate directory for them. Let me know if that's not the case.
- I pictured adding example extension projects files to this test to become part of the workflow of adding extensions to Scratch. Does that sound right to you all and is there documentation I can update with this testing info once it's merged in?


nomadtechie:
@ktbee have you considered using something like `fs.readDir` to get the file paths of all the extensions which need to be loaded, and then just looping through one by one. The advantage if using this dynamic loop is you do not need to copy the extensions, and your tests will be valid and automatically cover for new extensions that get added to the project.

ktbee:
Thanks @nomadtechie for the suggestion! It pointed me in a good direction in terms of making this test dynamic and also much simpler. The most recent changes I've pushed get a listing of the files in the `load-extensions` directory and determine the extension they're intended to test from parsing the name (ie `const ext = file.split('-')[0]`)

kchadha (approved):
LGTM!

The JS integration test looks good! I haven't looked in the individual test fixtures, but I suspect they are just adding a block or more with extension mentioned in the name.

To answer some of your questions:

> * Should I use more complicated or real projects to test extension loading?

I think it's perfectly reasonable to have minimal repro cases like the ones you've added.

I was actually wondering if it would be possible for you to add tests for the other extensions we have  (even the ones that are new in 3.0). I don't think it should be considered a blocker for this PR to go in, though.

> * If I should use real projects to test extensions, are there any you recommend I use? Perhaps the most popular ones that use a certain extension?

Thanks for adding Dolphins 3D!

> * Does the way I'm saving these test projects make sense? There are a lot of them, so I figured we'd want a separate directory for them. Let me know if that's not the case.

The separate directory + dynamic way of loading all the fixtures the same way seems good we are doing similar things in other places like scratch-parser (and this might even be a better approach than the one we're using there).

> * I pictured adding example extension projects files to this test to become part of the workflow of adding extensions to Scratch. Does that sound right to you all and is there documentation I can update with this testing info once it's merged in?

I think that sounds like a good idea! @thisandagain, I'm wondering if you have any ideas of where this documentation should live?

ktbee:
Thanks for the feedback @kchadha ! I ended up being able to add example tests for all of the Scratch 2 and 3 extensions except for `translation` and `videoSensing`, which seem to need more of the Scratch ecosystem stubbed out for them to run without an error during testing. I've added this and implementation information in the README for the `test/fixtures/loading-extensions` directory. I'll make a follow up issue to add them as well.

PR #1701, opened 2018-10-25
Correct the equality operator such that Infinity is equal to Infinity

### LINKED ISSUE #1699: Scratch 3's treatment of ±Infinity isn't the same as Scratch 2
### Expected Behavior

In Scratch 2 the following expressions are all true:
< ( 1 / 0 ) = [Infinity] > where the right side is a string
< (var) = [Infinity] > when var was set to 1/0 (i.e. Infinity) and right side is a string
< [list v] contains [Infinity] > when it contains a numerical infinity (e.g. from 1/0)
< (var) = (var) > when var is ±Infinity (numerical, e.g. from ±1/0)
etc...

### Actual Behavior

In Scratch 3 they are all false.
The last one above is particularly unexpected.

### Steps to Reproduce

Test the expressions given above in Scratch 2 vs 3.

N/A

joker314:
### Resolves

This PR resolves #1699 

### Proposed Changes

In this pull request, we make the utility method `Cast.compare` behave more consistently with Scratch 2.0.

The `<(...) = (...)>` block asks this utility how its inputs relate to each other (smaller/equal/greater). If it gets a return value of 0, it considers them equal.

The algorithm Scratch 3.0 uses for number-like strings (like case-sensitive `Infinity` and `-Infinity`) is to subtract one argument from the other. If they're the same, the return value is expected to be zero.

However, when we pass two infinities of the same sign, there isn't an intuitive answer JS can give. Is the difference between Infinity and Infinity, zero or infinity?

Hence, we match the 2.0 algorithm (where we use `<`, `>` and `==`) to resolve the issue. However, to avoid duplicate code, we use the same logic for numbers and strings.

### Reason for Changes

We make these changes in an effort to make more projects run in a similar way in Scratch 2.0 and Scratch 3.0.

### Test Coverage

We introduce two automated tests:
- `Infinity = "Infinity"` - this is what the issue was talking about
- `NaN = NaN` - while this worked already, it seemed good to have this in a unit test

In addition, we verify the issue has been resolved manually.

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -116,18 +116,15 @@ class Cast {
         if (isNaN(n1) || isNaN(n2)) {
             // At least one argument can't be converted to a number.
             // Scratch compares strings as case insensitive.
-            const s1 = String(v1).toLowerCase();
-            const s2 = String(v2).toLowerCase();
-            if (s1 < s2) {
-                return -1;
-            } else if (s1 > s2) {
-                return 1;
-            }
-            return 0;
+            n1 = String(v1).toLowerCase();
+            n2 = String(v2).toLowerCase();
         }
-        // Compare as numbers.
-        return n1 - n2;
-
+        if (n1 < n2) {
+            return -1;
+        } else if (n1 > n2) {
+            return 1;
+        }
+        return 0;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


apple502j:
👎 Check "help wanted" tag before making PRs!

thisandagain:
Because of the tricky nature of getting this exactly right in each circumstance we are going to keep this internal.

joker314:
Hi @thisandagain , I actually don't think I agree:

I just wanted to note that this issue was not about Scratch's casting (it was doing that correctly).

Instead, it was about a collapsing of three lines into one from 2.0 to 3.0. This seemed to make the code more readable; but it failed for Infinity.

Hence, I don't think this is actually going to be any wider than the `compare` method. Nonetheless, perfectly happy for this to be internally fixed!

PR #1702, opened 2018-10-27
Many fixes about Text2Speech

### LINKED ISSUE #1620: Text to Speech extension stops speaking after # or &
### Expected Behavior

The text to speech extension should either attempt to speak any character it encounters, or skip it.

### Actual Behavior

If there's a hashtag or ampersand symbol (# or &), the speech stops at that point.

### Steps to Reproduce

Add a `speak` block and enter something like "one # two", and it will only say "one".

Note that after a quick exploration with @picklesrus it appears we simply need to use encodeURIComponent instead of encodeURI.

### LINKED ISSUE #1673: "Set language" menu in Text to Speech extension should accept dropped language names
### Expected Behavior

In the Text to Speech extension, the `set language` block should let you drop a block that reports a language name onto it. This should work for language names in any language. For example, you should be able to use the “language” reporter from the Translate extension to set the Text to Speech language.

### Actual Behavior

Dropped arguments in the `set language` block are generally ignored (technically right now you can drop in scratch locale codes, but that's not how it should work!).

### LINKED ISSUE #1678: Translate languages for Set language block (TTS Extension)
### Expected Behavior
Language names should be in Japanese when I use Scratch in Japanese.

### Actual Behavior
It says "Japanese" instead of "日本語".

### Steps to Reproduce
1. Load TTS Extension
2. Set the language to japanese

On Windows 10 Firefox 62

### LINKED ISSUE #1679: Kitten doesn't work in Japanese
### Expected Behavior
With these blocks Scratch should speak "meow".
image

### Actual Behavior
It just says "こんにちは".

### Steps to Reproduce
Load TTS Extension and add those blocks.
`こんにちは`

On Windows 10 Firefox 62

apple502j:
### Resolves
Resolves #1673 
Resolves LLK/scratch-gui#3425
Resolves #1620 
Resolves #1678 
Resolves #1679

Resolves problem which default language is English in ja-Hira (no issue)
Resolves todo "Translate meow"
Resolves a problem which TTS requested to API even the input is blank (no issue)

### Proposed Changes
**NOTE: It makes LANGUAGE_INFO revered! We used to use `{name: code}` but from now it's `{code: name}`**

* Translated language names with formatMessage. For translate extension, API did that. However, TTS API doesn't do that, so I implemented it via source code.
* Key-value swap is because keys can't be variables
* SCRATCH_LOCALES for dropped languages
* setCurrentLanguage:
** ja-Hira -> ja
** check from LANGUAGE_INFO and SCRATCH_LOCALES
* if the input is blank then return without calling API
* Translate meow with formatMessages
* Remove ASCII symbols. non-ASCII might be a todo.
* Use string length to replace with meow
* Use encodeURIComponent

### Reason for Changes
See each issues.

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -151,21 +151,103 @@ class Scratch3Text2SpeechBlocks {
      * An object with language names mapped to their language codes.
      */
     get LANGUAGE_INFO () {
+        const messages = {
+            danish: formatMessage({
+                id: 'text2speech.lang.danish',
+                default: 'Danish',
+                description: 'Danish language.'
+            }),
+            dutch: formatMessage({
+                id: 'text2speech.lang.dutch',
+                default: 'Dutch',
+                description: 'Dutch language.'
+            }),
+            english: formatMessage({
+                id: 'text2speech.lang.english',
+                default: 'English',
+                description: 'English language.'
+            }),
+            french: formatMessage({
+                id: 'text2speech.lang.french',
+                default: 'French',
+                description: 'French language.'
+            }),
+            german: formatMessage({
+                id: 'text2speech.lang.german',
+                default: 'German',
+                description: 'German language.'
+            }),
+            icelandic: formatMessage({
+                id: 'text2speech.lang.icelandic',
+                default: 'Icelandic',
+                description: 'Icelandic language.'
+            }),
+            italian: formatMessage({
+                id: 'text2speech.lang.italian',
+                default: 'Italian',
+                description: 'Italian language.'
+            }),
+            japanese: formatMessage({
+                id: 'text2speech.lang.japanese',
+                default: 'Japanese',
+                description: 'Japanese language.'
+            }),
+            polish: formatMessage({
+                id: 'text2speech.lang.polish',
+                default: 'Polish',
+                description: 'Polish language.'
+            }),
+            portuguese_brazilian: formatMessage({
+                id: 'text2speech.lang.portuguese_brazilian',
+                default: 'Portuguese (Brazilian)',
+                description: 'Portuguese (Brazilian) language.'
+            }),
+            portuguese: formatMessage({
+                id: 'text2speech.lang.portuguese',
+                default: 'Portuguese (Europian)',
+                description: 'Portuguese (Europian) language.'
+            }),
+            russian: formatMessage({
+                id: 'text2speech.lang.russian',
+                default: 'Russian',
+                description: 'Russian language.'
+            }),
+            spanish: formatMessage({
+                id: 'text2speech.lang.spanish',
+                default: 'Spanish (Europian)',
+                description: 'Spanish (Europian) language.'
+            }),
+            spanish_419: formatMessage({
+                id: 'text2speech.lang.spanish_419',
+                default: 'Spanish (Latin American)',
+                description: 'Spanish (Latin American) language.'
+            })
+        };
         return {
-            'Danish': 'da',
-            'Dutch': 'nl',
-            'English': 'en',
-            'French': 'fr',
-            'German': 'de',
-            'Icelandic': 'is',
-            'Italian': 'it',
-            'Japanese': 'ja',
-            'Polish': 'pl',
-            'Portuguese (Brazilian)': 'pt-br',
-            'Portuguese (European)': 'pt',
-            'Russian': 'ru',
-            'Spanish (European)': 'es',
-            'Spanish (Latin American)': 'es-419'
+            'da': messages.danish,
+            'nl': messages.dutch,
+            'en': messages.english,
+            'fr': messages.french,
+            'de': messages.german,
+            'is': messages.icelandic,
+            'it': messages.italian,
+            'ja': messages.japanese,
+            'pl': messages.polish,
+            'pt-br': messages.portuguese_brazilian,
+            'pt': messages.portuguese,
+            'ru': messages.russian,
+            'es': messages.spanish,
+            'es-419': messages.spanish_419
+        };
+    }
+
+    /**
+     * List of languages which uses different names on Scratch/Translate extension
+     */
+    get SCRATCH_LOCALES () {
+        return {
+            Portuguese: 'pt',
+            Spanish: 'es'
         };
     }
 
@@ -185,6 +267,7 @@ class Scratch3Text2SpeechBlocks {
             'is': 'is-IS', // Icelandic
             'it': 'it-IT', // Italian
             'ja': 'ja-JP', // Japanese
+            'ja-Hira': 'ja-JP', // Japanese (Hiragana)
             'pl': 'pl-PL', // Polish
             'pt-br': 'pt-BR', // Portuguese (Brazilian)
             'pt': 'pt-PT', // Portuguese (European)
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_text2speech/index.js
```diff
@@ -458,21 +550,29 @@ class Scratch3Text2SpeechBlocks {
...
-        // @todo localize this?
         if (state.voiceId === KITTEN_ID) {
-            words = words.replace(/\w+/g, 'meow');
+            const meow = formatMessage({
+                id: 'text2speech.meow',
+                default: 'meow',
+                description: 'Meow call used for kitten voice.'
+            });
+            words = meow.repeat(Math.ceil(
+                words.replace(/[\x20-\x2f\x3a-\x40\x7b-\x7e]/g, '').length / 4
```
joker314, line 567: This works for Japanese. Yet now this doesn't work for English!

What we want is:
- for Japanese (etc) to count by character
- for English (etc) to count by word

('meow' is a word, of course, and therefore must replace words)


It's also probably important to perform the division by 2 (like you did here -- though you divided by two twice, since you want _two_ characters to become one meow, I think(?)) for non-Japanese characters that are still represented with a low surrogate and a high surrogate pair (like emoji)

The solution for this is probably a little more complicated, therefore?

joker314 (commented):
Woah! Loads of fixes -- awesome! Just some questions/comments

src/extensions/scratch3_text2speech/index.js
joker314: Note: cats can no longer make dramatic&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;pauses: "meow!meow" is pronounced differently from "meowmeow", but this change strips punctuation (which I think is required by #LLK/scratch-gui#3425, yet the synth server doesn't seem to pronounce the punctuation when it's at the end/middle of words, yet it does pause)

Haven't had a chance to review the other changes: but they look super good from a glance, nice job!

joker314: On the other hand... splitting by syllables in English (etc) does makes some sense; maybe we should keep it like this. What are your thoughts?

src/extensions/scratch3_text2speech/index.js
```diff
@@ -458,21 +550,29 @@ class Scratch3Text2SpeechBlocks {
...
+            ));
         }
 
         // Build up URL
         let path = `${SERVER_HOST}/synth`;
         path += `?locale=${this.localeToPolly(this.getCurrentLanguage())}`;
         path += `&gender=${gender}`;
-        path += `&text=${encodeURI(words.substring(0, 128))}`;
+        path += `&text=${encodeURIComponent(words.substring(0, 128))}`;
```
joker314, line 575: :+1: #1620

src/extensions/scratch3_text2speech/index.js
apple502j: It's too hard to check if a language **always** uses whitespace to divide words.
And The kitten pronounces `antidisestablishmentarianism` and `hi` as `meow` - why is 28-letter word equal to 2-letter word?

src/extensions/scratch3_text2speech/index.js
```diff
@@ -366,6 +449,15 @@ class Scratch3Text2SpeechBlocks {
         // Only set the language if it is in the list.
         if (this.isSupportedLanguage(languageCode)) {
             stage.textToSpeechLanguage = languageCode;
+        } else if (languageCode === 'ja-Hira') {
+            // Japanese (Hiragana) is equal to Japanese here
+            stage.textToSpeechLanguage = 'ja';
```
Kenny2github, line 454: A lot of this seems to be specifically aimed at Hiragana. Are there any other languages with writing systems that might have the same problem as Hiragana does for Japanese?

src/extensions/scratch3_text2speech/index.js
Kenny2github: "an-ti-dis-es-ta-blish-men-tar-i-an-i-sm" 12 syllables vs "hi" 1 syllable is the more pressing question... unfortunately I don't think there's any good way to split words by syllable so splitting by words is the next best thing.

src/extensions/scratch3_text2speech/index.js
apple502j: I don't think so - Chinese might be(hant vs hans), but TTS doesn't support it

joker314: Maybe it would be a good idea to change `isSupportedLanguage` such that it is aware of the prefixes, and performs the conversion automatically for every language of the form `xx-yy`. Alternatively, you could add an easy-to-extend array/object of such languages and search through it (for now, it would only have one entry).

Perhaps, in the future, the TTS extension might support more languages.

asqwde: I hope it does

src/extensions/scratch3_text2speech/index.js
```diff
@@ -151,21 +151,103 @@ class Scratch3Text2SpeechBlocks {
      * An object with language names mapped to their language codes.
      */
     get LANGUAGE_INFO () {
+        const messages = {
+            danish: formatMessage({
+                id: 'text2speech.lang.danish',
+                default: 'Danish',
+                description: 'Danish language.'
```
chrisgarrity, line 158: These descriptions probably won't be enough context for the translators. It should probably be something like: 'How to refer to the Danish language in your language. '

src/extensions/scratch3_text2speech/index.js
```diff
@@ -151,21 +151,103 @@ class Scratch3Text2SpeechBlocks {
...
+                description: 'Japanese language.'
+            }),
+            polish: formatMessage({
+                id: 'text2speech.lang.polish',
+                default: 'Polish',
+                description: 'Polish language.'
+            }),
+            portuguese_brazilian: formatMessage({
+                id: 'text2speech.lang.portuguese_brazilian',
+                default: 'Portuguese (Brazilian)',
```
chrisgarrity, line 202: We should avoid things like parens in the name. The default should be 'Brazilian Portuguese'

src/extensions/scratch3_text2speech/index.js
```diff
@@ -151,21 +151,103 @@ class Scratch3Text2SpeechBlocks {
...
+                description: 'Portuguese (Brazilian) language.'
+            }),
+            portuguese: formatMessage({
+                id: 'text2speech.lang.portuguese',
+                default: 'Portuguese (Europian)',
+                description: 'Portuguese (Europian) language.'
```
chrisgarrity, line 208: In the interface the 'European' versions of Spanish and Portuguese are not referred to that way. They are simply 'Spanish' and 'Portuguese' (Also European has a spelling error).

src/extensions/scratch3_text2speech/index.js
```diff
@@ -151,21 +151,103 @@ class Scratch3Text2SpeechBlocks {
...
+                description: 'Russian language.'
+            }),
+            spanish: formatMessage({
+                id: 'text2speech.lang.spanish',
+                default: 'Spanish (Europian)',
+                description: 'Spanish (Europian) language.'
+            }),
+            spanish_419: formatMessage({
+                id: 'text2speech.lang.spanish_419',
+                default: 'Spanish (Latin American)',
```
chrisgarrity, line 222: The default should also be 'Latin American Spanish'

chrisgarrity (changes requested):
It also feels like there should be a way to have a consistent way to handle hiragana along with other languages that have multiple forms like Spanish and Portuguese.

ericrosenbaum:
Hi @apple502j sorry about the delay on this one. This PR tries to address several issues at once, and we would prefer to fix some of them in a different way. I'm going to need to make my own PRs to address the issues individually. Your code and the conversations here are a useful starting point though! I'll keep this open for now.

PR #1703, opened 2018-10-29
Cloud IO

kchadha:
### Proposed Changes

Adds a cloud IO device to send and receive cloud variable updates. This is a work in progress towards adding cloud variables to Scratch 3.0. With these changes, scratch 2.0 projects with cloud variables can be loaded in Scratch 3.0 when connected to the project page and the user can send and receive updates to cloud variables in the project. This PR does not allow creating new cloud variables or include any rate limiting logic or any validation of how many cloud variables are in the project.

#### Related PRs

LLK/scratch-gui#3539
LLK/scratch-www#2245

### Reason for Changes

Cloud variables!

### Test Coverage

Added unit tests for the new io device.

### DIFF
--- a/src/io/cloud.js
+++ b/src/io/cloud.js
@@ -1,0 +1,98 @@
+/**
+ * Cloud IO Device responsible for sending and receiving messages from
+ * cloud provider (mananging the cloud server connection) and interacting
+ * with cloud variables in the current project.
+ */
+
+const Variable = require('../engine/variable');
+const log = require('../util/log');
+
+class Cloud {
+    constructor () {
+        /**
+         * Reference to the cloud data provider, responsible for mananging
+         * the web socket connection to the cloud data server.
+         * @type{!object}
+         */
+        this.provider = null;
+
+        /**
+         * Reference to the stage target which owns the cloud variables
+         * in the project.
+         * @type{!Target}
+         */
+        this.stage = null;
+    }
+
+    /**
+     * Set a reference to the cloud data provider.
+     * @param {object} provider The cloud data provider
+     */
+    setProvider (provider) {
+        this.provider = provider;
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/io/cloud.js
```diff
@@ -0,0 +1,98 @@
...
+     * cloud variables in the project.
+     * @param {Target} stage The stage target
+     */
+    setStage (stage) {
+        this.stage = stage;
+    }
+
+    /**
+     * Handle incoming data to this io device.
+     * @param {object} data The data to process
```
rschamp: Would you please document the shape data is expected to be?  I gave it a try just to demonstrate the format for objects. I don't know if this syntax is actually valid though — you may need to use `typedef` to document and refer to the shape of `varUpdate` (but then you could reuse it below)
```suggestion
     * @param {object} data The data to process
     * @param {object} data.varUpdate The update object for a variable
     * @param {string} data.varUpdate.name The name of the variable to update
     * @param {number} data.varUpdate.value The value of the variable to update
```

src/io/cloud.js
```diff
@@ -0,0 +1,98 @@
...
+    requestUpdateVariable (name, value) {
+        if (this.provider) {
+            this.provider.updateVariable(name, value);
+        }
+    }
+
+    /**
+     * Update a cloud variable in the runtime based on the message received
+     * from the cloud provider.
+     * @param {object} varUpdate Data describing a cloud variable update received
```
rschamp: Similar request as above.

rschamp (changes requested):
Looks great! Just some more thorough docs would be nice.

kchadha:
@rschamp I added docs for the different types and confirmed that they look reasonable in the generated docs

rschamp (approved):

PR #1709, opened 2018-10-30
Shortcut Cast.toNumber if given a number

### LINKED ISSUE #1705: Review Cast.toNumber performance
### Expected Behavior

Cast.toNumber should perform its job as quickly as possible with reasonable code complexity.

### Actual Behavior

Cast.toNumber likely loses some performance by relying on `Number()` to do most of its work. `Number()` forces whatever value is passed to be coerced into a Number even when it may already be a number.

### Steps to Reproduce

Use code like testing `if (typeof value === 'number') return value;` to shortcut out of Cast.toNumber and measure its performance across different platforms to determine if this is a reasonable location for some performance work.

mzgoddard:
### Resolves

Close #1705 

### Proposed Changes

Check if `typeof value === 'number'` and do a separate isNaN check to confirm its not NaN.

### Reason for Changes

If toNumber is called on a number avoiding passing the number to Number() can provide a small performance improvement.

### Numbers

This change alone does not improve the blocks per second or fps in Scratch due to how the Scratch VM runs. Scratch iterates all threads before checking if it has used too much time instead of checking after each thread. I mention the later because that would be more likely to show a difference in blocks per second or fps (and we don't use that method because it would make projects nondeterministic).

With that in mind for this set of performance numbers I've recorded the percent of time spent in Cast.toNumber with browser inspector performance panels.

Together with other changes this will help speed up Scratch.

|  Project, Platform | before (% of time) | after (% of time) | Change |
|  ------ | ------ | ------ | ------ |
|  89811578, 2017 MBP 15" Chrome | 1.50% | 1.04% | 0.46% |
|  89811578, 2017 MBP 15" Firefox | 1.79% | 1.16% | 0.63% |
|  89811578, 2017 MBP 15" Safari | 2.00% | 2.00% | 0.00% |
|  89811578, 2017 SMS Chromebook | 1.58% | 0.90% | 0.68% |
|  89811578, 2018 Raspberry Pi 3B+ | 1.33% | 1.12% | 0.21% |

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -1,6 +1,14 @@
 const Color = require('../util/color');
 
 /**
+ * Store and possibly polyfill Number.isNaN. Number.isNaN can save time over
+ * self.isNaN by not coercing its input. We need to polyfill it to support
+ * Internet Explorer.
+ * @const
+ */
+const _NumberIsNaN = Number.isNaN || isNaN;
+
+/**
  * @fileoverview
  * Utilities for casting and comparing Scratch data-types.
  * Scratch behaves slightly differently from JavaScript in many respects,
@@ -25,14 +33,14 @@ class Cast {
         if (typeof value === 'number') {
             // Scratch treats NaN as 0, when needed as a number.
             // E.g., 0 + NaN -> 0.
-            if (isNaN(value)) {
+            if (_NumberIsNaN(value)) {
                 return 0;
             }
             return value;
         }
 
         const n = Number(value);
-        if (isNaN(n)) {
+        if (_NumberIsNaN(n)) {
             // Scratch treats NaN as 0, when needed as a number.
             // E.g., 0 + NaN -> 0.
             return 0;
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -20,6 +20,17 @@ class Cast {
      * @return {number} The Scratch-casted number value.
      */
     static toNumber (value) {
+        // If value is already a number we don't need to coerce it with
+        // Number().
+        if (typeof value === 'number') {
+            // Scratch treats NaN as 0, when needed as a number.
+            // E.g., 0 + NaN -> 0.
+            if (isNaN(value)) {
+                return 0;
+            }
+            return value;
+        }
+
         const n = Number(value);
         if (isNaN(n)) {
             // Scratch treats NaN as 0, when needed as a number.


kchadha (approved):
LGTM!

src/util/cast.js
```diff
@@ -20,6 +20,17 @@ class Cast {
      * @return {number} The Scratch-casted number value.
      */
     static toNumber (value) {
+        // If value is already a number we don't need to coerce it with
+        // Number().
+        if (typeof value === 'number') {
+            // Scratch treats NaN as 0, when needed as a number.
+            // E.g., 0 + NaN -> 0.
+            if (isNaN(value)) {
```
joker314: Is `Number.isNaN(value)` more efficient for a similar reason? `isNaN` still implicitly coerces to a number.

kchadha: @joker314, good catch!

kchadha (changes requested):
@mzgoddard, looks like lint is failing because of the naming convention for `Number_isNaN`

src/util/cast.js
mzgoddard: @joker314 That's a great point.

I made the change (with back compat support in mind for Internet Explorer which does not provide Number.isNaN, in which case self.isNaN is a safe backup) and recorded some new numbers.

|  Project, Platform | before (% of time) | after (% of time) | Change |
|  ------ | ------ | ------ | ------ |
|  89811578, 2017 MBP 15" Chrome | 1.50% | 0.93% | 0.57% |
|  89811578, 2017 MBP 15" Firefox | 1.79% | 1.24% | 0.55% |
|  89811578, 2017 MBP 15" Safari | 2.00% | 1.40% | 0.60% |
|  89811578, 2017 SMS Chromebook | 1.58% | 0.90% | 0.68% |
|  89811578, 2018 Raspberry Pi 3B+ | 1.33% | 0.95% | 0.38% |

Chrome sees a minor change with it.
Looks like I could have recorded my Firefox numbers better.
And Safari sees an improvement now.

mzgoddard:
Lint fixed.

src/util/cast.js
```diff
@@ -1,5 +1,13 @@
 const Color = require('../util/color');
 
+/**
+ * Store and possibly polyfill Number.isNaN. Number.isNaN can save time over
+ * self.isNaN by not coercing its input. We need to polyfill it to support
+ * Internet Explorer.
```
kchadha: Scratch 3.0 does not support IE, so we can take out that part of the comment.

kchadha (changes requested):
One last minor comment, and then after that LG.

kchadha (approved):
Actually, I think it's fine to leave the comment about IE as is. This PR LGTM.

src/util/cast.js
joker314: I figure I'd construct a table to show the change for just this commit. (I think I worked this out correctly?)

<table>
<tr><th>Project, Platform</th><th>isNaN (%)</th><th>Number.isNaN (%)</th><th>Change of Change (%)</tr>
<tr><td>89811578, 2017 MBP 15" Chrome</td><td>0.46</td><td>0.57</td><td><strong>+0.11</strong></td></tr>

<tr><td>89811578, 2017 MBP 15" Firefox</td><td>0.63</td><td>0.55</td><td><strong>-0.08</strong></td></tr>

<tr><td>89811578, 2017 MBP 15" Safari</td><td>0.00</td><td>0.60</td><td><strong>+0.60</strong></td></tr>

<tr><td>89811578, 2017 SMS Chromebook</td><td>0.68</td><td>0.68</td><td><strong>±0.00</strong></td></tr>

<tr><td>89811578, 2018 Raspberry Pi 3B+</td><td>0.21</td><td>0.38</td><td><strong>+0.17</strong></td></tr>
</table>

PR #1711, opened 2018-10-30
Import cloud var info and enforce cloud var limit

kchadha:
### Proposed Changes

Tracks whether a project uses cloud data in the runtime, and enforces a cloud data limit on loaded projects.

Also fixes a bug where cloud variables were not getting serialized correctly in 3.0 projects.

### Reason for Changes

Towards adding cloud variables to 3.0.

### Test Coverage

Added integration and unit tests for the cloud variable limit system and loading cloud variable data from 2.0 and 3.0 projects.


src/serialization/sb2.js
```diff
@@ -470,12 +470,21 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
...
+                // because it will update the cloud variable limit counter.
+                runtime.canAddNewCloudVariable();
             const newVariable = new Variable(
                 getVariableId(variable.name, Variable.SCALAR_TYPE),
                 variable.name,
                 Variable.SCALAR_TYPE,
-                variable.isPersistent
+                isCloud
             );
+            if (isCloud && !runtime.hasCloudData) runtime.hasCloudData = true;
```
kchadha: I ended up deciding to just use this check (here and in sb3) to set the runtime `hasCloudData` flag since it is more reliable and @thisandagain said that we're not using the existing sb2 flag for research.

test/integration/cloud_variables_sb2.js
```diff
@@ -0,0 +1,152 @@
+const path = require('path');
```
kchadha, line 1: This file and the analogous sb3 file are the same tests just running on sb2s vs. sb3s.

test/unit/engine_runtime.js
```diff
@@ -127,3 +127,38 @@ test('Project loaded emits runtime event', t => {
...
+    for (let i = 0; i < 8; i++) {
+        t.equal(rt.canAddNewCloudVariable(), true);
+    }
+
+    // We should be at the cloud variable limit now
+    t.equal(rt.canAddNewCloudVariable(), false);
+
+    // Removing a cloud variable should allow the addition of exactly one more
+    // when we are at the cloud variable limit
+    rt.removeExistingCloudVariable();
```
kchadha: Right now, this unit test is the only place we're using this function since creating/deleting cloud variable functionality doesn't exist yet.

src/engine/runtime.js
```diff
@@ -71,6 +71,47 @@ const ArgumentTypeMap = (() => {
...
+ * and remove an existing cloud variable.
+ * These are to be called whenever attempting to create or delete
+ * a cloud variable.
+ * @return {CloudDataManager} The functions to be used when adding or removing a
+ * cloud variable.
+ */
+const cloudDataManager = () => {
+    let cloudVariableLimit = 8;
+
+    const canAddNewCloudVariable = () => {
```
rschamp: `canAdd_` causing a decrement of the limit is unexpected since it reads like a query but in fact is an action. I think we should split this into `canAddNewCloudVariable` and a new `addCloudVariable` to match `removeExistingCloudVariable`.

src/engine/runtime.js
```diff
@@ -71,6 +71,47 @@ const ArgumentTypeMap = (() => {
...
+        if (cloudVariableLimit > 0) {
+            cloudVariableLimit--;
+            return true;
+        }
+        return false;
+    };
+
+    const removeExistingCloudVariable = () => {
```
rschamp: I think this would be more accurate as just `removeCloudVariable` (which could match `addCloudVariable` better) — it doesn't check for existence so it's a little misleading, if not over-specific.

src/serialization/sb2.js
rschamp: If we split out the query from the decrement, then we'd need to clean up here so this would become:
```suggestion
            if (isCloud) {
                runtime.addCloudVariable();
                runtime.hasCloudData = true;
           }
```

src/serialization/sb3.js
```diff
@@ -900,12 +900,21 @@ const parseScratchObject = function (object, runtime, extensions, zip) {
...
+                // because it will update the cloud variable limit counter.
+                runtime.canAddNewCloudVariable();
             const newVariable = new Variable(
                 varId, // var id is the index of the variable desc array in the variables obj
                 variable[0], // name of the variable
                 Variable.SCALAR_TYPE, // type of the variable
-                (variable.length === 3) ? variable[2] : false // isPersistent/isCloud
+                isCloud
             );
+            if (isCloud && !runtime.hasCloudData) runtime.hasCloudData = true;
```
rschamp: Same comment as above

rschamp (changes requested):
Just a few nits about naming and state.

src/engine/runtime.js
kchadha: Yeah, I was thinking about that too. What I did like about having them both in one function was that you couldn't forget to update the limit when setting the `isCloud` flag on the variable. Is there something else we can name this function to keep that functionality tightly coupled?

rschamp (changes requested):
Thanks for all the tests! They are failing though. It looks good otherwise.

kchadha:
@rschamp, I fixed up the tests. I was accidentally calling two of the cloud data manager functions during initialization. I have added unit tests that now checks that those are actually functions.

rschamp (approved):
The system is working!

Issue #1682, opened 2018-10-19
PicoBoard projects fail to load when using reporter blocks


ktbee:
### Expected Behavior

PicoBoard projects aren't supported in Scratch 3.0, so its blocks shouldn't load.

### Actual Behavior

Project fails to load entirely, just get error that "Oops! Something went wrong" on the screen, and this error in the console:

```
raven.js:55 Error: Failed to load project from server [id=#116599254]: TypeError: Cannot set property 'parent' of null
```

### Steps to Reproduce

1) Find a project that uses PicoBoard reporter blocks in non-PicoBoard blocks (example: 116599254)
2) Try to load project, observe it fails to load (https://llk.github.io/scratch-gui/develop/#116599254)
3) Remove reporter blocks from non-PicoBoard blocks (example: 254072184)
4) Observe it loads successfully and doesn't display PicoBoard blocks (https://llk.github.io/scratch-gui/develop/#254072184)

### Operating System and Browser

Mac OS 10.13.4 Chrome 69.0.3497.100

ktbee:
This bug is potentially similar to https://github.com/LLK/scratch-vm/pull/1524

ktbee:
closed via #1716 !

PR #1716, opened 2018-10-31
Handle unknown opcode in input

spectranaut:
### Resolves

https://github.com/LLK/scratch-vm/issues/1682

### Proposed Changes

Fix import error when block has an input with an unknown opcode.

### Reason for Changes

#1254 skips unknown opcodes when importing blocks from sb2 projects. However, that PR missed the case where an input is undefined. This PR fixes projects that were failing to load because of that issue.

### Test Coverage

This is my first issue, I'm not sure how to test other than manually with this project: https://llk.github.io/scratch-gui/develop/#116599254

I added a unit test called "unknown-opcode-as-reporter-block.js"

### How to test this

1. In Scratch 2.0, drag out a block with an input
2. Put a block with unknown opcodes in the input. Examples are the "picoboard" extension blocks, such as "sensor value"
3. Import the project into 3.0
4. The project should be able to load and there will be a default value (shadow) instead of the input. There should be warnings about skipped opcodes in the console.


src/serialization/sb2.js
```diff
@@ -843,6 +843,7 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
     for (let i = 0; i < blockMetadata.argMap.length; i++) {
         const expectedArg = blockMetadata.argMap[i];
         const providedArg = sb2block[i + 1]; // (i = 0 is opcode)
+
```
kchadha: Nit: stray newline

src/serialization/sb2.js
```diff
@@ -866,32 +867,35 @@ const parseBlock = function (sb2block, addBroadcastMsg, getVariableId, extension
...
+                    innerBlocks = parsedBlockDesc[0] ? [parsedBlockDesc[0]] : [];
                     // Update commentIndex
                     commentIndex = parsedBlockDesc[1];
                 }
                 parseState.expectedArg = parentExpectedArg;
-                // Check if innerBlocks is an empty list.
-                // This indicates that all the inner blocks from the sb2 have
+
+                // Obscures any shadow.
+                shadowObscured = true;
```
kchadha: You should move this line below the `if (innerBlocks.length > 0)` block below since it's not being used till later

kchadha (changes requested):
A few minor comments, and then it's good to go!

spectranaut:
addressed your feedback, @kchadha ! :)

kchadha (approved):
LGTM!

PR #1719, opened 2018-11-01
Handle images uploaded as sprites

rschamp:
### Resolves

Resolves LLK/scratch-gui#3582

### Reason for Changes

The sprite upload code in GUI serializes asset data, flattening it. We need to rehydrate the asset so that it can be used.


kchadha:
See comments for this PR on the gui version of this hotfix: LLK/scratch-gui#3603

PR #1722, opened 2018-11-02
Handle coordinate precision the same as Scratch 2

### LINKED ISSUE #1242: Compatibility - number precision is handled differently in Scratch 2 and 3
### Expected Behavior

In [this project](https://scratch.mit.edu/projects/229606968/), the sprite should slide to the x and y positions defined in the sprite variables, then move to another random position.

### Actual Behavior

The sprite slides to the x and y position defined in the sprite variables, but never escapes the repeat until loop.

### Steps to Reproduce

Load the project in [the Scratch 3.0 preview](https://llk.github.io/scratch-gui/develop/#229606968), and press the green flag. After a while, click the 'y value difference' variable block in the workspace. It will be stuck at a value very close but not equal to zero (Scratch 2 rounds this to 0).

macOS 10.13.5, Firefox Nightly 62, Chrome 67, Safari 11.1.1

ktbee:
### Resolves

Fixes #1242 

### Proposed Changes

Instead of only rounding the coordinates to 8 places, we use [the same logic as Scratch 2](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/primitives/MotionAndPenPrims.as#L225-L230) to check whether the difference is negligible between the current coordinate and the rounded version of the coordinate. If the difference is less than a certain threshold, use the rounded version. This PR also removes `_roundCoord`, which was a previous approach to rounding coordinates. 

### Reason for Changes

[`_roundCoord` previously limited coordinates to 8 decimal places](https://github.com/LLK/scratch-vm/pull/1160), but it used different logic to limit coordinate precision than Scratch 2 did. The difference in logic however causes incompatibility for some projects, like https://llk.github.io/scratch-gui/develop/#229606968. 

### Test Coverage

I've added a unit test for the motion block's limited coordinates in `test/unit/blocks_motion.js` that uses some of the same logic as this test Scratch project: https://llk.github.io/scratch-gui/develop/#223591096

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -270,7 +270,7 @@ class Scratch3MotionBlocks {
         return util.target.direction;
     }
 
-    // This corresponds to snapToInteger in Scratch 2's
+    // This corresponds to snapToInteger in Scratch 2
     limitPrecision (coordinate) {
         const rounded = Math.round(coordinate);
         const delta = coordinate - rounded;
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -272,8 +272,8 @@ class Scratch3MotionBlocks {
 
     limitPrecision (coordinate) {
         const rounded = Math.round(coordinate);
-        const epsilon = coordinate - rounded;
-        const preciseCoord = (Math.abs(epsilon) < 1e-7) ? rounded : coordinate;
+        const delta = coordinate - rounded;
+        const preciseCoord = (Math.abs(delta) < 1e-7) ? rounded : coordinate;
 
         return preciseCoord;
     }
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -274,9 +274,9 @@ class Scratch3MotionBlocks {
     limitPrecision (coordinate) {
         const rounded = Math.round(coordinate);
         const delta = coordinate - rounded;
-        const preciseCoord = (Math.abs(delta) < 1e-9) ? rounded : coordinate;
+        const limitedCoord = (Math.abs(delta) < 1e-9) ? rounded : coordinate;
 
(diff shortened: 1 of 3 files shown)


src/blocks/scratch3_motion.js
```diff
@@ -259,16 +259,25 @@ class Scratch3MotionBlocks {
...
     getY (args, util) {
-        return util.target.y;
+        return this.limitPrecision(util.target.y);
     }
 
     getDirection (args, util) {
         return util.target.direction;
     }
+
+    // This corresponds to snapToInteger in Scratch 2's
```
kchadha: Nit: `Scratch 2's` -> `Scratch 2`?

ktbee: Good catch, fixed!

kchadha:
@ktbee, the code looks good! However, I wasn't able to see a difference in behavior of the test project from the issue being fixed: project 229606968

ktbee:
Thanks for taking a look @kchadha ! I should have mentioned in my PR, the example project 229606968 is fixed, but it's not 100% obvious it's fixed because the say bubble's number precision is limited to two places, so it's not showing that the difference between coordinates is reducing in the same way it does in Scratch 2. If you take a look at that project with this fix and wait at ~ 7 seconds after the point hits zero, you'll see that it eventually moves to another point again.

kchadha:
@ktbee, that makes sense! I'll confirm one more time that I'm seeing the change you mention and then I think this PR is good to go.

I was confused by the say bubble rounding (as you mentioned). I just confirmed with @thisandagain, that the say bubble rounding was an intentional change from 2.0 -> 3.0 and it should be documented in the [vm compatibility wiki page](https://github.com/LLK/scratch-vm/wiki/Compatibility). =]

kchadha (approved):
LGTM

PR #1727, opened 2018-11-05
Add unloaded extensions on block sharing

paulkaplan:
Fixes an issue where trying to share blocks from the backpack that include not-previously-loaded extensions does not load them, causing mayhem.

@kchadha added unit tests for the helper and new functionality in shareBlocksToTarget. Let me know if the refactoring was done the way you were expecting.

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -274,7 +274,7 @@ const compressInputTree = function (block, blocks) {
 /**
  * Get non-core extension ID for a given sb3 opcode.
  * @param {!string} opcode The opcode to examine for extension.
- * @return {?string} The extension ID, if it exists/is not a core extension.
+ * @return {?string} The extension ID, if it exists and is not a core extension.
  */
 const getExtensionIdForOpcode = function (opcode) {
     const index = opcode.indexOf('_');
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -272,6 +272,19 @@ const compressInputTree = function (block, blocks) {
 };
 
 /**
+ * Get non-core extension ID for a given sb3 opcode.
+ * @param {!string} opcode The opcode to examine for extension.
+ * @return {?string} The extension ID, if it exists/is not a core extension.
+ */
+const getExtensionIdForOpcode = function (opcode) {
+    const index = opcode.indexOf('_');
+    const prefix = opcode.substring(0, index);
+    if (CORE_EXTENSIONS.indexOf(prefix) === -1) {
+        if (prefix !== '') return prefix;
+    }
+};
+
+/**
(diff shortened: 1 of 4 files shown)


src/serialization/sb3.js
```diff
@@ -271,6 +271,19 @@ const compressInputTree = function (block, blocks) {
     return block;
 };
 
+/**
+ * Get non-core extension ID for a given sb3 opcode.
+ * @param {!string} opcode The opcode to examine for extension.
+ * @return {?string} The extension ID, if it exists/is not a core extension.
+ */
+const getExtensionIdForOpcode = function (opcode) {
```
kchadha, line 279: I think this should probably be a utility function instead of living in `sb3.js`. I think it's used in a bunch of places including `sb2.js`, so it would be useful to abstract it out further.

src/serialization/sb3.js
```diff
@@ -271,6 +271,19 @@ const compressInputTree = function (block, blocks) {
     return block;
 };
 
+/**
+ * Get non-core extension ID for a given sb3 opcode.
+ * @param {!string} opcode The opcode to examine for extension.
+ * @return {?string} The extension ID, if it exists/is not a core extension.
```
kchadha: Nit: replace the `exists/is` with `exists and is`

test/unit/serialization_sb3.js
```diff
@@ -221,3 +221,14 @@ test('deserializeBlocks on already deserialized input', t => {
             t.end();
         });
 });
+
+test('getExtensionIdForOpcode', t => {
+    t.equal(sb3.getExtensionIdForOpcode('wedo_loopy'), 'wedo');
+
+    // does not consider CORE to be extensions
+    t.false(sb3.getExtensionIdForOpcode('control_loopy'));
```
kchadha, line 229: Could you also add a test for opcodes without `_` in the name?

kchadha (commented):
Minor comments in the code changes.

src/serialization/sb3.js
paulkaplan: Is the list of `CORE_EXTENSIONS` necessarily the same between sb3/sb2? Or should the utility consume that list?

paulkaplan: Given the CORE_EXTENSIONS stuff, I think this probably should live in sb3. Either that or we need to change how we have 3 declarations of a constant called "CORE_EXTENSIONS", some of which are different.

paulkaplan:
@kchadha fixed comments other than the one about the location of the utility. We can talk about that IRL if you want.

kchadha:
@paulkaplan, given the question you raised and the fact that the CORE_EXTENSIONS constant is spread out, we can defer moving the utility for later.

kchadha (approved):

PR #1730, opened 2018-11-06
Make volume a sprite-specific monitor

ericrosenbaum:
### Resolves

Part of https://github.com/LLK/scratch-gui/issues/3656

### Proposed Changes

Set the correct properties for the volume monitor.

### Reason for Changes

The volume monitor should be sprite-specific.


kchadha (approved):
LGTM

Issue #1412, opened 2018-08-01
Changing volume doesn't affect notes that are already playing


TheLogFather:
### Expected Behavior

In Scratch 2, changing volume in a sprite/clone has an effect on notes that are already playing in that sprite/clone.

### Actual Behavior

In Scratch 3 the volume change has no effect on notes that have already started.
(Note, however, that it does appear to affect sounds that are playing.)

### Steps to Reproduce

Create a simple script with a "repeat (10)" loop containing "change volume by (-10)".
If volume is not already full, click on a block "set volume to (100)".
Click on a block "set instrument to [(3) Organ]".
Click on a block "play note (60) for (5) beats".
Straight away, click on the repeat loop script.

In Scratch 2 the volume of the organ note decays to nothing due to the volume change.
In Scratch 3 the organ note remains the same volume until it finishes (early, due to https://github.com/LLK/scratch-vm/issues/1411).

### Operating System and Browser

N/A

PR #1735, opened 2018-11-06
Apply sprite's volume to music extension notes and drums

### LINKED ISSUE #1407: Pan effect works incorrectly with music extension
### Expected Behavior

The pan effect should correctly apply to sounds played by the music extension. In particular, setting the pan to 100 should cause the sound from a "play note" block to play only in the right channel. The left channel should be silent.

### Actual Behavior

The pan effect on sounds played with the "play note" and "play drum" blocks is incorrect in a weird way. For example, run this stack:

screen shot 2018-07-31 at 2 34 19 pm

If you use headphones, and listen to only the left side (which should be silent), you'll hear the very beginning of each note.

### LINKED ISSUE #1732: Volume is not applied until after the beginning of each music extension note
### Expected Behavior

When you set the volume, then play a note using the "play note" block, the note should play at that volume.

### Actual Behavior

Currently, we appear to be applying the volume a short time after the note starts. As a result, if you set the volume to 0, then play notes, you hear a very short, full volume version of each note. 

At intermediate volumes, you hear a very short, full volume version of the note, followed by the rest of it at the correctly set volume.

I believe this is essentially the same bug as this one about the pan effect: https://github.com/LLK/scratch-vm/issues/1407

### Steps to Reproduce

Add the music extension, and listen to the result of each stack:
screen shot 2018-11-06 at 11 09 10 am

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1732
Resolves https://github.com/LLK/scratch-vm/issues/1407

### Proposed Changes

Don't use the audio engine's effects chain feature in the music extension. Instead, create gain nodes and use those to apply the sprite's volume setting to each note or drum sound when it starts playing. The volume is set instantaneously, before the sound starts. This prevents the bug in #1732. 

This also resolves #1407 by disabling the pan effect for the music extension. 

### Reason for Changes

Previously, we were using the effects chain to apply the volume, but it has a built in transition time (to prevent clicking sounds due to sudden transitions). The unintended consequence was that every note or drum sound was played initially at 100% volume, and then quickly transitioned to the set volume, so even when the volume is 0%, you would hear the onset of the note or drum sound. This change sets the volume instantaneously before playback.

Note that this change does not deal with https://github.com/LLK/scratch-vm/issues/1412, because the volume does not get updated after the sound starts playing.

We also decided to disable the pan effect for the music extension. We are generally moving toward an approach of not "crossing the streams", meaning we don't want settings controlled by blocks outside the extension (such as the "set pan effect" and "set pitch effect" blocks) to interact with the extension. Volume affecting the music extension is an exception, but it's necessary because it's a pre-existing feature.

### Testing

Compare the sound of this recently featured note blocks project before and after the change: 257504394

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -985,18 +985,18 @@ class Scratch3MusicBlocks {
             player.take();
         }
 
-        const chain = engine.createEffectChain();
-        chain.setEffectsFromTarget(util.target);
-
         // Set its pitch.
         const sampleNote = sampleArray[sampleIndex];
         const notePitchInterval = this._ratioForPitchInterval(note - sampleNote);
 
-        // Create a gain node for this note, and connect it to the sprite's
-        // simulated effectChain.
+        // Create gain nodes for this note's volume and release, and chain them
+        // to the output.
         const context = engine.audioContext;
+        const volumeGain = context.createGain();
+        volumeGain.gain.setValueAtTime(util.target.volume / 100, engine.currentTime);
         const releaseGain = context.createGain();
-        releaseGain.connect(chain.getInputNode());
+        volumeGain.connect(releaseGain);
+        releaseGain.connect(engine.getInputNode());
 
         // Schedule the release of the note, ramping its gain down to zero,
         // and then stopping the sound.
@@ -1018,7 +1018,7 @@ class Scratch3MusicBlocks {
         player.play();
         // Connect the player to the gain node.
         player.connect({getInputNode () {
-            return releaseGain;
+            return volumeGain;
         }});
         // Set playback now after play creates the outputNode.
         player.outputNode.playbackRate.value = notePitchInterval;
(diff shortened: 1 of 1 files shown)


apple502j:
Question: Should TTS also use this value?

ericrosenbaum:
> Question: Should TTS also use this value?

The note I made above explains why we have chosen not to do that:

>we don't want settings controlled by blocks outside the extension (such as the "set pan effect" and "set pitch effect" blocks) to interact with the extension. Volume affecting the music extension is an exception, but it's necessary because it's a pre-existing feature.

apple502j:
NOTE Re-open #1412 after merging! This closes that issue because of "does not **resolve**"

ericrosenbaum:
> NOTE Re-open #1412 after merging! This closes that issue because of "does not **resolve**"

@apple502j good catch- I fixed it by changing the description here

thisandagain (approved):
Looks good! Thanks @ericrosenbaum

PR #1737, opened 2018-11-06
Text layer

fsih:
### Resolves
Fixes https://github.com/LLK/scratch-vm/issues/672

On a high level, loadBitmap_ now calls fetchBitmapCanvas_ instead of loading assets itself.  fetchBitmapCanvas_ is based on towerofnix's work and now performs both bitmap resolution adjustment and gluing together of assets.

### Proposed Changes
Acknowledge the text layer in sb2 files
When we encounter one, merge it with the base layer and save as a new asset. Scratch 3 will forget about the editable properties of the text.

### Reason for Changes
Text layer was being ignored before, causing old files first created in Scratch 1 to lose their text

### Test Coverage
Tested projects 124310159, 173088631, others mentioned in https://github.com/LLK/scratch-vm/issues/672


fsih:
Why are reviews not required to merge on VM?

cwillisf:
@fsih I think this repository was created before we knew about that setting, or maybe even before that setting existed(?), and we just never went back and added the setting.

src/import/load-costume.js
```diff
@@ -30,76 +30,149 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+            delete costume.textLayerMD5;
+            delete costume.textLayerAsset;
+        });
+};
+
+const loadBitmap_ = function (costume, runtime, rotationCenter) {
+    return fetchBitmapCanvas_(costume, runtime, rotationCenter).then(fetched => new Promise(resolve => {
+        rotationCenter = fetched.rotationCenter;
+
+        const saveAssetToStorage = function (dataURI) {
```
cwillisf: I would name this function something else, like `makeAsset` or `createAsset`, since it doesn't actually "save" anything to storage. (I guess this was probably calling `cache` at some point in the past?)

src/import/load-costume.js
```diff
@@ -110,14 +183,14 @@ const loadBitmap_ = function (costume, costumeAsset, runtime, rotationCenter) {
  * @property {number} rotationCenterX - the X component of the costume's origin.
  * @property {number} rotationCenterY - the Y component of the costume's origin.
  * @property {number} [bitmapResolution] - the resolution scale for a bitmap costume.
- * @param {!Asset} costumeAsset - the asset of the costume loaded from storage.
+ * @param {!Asset} costume.asset - the asset of the costume loaded from storage.
```
cwillisf: This should use `@property`

src/serialization/deserialize-assets.js
```diff
@@ -106,13 +110,44 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
...
 
-    return costumeFile.async('uint8array').then(data => storage.createAsset(
-        assetType,
-        // TODO eventually we want to map non-png's to their actual file types?
-        costumeFormat,
-        data,
-        assetId
-    ));
+    // textLayerMD5 exists if there is a text layer, which is a png of text from Scratch 1.4
+    // that was opened in Scratch 2.0. In this case, set costume,textLayerAsset.
```
cwillisf: typo: `costume,textLayerAsset` should be `costume.textLayerAsset`

src/import/load-costume.js
```diff
@@ -30,76 +30,149 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+ *     assetMatchesBase is true if the asset matches the base layer; false if it required adjustment
+ */
+const fetchBitmapCanvas_ = function (costume, runtime, rotationCenter) {
+    if (!costume || !costume.asset) {
+        return Promise.reject('Costume load failed. Assets were missing.');
+    }
     return new Promise((resolve, reject) => {
-        const imageElement = new Image();
+        const baseImageElement = new Image();
+        const textImageElement = new Image();
```
cwillisf: Would it make sense to delay `textImageElement = new Image()` until & unless we need it? It looks like this could move down into the `if (costume.textLayerAsset)` block and avoid creating an unnecessary extra `Image` per costume. It's probably not a big deal but if it's as easy a change as it looks then it might be worth it to decrease GC noise, especially for memory-constrained devices loading projects with many costumes.

cwillisf (changes requested):
I picked a few nits here and there but for the most part this looks great! I'm glad you were able to build on @towerofnix's work. I felt kinda sad when I closed #915!

Also: please delete `src/serialization/.sb2.js.swp` -- looks like a `vim` swap file accidentally made it into one of the commits.

src/import/load-costume.js
fsih: Sure, I had just left it that way because I thought it was a bit easier to read without the if blocks.

fsih:
I'm hitting a wall with writing the integration test. I added functions to fake-renderer so that the test wouldn't skip loadCostumeFromAsset due to no rendering module being present, but now it fails because it doesn't know how to instantiate Image. Is there a better way I should be going about this? @kchadha @cwillisf

cwillisf:
Unfortunately I don't think we'll be able to truly test rendering in this repository without something like the Jest/Selenium setup we have in scratch-gui.

It does look like you could test `deserializeCostume` without a DOM, but that might be the only low-hanging fruit. If you add a way to mock `loadCostume` and `loadSound` you might be able to make tests for `parseScratchObject` in `sb2.js` and `sb3.js`.

Another approach would be to put a text-layer load & render test into scratch-gui, where you could actually do the render and see if it turns out alright.

fsih:
Trying to put the test in gui sounds like a good idea!

One more question about this PR, there are a lot of places where I catch an error only to return reject(e). Is that necessary, or will promises handle errors in chains of thens for you?

cwillisf:
In general I believe not catching the exception would be equivalent but there's at least one subtlety to look out for:
https://stackoverflow.com/questions/33445415/javascript-promises-reject-vs-throw

cwillisf (dismissed):
👍

fsih:
Thanks!

I'm going to wait for @kchadha to take a quick look before putting it in, just to make sure I got all the things we discussed doing.

src/import/load-costume.js
```diff
@@ -30,76 +30,151 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+ * @param {?object} rotationCenter - optionally passed in coordinates for the center of rotation for the image. If
+ *     none is given, the rotation center of the costume will be set to the middle of the costume later on.
+ * @property {number} costume.bitmapResolution - the resolution scale for a bitmap costume.
+ * @returns {?Promise} - a promise which will resolve to an object {canvas, rotationCenter, assetMatchesBase},
+ *     or reject on error.
+ *     assetMatchesBase is true if the asset matches the base layer; false if it required adjustment
+ */
+const fetchBitmapCanvas_ = function (costume, runtime, rotationCenter) {
+    if (!costume || !costume.asset) {
+        return Promise.reject('Costume load failed. Assets were missing.');
```
kchadha, line 49: Is this a case that we expect or just general error handling? If it's a case we expect in any sort of normal execution, maybe add a comment about which situations would lead to this..

src/import/load-costume.js
```diff
@@ -30,76 +30,151 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
             // eslint-disable-next-line no-use-before-define
             removeEventListeners();
-            reject('Image load failed');
+            reject('Costume load failed. Asset could not be read.');
         };
         const onLoad = function () {
-            // eslint-disable-next-line no-use-before-define
-            removeEventListeners();
-            resolve(imageElement);
+            if (loadedOne) {
```
kchadha, line 69: Maybe add a quick comment about what loadedOne represents. It was a little confusing at first. =D

src/import/load-costume.js
```diff
@@ -30,76 +30,151 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+        canvas.height = baseImageElement.height;
+
+        const ctx = canvas.getContext('2d');
+        ctx.drawImage(baseImageElement, 0, 0);
+        if (textImageElement) {
+            ctx.drawImage(textImageElement, 0, 0);
+        }
+        if (scale !== 1) {
+            if (!runtime.v2BitmapAdapter) {
+                return Promise.reject('No V2 Bitmap adapter present.');
```
kchadha: Does it make sense to move this check/fail earlier so that we don't do the canvas draw image stuff unnecessarily?

src/import/load-costume.js
```diff
@@ -30,76 +30,151 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+            rotationCenter: rotationCenter,
+            // True if the asset matches the base layer; false if it required adjustment
+            assetMatchesBase: scale === 1 && !textImageElement
+        };
+    })
+        .catch(e => Promise.reject(e))
+        .finally(() => {
+            // Clean up the costume object
+            delete costume.textLayerMD5;
+            delete costume.textLayerAsset;
```
kchadha, line 134: 🎉

src/import/load-costume.js
```diff
@@ -30,76 +30,151 @@ const loadVector_ = function (costume, costumeAsset, runtime, rotationCenter, op
...
+            costume.assetId = costume.asset.assetId;
+            costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
+        };
+
+        if (!fetched.assetMatchesBase) {
+            updateCostumeAsset(fetched.canvas.toDataURL());
+        }
+        resolve(fetched.canvas);
+    }))
+        .catch(e => Promise.reject(e))
```
kchadha: Oh wow, I didn't know you could put a catch in the middle of a then chain... does it make a difference being in the middle rather than at the end?

src/serialization/deserialize-assets.js
```diff
@@ -79,7 +81,9 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
             costume.asset.dataFormat,
             new Uint8Array(Object.keys(costume.asset.data).map(key => costume.asset.data[key])),
             costume.asset.assetId
-        ));
+        )).then(asset => {
+            costume.asset = asset;
```
kchadha, line 85: ~I'm a little confused about this addition...~

kchadha: Nvm, I figured it out. =]

src/serialization/deserialize-assets.js
```diff
@@ -106,13 +110,44 @@ const deserializeCostume = function (costume, runtime, zip, assetFileName) {
...
-        costumeFormat,
-        data,
-        assetId
-    ));
+    // textLayerMD5 exists if there is a text layer, which is a png of text from Scratch 1.4
+    // that was opened in Scratch 2.0. In this case, set costume.textLayerAsset.
+    let textLayerFilePromise;
+    if (costume.textLayerMD5) {
+        textLayerFileName = textLayerFileName ? textLayerFileName :
+            `${costume.textLayerMD5}.png`;
```
kchadha: Does textLayer info behave the same way that baseLayer info does? E.g. in sb2 files, there's a baseLayerMD5 which is the asset ID and there's the actual file name which is something like 0.png, does the textLayer also have a similar concept or is its file name assetID + extension?

Mostly I'm wondering if this `textLayerFileName` param is necessary.

kchadha (commented):
The code looks pretty good! I had some questions/comments in the code review.

src/import/load-costume.js
fsih: This should never happen; loadCostumeFromAsset should always be called with an asset loaded. But if it ever does it should fail cleanly instead of the weird thing it was doing before.

src/import/load-costume.js
fsih: This is another thing that shouldn't happen, except in tests (like no renderer attached.) I've moved it to the top of the function.

src/import/load-costume.js
fsih: The catch documentation says,

> It behaves the same as calling Promise.prototype.then(undefined, onRejected) (in fact, calling obj.catch(onRejected) internally calls obj.then(undefined, onRejected)).

This is what I was asking about in comment https://github.com/LLK/scratch-vm/pull/1737#issuecomment-438817340 , I think I don't actually need the catch here, but I don't understand how promise rejection works. Say the promise right before the catch here rejects. Then does the rejection bubble all the way up the chain automatically? Or do I need to manually pass the rejection back to the previous then, using this catch, otherwise it will return Promise.resolve(null)?

src/serialization/deserialize-assets.js
fsih: I think textLayerFileName works exactly the same way as assetFileName. I'm not sure what you mean about it not being necessary. Is there somewhere else I should be getting it?

src/import/load-costume.js
kchadha: I think the rejection should bubble up the chain automatically. Is there a test we can set up that ensures this?

src/serialization/deserialize-assets.js
kchadha: The reason `assetFileName` is provided as an argument to this function and the reason that 
https://github.com/LLK/scratch-vm/pull/1737/files#diff-7fd991d2b1bb7163e8a10acfcad7c3e3R68
exists is because sb2s and sb3s have different file name structures. The call to `deserializeCostume` from the sb3 deserialization code does not provide the `assetFileName` argument, so the `or` case in that line is for the `sb3` case.

Since we only expect the text layer special case to be part of the sb2 deserialization process, we can probably just have the sb2 import code add the filename to the costume as a temporary property that we look for here and then delete it when this function successfully deserializes the costume. 

The downside of that is that we could refactor the code in deserializeCostume later down the line and forget to delete the properties, so I'm also okay to leave the argument as is, but I think this line (e.g. ``textLayerFileName = textLayerFileName ? textLayerFileName : `${costume.textLayerMD5}.png`;``) should be removed.

kchadha (commented):
@fsih, I answered some of the questions you had about my previous comments.

src/import/load-costume.js
fsih: I want to add a test for loadCostume but I can't in VM because tests in VM don't recognize browser elements. Hopefully I can add one in GUI.

I made a little JS fiddle that shows that the error bubbles up through thens, even if one of the thens in the chain explicitly catches the error and chooses to ignore it... so I'll remove these catches that do nothing.
https://jsfiddle.net/c298frvb/

kchadha (approved):
LGTM! Thanks for making those last changes.

PR #1740, opened 2018-11-07
Fix flacky test: order-changes-backwards-2-broadcast-and-wait.sb2

### LINKED ISSUE #1677: order-changes-backwards-2-broadcast-and-wait.sb2 test is flaky
### Expected Behavior
Tests should not be flaky

### Actual Behavior
This test failed 3 out of 4 times on Travis
Tested locally and had 5/5 success rate

```
not ok 12 - ./test/integration/execute.js # time=7807.039ms
 ---
 timeout: 240000
 file: ./test/integration/execute.js
 command: /home/travis/.nvm/versions/node/v10.12.0/bin/node
 args:
   - ./test/integration/execute.js
 stdio:
   - 0
   - pipe
   - 2
 cwd: /home/travis/build/LLK/scratch-vm
 exitCode: 1
 ...

```
The cause is this subtest:

```
order-changes-backwards-2-broadcast-and-wait.sb2
not ok 15 - order-changes-backwards-2-broadcast-and-wait.sb2 # time=102.42ms
     ---
     failures:
       - tapError: incorrect number of tests
     ...
```

spectranaut:
### Resolves

#1677

### Proposed Changes

The test fails because "END" is said after waiting a set amount of time indicating the end of the test. On travis, where resources are limited, the test can take longer to run than the period allowed. This PR changes the "END" statement to deterministically be stated after the last test is run.

### Reason for Changes

The test is flacky and randomly fails for the wrong reason.

### Test Coverage

I submited this test on the bocoup branch and ran travis CI 10 times, you can see the results here: https://github.com/bocoup/scratch-vm/pull/2 . The failing tests are unrelated, this particular test never fails.

To inspect the logic, you can look here: https://scratch.mit.edu/projects/259897810


kchadha (commented):
What if instead of having two `say end` blocks, one that *should* run if the test passes and one that *should* run if the test fails, we had each `when I receive` hat that runs broadcast the value of `order`, and have a `when I receive` hat on the stage that receives the message 4 (which should always be the last value of order regardless of whether the test passes or fails), and then has the `say end` block execute...?

I think this would make the test slightly easier to reason about and then we still `say end` regardless of how exactly the test fails (e.g. if it fails in a way we don't expect).

@mzgoddard, I think you might have written the original test, does this sound like a good plan?

kchadha:
Marking this as needs work until the above question gets answered.

spectranaut:
hi @kchadha -- I just added a commit implementing the change you suggested, if I understand it correctly. You can see the changes in this scratch project: https://scratch.mit.edu/projects/262822442/

I'm not totally convinced this is the best way forward -- the reason why is this: this is a test of the execution order of `when I receive` hats with the use of `broadcast and wait`.  Introducing an additional `broadcast` block makes the test rely on extra functionality not being tested by this test.

In my original fix, I tried to use no additional functionality when adding the two `say end` blocks. It's a little harder to reason and to read, but the goal is to reply on as little as possible when writing the test. If you do make the headache of reasoning, my original test is correct, an `end` WILL BE said IF AND ONLY IF the test has run it's course.

Does that make sense? Ultimately I'm happy for you or @mzgoddard to prefer either method of saying "end".

mzgoddard:
Reviewing the test I think it can be simplified. We can get rid of Sprite3 and move the `say "plan 4"` and `say "end"` to the Stage. Put `say "plan 4"` before the wait in the Stage. And put `say "end"` after the last `broadcast ... and wait`. That'll also let us get rid of the new `broadcast [order]`.

Sorry I didn't think about that earlier, and avoided this issue when I originally created the test.

mzgoddard:
It'd help for me to remember that stage cannot technically `say ...` things. So instead move the `broadcast [order]` to the Stage or replace it with `broadcast "end"`. Leave `say "plan 4"` where it is.

kchadha:
@mzgoddard, I also didn't remember that the stage can't say things. Oops! 😅

spectranaut:
Ok I did what @mzgoddard suggested -- there is another "broadcast" on the stage with the string "end", which the third sprite waits on then says "end"!

kchadha (approved):
The test looks great, thanks for fixing this!

Issue #1715, opened 2018-10-31
Add tests to confirm the translation and videoSensing extensions load


ktbee:
### Expected Behavior

We have tests that check that all extensions load thanks to #1700 

### Actual Behavior

We have tests for loading most extensions in `test/integration/load-extensions.js` with test projects in `test/fixtures/load-extensions`, but not for the `translation` and `videoSensing` extension. 

### Steps to Reproduce

If you add an example test project file to `test/fixtures/load-extensions` and run the test, you will get an error that some function or value is undefined. These extensions seem to need more Scratch infrastructure stubbed out than the others, and so some thought might need to go into the simplest way to do that for multiple extensions .

PR #1741, opened 2018-11-07
Set video state after project load

### LINKED ISSUE #1062: Load video transparency state from stage property when a project gets loaded

### LINKED ISSUE #1635: Load videoOn and videoAlpha state from sb2 project's initial values

ktbee:
### Resolves

Fixes #1635, fixes #1062, partially #1715 with a test for whether the `videoSensing` extension loads

### Proposed Changes

This PR does a few things:
- Checks the stage target after the project has loaded and sets the project's `videoOn` and `videoTransparency` accordingly. 
- Makes the check for `this._drawable` more specific in `video.js`. Since its default value is `-1`, it evaluated to truthy even when it shouldn't have. This was causing some issues when I was writing a test for the video state since `(!this._drawable)` evaluated to true and the video device thought a renderer should be present. 
- Arranges the `load-extensions` test project files in subdirectories now that we have multiple tests in `load-extensions.js`

### Reason for Changes

Previously the check for the stage's video properties was done before targets were installed, so the stage target's info wasn't available. Now with the `PROJECT_LOADED` event, we can wait until later to check the stage and get the info we need to correctly set the project's video values. 

### Test Coverage

I've added some tests for whether the video is on and the video's transparency in `load-extensions.js`. There doesn't appear to be a flag or property indicating the video io device has been turned on in the same way that `_ghost` stores the transparency value, so I used the definition of the `mirror` property as a proxy for that information. Let me know if I should check whether the video is on another way though!


src/extensions/scratch3_video_sensing/index.js
```diff
@@ -93,20 +93,22 @@ class Scratch3VideoSensingBlocks {
         this._lastUpdate = null;
 
         if (this.runtime.ioDevices) {
+            // Configure the video device with values from a globally stored
+            // locations.
```
kchadha: > from **a** globally stored location**s**

Nit: typo

src/io/video.js
```diff
@@ -126,7 +126,7 @@ class Video {
      */
     setPreviewGhost (ghost) {
         this._ghost = ghost;
-        if (this._drawable) {
+        if (this._drawable !== -1) {
```
kchadha, line 136: Could you add a comment above this line about what -1 represents (I assume this._drawable === -1 represents that the video is off, right?)

kchadha (changes requested):
@ktbee, I left some minor comments in the code review. I also noticed while testing that when I upload the test project that has the video state off, the green camera light on my computer turns on briefly even though the video never comes on the stage. Do you think it's possible to fix this issue in this same PR? I'm wondering if defaulting to OFF instead of ON in the code linked below might fix the issue, but I haven't thought a lot about what other implications this might have... From just my first reaction, I think it makes sense for the video to default to OFF if the stage target isn't present...

https://github.com/LLK/scratch-vm/blob/develop/src/extensions/scratch3_video_sensing/index.js#L182

ktbee:
@kchadha thanks for noticing that, I agree it would be best not to have the camera turn on at all if you're not expecting it to! I was able to prevent this by changing the line you pointed out and by enabling the video based on the project's `videoState` in the `getInfo` function, whereas previously it enabled it on all cases. 

One thing that was tricky is that these changes prevented the camera from turning on when a user first adds the video sensing extension in the UI, which is something I believe we do want. I tracked it down to an inconsistent default `videoState` value in the scratch-gui, and I have a PR in to fix that: https://github.com/LLK/scratch-gui/pull/3817 . So we should probably wait to merge this PR into the VM until after this is updated in the GUI.

ktbee:
Just a heads up, the debugging I was doing had to do with a larger Travis CI test issue where the CI's node version wasn't compatible with some existing tests. Should be all set now to take a look @kchadha !

src/extensions/scratch3_video_sensing/index.js
```diff
@@ -92,21 +92,16 @@ class Scratch3VideoSensingBlocks {
          */
         this._lastUpdate = null;
 
+
```
kchadha, line 101: Nit: stray newline

kchadha (changes requested):
From our offline discussions, proposed changes on this and the related scratch-gui PR are:

- remove extension related properties from the default project completely (`tempo`, `videoTransparency`, and `videoState`, `volume` is not an extension related property)
- change `globalVideoState` getter in video extension back to defaulting to ON if either stage is not present or stage.videoState is not defined, otherwise use stage.videoState
- remove sb2 import code above which explicitly sets the stage’s video state to OFF if `videoOn` is not specified in the project being imported. We can just leave that property out of the project and have the videoState default to on as defined above
- refactor `installTargets` so we populate `this.runtime.targets` before loading extensions instead of after

ktbee:
@kchadha, @cwillisf, @mzgoddard, I've implemented our changes that we discussed offline, namely that we're not changing the install order for targets and extensions for now and are instead going with a workaround for loading the correct video extension state until we have a more defined system for passing state to extensions (#1789). Here are the details of my workaround:

- When the video sensing extension is loaded, add a `PROJECT_LOADED` event in the constructor that updates the `videoState` and `videoTransparency` display with the stage target values once the project has loaded
- Update the video display properties in the `getInfo` method in the video sensing extension. This will turn the video on and set the default transparency unless the stage target says otherwise. Calling this in `getInfo` allows the video to turn on automatically when you first add the extension, but doesn't turn it on if it has been disabled by the project's properties
- This solution depends on https://github.com/LLK/scratch-gui/pull/3817 for removing the default `videoState` and `videoTransparency` values in the default sb3 project so it doesn't create conflicting state (previously it was set to `off`, which prevented the video from turning on when adding the extension for the first time)

ktbee:
Also I'll be sure to squash all of these commit messages before merging this :)

kchadha (changes requested):
@ktbee, as discussed offline, currently what I’m seeing is that if you upload a project from 2.0 that doesn’t have the videoOn property, then adding the video extension to that imported project does not turn video on on the stage (it should)

E.g.
1. Create a 2.0 project that has nothing to do with video (e.g. When green flag > move 10 steps)
2. Import that project into 3.0
3. In 3.0, add the video extension to the project; Note that the video doesn’t turn on

I tried commenting out the line in serialization/sb2 where you default to OFF if the videoOn property is provided but falsey.

This fixes the above issue but then introduces the issue where we are no longer correctly turning video state off if the project has actual video blocks on the workspace but the video state is off. E.g. https://scratch.mit.edu/projects/255933884/

ktbee:
Thanks for catching that scenario @kchadha! I hadn't realized that I hadn't tried to add an extension to an sb2 project loaded in sb3 yet. I'm curious what you think of [this solution](https://github.com/LLK/scratch-vm/pull/1741/commits/7176917d1ab2b5b6ba0d951f3e9f194596771bcc) to use a flag to determine the first time the video sensing extension is installed, and set the default video properties ("on", transparency at 50%) then. It relies on some particular timing, let me know if you think it's too fragile. Here is a timeline of how video state is updated in this PR:

1) `InstallTargets` runs in the virtual machine 
	- Loads video sensing extension if it exists on the saved project. Video property state will have been set in serialization
	- When the extension is loaded, it adds an event to update the video device’s on/off state and transparency when project is loaded and sets the `firstInstall` flag to true

2) `getInfo` is called as part of loading the extension
 	- Runs when extension is installed, so the `firstInstall` flag is true and the extension turns on video and sets transparency to 50. Then the `firstInstall` flag is set to false

3) `loadProject` runs in the virtual machine
 	- Emits event that triggers update of video device display based on values that are set in the extension’s state, which initially came from serialization. For projects that have video blocks but an off video state, it resets the video values to the ones stored on the project.
 	- Only a factor if the extension is loaded part of a installing targets and before the project has finished loading. Otherwise if the extension is added after the project has loaded (ie using the UI), this event will have already happened and won't affect the video state.

The downside I see is that projects that have video blocks and an off state end up having their video display properties updated a couple times. However it happens fast enough that I haven't seen my camera light turn on, which might be alright for a workaround.

kchadha (commented):
There is a merge conflict with one of the files, but after resolving that, LGTM.

I did notice one issue which I filed separately (#1817) because I don't think it's worth fixing right now, and it's possible we don't want to fix it at all...

kchadha (approved):
Oops looks like I was looking at an old version of the PR page. 

LGTM.

PR #1742, opened 2018-11-08
Multi monitors

### LINKED ISSUE #1218: How will we support Scratch 2.0 projects that have multiple "current ()" monitors?
In Scratch 2.0, it was possible to have monitors for "current second," "current minute," "current hour" etc. simultaneously. In 3.0, they will all share 1 monitor. We should figure out how projects with multiple monitors will behave in 3.0.

https://scratch.mit.edu/projects/196922245/

- When I open the project in 3.0, will all of the different monitors still be present, or will they all merge into 1 monitor?
- If I change the value on the dropdown on the "current ()" block, will all of the monitors change? One of them? Will a new monitor appear?

### LINKED ISSUE #1717: Day of Week option in Current sensing block doesn't import from sb2 properly
Scratch 3.0 expects the value of the menu to be "DAYOFWEEK" and scratch 2 provides it as "day of week".  The spaces seem to be the problem.

Notice that the value of Current day of week monitor is 0 on a project imported from sb2: 
https://beta.scratch.mit.edu/#258701696

cc/ @kchadha

kchadha:
### Resolves

Resolves #1717 
Resolves LLK/scratch-gui#1218

### Proposed Changes

Allow multiple monitors for blocks with parameters. 

In Scratch 3.0, we currently have three examples of reporter blocks that are monitor-able (have a checkbox in the flyout) and also have a parameter which determines what is being monitored. The  changes in this PR make it so that monitors with the same reporter but different values for the parameter can be shown simultaneously (e.g. have a monitor for current year as well as current minute showing on the stage).

Also resolves a bug where image was not importing properly into 3.0.

### Reason for Changes

The above was previously not working.

### Test Coverage

- [ ] 1. Show monitors for all of the options for the `(current [v year])` block.
- [ ] 1.5 Observe that day of the week has the correct value where Sunday = 1.
- [ ] 2. Selectively hide some of these monitors.
- [ ] 3. Rinse and Repeat.
- [ ] 4. Repeat steps 1-3 for the `(costume [v number])` and the `(backdrop [v number])` blocks

- [ ] Import 260563473.

cc/ @picklesrus


paulkaplan (approved):
LGTM... so strange though... thank you for documenting.

src/engine/blocks.js
```diff
@@ -559,11 +569,36 @@ class Blocks {
...
+            if (block.fields && Object.keys(block.fields).length > 0 &&
+                block.opcode !== 'data_variable' && block.opcode !== 'data_listcontents') {
+
+                // This block has an argument which needs to get separated out into
+                // multiple monitor blocks with ids based on the selected argument
+                const newId = getMonitorIdForBlockWithArgs(block.id, block.fields);
+                // Note: we're not just constantly creating a longer and longer id everytime we check
+                // the checkbox because we're using the id of the block in the flyout as the base
+
+                // check if a block withl the new id already exists, otherwise create
```
paulkaplan: with*

kchadha:
This PR also fixed #1513 and #1628.

PR #1747, opened 2018-11-09
Always ignore video sensing monitor when importing from 2.0 and import visible extension monitors.

### LINKED ISSUE #1745: Visible extension monitors don't load when imported from 2.0 when there are no blocks from that extension in the project
### Expected Behavior

Visible monitors should load.

### Actual Behavior

Visible extension monitors don't load when imported from 2.0 when there are no blocks from that extension in the project (e.g. a project that has a visible video monitor but doesn't have any video blocks in the project)

### Steps to Reproduce

Load 260787219 or 260788039 in Scratch 3.0 and observe that there is no monitor on the stage.

kchadha:
### Resolves

Resolves #1745 (see comment in that issue about resolution decision for the video sensing monitor).

### Proposed Changes

Fixes issue where visible extension monitors weren't loading when importing from 2.0. Ignores video sensing monitor when importing monitors.

### Test Coverage

Added integration tests.

cc/ @picklesrus, @ktbee


test/integration/sb2-import-extension-monitors.js
```diff
@@ -0,0 +1,107 @@
...
+    vm.clear();
+    vm.setCompatibilityMode(false);
+    vm.setTurboMode(false);
+    vm.loadProject(visibleTempoMonitorProject).then(() => {
+        t.equal(vm.extensionManager.isExtensionLoaded('music'), true);
+        t.equal(vm.runtime._monitorState.size, 1);
+        t.equal(vm.runtime._monitorState.has('music_getTempo'), true);
+        t.equal(vm.runtime._monitorState.get('music_getTempo').visible, true);
+        t.end();
+        process.nextTick(process.exit); // This is needed because this is the end of the last test in this file!!
```
paulkaplan: Just FYI you can replace this with `tap.tearDown(() => process.nextTick(process.exit));` at the top of this file, to the same effect but without the test order dependency.

src/serialization/sb2.js
```diff
@@ -245,17 +245,42 @@ const globalBroadcastMsgStateGenerator = (function () {
...
     const extID = opcode.substring(0, extIndex);
 
-    // All non-core extensions should be added by blocks at this point
-    // We can assume this is an unintended monitor and skip parsing if it belongs to a non-core extension
-    if (CORE_EXTENSIONS.indexOf(extID) === -1) {
-        if (extID !== '') return;
+    if (extID === 'videoSensing') {
+        return;
+    } else if (CORE_EXTENSIONS.indexOf(extID) === -1 && extID !== '' &&
+        !extensions.extensionIDs.has(extID) && !object.visible) {
```
paulkaplan, line 272: This line sets up a hard ordering requirement between block loading/extension finding and monitor parsing. I don't know if there is something we can do to enforce this so that it doesn't get shuffled in the future, but it might be worth at least noting in the JSdoc at the top of parseMonitorObject that calling this before you've loaded the projects blocks will result in undefined behavior.

src/serialization/sb2.js
```diff
@@ -245,17 +245,42 @@ const globalBroadcastMsgStateGenerator = (function () {
...
+        // Don't import this monitor if it refers to a non-core extension that
+        // doesn't exist anywhere else in the project and it isn't visible.
+        // This should only apply to the tempo block at this point since
+        // there are no other sb2 blocks that are now extension monitors.
+        return;
     }
 
+
+    // if (!object.visible) {
```
paulkaplan: Did you intend to leave this chunk commented out?

src/serialization/sb2.js
```diff
@@ -245,17 +245,42 @@ const globalBroadcastMsgStateGenerator = (function () {
...
-    if (CORE_EXTENSIONS.indexOf(extID) === -1) {
-        if (extID !== '') return;
+    if (extID === 'videoSensing') {
+        return;
+    } else if (CORE_EXTENSIONS.indexOf(extID) === -1 && extID !== '' &&
+        !extensions.extensionIDs.has(extID) && !object.visible) {
+        // Don't import this monitor if it refers to a non-core extension that
+        // doesn't exist anywhere else in the project and it isn't visible.
+        // This should only apply to the tempo block at this point since
+        // there are no other sb2 blocks that are now extension monitors.
```
paulkaplan, line 276: Is it worth throwing/asserting/logging this assumption?

paulkaplan (changes requested):
Code and tests LGTM.  Couple of minor comments/questions.

paulkaplan (approved):

paulkaplan:
Woops accidentally marked needs changes. I think it would be good to remove the commented code if it isn't needed, and to switch the way the tests exit, but those do not require re-review!

src/serialization/sb2.js
kchadha: Sounds good!

src/serialization/sb2.js
kchadha: That was more of a note for the reader rather than an assumption that needs to be logged/asserted. Originally I was planning to explicitly check for the tempo monitor, but I ended up making this code more generic so it should be able to handle other monitors that have become extension monitors if someone  somewhere tries to use this code for slightly augmented sb2s.

src/serialization/sb2.js
kchadha: Fixed

PR #1750, opened 2018-11-12
Add method for collecting all targets' assets

### LINKED ISSUE #1601: Add API for getting unsaved ("dirty") assets to be saved
Currently we can save a single .sb3 archive, but we want an easy way to save the SB3 JSON separately from each of the assets that needs to be saved.

rschamp:
### Resolves

Resolves #1601

### Proposed Changes

Adds a VM getter that returns all assets stored on all targets.

### Reason for Changes

Towards incremental saving, this gives a way for the GUI to save all the assets individually.  For now, it is the responsibility of the consumer to determine if these assets should be saved or not. Otherwise the VM would need to be responsible for saving, which has been out of its scope.  Does not include the project JSON since that would require knowledge of a project ID, which the VM has not been responsible for so far.

### Test Coverage

Added unit tests for the function as well as the getter on the VM.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -18,7 +18,7 @@ const Variable = require('./engine/variable');
 
 const {loadCostume} = require('./import/load-costume.js');
 const {loadSound} = require('./import/load-sound.js');
-const {collectAssets, serializeSounds, serializeCostumes} = require('./serialization/serialize-assets');
+const {serializeSounds, serializeCostumes} = require('./serialization/serialize-assets');
 require('canvas-toBlob');
 
 const RESERVED_NAMES = ['_mouse_', '_stage_', '_edge_', '_myself_', '_random_'];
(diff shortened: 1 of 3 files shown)


src/virtual-machine.js
```diff
@@ -336,6 +336,13 @@ class VirtualMachine extends EventEmitter {
         });
     }
 
+    /*
+     * @type {Array<object>} Array of all costumes and sounds currently in the runtime
+     */
+    get assets () {
+        return collectAssets(this.runtime);
+    }
```
cwillisf, line 348: Since `collectAssets` reaches into the runtime to do its work and this looks like the only place it's called from, it seems like it would make sense to merge the guts of `collectAssets` into this accessor or into the runtime itself.

cwillisf (approved):
This looks fine to me, though I commented on a potential simplification...

src/virtual-machine.js
rschamp: Good point. It used to do more — you could filter by a target — but then I decided to only make it do what we need.

PR #1755, opened 2018-11-13
Create cloud variables

kchadha:
### Proposed Changes

Add ability to create cloud variables, requesting the cloud provider to create a cloud variable on the server and waiting for confirmation from the cloud provider to update the actual isCloud flag and cloud variable limit.

### Reason for Changes

Creating cloud variables!

### Test Coverage

Testing instructions included in the related gui PR.

### Related PRs
- LLK/scratch-gui#3749
- LLK/scratch-blocks#1782

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -233,11 +233,16 @@ class Target extends EventEmitter {
      * @param {string} id Id of variable
      * @param {string} name Name of variable.
      * @param {string} type Type of variable, '', 'broadcast_msg', or 'list'
+     * @param {boolean} isCloud Whether the variable to create has the isCloud flag set.
+     * Additional checks are made that the variable can be created as a cloud variable.
      */
-    createVariable (id, name, type) {
+    createVariable (id, name, type, isCloud) {
         if (!this.variables.hasOwnProperty(id)) {
             const newVariable = new Variable(id, name, type, false);
             this.variables[id] = newVariable;
+            if (isCloud && this.isStage) {
+                this.runtime.ioDevices.cloud.requestCreateCloudVariable(newVariable);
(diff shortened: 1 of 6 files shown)


src/engine/target.js
```diff
@@ -233,11 +233,16 @@ class Target extends EventEmitter {
...
      * @param {string} type Type of variable, '', 'broadcast_msg', or 'list'
+     * @param {boolean} isCloud Whether the variable to create has the isCloud flag set.
+     * Additional checks are made that the variable can be created as a cloud variable.
      */
-    createVariable (id, name, type) {
+    createVariable (id, name, type, isCloud) {
         if (!this.variables.hasOwnProperty(id)) {
             const newVariable = new Variable(id, name, type, false);
             this.variables[id] = newVariable;
+            if (isCloud && this.isStage) {
```
paulkaplan, line 243: Is it strange to include the stage requirement here as well as elsewhere? Should it just be the callers responsibility?

paulkaplan (changes requested):
Changes look ok, but can you include a test for the new functionality in io/cloud? And add a new test for the new param to createVariable on target (functionality is covered by existing test, but not the cloudness).

src/engine/target.js
kchadha: Yeah, I figured it would be good to check here anyways because "anyone" can call target.createVariable..., but we never want cloud variables to be local..

kchadha:
@paulkaplan Just added/updated the unit tests. =]

paulkaplan (approved):

paulkaplan:
Good to merge apart from that lint error

Issue #1756, opened 2018-11-14
"Play note" blocks in imported projects do not have the note picker


ericrosenbaum:
### Expected Behavior

All "play note" blocks should display the note picker field when you click the note input.

### Actual Behavior

"Play note" blocks in projects imported from 2.0 do not display the note picker field.

PR #1757, opened 2018-11-14
Use note picker in imported play note blocks

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1756

### Proposed Changes

"Play note"  blocks in projects imported from 2.0 should use the note picker field.

Also, obscured shadows in imported blocks with the note picker should default to 60.

### Test Coverage

Test 2.0 projects with note blocks, such as:
261062957
187641716


paulkaplan:
As mentioned offline, lets also set the default value for obscured shadows on this field type to 60

paulkaplan (approved):
LGTM

PR #1759, opened 2018-11-14
Make the wait block's logic and behavior more consistent with Scratch 2

ktbee:
### Resolves

Fixes [the seagull project](https://llk.github.io/scratch-gui/develop/#169401431) in #1478, where the "BREATH" sprite doesn't disappear as expected. It also includes some compatibility fixes for [a space invaders project](https://scratch.mit.edu/projects/252324412/) and a [project that tracks the Scratch loop](https://scratch.mit.edu/projects/259055091)

### Proposed Changes

- This PR replaces the wait block's use of JavaScript promises with logic that keeps track of how much time has elapsed with the Timer utility. This makes the wait block's timing more consistent with how [it was implemented in Scratch 2](https://github.com/LLK/scratch-flash/blob/develop/src/interpreter/Interpreter.as#L604-L609) and avoids race conditions with the JS event loop.
- Adds a timer and timer-related logic to blocks' utility functions
- Adds a time server to the runtime to make available a millisecond clock that updates as threads are stepped. This is useful when a block's timing needs to mimic logic for [Sprite 2's `currentMSecs` timestamp.](https://github.com/LLK/scratch-flash/blob/72e4729b8189d11bbe51b6d94144b0a3c392ac9a/src/interpreter/Interpreter.as#L245)

### Reason for Changes

In the example seagulls project, the "BREATHE" sprite doesn't disappear because of a race condition with its wait blocks. The hide block executes before the show block. Using the timer utility and timestamps makes the wait times more consistent.

We also needed a milliseconds clock that keeps the same timestamp for an entire step of the Sequencer, [as in Scratch 2.](https://github.com/LLK/scratch-flash/blob/72e4729b8189d11bbe51b6d94144b0a3c392ac9a/src/interpreter/Interpreter.as#L245) This cached value is what allowed the enemy ships in [this Space Invaders project](https://scratch.mit.edu/projects/252324412/) to move in unison. Without this cached value, my wait block timer fix made the ships even more out of sync. This millisecond timestamp is passed as a custom `nowObj` for the wait block's timer. 

### Test Coverage

I've added tests for the wait block in `blocks_control.js`.

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -1,4 +1,5 @@
 const Cast = require('../util/cast');
+const Timer = require('../util/timer');
 
 class Scratch3ControlBlocks {
     constructor (runtime) {
@@ -38,6 +39,43 @@ class Scratch3ControlBlocks {
             control_clear_counter: this.clearCounter,
             control_all_at_once: this.allAtOnce
         };
+    }
+
+    /**
+     * Check the stack timer, and if its time is not up yet, yield the thread.
+     * @param {object} util - utility object provided by the runtime.
+     * @private
+     */
+    _checkStackTimer (util) {
+        const timeElapsed = util.stackFrame.timer.timeElapsed();
+        if (timeElapsed < util.stackFrame.duration) {
+            util.yield();
+        }
+    }
+
+    /**
+     * Check if the stack timer needs initialization.
+     * @param {object} util - utility object provided by the runtime.
+     * @return {boolean} - true if the stack timer needs to be initialized.
+     * @private
+     */
+    _stackTimerNeedsInit (util) {
+        return !util.stackFrame.timer;
+    }
+
+    /**
+     * Start the stack timer and the yield the thread if necessary.
+     * @param {object} util - utility object provided by the runtime.
+     * @param {number} duration - a duration in seconds to set the timer for.
+     * @private
+     */
+    _startStackTimer (util, duration) {
+        duration = Math.max(0, 1000 * Cast.toNumber(duration));
+
+        util.stackFrame.timer = new Timer();
+        util.stackFrame.timer.start();
+        util.stackFrame.duration = duration;
+        util.yield();
     }
 
     getHats () {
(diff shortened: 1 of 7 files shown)


Mr-Dave2:
For your testing, I believe that this is another project where the timing of the wait blocks must be precise or the invader sprites get out of line.  
https://beta.scratch.mit.edu/#252324412

This was mentioned in the forums as an issue here: 
https://scratch.mit.edu/discuss/topic/308114/?page=13#post-3293911

ktbee:
@Mr-Dave2 thank you for the example project! It looks like the invaders' timing is more out of sync with my fix, so I'm going to investigate this a little more. Thanks again for letting me know.

Mr-Dave2:
This may be less useful but I'm adding another comment about the difference in the way a wait(0) block works.  
This program has two small sprites that constantly move and change costumes in a forever loop.  A slider variable can be used to also include a wait(0) in those motion loops. The stage has a loop to calculate FPS using a technique I've seen other people use.  

When run under Scratch 2, the FPS counter shows 30 both with and without the wait(0) block.

When run under Scratch 3, the FPS counter shows 30 without the wait(0) block but goes over 50,000 when the wait(0) block is included.

Scratch 2 link: https://scratch.mit.edu/projects/259055091/
Scratch 3 link: https://beta.scratch.mit.edu/#259055091

I don't have an example of how this could affect an actual project.

src/blocks/scratch3_control.js
```diff
@@ -107,13 +144,17 @@ class Scratch3ControlBlocks {
...
+        const stackFrame = util.stackFrame;
         const duration = Math.max(0, 1000 * Cast.toNumber(args.DURATION));
-        return new Promise(resolve => {
-            setTimeout(() => {
-                resolve();
-            }, duration);
-        });
+
+        if (this._stackTimerNeedsInit(stackFrame)) {
+            this._startStackTimer(stackFrame, duration);
```
towerofnix: Should the `const duration` block be placed inside the `if ()` here? It isn't used outside the scope of the `if`, so there'd presumably be a performance gain by only calculating the duration inside the `if`, when it's needed.

ktbee: Good call! I'll make that update, thanks @towerofnix

ktbee:
Okay, I think we've covered our compatibility bases on this one! Though let me know if you find other tricky projects or examples @Mr-Dave2. I have fixes now in place for the Space Invaders project and the "Wait 0 yield timing" project. 

@kchadha this implementation in this PR has changed a bit, I'll update the original comment with my proposed changes.

kchadha (commented):
The code LGTM, @rschamp, @thisandagain, @cwillisf, I'm curious to get thoughts from one of you since you are probably more familiar with the sequencer timing etc.

thisandagain:
This looks good! Thanks @ktbee.

thisandagain (approved):

PR #1760, opened 2018-11-14
Added an option to provide Scratch Link with a Organizationally Uniqu…

knandersen:
### Resolves

Currently Scratch VM wants to discover Bluetooth devices, it is only possible to provide a major and minor device class as a search parameter. As these device classes are relatively generic, while searching for a specific device, other devices that share the same major and minor device class will appear. 

In this case, the EV3 extension has major class Toy (8) and minor class Robot (1), and it is highly likely that other companies will have products that use this exact combination of major and minor device class.

### Proposed Changes

This change adds an optional Organisationally Unique Idenfitier (OUI)-prefix as information for Scratch Link to further filter Bluetooth-devices, so that only EV3 Programmable Bricks when using the EV3 extension.

### Reason for Changes

To avoid users from being shown other devices than those they intend to connect to.

### Test Coverage

No tests have been added.


cwillisf (approved):
👍

thisandagain:
@cwillisf @knandersen Can you folks resolve the conflicts and get this merged or closed?

knandersen:
@thisandagain done!

knandersen:
Based on the problems we’ve experienced on Windows re: supplying a filter that Scratch Link doesn’t understand, we should probably hold off on merging this in until the Windows-part of the OUI-prefix enhancement gets implemented. Thoughts @cwillisf ?

cwillisf:
If I remember correctly we decided that OUI-based filtering is not the best option, and we're filtering based on a name prefix instead. @knandersen please re-open this if you disagree or otherwise think we should bring the idea of OUI filtering back :)

PR #1763, opened 2018-11-18
Various fixes to extension disconnect errors

### LINKED ISSUE #1590: Consider adding a way to detect that wedo2 has disconnected?
We should consider whether to add a way to detect that a wedo2 has become disconnected (e.g. because the peripheral has gone out of range, or the battery has died or been removed). One way would be to send a request at some interval (e.g. to check the battery level), which would error in those cases.

evhan55:
### Resolves

- Resolves #1590: Consider adding a way to detect that wedo2 has disconnected?
- Fixes disconnect error handling in existing hardware extensions to correctly clear timeouts and other data

### Notes

- In implementing this fix for LEGO WeDo2, I noticed that error handling wasn't quite complete among all the hardware extensions.  This PR changes BLE/BT sockets to take in `disconnectCallback`s in order to properly have hardware extensions clear any remaining timeoutIDs on the window, clear locally stored data, etc.  As a result of those enhancements, this PR requires testing all three extensions.
- Disconnect alerts might look too large, that's being addressed in another PR
- Once this lands, then https://github.com/LLK/scratch-gui/pull/3922 can land as well

### Testing LEGO WeDo 2
1. Load extension and connect LEGO WeDo2
2. Turn off LEGO WeDo2 by pressing button for 3 seconds
3. In 5 seconds or less, should see disconnect alert and status button should turn orange

### Testing EV3
1. Load extension and connect EV3
2. Turn off EV3
3. In a few seconds, should see disconnect alert and status button should turn orange

### Testing micro:bit
1. Load extension and connect micro:bit
2. Turn off micro:bit
3. In 5 seconds, should see disconnect alert and status button should turn orange


src/extensions/scratch3_microbit/index.js
```diff
@@ -317,7 +317,11 @@ class MicroBit {
 
         // cancel disconnect timeout and start a new one
         window.clearInterval(this._timeoutID);
-        this._timeoutID = window.setInterval(this.disconnect, BLETimeout);
+        this._timeoutID = window.setInterval(
+            // send an error to the BLE socket
+            this._ble._sendDisconnectError.bind(this._ble, 'micro:bit stopped receiving data'),
```
evhan55: This is a slight hack, but better than before.  The micro:bit extension sends an error to the BLE socket with an internal error string (that doesn't need to be translated), and then that triggers all the correct disconnect error behavior.

src/extensions/scratch3_microbit/index.js
```diff
@@ -317,7 +317,11 @@ class MicroBit {
 
         // cancel disconnect timeout and start a new one
         window.clearInterval(this._timeoutID);
-        this._timeoutID = window.setInterval(this.disconnect, BLETimeout);
+        this._timeoutID = window.setInterval(
+            // send an error to the BLE socket
+            this._ble._sendDisconnectError.bind(this._ble, 'micro:bit extension stopped receiving data'),
```
evhan55: This might look slightly confusing, but it is better than before. The micro:bit extension sends an error to the BLE socket with an internal error string (that doesn't need to be translated), and then that triggers all the correct disconnect error behavior for both the micro:bit extension and the BLE socket.

cwillisf: It might make sense to create a constant for this instead of using a magic ~number~string. Also, if this is caught by the BLE class it should probably not include "micro:bit" in the string... maybe "BLE extension stopped receiving data"?

ericrosenbaum:
Starting to test this now.  Weirdly, it seems that _checkBatteryLevel does not cause an error if you actually disconnect the battery. It does work exactly as you describe (i.e. if you power down the WeDo with its button), so I guess this is a separate problem. I'll file an issue to investigate that separately.

src/extensions/scratch3_wedo2/index.js
```diff
@@ -757,6 +779,20 @@ class WeDo2 {
...
+     * Check the battery level on the WeDo 2.0. If the WeDo 2.0 has disconnected
+     * for some reason, the BLE socket will get an error back and automatically
+     * close the socket.
+     */
+    _checkBatteryLevel () {
+        this._ble.read(
+            BLEService.DEVICE_SERVICE,
+            BLECharacteristic.LOW_VOLTAGE_ALERT,
+            false,
+            this._onMessage // will ignore result
```
cwillisf: Since `optStartNotifications` is false I think it would make more sense to just leave out the last two parameters here.

ericrosenbaum:
I tested on Windows 10 version 1703, but there are other problems that interact with these changes, so probably we should just ignore that until we have 1703 working properly.

ericrosenbaum:
I found one problem (platform independent) that will need fixing- the disconnect button in the connection modal does not fully work. It disconnects the device, but the status indicator in the blocks menu does not update. To repro:

- Connect to a microbit
- Close the connection modal
- Re-open the connection modal by clicking the status indicator
- Click the "disconnect" button in the connection modal
- Note that the microbit is disconnected (scrolling its name) but that status indicator in the blocks menu still shows connected

cwillisf (changes requested):
Some minor nitpicks but otherwise the code looks fine to me. I don't immediately see the cause of the problem that @ericrosenbaum mentioned, though :/

evhan55:
> I found one problem (platform independent) that will need fixing- the disconnect button in the connection modal does not fully work. It disconnects the device, but the status indicator in the blocks menu does not update. To repro:
> 
> * Connect to a microbit
> * Close the connection modal
> * Re-open the connection modal by clicking the status indicator
> * Click the "disconnect" button in the connection modal
> * Note that the microbit is disconnected (scrolling its name) but that status indicator in the blocks menu still shows connected

Sounds good, will investigate, thank you!

src/extensions/scratch3_microbit/index.js
evhan55: Thanks! This is being sent from within the microbit class in this case, but a more generic string probably makes sense anyway, and making it a constant, too. 👍

src/extensions/scratch3_wedo2/index.js
evhan55: Oh I see, thanks

evhan55:
@ericrosenbaum Thank you for catching the status icon error!  I have addressed it, and also created a new PR in gui to match (which you will need to test for the correct behavior):

https://github.com/LLK/scratch-gui/pull/3922

Edit:
Just FYI, I tested all these scenarios on both those branches:

EV3:
- connect EV3
- turn off Scratch Link, turn off EV3, turn off bluetooth > all give alert + icon goes orange
- connect EV3
- click disconnect in modal > icon goes orange

micro:bit:
- connect micro:bit
- turn off Scratch Link, turn off micro:bit, turn off bluetooth > all give alert + icon goes orange
- connect micro:bit
- click disconnect in modal > icon goes orange

WeDo2:
- connect WeDo2
- turn off Scratch Link, turn off WeDo2, turn off bluetooth > all give alert + icon goes orange
- connect WeDo2
- click disconnect in modal > icon goes orange

cwillisf (approved):

cwillisf:
The code changes look good to me. I'm 👍 as long as you and @ericrosenbaum are both happy too :)

ericrosenbaum:
I was testing using this branch (also using https://github.com/LLK/scratch-gui/pull/3922) by variously connecting, disconnecting, reconnecting, and got into a bad state. I believe I had:

- connected, disconnected, reconnected EV3 and microbit
- connected, disconnected, reconnected WeDo
- powered down WeDo with the button

At that point, all three extensions have green status indicators, but the WeDo is powered off, and the microbit and EV3 are not responding to commands. I had this error in the console:
```
Uncaught TypeError: this._characteristicDidChangeCallback is not a function
    at BLE.didReceiveCall (lib.min.js:248740)
    at BLE._handleRequest (lib.min.js:257033)
    at BLE._handleMessage (lib.min.js:256987)
    at BLE._onSocketMessage (lib.min.js:256868)
    at WebSocket.JSONRPCWebSocket._this._ws.onmessage
```

I'll try to repro this in a simpler way.

update: I can't reproduce the exact scenario I saw above, but I found a probably-related issue. The WeDo quickly (<1 sec) stops receiving sensor input, and that's what triggers the error above. To repro:

- Connect a WeDo, with a distance sensor attached
- Check the "distance" monitor 
- Move your hand in front of the distance sensor. Note that it quickly stops responding, and there are repeating console errors.

cwillisf:
@ericrosenbaum is this problem new? If not, I suggest we should merge this for the things it does fix, and to help make sure the code doesn't get out of date relative to the rest of `scratch-vm`.

ericrosenbaum:
@cwillisf the problem I'm describing, where the WeDo extension has no sensor input, does not exist on develop (if it were, the extension would be quite broken). As far as I can tell this PR causes it, but I'm not sure how.

cwillisf:
Oh, sorry... I got distracted by the top of your comment with the connect & disconnect instructions and missed the "probably-related" issue at the bottom. That does sound serious :(

evhan55:
@ericrosenbaum @cwillisf I can investigate!  Will reassign it back to myself.

src/extensions/scratch3_wedo2/index.js
```diff
@@ -757,6 +779,20 @@ class WeDo2 {
...
+     * Check the battery level on the WeDo 2.0. If the WeDo 2.0 has disconnected
+     * for some reason, the BLE socket will get an error back and automatically
+     * close the socket.
+     */
+    _checkBatteryLevel () {
+        this._ble.read(
+            BLEService.DEVICE_SERVICE,
+            BLECharacteristic.LOW_VOLTAGE_ALERT,
+            false,
+            this._onMessage // TODO: make optional
```
evhan55: I had to put this back in order to not clear the `characteristicDidChangeCallback` in the BLE socket.

evhan55:
@ericrosenbaum @cwillisf :
Thanks Eric for the catch!  I believe it was only related to the WeDo, but let me know if you find otherwise.

I discovered that by not setting the callback on the new battery level 'read' call to the BLE socket, I was inadvertently clearing the `characteristicDidChangeCallback` for the BLE socket and causing the errors Eric was seeing.  I put the callback back in to the battery read, with a note to perhaps make that optional in the future.

I rebased this code, tested all three devices again, and this time tested both input and output every time a device was connected to make sure.  I marked as 'needs review' again, and will be happy to investigate any further bugs that come up.  I do still believe this PR is needed at some point, so I am happy to keep fine tuning it.

evhan55:
@ericrosenbaum 

One more thing, I found another issue while doing all this testing, which also happens currently on beta.scratch.mit.edu, so I don't think this PR introduces it:

Sometimes (not always), when clicking 'Disconnect' in the connection modal, Scratch successfully disconnects from the WeDo and no data is sent or received, but the WeDo remains powered/looks connected.  This seems to happen on develop right now, I filed an issue:

https://github.com/LLK/scratch-vm/issues/1845

src/extensions/scratch3_microbit/index.js
```diff
@@ -317,7 +327,11 @@ class MicroBit {
 
         // cancel disconnect timeout and start a new one
         window.clearInterval(this._timeoutID);
-        this._timeoutID = window.setInterval(this.disconnect, BLETimeout);
+        this._timeoutID = window.setInterval(
+            // send an error to the BLE socket
+            this._ble._sendDisconnectError.bind(this._ble, BLEDataStoppedError),
```
ericrosenbaum: per discussion we decided to make this a public function and rename to `triggerDisconnectError` (or similar)- and add JSDOC. And also the corresponding change in bt.js

ericrosenbaum: Line 302 (in `_onConnect`) also needs the change to call this function on `_ble` instead of just `disconnect`

ericrosenbaum:
Lets add explicit info in the comment for each disconnect event in runtime about the event's intended side effects. PERIPHERAL_DISCONNECT: updates the status button in the blocks menu. PERIPHERAL_DISCONNECT_ERROR: triggers an alert in the GUI.

src/extensions/scratch3_wedo2/index.js
ericrosenbaum: Per discussion, lets instead leave this parameter null, and in ble.js (and bt.js?), check for null before trying to call the callback

src/extensions/scratch3_microbit/index.js
evhan55: I have chosen to name it `handleDisconnectError` and will also update two other error handlers in both BT/BLE to also use "handle" in their name (as opposed to "send")

src/extensions/scratch3_wedo2/index.js
evhan55: bt.js doesn't seem to need it because `_messageCallback` there seems to be required in the constructor.  Leaving bt.js as is for now with respect to this, and can handle if it comes up later.

evhan55:
@ericrosenbaum 
Thank you for all the feedback.  I incorporated all suggestions, plus made two more changes of my own:
- renamed the events to `PERIPHERAL_DISCONNECTED` for intentional user disconnect events, and `PERIPHERAL_CONNECTION_LOST_ERROR` for errors that cause disconnects
- I had to add another null check to the characteristicDidChange flag in the BLE socket, I think it's necessary for this PR, although not 100% confident, but would rather fix it in a later PR if it introduces some latent bug

I have tested on Windows:
- EV3:
1. close Scratch Link
2. turn off Bluetooth
3. turn off device
4. disconnect via dialog
5. all while testing I/O in between steps
- microbit: same
- WeDo2: same

It would be great if you could test on Mac, thank you!

ericrosenbaum (approved):
Tested all the disconnection type on all three devices and it looks good. Thanks for your patience on this one @evhan55!

PR #1770, opened 2018-11-20
Fix monitor tests to count only monitor threads

### LINKED ISSUE #1769: SB3 Monitors test fails on Node 10 on Travis CI (and Linux in general)
### Expected Behavior

Unit tests should pass on Travis CI.

### Actual Behavior

One of the tests in `test/integration/monitors_sb3.js` fails on Travis CI and in other Linux environments:
```
test/integration/monitors_sb3.js .................. 122/123
  importing sb2 project with monitors
  not ok should be equal
    +++ found
    --- wanted
    -17
    +18
    compare: ===
    at:
      line: 21
      column: 11
      file: test/integration/monitors_sb3.js
      type: VirtualMachine
      function: vm.on.e
    stack: |
      VirtualMachine.vm.on.e (test/integration/monitors_sb3.js:21:11)
      VirtualMachine.getPlaygroundData (src/virtual-machine.js:200:14)
      Timeout.setTimeout (test/integration/monitors_sb3.js:259:20)
    source: |
      t.equal(vm.runtime._lastStepDoneThreads.length, 17);

total ............................................. 122/123


  122 passing (4s)
  1 failing
```

### Steps to Reproduce

On a Linux or WSL system:
1. Install Node 10 (tested with Node 10.13.0)
2. Clone this repository
3. `npm i`
4. `./node_modules/.bin/tap test/integration/monitors_sb3.js`

This seems to happen only on Linux, including both Travis CI and WSL.

cwillisf:
### Resolves

Resolves #1769 

### Proposed Changes

The monitor tests now count only monitor threads, not all threads, when checking whether the appropriate number of monitor updates ran in the last step.

### Reason for Changes

These tests should not fail if there happens to have been a non-monitor thread which ran in the last step. For example, the SB3 version of this test runs a "hide variable" block; if that block runs in the last VM step before the comparison then the test would previously fail.

### DIFF
--- a/test/integration/monitors_sb2.js
+++ b/test/integration/monitors_sb2.js
@@ -17,7 +17,12 @@ test('importing sb2 project with monitors', t => {
         // All monitors should create threads that finish during the step and
         // are revoved from runtime.threads.
         t.equal(threads.length, 0);
-        t.equal(vm.runtime._lastStepDoneThreads.length, 8);
+
+        // we care that the last step updated the right number of monitors
+        // we don't care whether the last step ran other threads or not
+        const lastStepUpdatedMonitorThreads = vm.runtime._lastStepDoneThreads.filter(thread => thread.updateMonitor);
+        t.equal(lastStepUpdatedMonitorThreads.length, 8);
+
         // There should be one additional hidden monitor that is in the monitorState but
         // does not start a thread.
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LGTM

PR #1774, opened 2018-11-20
VM changes for the sensing_of block.  This handles lists properly (by…

### LINKED ISSUE #1762: Sensing of block will evaluate a list in the first menu.
This isn't quite an issue yet, but it is about to be once #1425 and https://github.com/LLK/scratch-gui/issues/2389 are fixed.

The problem is this code doesn't check the variable types and it probably should:
https://github.com/LLK/scratch-vm/blob/develop/src/blocks/scratch3_sensing.js#L316

### Expected Behavior

Variables should show up in the first menu of the sensing of block (they don't yet, but I'm working on it #1425). Once I add them, you can get a list in there and it will also evaluate:
screen shot 2018-11-14 at 7 11 14 pm

### Steps to Reproduce

Getting the list in their is slightly convoluted:
1. make a project with 2 sprites
2. in sprite one add variable 'pizza'
3. in sprite 2, drag out a sensing of block to the workspace
4. select sprite 1 in the right menu
5. select the variable 'pizza' in the first sensing block menu
6. Go back to sprite 1 and add a list 'pizza'. Add some items to the list.
7. Remove the variable 'pizza'
8. Go back to Sprite 2 and click on the sensing block. (which should still have pizza selected)

In scratch 2 this evaluates to 0 (though the steps to get the list selected are slightly different)

### LINKED ISSUE #2389: ___ of ___ block should not have backdrop # or backdrop name options
### Expected Behavior

The `___ of ____` sensing block should not have `backdrop #` or `backdrop name` options in the dropdown and match scratch 2.0

### Actual Behavior

It has those extra options
screen shot 2018-06-18 at 4 45 25 pm

### Steps to Reproduce

Choose Sensing from the Category Menu
Locate the ___ of ___ block (starts as X position of Stage)
Select the first dropdown on the selected block
Note: it has `backdrop #` and `backdrop name` as options

Chrome on Mac

### LINKED ISSUE #2390: ___ of ___ sensing block should have different options for stage
### Expected Behavior

If the stage is selected in the second drop down of the ___ of ___ sensing block, the first drop down should not show `x position`, `y position`, `direction`, `costume #`, `costume name`, or `size`

### Actual Behavior

All of those options appear
screen shot 2018-06-18 at 5 05 41 pm

### Steps to Reproduce

Set the ___ of ___ sensing block to the stage
Click on the first drop down

Chrome on Mac

picklesrus:
… ignoring them like Scratch2 and makes the attribute menu update based on what was chosen in the target menu.  See issues for context.

### Resolves

Resolves #1762 and resolves LLK/scratch-gui#2389 and resolves LLK/scratch-gui#2390.


src/blocks/scratch3_sensing.js
```diff
@@ -313,12 +313,11 @@ class Scratch3SensingBlocks {
             }
         }
 
-        // Variables
+        // Local variables (not lists).
```
kchadha: I think the word "Local" here is misleading and should probably just say "Target's variables" or something

src/engine/blocks.js
```diff
@@ -563,6 +563,16 @@ class Blocks {
                 if (!optRuntime){
                     break;
                 }
+                // The selected item in the sensing of block menu needs to change based on the
+                // selected target.  Set it to the first item in the menu list.
+                if (block.opcode === 'sensing_of_object_menu') {
+                    if (block.fields.OBJECT.value === '_stage_') {
+                        this._blocks[block.parent].fields.PROPERTY.value = 'backdrop #';
+                    } else {
+                        this._blocks[block.parent].fields.PROPERTY.value = 'x position';
```
kchadha, line 576: I think the only problem I have with this is that we could decide to change the contents of the menu and maybe add an attribute to the top of the menu or re-arrange the menu and this would still be hardcoded to `x position`. Maybe we can keep track of the sprite menu options and the stage menu options in exported constants in the sensing category file `src/blocks/scratch3_sensing.js` and have the scratch-gui code also use those same constants (we might have to re-export them in the top level virtual machine file as well)? This does seem a bit like over kill though....

src/engine/runtime.js
```diff
@@ -597,6 +597,14 @@ class Runtime extends EventEmitter {
         return 'BLOCKSINFO_UPDATE';
     }
 
+    /**
+     * Event name for reporting that block was updated and needs.
```
kchadha: Incomplete comment?

test/unit/blocks_sensing.js
```diff
@@ -229,6 +229,33 @@ test('loud? boolean', t => {
     t.end();
 });
 
+test('get attribute of sprite variable', t => {
+    const rt = new Runtime();
+    const sensing = new Sensing(rt);
+    const s = new Sprite();
+    const target = new RenderedTarget(s, rt);
+    const thing = {
```
kchadha: maybe rename this to something more descriptive?

test/unit/blocks_sensing.js
```diff
@@ -229,6 +229,33 @@ test('loud? boolean', t => {
...
+        name: 'cars',
+        value: 'trucks'
+    };
+    rt.getSpriteTargetByName = () => target;
+    target.lookupVariableByNameAndType = () => thing;
+    t.equal(sensing.getAttributeOf({PROPERTY: 'cars'}), 'trucks');
+
+    t.end();
```
kchadha, line 249: I think these tests would be more useful if you mocked out the actual variable and added it to the target (and then called getAttributeOf with both the property and the object set, what do you think?

test/unit/blocks_sensing.js
```diff
@@ -229,6 +229,33 @@ test('loud? boolean', t => {
...
+});
+test('get attribute of variable that does not exist', t => {
+    const rt = new Runtime();
+    const sensing = new Sensing(rt);
+    const s = new Sprite();
+    const stage = new RenderedTarget(s, rt);
+    rt.getTargetForStage = () => stage;
+    stage.lookupVariableByNameAndType = () => null;
+    t.equal(sensing.getAttributeOf({PROPERTY: 'stage'}), 0);
```
kchadha: The non-existent variable should have a less confusing name because it's not immediately clear what's happening here.

kchadha (changes requested):
I think general approach is reasonable, with a comment about hard coding the menu option names.
However, I'm not sure that's worth changing if it ends up being a lot more work...

Also left some comments about the added unit tests.

src/blocks/scratch3_sensing.js
picklesrus: Fixed!

src/engine/blocks.js
picklesrus: I agree that is isn't the best. Trying to prioritize though.. How about I file a tech debt issue for tracking?

src/engine/runtime.js
picklesrus: whoops. Fixed.

test/unit/blocks_sensing.js
picklesrus: Sorry. Changed to variable.

test/unit/blocks_sensing.js
picklesrus: Agreed. Mostly copy/paste error.

test/unit/blocks_sensing.js
picklesrus: I'm torn. IIUC, I think doing that means testing the functionality of lookupVariableByNameAndType, which should (at least theoretically) be tested somewhere else.  Maybe it isn't though?  To me, it feels less like a unit test if I do it the way you suggested, but probably would increase our test coverage numbers.

picklesrus:
Did most things, 2 real questions/comments left - one about sharing the menu constants and one about what to mock in the tests.

test/unit/blocks_sensing.js
kchadha: Part of what we're testing is whether you're providing things like `lookupVariableByNameAndType` the right arguments inside of `getAttributeOf`. Stubbing out `lookupVariableByNameAndType` doesn't get at that and thus isn't really testing whether the logic inside of `getAttributeOf` is correct. I don't think changing it to remove the `lookupVariableByNameAndType` stub would make this a duplicate test of the `lookupVariableByNameAndType` functionality but rather a test of whether the `getAttributeOf` logic functions correctly as a whole...

IMO, it's still a unit test, just testing `getAttributeOf` as one whole unit. Even if we change the internals of `getAttributeOf` and maybe even stop using `lookupVariableByNameAndType`, I would still expect its unit test to work assuming we're not changing the fundamental functionality of what `getAttributeOf` does.

Thoughts?

kchadha: Now I'm a little torn too haha.

picklesrus: Went back to not mocking lookupVariableByNameAndType. I couldn't add variables with target.createVariable because I couldn't figure out how to have it set a value too so I just smashed it right into the dictionary... But, maybe I'm just missing something.

kchadha: You can set a value after creating the variable by `target.variables[<varId>].value = <thing you want>`

Putting the variable in the dictionary is fine too.

picklesrus:
Closing this in favor of https://github.com/LLK/scratch-vm/pull/1786 because github is being weird about reading the diff :(

PR #1775, opened 2018-11-21
Cloud rename delete

kchadha:
### Proposed Changes

Allows renaming and deleting cloud variables and requests the cloud provider to rename/delete the cloud variable on the server.

### Reason for Changes

Exists in 2.0.

### Test Coverage

Added unit tests for new functionality.

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -241,7 +241,7 @@ class Target extends EventEmitter {
             const newVariable = new Variable(id, name, type, false);
             this.variables[id] = newVariable;
             if (isCloud && this.isStage) {
-                this.runtime.ioDevices.cloud.requestCreateCloudVariable(newVariable);
+                this.runtime.ioDevices.cloud.requestCreateVariable(newVariable);
             }
         }
     }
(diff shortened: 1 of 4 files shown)


src/engine/target.js
```diff
@@ -314,8 +319,12 @@ class Target extends EventEmitter {
      */
     deleteVariable (id) {
         if (this.variables.hasOwnProperty(id)) {
+            const variable = this.variables[id];
```
paulkaplan: This is a minor thing, but I think holding on to a "variable" object that gets deleted feels like it could lead to some strange bugs/brain confusion editing this code in the future. If all you need is the name / cloudyness, maybe just take out those fields as "deletedVarName" and "deletedVariableWasCloud" or something?

paulkaplan (dismissed):
LGTM other than the minor issue about holding onto a deleted object. Approved with or without that change, up to you.

src/engine/target.js
kchadha: Makes sense!

kchadha:
```suggestion
            // Get info about the variable before deleting it
            const deletedVariableName = this.variables[id].name;
            const deletedVariableWasCloud = this.variables[id].isCloud;
```

src/engine/target.js
```diff
@@ -314,8 +319,12 @@ class Target extends EventEmitter {
...
             delete this.variables[id];
             if (this.runtime) {
+                if (variable.isCloud && this.isStage) {
```
kchadha:
```suggestion
                if (deletedVariableWasCloud && this.isStage) {
```

src/engine/target.js
```diff
@@ -314,8 +319,12 @@ class Target extends EventEmitter {
...
+                    this.runtime.ioDevices.cloud.requestDeleteVariable(variable.name);
```
kchadha:
```suggestion
                    this.runtime.ioDevices.cloud.requestDeleteVariable(deletedVariableName);
```

paulkaplan (approved):

PR #1779, opened 2018-11-26
Add "project changed" event

rschamp:
And emit it whenever we think the project has changed. Try to not emit it when a change has happened internally that shouldn't affect the serialized project.

This iteration fires the event too frequently, e.g., when switching sprites. This is meant as a simple initial implementation that can be improved.

/cc @benjiwheeler @cwillisf @kchadha

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -502,6 +502,14 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Event name for report that a change was made that can be saved
+     * @const {string}
+     */
+    static get PROJECT_CHANGED () {
+        return 'PROJECT_CHANGED';
+    }
+
+    /**
      * Event name for targets update report.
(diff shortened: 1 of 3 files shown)


src/engine/runtime.js
```diff
@@ -1737,7 +1745,6 @@ class Runtime extends EventEmitter {
         // Script glows must be cleared.
         this._scriptGlowsPreviousFrame = [];
         this._updateGlows();
-        this.requestTargetsUpdate(editingTarget);
```
rschamp, line 1740: This is ok to do because everywhere that calls `runtime.setEditingTarget` already also calls `emitTargetsUpdate` itself. It is necessary to remove this here so we can more carefully control when the project changed event is also called with the targetsUpdate event.

paulkaplan (approved):
LGTM

PR #1782, opened 2018-11-27
Makey Makey extension

ericrosenbaum:
Hey cool, an extension for the Makey Makey!

**Wait... what?**

Uh, hm. Good point. Technically, Makey Makey already just works with Scratch 3.0. You can use the existing "when key pressed" block. 

**So, like, why?**

- Makey Makey is already used with Scratch by lots of people. We thought putting it into the extension library would help give it some more visibility in the Scratch community.
- The extension's "when key pressed" block has a shorter menu, containing only the 11 keys available on the Makey Makey. So it's a tiny bit easier to use.
- Also we added a cool new block!

**Woah what's the new block?**

It's "when [left up right] pressed in order". It lets you detect that a sequence of keyboard keys was pressed. The menu contains several other sequences. These can be used, for example, to add secret codes to games and other projects (unlock the secret level!).

Some people have also experimented with a kind of "gesture" sensing using makey makey: put two or more inputs (such as graphite shapes or foil patches) in a row, and swipe your finger across them to make something happen. This is an advanced makey makey technique, but it's cool, and this block makes it much easier to try.

You can also drop a variable onto it with space-separated key names, to create your own sequence (such as "up down left right" or "a b c d"). 

screen shot 2018-11-27 at 10 14 25 am


benjiwheeler (commented):
Exciting, I love this.

src/extensions/scratch3_makeymakey/index.js
```diff
@@ -0,0 +1,372 @@
...
+         * A toggle that alternates true and false each frame, so that an
+         * edge-triggered hat can trigger on every other frame.
+         * @type {boolean}
+         */
+        this.frameToggle = false;
+
+        // Set an interval that toggles the frameToggle every frame.
+        setInterval(() => {
+            this.frameToggle = !this.frameToggle;
+        }, this.runtime.currentStepTime);
```
rschamp, line 84: We need a step event...

src/extensions/scratch3_makeymakey/index.js
```diff
@@ -0,0 +1,372 @@
...
+            }
+        }
+    }
+
+    /*
+     * Add a key sequence to the set currently being checked on each key press.
+     * @param {string} sequenceString a string of space-separated KEY_IDs.
+     * @param {array} sequenceArray an array of KEY_IDs.
+     */
+    addSequence (sequenceString, sequenceArray) {
```
rschamp, line 345: Why doesn't it matter that there is `addSequence` but not `removeSequence`? What happens after you don't listen for a sequence anymore? What if someone put a variable into the hat block and then changed that variable in a forever loop?

rschamp (approved):
This looks good to me! I have a question you probably already considered.

src/extensions/scratch3_makeymakey/index.js
ericrosenbaum: that would be handy!

src/extensions/scratch3_makeymakey/index.js
ericrosenbaum: I thought about it (only a little bit, really) and decided that it would be a rare and weird situation that is (I think) no more harmful than other things that you can do, likely only on purpose, to mess up Scratch, such as storing tons of data in variables and lists. Then again, we should find a way to fix it! I'll make a follow-up issue to consider different options.

PR #1783, opened 2018-11-27
Emit a RUNTIME_STARTED event to track if the vm has been started

paulkaplan:
Needed to keep track of whether the VM has been started.


rschamp (approved):

rschamp:
It looks like it's failing for something unrelated?

paulkaplan:
@rschamp i have no idea why that was happening...

PR #1784, opened 2018-11-27
Make setLocale consistently refresh blocks

cwillisf:
### Resolves

Resolves LLK/scratch-gui#3438

### Proposed Changes

This change makes the VM's `setLocale` method call `refreshBlocks` even when not changing the language, though it still skips an unnecessary call to `formatMessage.setup` in this case.

### Reason for Changes

The GUI relies on `refreshBlocks` to ensure that the toolbox has updated contents when switching between the community view and the "see inside" view.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -1030,10 +1030,9 @@ class VirtualMachine extends EventEmitter {
      *     updated for a new locale (or empty if locale hasn't changed.)
      */
     setLocale (locale, messages) {
-        if (locale === formatMessage.setup().locale) {
-            return Promise.resolve();
+        if (locale !== formatMessage.setup().locale) {
+            formatMessage.setup({locale: locale, translations: {[locale]: messages}});
         }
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM given discussions offline about making sure that this does not cause us to unnecessarily refresh the workspace or blocks, but that doesn't seem like the case here.

I still think we should get @chrisgarrity's review to make sure these changes are sound.

chrisgarrity (approved):
Looks good. I'm still not clear on why vm.setLocale is always getting called when switching between community view and see inside, but assuming that it is, this should force the blocks to render in the current locale.

cwillisf:
> I'm still not clear on why vm.setLocale is always getting called when switching between community view and see inside

The blocks container (`src/container/blocks.jsx`) calls `setLocale` from `componentDidMount` unless the container is not visible. Since we basically rebuild the whole UI when we switch, the component ends up getting remounted at that time.

PR #1797, opened 2018-12-03
use Object.keys instead of for ... in, because with Opal.

takaokouji:
### Proposed Changes

use Object.keys instead of for ... in in Blocks.getAllVariableAndListReferences.

### Reason for Changes

If added some methods to Array.prototype by some libraries like Opal, it's failed.

### Test Coverage

I checked scratch-gui with this changes. Drag and drop blocks from a sprite to another sprite, and it's done.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -799,7 +799,7 @@ class Blocks {
     getAllVariableAndListReferences (optBlocks) {
         const blocks = optBlocks ? optBlocks : this._blocks;
         const allReferences = Object.create(null);
-        for (const blockId in blocks) {
+        Object.keys(blocks).forEach(blockId => {
             let varOrListField = null;
             let varType = null;
             if (blocks[blockId].fields.VARIABLE) {
@@ -823,7 +823,7 @@ class Blocks {
                     }];
                 }
             }
-        }
+        });
         return allReferences;
     }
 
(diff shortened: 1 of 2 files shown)


takaokouji:
I'm developing Smalruby 3 ( https://smalruby.jp/smalruby3-gui ) based scratch-gui. I hope it will be bridge of Ruby and Scratch. It's use Opal, so I made this PR. Please merge this.

paulkaplan:
Hi @takaokouji. thanks for the interest, that looks like a cool project. Could you explain a bit more about the need to move away from for/in? It is quite a common JS idiom used in many places in our codebase, and we would need to introduce a lint rule or something to prevent it from coming back.

Also sorry about the delay in reviewing, we are gearing up for the 3.0 launch and so are trying to stabilize things.

takaokouji:
@paulkaplan 

> Could you explain a bit more about the need to move away from for/in?

OK. I think to move away only for Array object. Almost your projects's `for ... in` are for Object, it's OK. However, sometime `for ... in` for Array object make trouble like my case that is with Opal. Opal is to add some method Array prototype.

So, please use `for ... of` or Array.prototype.forEach instead of `for ... in` for Array object.

takaokouji:
see https://eslint.org/docs/rules/guard-for-in .

PR #1801, opened 2018-12-03
Fix usage of loadCostumeFromAsset.

paulkaplan:
Fixes the issue where duplicating a sprite did not correctly assign a new skinId for the renderer. It was because the renderer didn't appear to be available, so a new skinId was not assigned.

Introduced by https://github.com/LLK/scratch-vm/pull/1737

Fixes https://github.com/LLK/scratch-gui/issues/3924


fsih (approved):
Ack! I'm surprised it didn't break more, actually.

Thanks for fixing!

PR #1803, opened 2018-12-03
Remove promise finally because we are having trouble with polyfills

paulkaplan:
@fsih's recommendation to avoid issues with Promise.finally availability on Edge.


fsih (approved):
LGTM! I tested out Chrome and Edge

Issue #1807, opened 2018-12-04
Task queue tests are flaky


rschamp:
### Expected Behavior

_Please describe what should happen_
`npm test` should consistently pass

### Actual Behavior

_Describe what actually happens_
The task queue unit tests sometimes pass, and sometimes fail. Most of the time on Travis they seem to fail.

rschamp:
Builds after #1804 still sometimes fail, so this is still an issue. Fixed temporarily with #1808.

PR #1804, opened 2018-12-04
Task Queue unit test stubbed timer

evhan55:
Potentially Resolves #1807: Task queue tests are flaky

Could be a solution instead of #1808: Skip task queue tests

Tested Locally:
- Ran `npm run env -- tap ./test/unit/util_task-queue.js` and `4/4` passed
- Ran `npm test` and `2487/2487` passed

PR #1808, opened 2018-12-04
Skip task queue tests

rschamp:
They are flaky. Pending resolution of #1807, skip them.


rschamp:
Fixed for real in #1804

rschamp:
#1804 didn't actually fix it, so doing this for now.

PR #1809, opened 2018-12-04
Serialize Infinity and NaN to 0

### LINKED ISSUE #1746: Numeric Infinity (1/0) turns into "0" when saving to sb2 format
### Expected Behavior

In scratch 2: You should be able to save the value (1/0) (constructed using the `() / ()` operator block). The value is listed as `Infinity` in the list or variable.

### Actual Behavior

In scratch 2: You cannot save the value (1/0) (constructed using the `() / ()` operator block). The value is listed as `Infinity`, then is turned into `0` in the sb2 creation.

### Steps to Reproduce

1. Either:
  a. Create a variable.
  b. Give the variable the value (1/0) (using the `() / ()` operator block) using the `set () to ()` block.
2. Or:
  a. Create a list.
  b. Add the value (1/0) (using the `() / ()` operator block) to the list using the `add () to ()` block.
3. Notice that the number shows up as `Infinity` in the visual representation of the list or variable.
4. Save the project and download the program.
5. Exit the window.
6. Notice:
   a. The value has turned to `0`
   b. If you inspect the project.json, the value there is `0` as well

See this project for an example: https://scratch.mit.edu/projects/260496681/#editor

This may or may not be a compatibility issue with scratch 3.

Debian 10 and Firefox 62

paulkaplan:
### Resolves

Resolves #1746 

### Proposed Changes

Serialize +/- Infinity and NaN to 0 when saving, same as Scratch2.

### Reason for Changes

Because Infinity and NaN get serialized to `null` which is not of type number, the parser validation fails for projects with variables set to Infinity or NaN.

Scratch2 handles this by serializing those values to 0. Do the same thing here.

### Test Coverage

Added a string util for this particular type of serialization, added unit tests to cover it and its usage in VM.

### DIFF
--- a/test/unit/util_string.js
+++ b/test/unit/util_string.js
@@ -63,3 +63,14 @@ test('unusedName', t => {
     );
     t.end();
 });
+
+test('stringify', t => {
+    const obj = {a: Infinity, b: NaN, c: -Infinity, d: 23, e: 'str'};
+    const parsed = JSON.parse(StringUtil.stringify(obj));
+    t.equal(parsed.a, 0);
+    t.equal(parsed.b, 0);
+    t.equal(parsed.c, 0);
+    t.equal(parsed.d, 23);
+    t.equal(parsed.e, 'str');
+    t.end();
+});
--- a/test/unit/util_string.js
+++ b/test/unit/util_string.js
@@ -65,12 +65,22 @@ test('unusedName', t => {
 });
 
 test('stringify', t => {
-    const obj = {a: Infinity, b: NaN, c: -Infinity, d: 23, e: 'str'};
+    const obj = {
+        a: Infinity,
+        b: NaN,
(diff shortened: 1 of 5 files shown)


test/unit/util_string.js
```diff
@@ -63,3 +63,14 @@ test('unusedName', t => {
     );
     t.end();
 });
+
+test('stringify', t => {
```
kchadha, line 67: Could you also add a test that the new stringify function handles nested objects correctly?

E.g. `{a: Infinity, b: {b1: Infinity, b2: 42}}`

kchadha (commented):
Requesting one more addition to the unit test for stringify, otherwise looks good!

paulkaplan:
@kchadha fixed

kchadha (approved):
LGTM!

PR #1815, opened 2018-12-04
Clear the pen layer when runtime dispose happens.

### LINKED ISSUE #1850: Pen layer not cleared if you switch projects
### Expected Behavior
If you load a saved (video motion) project from your computer after playing a project that uses pen to draw, the camera should turn on and be visible. Anything drawn by pen in the previous project should be cleared.

### Actual Behavior

First of all, any backdrop is in front of the video layer, so you cannot see the video even though it's active and can be sensed. After removing the backdrop, the video is visible along with any previous pen layer.

### Steps to Reproduce
* load https://scratch.mit.edu/projects/116093633/ and run it to generate trails on the pen layer
* upload a video motion project from a local file (e.g. https://scratch.mit.edu/projects/217939203/)
* Video is running, but is not visible
* edit the backdrop to create a new transparent backdrop
screen shot 2018-04-23 at 12 36 41 pm

Mac OS 10.13.3 Chrome 66

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/1850

### Proposed Changes

Add a RUNTIME_DISPOSED event that the pen extension listens for and clears the pen layer when it occurs. 

No attempt is made in this PR to unload the extension blocks or anything like that. 

Repro that now works:
1. Add pen extension
2. Move the cat around and stamp a bunch of times
3. "File -> New"
4. See that the pen layer is cleared.

### DIFF
--- a/test/unit/engine_runtime.js
+++ b/test/unit/engine_runtime.js
@@ -239,3 +239,14 @@ test('setCompatibilityMode does not restart if it was not running', t => {
     t.equal(started, false);
     t.end();
 });
+
+test('Disposing the runtime emits an event', t => {
+    let disposed = false;
+    const rt = new Runtime();
+    rt.addListener('RUNTIME_DISPOSED', () => {
+        disposed = true;
+    });
+    rt.start();
+    t.equal(disposed, true);
+    t.end();
+});
--- a/test/unit/engine_runtime.js
+++ b/test/unit/engine_runtime.js
@@ -246,7 +246,7 @@ test('Disposing the runtime emits an event', t => {
     rt.addListener('RUNTIME_DISPOSED', () => {
         disposed = true;
     });
-    rt.start();
+    rt.dispose();
     t.equal(disposed, true);
     t.end();
 });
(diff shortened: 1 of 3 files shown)


test/unit/engine_runtime.js
```diff
@@ -239,3 +239,14 @@ test('setCompatibilityMode does not restart if it was not running', t => {
...
     t.end();
 });
+
+test('Disposing the runtime emits an event', t => {
+    let disposed = false;
+    const rt = new Runtime();
+    rt.addListener('RUNTIME_DISPOSED', () => {
+        disposed = true;
+    });
+    rt.start();
```
kchadha:
```suggestion
    rt.dispose();
```

kchadha: I tried to commit the above suggestion but turns out I don't have permission to push to your fork...

kchadha (changes requested):
Failing test, but after that LG.

paulkaplan:
Thanks for that! I was being too hasty! Included your suggestion.

kchadha (approved):
LGTM

PR #1818, opened 2018-12-05
Clear out the blocks in dispose. Fixes #1758 where old monitored vari…

### LINKED ISSUE #1758: Monitors (and monitored variables) from old projects are kept around across project loads
### Expected Behavior

Loading a new project should not keep around variables from an old project.

### Actual Behavior

Loading a new project (via file upload) after having a project with monitors checked previously, 
keeps around the variables from the old project in the new project. This is because the monitorBlocks container is not getting cleared out when clearing the VM runtime.

### Steps to Reproduce

1. Create a new project
2. Create some variables and monitor them
3. Load a new project from the filesystem
4. Switch sprites and switch back or switch between the costume tab and the code tab
5. Observe that the variables from your old project are in the toolbox

### Proposed Solution
Remove knowledge of previous monitors when clearing the runtime

Thanks to @summerscar for finding this issue!

picklesrus:
…ables were getting kept across project loads

### Resolves

 Fixes #1758

### Proposed Changes
The original approach of clearing out all of monitor blocks didn't work (see the note below) because we recycle stage monitor blocks.

This way only deletes sprite specific monitors and global variables but leaves stage monitors (like backdrop number) alone.

### Test Coverage

Added integration test and verified that it fails without the change to dispose.


kchadha (dismissed):
Hmmm it looks like there's something weird going on with the monitor blocks that have a menu.

You can observe via: 
1. Create a monitor for backdrop number
2. Create a monitor for backdrop name; observe that the backdrop name monitor appears empty
3. Change the selected option in the menu of the _costume_ block in the toolbox; observe that the backdrop name monitor got filled in

It's not immediately clear what's going on here, but I think we should investigate and then decide what to do about it.

picklesrus:
Fixed @kchadha's bug (great fine!). Since we ended up working together and tried multiple approaches, looking for a new set of eyes on this PR.

paulkaplan (approved):
Meeps, needing to tip-toe around certain recycled blocks makes me scared. But if it works! 

We might want to file an "ergonomics" issue for clarifying what "dispose" means for the runtime, since it is becoming pretty specialized. It doesn't, for instance, stop the stepping loop for threads, which may or may not be an oversight...

kchadha:
@paulkaplan, that sounds like a good idea. Yeah, runtime dispose becoming suddenly populated with a lot of different things is definitely worrisome and something to keep an eye on. This was a great example of us not realizing that we're making assumptions in other places that certain things (e.g. stage monitor blocks in the monitor block container) are going to stick around and things break if they are cleared out. It would be good to track these kinds of things in `ergonomics` or `tech debt` issues.

picklesrus:
Filed #1829 and labeled it both tech debt and ergonomics. If there are other dispose-y things that show up soon, please pile them on.

PR #1821, opened 2018-12-05
Fix an issue where the targetId of a monitor could be reassigned

### LINKED ISSUE #1623: Extraneous local variable created via Hide/Show
### Expected Behavior

Local variables that are not defined or referenced in a sprite should not be dynamically created for that sprite.

### Actual Behavior

It's possible for stacks in a sprite to create local variables in another sprite.

### Steps to Reproduce

Add a new sprite (such as "Ben") to a new project.
Create a local variable in Ben (such as "Local").
Create this stack in Ben:  When GF Clicked; Hide variable "Local"; Show variable "Local".
Switch to Sprite1 (there should be no stacks and the only variable is the default "my variable".
Click the GF twice.

Notice that the monitor for "Local" now says "Sprite1: Local".  

Switch to Ben and then switch back to Sprite1.

Notice that "Local" now appears in the list of variables for Sprite1.

Windows 7 / Chrome Version 69.0.3497.100 (Official Build) (64-bit)

Here's a small [test case.](https://beta.scratch.mit.edu/compatibility-testing.html#249245634)  [Scratch 2 link](https://scratch.mit.edu/projects/249245634) and [Scratch 3 link](https://beta.scratch.mit.edu#249245634).

and the problem was reported in the forums [here:](https://scratch.mit.edu/discuss/topic/308114/?page=12#post-3263867)

paulkaplan:
This caused a local variable monitor to try running its thread on a different sprite, causing a new local variable to be created with the same monitor id.

### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1623

### Proposed Changes

The change is twofold
1. Fix the way `isSpriteLocalVariable` check works, it does not depend on the editing target at the time.
3. Prevent reassigning `targetId` on blocks, to allow checkbox to be "checked" (via hide/show block) 

Consternation ensued about whether the hide/show should be emitting checkbox events to the runtime (it probably shouldn't be).

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -625,18 +625,22 @@ class Blocks {
             // Variable blocks may be sprite specific depending on the owner of the variable
             let isSpriteLocalVariable = false;
             if (block.opcode === 'data_variable') {
-                isSpriteLocalVariable = !optRuntime.getEditingTarget().isStage &&
-                    optRuntime.getEditingTarget().variables[block.fields.VARIABLE.id];
+                isSpriteLocalVariable = !(optRuntime.getTargetForStage().variables[block.fields.VARIABLE.id]);
             } else if (block.opcode === 'data_listcontents') {
-                isSpriteLocalVariable = !optRuntime.getEditingTarget().isStage &&
-                    optRuntime.getEditingTarget().variables[block.fields.LIST.id];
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM

PR #1822, opened 2018-12-05
Make duplicateVariable deep copy the array instead of keeping reference.

### LINKED ISSUE #1798: Clones inherit "For this sprite only" list
### Expected Behavior
Like variables are, "for this sprite only" lists should be independent on each clone.

### Actual Behavior
They share the same list.

### Steps to Reproduce
I made a simple project to test it. 
image
[LocalListBug.zip](https://github.com/LLK/scratch-vm/files/2638846/LocalListBug.zip)
(This is a sb3 file. Change extension to `sb3`. )
Press space key. On Scratch 2.0, the parent always says "0".
On Scratch 3.0, both the parent and the clone say the same number.

Windows 7 / Firefox 63

Thanks to videobob for [finding this strange bug](https://scratch.mit.edu/discuss/post/3333723/)!

paulkaplan:
### Resolves

Fixes https://github.com/LLK/scratch-vm/issues/1798

### Proposed Changes

Make duplicateVariable deep copy the array instead of keeping reference.

### Reason for Changes

Previously it looked like clones were sharing private lists, but it was actually just a bug in the way the values were copied while duplicating

### Test Coverage

Added unit test

### DIFF
--- a/test/unit/engine_target.js
+++ b/test/unit/engine_target.js
@@ -323,6 +323,18 @@ test('duplicateVariable creates a new variable with a new ID by default', t => {
     t.end();
 });
 
+test('duplicateVariable creates new array reference for list variable.value', t => {
+    const target = new Target();
+    const arr = [1, 2, 3];
+    target.createVariable('a var ID', arr, Variable.LIST_TYPE);
+    const originalVariable = target.variables['a var ID'];
+    const newVariable = target.duplicateVariable('a var ID');
+    // Values are deeply equal but not the same object
+    t.deepEqual(originalVariable.value, newVariable.value);
+    t.notEqual(originalVariable.value, newVariable.value);
+    t.end();
+});
+
 test('duplicateVariable creates a new variable with a original ID if specified', t => {
     const target = new Target();
     target.createVariable('a var ID', 'foo', Variable.SCALAR_TYPE);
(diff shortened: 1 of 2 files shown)


test/unit/engine_target.js
```diff
@@ -323,6 +323,18 @@ test('duplicateVariable creates a new variable with a new ID by default', t => {
     t.end();
 });
 
+test('duplicateVariable creates new array reference for list variable.value', t => {
+    const target = new Target();
+    const arr = [1, 2, 3];
+    target.createVariable('a var ID', arr, Variable.LIST_TYPE);
```
kchadha: Hmmm not sure what's happening here but the second argument to `createVariable` should be a string not an array..

kchadha:
```suggestion
    target.createVariable('a var ID', 'arr', Variable.LIST_TYPE);
```

test/unit/engine_target.js
```diff
@@ -323,6 +323,18 @@ test('duplicateVariable creates a new variable with a new ID by default', t => {
...
+    const originalVariable = target.variables['a var ID'];
```
kchadha, line 330:
```suggestion
    const originalVariable = target.variables['a var ID'];
    originalVariable.value = arr;
```

kchadha (commented):
The unit test has a typo and a missing value setting line. LG after that.

PR #1825, opened 2018-12-06
Emit PROJECT_START event publicly on green flag click.

paulkaplan:
re-emit the PROJECT_START event publicly on the VM.

Eventually part of https://github.com/LLK/scratch-www/issues/1858


kchadha (approved):
LGTM

PR #1831, opened 2018-12-09
Clear ask bubble when stopForTarget

apple502j:
### Resolves
Resolves https://github.com/LLK/scratch-gui/issues/3932

### Proposed Changes
Say nothing when `stop other scripts`

### Reason for Changes
To clear the bubble

### DIFF
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -143,7 +143,7 @@ class Scratch3SensingBlocks {
         ));
 
         if (currentlyAsking) {
-            this.runtime.emit('SAY', stopTarget, 'say', null);
+            this.runtime.emit('SAY', stopTarget, 'say', '');
             if (this._questionList.length > 0) {
                 this._askNextQuestion();
             } else {
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -143,11 +143,8 @@ class Scratch3SensingBlocks {
         ));
 
         if (currentlyAsking) {
-            if (this._questionList.length > 0) {
-                this._askNextQuestion();
-            } else {
-                this.runtime.emit('QUESTION', null);
-            }
+            this.runtime.emit('SAY', stopTarget, null, '');
+            this.runtime.emit('QUESTION', null);
         }
     }
 
--- a/src/blocks/scratch3_sensing.js
+++ b/src/blocks/scratch3_sensing.js
@@ -143,8 +143,12 @@ class Scratch3SensingBlocks {
         ));
 
         if (currentlyAsking) {
-            this.runtime.emit('SAY', stopTarget, null, '');
-            this.runtime.emit('QUESTION', null);
+            this.runtime.emit('SAY', stopTarget, 'say', null);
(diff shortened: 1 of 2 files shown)


kchadha (changes requested):
@apple502j, thank you for submitting this pull request!

This fix does not take into account that there could be other questions queued up to be asked after the scripts in that particular sprite are stopped.

For example you could have a project where two different sprites want to ask and wait when the green flag is clicked. "Running `stop other scripts in this sprite` in the first sprite asking the question should not completely clear out the ask bubble for other sprites waiting to ask a question.

See 269439546 for an example project where your changes don't work the same way as scratch 2.0. Click the green flag, and then click on the balloon.

src/blocks/scratch3_sensing.js
```diff
@@ -143,6 +143,7 @@ class Scratch3SensingBlocks {
         ));
 
         if (currentlyAsking) {
+            this.runtime.emit('SAY', stopTarget, 'say', null);
```
kchadha: This line causes the error below and the say bubble does not actually go away. The `null` should be `''` instead.

```
Uncaught TypeError: Cannot read property 'normalize' of null
    at SVGTextWrapper.wrapText (text-wrapper.js:40)
    at SVGTextWrapper.wrapText (svg-text-wrapper.js:114)
    at SVGTextBubble._buildTextFragment (svg-text-bubble.js:178)
```

kchadha:
```suggestion
            this.runtime.emit('SAY', stopTarget, 'say', '');
```

kchadha (changes requested):
There is a minor bug in the code, but after that is resolved, I think this is good to go.

kchadha (approved):
@apple502j, I found another subtle bug, but I think I will file that as a separate issue.

If a sprite is currently asking, but then some other code triggers that same sprite to "say" something (e.g. `when [space] key pressed > say hello`), stopping "other scripts in this sprite" has different behavior in Scratch 2.0 and 3.0. In 2.0, the 'hello' bubble is not cleared. With these changes, in 3.0, the 'hello' bubble is cleared because we are not differentiating between say bubbles and ask bubbles, and technically the sprite was still running the "ask and wait" block when we ran the stop block.

-- Approving this PR since this is closer than what we had previously to the behavior of Scratch 2.0 and the above bug is more of an edge case.

kchadha:
Filed the issue I described in #1848

kchadha (changes requested):
Oops, it looks like there are two unit tests that are failing since the new code causes a function called `getCustomState` on the target to get called, but we're using a fake target object in the tests. 

You can fix this by adding `getCustomState: () => ({})` to the mock target object in the following lines of code: 

https://github.com/LLK/scratch-vm/blob/dfd47358719186027bdd2b0d003dcab61c362f7c/test/unit/blocks_sensing.js#L70

https://github.com/LLK/scratch-vm/blob/dfd47358719186027bdd2b0d003dcab61c362f7c/test/unit/blocks_sensing.js#L97

kchadha (approved):
LGTM! Thank you for making those changes!

PR #1834, opened 2018-12-11
Fix issue where edge-activated hats only run on one sprite after duplicating the sprite

### LINKED ISSUE #1495: Edge-triggered hat only triggers in duplicated sprite, not original
### Expected Behavior

Hat blocks should always trigger as expected, even after duplicating sprites.

### Actual Behavior

Using the hat blocks in the micro:bit extension, which are edge-triggered hats, if I create a stack triggered by the "when button pressed" hat, and then duplicate the sprite, the hat only fires when I press the button in the newly duplicated sprite, not the original.

This problem also affects the core hat block "when [loudness/timer] > 10".

### Steps to Reproduce

- Add a stack like `when loudness > 10 / go to random position`
- Duplicate the sprite
- Note that the hat only triggers in the duplicated sprite

hat-duplication-bug-loudness

OR

- Add the micro:bit extension
- Make a stack like `when button A pressed / turn 15 degrees`
- Press the micro:bit button the see it work
- Duplicate the sprite
- Press the micro:bit button, and only the duplicate sprite turns

hat-duplication-bug

### LINKED ISSUE #1507: Edge triggered hats can trigger continuously after duplication
### Expected Behavior

An edge-triggered hat (such as "when tilted left") should trigger a stack to run once, at the moment the tilt value crosses the threshold in that axis.

### Actual Behavior

If you duplicate a sprite with an edge triggered hat, then change the input on the duplicated hat, both hats will now trigger 30 times per second, instead of once.

In the gif below, at first with just one sprite, it has the correct behavior: each button press causes the sprite to turn 15 degrees. After duplicating and switching the second sprite's hat to "when B button pressed," now each sprite turns continuously when you press its button. 

hat-duplicate-bug

### Steps to Reproduce

- Add the microbit extension
- Build the stack `when A button pressed / turn 15 degrees`
- Note that pressing the button causes the sprite to turn 15 degrees, running the stack once
- Duplicate the sprite
- In the duplicate, change the input on the hat to B
-  Note that pressing either button causes the corresponding sprite to turn continuously

kchadha:
### Resolves

Resolves #1495
Resolves #1507 

### Proposed Changes

Duplicating a sprite does not duplicate block IDs which is a problem for how edge triggered hats work because there is a global map on the runtime which maps edge triggered hat block ids to their values, causing the issue described in #1495.

Resolve this issue by making the key on the runtime's _edgeTriggeredHatValues map more unique by including the targetID of the thread the block belongs to. This should make keys in that map unique for duplicated sprites as well as sprite clones.

### Test Coverage

Tested manually with sprite duplication as well as sprite clones. Added integration tests for duplicating/cloning a sprite with an edge activated hat.


ericrosenbaum:
It's worth checking whether this also addresses https://github.com/LLK/scratch-vm/issues/1507

kchadha:
@ericrosenbaum, good catch! I just tested the repro for #1507, and I believe that issue is also resolved by this PR.

src/engine/runtime.js
```diff
@@ -1233,12 +1233,14 @@ class Runtime extends EventEmitter {
...
      * @param {!string} blockId ID of hat to store value for.
+     * @param {!string} threadTargetId Target ID for the thread that the block belongs to
      * @param {*} newValue Value to store for edge-activated hat.
      * @return {*} The old value for the edge-activated hat.
      */
-    updateEdgeActivatedValue (blockId, newValue) {
-        const oldValue = this._edgeActivatedHatValues[blockId];
-        this._edgeActivatedHatValues[blockId] = newValue;
+    updateEdgeActivatedValue (blockId, threadTargetId, newValue) {
+        const blockAndTargetId = `${blockId}_${threadTargetId}`;
```
mzgoddard: Since this information is now target specific should it be moved onto the Target? I mention this because that would be faster than creating this string for hat blocks that run every frame tick and because it seems weird to have Target specific data on the runtime.

test/integration/hat-threads-run-every-frame.js
```diff
@@ -103,6 +104,100 @@ test('edge activated hat thread not added twice', t => {
...
+ * runtime's _edgeActivatedHatValues map.
+ */
+test('edge activated hat should trigger for both sprites when sprite is duplicated', t => {
+
+    // Project that is similar to loudness-hat-block.sb2, but has code on the sprite so that
+    // the sprite can be duplicated
+    const projectWithSpriteUri = path.resolve(__dirname, '../fixtures/edge-triggered-hat.sb3');
+    const projectWithSprite = readFileToBuffer(projectWithSpriteUri);
+
+    const vm = new VirtualMachine();
```
mzgoddard, line 119:
```js
    const vm = new VirtualMachine();
    vm.attachStorage(makeTestStorage());

    // Start VM, load project, and run
    t.doesNotThrow(() => {
        vm.clear();
        vm.setCompatibilityMode(false);
        vm.setTurboMode(false);

        vm.loadProject(project).then(() => {
```

Is used by a lot of tests. Would it make sense to write this chunk as a test utility like `makeTestStorage` is?

test/integration/hat-threads-run-every-frame.js
```diff
@@ -103,6 +104,100 @@ test('edge activated hat thread not added twice', t => {
...
+            t.equal(vm.runtime.threads.length, 0);
+
+            vm.runtime._step();
+            t.equal(vm.runtime.threads.length, 1);
+            checkIsHatThread(t, vm, vm.runtime.threads[0]);
+            t.assert(vm.runtime.threads[0].status === Thread.STATUS_RUNNING);
+            // Run execute on the thread to populate the runtime's
+            // _edgeActivatedHatValues object
+            execute(vm.runtime.sequencer, vm.runtime.threads[0]);
+            t.equal(Object.keys(vm.runtime._edgeActivatedHatValues).length, 1);
```
mzgoddard: I wonder if there should be a formal (internal?) API for asking about a edge hat value. (Probably on target since they are target specific now?)

mzgoddard: Also maybe it should test that the hat value is specifically `undefined` or some other specific other?

test/integration/hat-threads-run-every-frame.js
```diff
@@ -103,6 +104,100 @@ test('edge activated hat thread not added twice', t => {
...
+ * runtime's _edgeActivatedHatValues map.
+ */
+test('edge activated hat should trigger for both sprites when sprite is cloned', t => {
+
+    // Project that is similar to loudness-hat-block.sb2, but has code on the sprite so that
+    // the sprite can be duplicated
+    const projectWithSpriteUri = path.resolve(__dirname, '../fixtures/edge-triggered-hat.sb3');
+    const projectWithSprite = readFileToBuffer(projectWithSpriteUri);
+
+    const vm = new VirtualMachine();
```
mzgoddard, line 170: More evidence for making this chunk a test utility?

mzgoddard (commented):
I think the edge activated hat values should be stored on the Target object. And there is a common test initialization pattern that could serve well as a utility function.

src/engine/runtime.js
kchadha: Yeah, @paulkaplan and I were also talking about that earlier. I think that makes sense and is an easy enough/low risk change.

test/integration/hat-threads-run-every-frame.js
kchadha: I think I'll probably make an ergonomics issue for this instead of adding it to this PR as it will affect nearly all of the integration tests.

test/integration/hat-threads-run-every-frame.js
kchadha: Mostly I just want to check about their separate existences.. after duplicating / cloning. There are other tests that make sure edge hats work properly

paulkaplan (approved):
LGTM

test/integration/hat-threads-run-every-frame.js
kchadha: Tracking in #1844

PR #1838, opened 2018-12-12
Add <is the pen down?> block

### LINKED ISSUE #1837: Add <is the pen down?> block
I was playing around in the ```lib.min.js``` file in the gui, and it only took me a few lines of code to include a ```<is the pen down?>``` block to the pen extension. This has been a long requested feature on the Scratch forums, and is incredibly easy to implement.

I haven't found any issues with it so far, and works correctly across all sprites AFAIK

**Block definition:**
```
{
    opcode: "isDown",
    text: l({
        id: "pen.isDown",
        default: "is the pen down?",
        description: "says if the pen is down"
    }),
    blockType: i.BOOLEAN
},
```
**Block function:**
```
{
    key: "isDown",
    value: function(e, t) {
        var A = t.target,
            n = this._getPenState(A);
        return n.penDown
    }
},
```
I'll open a PR to the extension ```src/extensions/scratch3_pen/index.js``` file in a few

ZenithRogue:
Resolves #1837

### DIFF
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -327,6 +327,15 @@ class Scratch3PenBlocks {
                     })
                 },
                 {
+                    opcode: "isDown",
+                    blockType: BlockType.BOOLEAN,
+                    text: formatMessage({
+                        id: 'pen.isDown',
+                        default: 'is the pen down?',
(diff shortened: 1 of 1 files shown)


thisandagain:
We will need to review your feature request before we consider an implementation PR.

Issue #1729, opened 2018-11-06
On matter of scoping and tail call optimization.


ghost:
As seen by this project, scratch 3.0 uses dynamic scoping as opposed to static scoping: https://scratch.mit.edu/projects/259998618/
This is an issue as it is (1) different from scratch 2.0, which may lead to incompatibility, and in addition, it makes it much harder for tail call optimization to be implemented.
Also, I would like to request that tail call optimization be implemented, as as scratch is meant to be a tool for starting programming, it should be able to facilitate use of many styles of programming.

towerofnix:
In my view, dynamic scoping is neat - but less important than tail-call optimization, because TCO probably enables more types of programming than dynamic scoping (which can partially be worked around via variables or lists), and dynamic scoping hypothetically causes incompatibilities. (But I'd be surprised if there exists a project which relies on the value of a custom block's nonexistent input that shares the name with one of a custom block calling that one to be 0.)

Edit: BTW, tied forum topic: https://scratch.mit.edu/discuss/topic/320438/

joker314:
Regarding dynamic scoping only:

<hr>

I would've thought dynamic scoping was removed in #1640, where an extra `return null;` is added to engine/threads.js

#### Old behaviour
Traverse the call stack until we find a custom hat which has a parameter with the correct name (dynamic scoping)

#### New behaviour
Traverse the call stack until we find **any** custom hat (static scoping)

<hr>

Procedures invoking broadcast scripts (invoking more broadcast scripts...) allows the broadcast scripts to access the parameters of the procedure which invoked it. This is dynamic scoping. I don't know 2.0 behaviour regarding this.

<hr>

The PR was merged 13 days ago. Maybe it hasn't been merged in by gui, hasn't reached production, and perhaps llk.github.io/scratch-vm isn't updated? Or that PR didn't solve the issue for some reason? Or I'm completely wrong?

joker314:
Problem has been idenitified by imfh and towerofnix on the forums.

When a custom block has no parameters, its frame's `params` object is `null`.

Back when Scratch 3.0 had dynamic scoping, we would--for efficiency--`continue` traversing the stack frames if we got `null`, because `null` means the custom procedure has no arguments, and hence can't contain an argument we're referencing. (thanks to imfh for discovering that non-`null` params work okay; to towerofnix for identifying the relevant code to remove)
[...]

PR #1839, opened 2018-12-12
fix: initialize stack frame params for all procedures

mzgoddard:
### Resolves

Related to https://github.com/LLK/scratch-vm/issues/1729

### Proposed Changes

Initialize parameters before pushing parameters onto a procedure's stack frame. This way all procedure stack frames define params. (Not all stack frames are for procedures. Some are for branching blocks like if or forever.)

### Reason for Changes

A [previous change](https://github.com/LLK/scratch-vm/pull/1640) fixed compatibility with Scratch 2, removing 3's unintentional scope leaking. This furthers that change so that procedures with no parameters will also not accidentally use values in other procedure stacks.

### Test Coverage

Add an additional execute integration test following the existing `procedures-nested-missing-number-param.sb2` test.


mzgoddard:
- Add some comments to the new methods

mzgoddard:
- Fix unit thread unit test that should call the new `initParams` function.

src/blocks/scratch3_procedures.js
```diff
@@ -38,6 +38,7 @@ class Scratch3ProcedureBlocks {
 
             const [paramNames, paramIds, paramDefaults] = paramNamesIdsAndDefaults;
 
+            util.initParams();
```
kchadha, line 44: Could you add a comment above this line indicating that this initializes the stack frame for the procedure call so that even procedures without parameters get their stack frames initialized.

```suggestion
            // Initialize params for the current stackFrame to {}, even if the procedure does
            // not take any arguments. This is so that `getParam` down the line does not look
            // at earlier stack frames for the values of a given parameter (#1729)
            util.initParams();
```

kchadha (commented):
LG but requesting documentation =]

kchadha (approved):
LGTM

kchadha:
Including this screen shot in here as a good test case:

image


This comes from the project mentioned in #1729

PR #1840, opened 2018-12-12
Use default builtin SVG asset when an SVG can't be loaded

ktbee:
### Resolves

A fix/better user experience for issues like https://github.com/LLK/scratch-gui/issues/3534

### Proposed Changes

Catch errors from parsing SVGs and load the default builtin SVG instead (depends on https://github.com/LLK/scratch-storage/pull/62 to successfully load the default SVG).

### Reason for Changes

Previously if you try to load a project with an SVG that Scratch 3 can't parse, it crashes the entire editor (example project: https://llk.github.io/scratch-gui/develop/#269985286). Previously Scratch 2 seemed to be less picky when loading SVGs (https://scratch.mit.edu/projects/269985286/). 

Loading #269985286 with these changes will switch the invalid SVG background with the default SVG asset, a grey box with a question mark.

### Test Coverage

I had some difficulty figuring out how to test that a default SVG gets loaded, especially since it's the DOMParser that errors in the original/example project and I wasn't sure if we wanted to add a Node package to simulate that. I'm open to suggestions though on how to approach this!

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -2,32 +2,34 @@ const StringUtil = require('../util/string-util');
 const log = require('../util/log');
 
 const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
-    let svgString = costume.asset.decodeText();
-    // SVG Renderer load fixes "quirks" associated with Scratch 2 projects
-    if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
-        log.error('No V2 SVG adapter present; SVGs may not render correctly.');
-    } else if (optVersion && optVersion === 2 && runtime.v2SvgAdapter) {
-        runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
-        svgString = runtime.v2SvgAdapter.toString();
-        // Put back into storage
-        const storage = runtime.storage;
-        costume.asset.encodeTextData(svgString, storage.DataFormat.SVG, true);
-        costume.assetId = costume.asset.assetId;
-        costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
-    }
-    // createSVGSkin does the right thing if rotationCenter isn't provided, so it's okay if it's
-    // undefined here
-    costume.skinId = runtime.renderer.createSVGSkin(svgString, rotationCenter);
-    costume.size = runtime.renderer.getSkinSize(costume.skinId);
-    // Now we should have a rotationCenter even if we didn't before
-    if (!rotationCenter) {
-        rotationCenter = runtime.renderer.getSkinRotationCenter(costume.skinId);
-        costume.rotationCenterX = rotationCenter[0];
-        costume.rotationCenterY = rotationCenter[1];
-        costume.bitmapResolution = 1;
-    }
+    return new Promise(resolve => {
+        let svgString = costume.asset.decodeText();
+        // SVG Renderer load fixes "quirks" associated with Scratch 2 projects
+        if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
+            log.error('No V2 SVG adapter present; SVGs may not render correctly.');
+        } else if (optVersion && optVersion === 2 && runtime.v2SvgAdapter) {
+            runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
+            svgString = runtime.v2SvgAdapter.toString();
+            // Put back into storage
+            const storage = runtime.storage;
+            costume.asset.encodeTextData(svgString, storage.DataFormat.SVG, true);
+            costume.assetId = costume.asset.assetId;
+            costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
+        }
+        // createSVGSkin does the right thing if rotationCenter isn't provided, so it's okay if it's
+        // undefined here
+        costume.skinId = runtime.renderer.createSVGSkin(svgString, rotationCenter);
+        costume.size = runtime.renderer.getSkinSize(costume.skinId);
+        // Now we should have a rotationCenter even if we didn't before
+        if (!rotationCenter) {
+            rotationCenter = runtime.renderer.getSkinRotationCenter(costume.skinId);
+            costume.rotationCenterX = rotationCenter[0];
+            costume.rotationCenterY = rotationCenter[1];
+            costume.bitmapResolution = 1;
+        }
 
-    return Promise.resolve(costume);
+        resolve(costume);
+    });
 };
 
 /**
@@ -212,7 +214,13 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
         rotationCenter = [costume.rotationCenterX, costume.rotationCenterY];
     }
     if (costume.asset.assetType.runtimeFormat === AssetType.ImageVector.runtimeFormat) {
-        return loadVector_(costume, runtime, rotationCenter, optVersion);
+        return loadVector_(costume, runtime, rotationCenter, optVersion)
+            .catch(() => {
+                // Use default asset if original fails to load
+                costume.assetId = runtime.storage.defaultAssetId.ImageVector;
+                costume.asset = runtime.storage.get(costume.assetId);
+                return loadVector_(costume, runtime);
+            });
     }
     return loadBitmap_(costume, runtime, rotationCenter, optVersion);
 };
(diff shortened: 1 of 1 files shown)


src/import/load-costume.js
```diff
@@ -7,7 +7,11 @@ const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
     if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
         log.error('No V2 SVG adapter present; SVGs may not render correctly.');
     } else if (optVersion && optVersion === 2 && runtime.v2SvgAdapter) {
-        runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
+        try {
+            runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
```
fsih, line 11: This code only runs on assets uploaded from sb2, but it's also possible to upload SVGs that contain errors outside of sb2, for example uploading a costume from file. In that case, the "Document does not appear to be SVG" error will be thrown by the line
`    costume.skinId = runtime.renderer.createSVGSkin(svgString, rotationCenter);`
below

ktbee: @fsih good point, thanks for catching that. I've moved everything into a try/catch block to make sure we load the default SVG if any of the SVG loading logic fails. Another approach I was thinking of was to use the default SVG at [the point of the original error](https://github.com/LLK/scratch-svg-renderer/blob/develop/src/svg-renderer.js#L84), though then we would need to add scratch-storage as a dependency to scratch-svg-renderer and it wouldn't allow us to update the costume's asset id from scratch-svg-renderer.

ktbee: @fsih good point, thanks for catching that. I've moved everything into a try/catch block to make sure we load the default SVG if any of the SVG loading logic fails. Another approach I was thinking of was to use the default SVG at [the point of the original error](https://github.com/LLK/scratch-svg-renderer/blob/develop/src/svg-renderer.js#L84), though then we would need to add scratch-storage as a dependency to scratch-svg-renderer and it wouldn't allow us to update the costume's asset id from scratch-svg-renderer.

fsih: I feel like it's not ideal to wrap so much code in a try. Could we limit the try/catch to places where we call v2SvgAdapter.toString?

scratch-svg-renderer shouldn't depend on scratch-storage. It should be limited to a util for performing operations on SVGs (so far it doesn't know about stuff like costumes and assets).

src/import/load-costume.js
```diff
@@ -211,8 +216,14 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
             typeof costume.rotationCenterY === 'number' && !isNaN(costume.rotationCenterY)) {
         rotationCenter = [costume.rotationCenterX, costume.rotationCenterY];
     }
-    if (costume.asset.assetType.runtimeFormat === AssetType.ImageVector.runtimeFormat) {
-        return loadVector_(costume, runtime, rotationCenter, optVersion);
+    if (costume.asset.assetType === AssetType.ImageVector) {
+        return loadVector_(costume, runtime, rotationCenter, optVersion)
+            .catch(() => {
+                // Use default asset if original fails to load
+                costume.assetId = runtime.storage.defaultAssetId.ImageVector;
```
fsih, line 220: Is it possible for defaultAssetId to be null in storage? It looks like it's {} by default. If setting that is required, could we check if it's missing and log an error below, so it's easier to catch (say line 257 where we check that storage exists)?

fsih (changes requested):

src/import/load-costume.js
ktbee: @fsih that makes sense. What do you think of how I have it set up now with wrapping the whole thing in a promise, since we're returning promises (either `reject` or `resolve`) anyways? This way any error from `v2SvgAdapter.toString` get passed up as rejected promises. Alternatively I can add multiple try/catch blocks, this just struck me as a little neater.

src/import/load-costume.js
ktbee: @fsih good question. Taking a look at [scratch-storage](https://github.com/LLK/scratch-storage/blob/29daf3f503f11a30a41106e7f5db0bc10bf26758/src/ScratchStorage.js#L16), the `defaultAssetId` data is populated soon after it's initialized to {}, so I think it's unlikely it will be null intentionally. However, that makes sense to me to add an error log in case this process fails for some reason. I added the error log a little later in the load costume process since not having the default asset isn't crucial until then. Let me know though if you think it should happen earlier.

src/import/load-costume.js
fsih: This looks good!

src/import/load-costume.js
fsih: I think it makes sense to log the error sooner, because then if an issue comes up we will see it, rather than needing to hit another error before we discover that one.

fsih (changes requested):
I tried running this and uploading corrupted SVGs as costumes and as sprites, and no new asset is appearing, and it still logs errors in the console. It's supposed to show a question mark right?

ktbee:
hmm @fsih I'm seeing those errors now as well, though previously no errors would be logged and the default  SVG would load (ie a grey block with a black question mark). I'm investigating what's changed.

ktbee:
@fsih I realized that I had accidentally unlinked my local version of scratch-storage and so my changes from https://github.com/LLK/scratch-storage/pull/62 weren't incorporated in my local environment. We need that PR for the default SVG to load and avoid errors. My scratch-storage PR was merged, so I thought it would now be available to the latest version of scratch-gui, but that doesn't seem to be the case. Do you know how that gets updated?

Also, I've moved the error message to happen earlier in `loadCostume`, if you want to take another look.

fsih:
@ktbee looks like your PR was deployed to npm 5 days ago:
https://travis-ci.org/LLK/scratch-storage/jobs/467591928
https://www.npmjs.com/package/scratch-storage

GUI is still linked to an older version of storage though:
https://github.com/LLK/scratch-gui/blob/develop/package.json#L110

That's surprising; greenkeeper usually finds dependency updates within a few hours, but I don't see a greenkeeper PR. The way to update manually would be to set scratch-storage to 1.2.1 in GUI at https://github.com/LLK/scratch-gui/blob/develop/package.json#L110

fsih:
When I link that storage version, it looks good when uploading an invalid SVG as a costume! It looks weird in the paint editor; we should probably make a bug for that. 
screen shot 2018-12-19 at 11 26 36

However, when I upload an invalid SVG as a sprite, the UI doesn't change, and it shows errors in the console:
invalidsprite

fsih:
Oddly, the image isn't registering as an SVG so it's trying to load as a bitmap
screen shot 2018-12-19 at 11 48 44

I don't know what's going on with the asset type not matching, but it's working for me if I add something like this in loadCostume
```
    if (costume.asset.assetType === AssetType.ImageVector) {
        return loadVector_(costume, runtime, rotationCenter, optVersion)
            .catch(() => {
                // Use default asset if original fails to load
                costume.assetId = runtime.storage.defaultAssetId.ImageVector;
                costume.asset = runtime.storage.get(costume.assetId);
                return loadVector_(costume, runtime);
            });
    } else if (costume.asset.assetType === AssetType.ImageBitmap) {
        return loadBitmap_(costume, runtime, rotationCenter, optVersion);
    } else {
        log.error(`Asset type not supported: ${costume.asset.assetType}`);
        // Use default asset if original fails to load
        costume.assetId = runtime.storage.defaultAssetId.ImageVector;
        costume.asset = runtime.storage.get(costume.assetId);
        return loadVector_(costume, runtime);
    }
```

fsih:
Ok cool, that asset type not matching issue looks like it was already solved in https://github.com/LLK/scratch-vm/pull/1830, so it should be fixed when you merge with the latest vm code.

fsih (changes requested):
1. Update storage dependency in gui
2. Merge with updates to VM code
3. Check to make sure uploading an invalid SVG as a sprite is correctly loading the question mark

ktbee:
@fsih thank you for the follow up steps and the review on https://github.com/LLK/scratch-gui/pull/4150. I assume it's good to merge in, but I thought I'd hold off until I get the "good to merge" label.

Locally I've been rebasing this branch in my scratch-vm repo to incorporate the changes in `develop`, so I think I should have the code for https://github.com/LLK/scratch-vm/pull/1830, however I'm still not able to upload an invalid SVG as a new sprite, so I'm digging into this now.

ktbee:
Okay! I think we're finally there! In my rebasing, I had accidentally reversed Karishma's SVG type fixes. Those are re-incorporated and I've confirmed that I can upload an invalid SVG as a sprite and I get the default grey box. Let me know what you think @fsih! 

I'll also file an issue to follow up on the funky appearance of the default SVG in scratch paint.

fsih (approved):
Awesome, thanks!

PR #1842, opened 2018-12-12
Serialize x and y for topLevel blocks even if they are 0

kchadha:
### Resolves

Fixes an issue where running `clean up` on a block workspace and then saving and loading that project caused the blocks to pile up on each other at 0,0.

### Proposed Changes

Fix falsey check so that we serialize a topLevel block's x and y position even if one or both of the values is 0 (or if otherwise falsey, default to 0).

### Test Coverage

Added unit test and tested manually.

Repro for issue:

1. Drag out a bunch of blocks, connect some stacks together
2. Right click the blocks workspace and click clean up
3. Save the project
4. Load the project (the project should now look the same as before, the blocks keeping their layout after the cleanup -- on develop, the serialized project would have had all the blocks lose their 'x' values, causing all the blocks to end up piled on top of each other on 0,0)


paulkaplan (approved):
LGTM

PR #1852, opened 2018-12-17
depend on scratch-sb1-converter and convert sb1 files to sb2

mzgoddard:
### Resolves

- Initial Scratch 1 file (`.sb` extension) support

### Proposed Changes

This depends on https://github.com/llk/scratch-sb1-converter/pull/12

Depend on scratch-sb1-converter. Use SB1File to check if the project being loaded is an `.sb` file. If so convert it to SB2 json and a (fake) zip.

### Reason for Changes

Supporting Scratch 1 files!

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "minilog": "3.1.0",
     "nets": "3.2.0",
     "scratch-parser": "4.3.3",
-    "scratch-sb1-converter": "github:mzgoddard/scratch-sb1-converter#build-to-sb2-",
+    "scratch-sb1-converter": "0.2.1",
     "scratch-translate-extension-languages": "0.0.20181205140428",
     "socket.io-client": "2.0.4",
     "text-encoding": "0.6.4",
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "minilog": "3.1.0",
     "nets": "3.2.0",
     "scratch-parser": "4.3.3",
+    "scratch-sb1-converter": "github:mzgoddard/scratch-sb1-converter#build-to-sb2-",
     "scratch-translate-extension-languages": "0.0.20181205140428",
     "socket.io-client": "2.0.4",
     "text-encoding": "0.6.4",
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -8,7 +8,7 @@ const ExtensionManager = require('./extension-support/extension-manager');
 const log = require('./util/log');
 const MathUtil = require('./util/math-util');
 const Runtime = require('./engine/runtime');
-const {SB1File} = require('scratch-sb1-converter');
+const {SB1File, ValidationError} = require('scratch-sb1-converter');
 const sb2 = require('./serialization/sb2');
 const sb3 = require('./serialization/sb3');
 const StringUtil = require('./util/string-util');
@@ -294,10 +294,13 @@ class VirtualMachine extends EventEmitter {
                 const sb1 = new SB1File(input);
                 const json = sb1.json;
                 json.projectVersion = 2;
-                console.log(json);
                 return resolve([json, sb1.zip]);
             } catch (e) {
-                console.error(e);
+                if (e instanceof ValidationError) {
+                    // The input does not validate as a Scratch 1 file.
+                } else {
+                    throw e;
+                }
             }
 
             // The second argument of false below indicates to the validator that the
@@ -442,7 +445,7 @@ class VirtualMachine extends EventEmitter {
 
         const runtime = this.runtime;
         const deserializePromise = function () {
-            let projectVersion = projectJSON.projectVersion;
+            const projectVersion = projectJSON.projectVersion;
             if (projectVersion === 2) {
                 return sb2.deserialize(projectJSON, runtime, false, zip);
             }
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -290,26 +290,33 @@ class VirtualMachine extends EventEmitter {
         }
 
         const validationPromise = new Promise((resolve, reject) => {
-            try {
-                const sb1 = new SB1File(input);
-                const json = sb1.json;
-                json.projectVersion = 2;
-                return resolve([json, sb1.zip]);
-            } catch (e) {
-                if (e instanceof ValidationError) {
-                    // The input does not validate as a Scratch 1 file.
-                } else {
-                    throw e;
-                }
-            }
-
             // The second argument of false below indicates to the validator that the
             // input should be parsed/validated as an entire project (and not a single sprite)
             validate(input, false, (error, res) => {
                 if (error) return reject(error);
                 resolve(res);
             });
-        });
+        })
+            .catch(error => {
+                try {
+                    const sb1 = new SB1File(input);
+                    const json = sb1.json;
+                    json.projectVersion = 2;
+                    return [json, sb1.zip];
+                } catch (sb1Error) {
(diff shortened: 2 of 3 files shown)


mzgoddard:
- Fix the incorrectly merged package-lock.json

package.json
```diff
@@ -43,6 +43,7 @@
     "minilog": "3.1.0",
     "nets": "3.2.0",
     "scratch-parser": "4.3.2",
+    "scratch-sb1-converter": "github:mzgoddard/scratch-sb1-converter#to-sb2",
```
mzgoddard: We'll change this to a version once it's published to npm and before merging this PR.

package.json
```diff
@@ -43,6 +43,7 @@
     "minilog": "3.1.0",
     "nets": "3.2.0",
     "scratch-parser": "4.3.2",
+    "scratch-sb1-converter": "github:mzgoddard/scratch-sb1-converter#build-to-sb2-",
```
kchadha: This should probably point somewhere else..?

src/virtual-machine.js
```diff
@@ -289,6 +290,16 @@ class VirtualMachine extends EventEmitter {
         }
 
         const validationPromise = new Promise((resolve, reject) => {
+            try {
+                const sb1 = new SB1File(input);
+                const json = sb1.json;
+                json.projectVersion = 2;
+                console.log(json);
```
kchadha: We probably shouldn't be logging the JSON anymore.

src/virtual-machine.js
```diff
@@ -289,6 +290,16 @@ class VirtualMachine extends EventEmitter {
...
+                return resolve([json, sb1.zip]);
```
kchadha: I think there should be a comment here that explains that `sb1.zip` doesn't actually include the `project.json` file, only the sounds and images (in case that is a point of confusion later down the line).

src/virtual-machine.js
```diff
@@ -289,6 +290,16 @@ class VirtualMachine extends EventEmitter {
...
+            } catch (e) {
+                console.error(e);
```
kchadha: Errors should probably get thrown like the rest of this code does.

At any rate, even if we're logging the error, you should use `log.error` here.

src/virtual-machine.js
```diff
@@ -431,7 +442,7 @@ class VirtualMachine extends EventEmitter {
 
         const runtime = this.runtime;
         const deserializePromise = function () {
-            const projectVersion = projectJSON.projectVersion;
+            let projectVersion = projectJSON.projectVersion;
```
kchadha: This change is unnecessary since projectVersion does not seem to be re-defined below.

kchadha (changes requested):
Some changes requested in the comments.

kchadha:
Sorry about the label confusion -- I don't remember adding the green `pr - good to merge` label...

Marking as needs work given the changes requested.

kchadha (approved):
LGTM!

src/virtual-machine.js
mzgoddard: I'll switch to `log.error`.

The errors that would happen here are validation errors that would indicate the passed data is not an SB1 File. If we threw those non-SB1 projects would not be able to load anymore.

We could change this to split the error handling behaviour on whether the error is a sb1-converter AssertionError (with a small change to sb1-converter to export its AssertionError). (Probably in a follow up PR?)

src/virtual-machine.js
mzgoddard: Nice catch. This is from any early change when I had the code using a different path to detect and transform to sb2.

src/virtual-machine.js
kchadha: Yeah, it sounds like without the change you're suggesting, we will be constantly logging errors that aren't actually errors (e.g. the normal use cases of loading sb2 and sb3 files)...

kchadha (changes requested):
I'm really confused about why github thinks I approved this PR...

mzgoddard:
- Make suggested changes

mzgoddard:
This PR now depends on https://github.com/LLK/scratch-sb1-converter/pull/20 and https://github.com/LLK/scratch-sb1-converter/pull/21.

mzgoddard:
- Rebase onto develop
- Bump sb1-converter dependency to 0.1.0

kchadha (approved):
LGTM

mzgoddard:
Depends on https://github.com/LLK/scratch-sb1-converter/pull/26.

mzgoddard:
- Change the validation attempt order. Attempt to validate as Scratch 2 or 3 and then failing that try to convert to Scratch 2.

This allows this change to be merged without https://github.com/LLK/scratch-sb1-converter/pull/27. However without that sb1-converter change Scratch 2/3 validation errors will instead likely appear as Scratch 1 AssertionErrors.

src/virtual-machine.js
```diff
@@ -295,6 +296,25 @@ class VirtualMachine extends EventEmitter {
                 if (error) return reject(error);
                 resolve(res);
             });
+        })
+        .catch(error => {
+            try {
+                const sb1 = new SB1File(input);
+                const json = sb1.json;
+                json.projectVersion = 2;
+                return [json, sb1.zip];
```
kchadha: I assume this line needs to be a resolve because this is all still happening as part of a Promise...

src/virtual-machine.js
```diff
@@ -295,7 +296,27 @@ class VirtualMachine extends EventEmitter {
...
                 resolve(res);
             });
-        });
+        })
+            .catch(error => {
+                try {
+                    const sb1 = new SB1File(input);
+                    const json = sb1.json;
+                    json.projectVersion = 2;
+                    return [json, sb1.zip];
```
kchadha: I assume this needs to be a Promise.resolve...?

```suggestion
                    return Promise.resolve([json, sb1.zip]);
```

src/virtual-machine.js
```diff
@@ -295,7 +296,27 @@ class VirtualMachine extends EventEmitter {
...
+                } catch (sb1Error) {
+                    if (sb1Error instanceof ValidationError) {
+                        // The input does not validate as a Scratch 1 file.
+                    } else {
+                        // The project appears to be a Scratch 1 file but it
+                        // could not be successfully translated into a Scratch 2
+                        // project.
+                        throw sb1Error;
```
kchadha:
```suggestion
                        return Promise.reject(sb1Error);
```

src/virtual-machine.js
```diff
@@ -295,7 +296,27 @@ class VirtualMachine extends EventEmitter {
...
+                    }
+                }
+                // Through original error since the input does not appear to be
+                // an SB1File.
+                throw error;
```
kchadha:
```suggestion
                return Promise.reject(error);
```

src/virtual-machine.js
```diff
@@ -295,7 +296,27 @@ class VirtualMachine extends EventEmitter {
...
+                    if (sb1Error instanceof ValidationError) {
+                        // The input does not validate as a Scratch 1 file.
+                    } else {
+                        // The project appears to be a Scratch 1 file but it
+                        // could not be successfully translated into a Scratch 2
+                        // project.
+                        throw sb1Error;
+                    }
+                }
+                // Through original error since the input does not appear to be
```
kchadha:
```suggestion
                // Throw original error since the input does not appear to be
```

kchadha (commented):
some suggestions in the code comments

PR #1856, opened 2018-12-17
Limit join result length to 10240

### LINKED ISSUE #922: 10240 limit for variables
### Expected Behavior
Variables never have over 10240 charactors in 2.0. If we can, it may be able to make the PC crashed

### Actual Behavior
It can have.

### Steps to Reproduce
program: see image
win7 homepremium 64bit SP1
firefox 58.0.1

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/922

### Proposed Changes

Limit the result of join the same way scratch-flash did.

### Reason for Changes

https://github.com/LLK/scratch-vm/issues/922

### Test Coverage

Added a unit test for it.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -91,7 +91,9 @@ class Scratch3OperatorsBlocks {
     }
 
     join (args) {
-        return Cast.toString(args.STRING1) + Cast.toString(args.STRING2);
+        // Limit result to 10240 characters: https://github.com/LLK/scratch-vm/issues/922
+        return (Cast.toString(args.STRING1) + Cast.toString(args.STRING2))
+            .slice(0, 10240);
     }
 
     letterOf (args) {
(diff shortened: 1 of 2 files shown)


kchadha (commented):
LG, but I'm wondering about performance...

How about:
```
const string1 = Cast.toString(args.STRING1);
const string1Length = string1.length;
if (string1Length >= 10240) return string1;
return string1 + Cast.toString(args.STRING2).slice(0, 10240 - string1Length);
```

It's more verbose but avoids unnecessary computation..

@mzgoddard, since you're moving back to performance related things, I'm wondering if you have thoughts on this?

picklesrus:
@kchadha had already reviewed - I think this just didn't get assigned back to @paulkaplan.

I took a very quick look at performance and doing the split is 65% slower than not, though I assumed the strings were always the following.
- 'banana'
- '4856838743587953448568387435879534485683874358795344856838743587953448568387435879534485683874358795344856838743587953448568387435879534485683874358795344856838743587953448568387435879534

I don't know if it would actually slow things down, but @kchadha's suggestions seems reasonable. Checking for length of those two string first did not seem to slow anything down.

apple502j:
I think it's too late to add this change - what do you think?

PR #1869, opened 2018-12-22
Always use English for kitten

### LINKED ISSUE #1858: Kitten 'meow' for Danish, Dutch and Icelandic doesn't sound right.
### Expected Behavior

With the Text-to-Speech extension, using the Kitten voice in Danish, Dutch, and Icelandic, the 'meow' doesn't sound right.

### Actual Behavior

In Danish, the meow sounds more like 'me-u' and feels truncated.
In Dutch, the meow sounds more like 'me-um'.
In Icelandic, the meow sounds more like 'me-uf'.

For an example of how 'meow' is pronounced in Danish, see: https://youtu.be/uxiXcAioZQc?t=5
The Google Translate-version of meows are also pretty good for reference: https://translate.google.com/#view=home&op=translate&sl=auto&tl=da&text=meow

### Steps to Reproduce

1. Load TTS-extension
2. Set voice to Kitten
3. Set language to Danish, Dutch or Icelandic.
4. Say any word

apple502j:
### Resolves
Resolves #1858 

### Proposed Changes
Always uses English for kitten (A pattern)

### Reason for Changes
`meow` sounds strange in some languages, and translating it is impossible with formatMessage.

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -457,6 +457,7 @@ class Scratch3Text2SpeechBlocks {
     speakAndWait (args, util) {
         // Cast input to string
         let words = Cast.toString(args.WORDS);
+        let locale = this.localeToPolly(this.getCurrentLanguage());
 
         const state = this._getState(util.target);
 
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_text2speech/index.js
```diff
@@ -466,11 +467,12 @@ class Scratch3Text2SpeechBlocks {
         // @todo localize this?
```
joker314: This is probably no longer a TODO?

joker314 (changes requested):

ericrosenbaum (approved):
😸😸 Looks good to me- thanks @apple502j !

PR #1871, opened 2018-12-26
Fix dynamic menus for string arrays

Affonso-Gui:
### Resolves

https://github.com/LLK/scratch-gui/issues/2891

### Proposed Changes

Fix the display of dynamic menus that return arrays of strings.

That is, with a function:
```
_initMenu() { return ['THIS', 'IS', 'A', 'TEST']; }
```
and a menu:
```
menus: { menu: '_initMenu' }
```
this PR makes this:
test_fail

look like this:
test_suceed


cwillisf (approved):
Thanks for fixing this!

PR #1877, opened 2018-12-28
Clamp ghost effect

apple502j:
### Resolves
Resolves #1875

### Proposed Changes
Clamps ghost effect using MathUtil

### Reason for Changes
It goes more than 100, or less than 0, which is not compatible

### Test Coverage
Use the script in #1875

PR #1878, opened 2018-12-28
Clamp ghost effect

### LINKED ISSUE #1875: Ghost effect not clamped to 0 or 100
### Expected Behavior

The actual value of the ghost effect should never be less than 0 or greater than 100 as is the case in 2.0.  Any attempt to set it to a number > 100 should set it to 100.  Similarly, any attempt to set it to a number < 0 should set it to 0.

### Actual Behavior

If you run "Change Ghost Effect by 10" repeatedly, the ghost effect can go above 100.  Similarly you can reduce the ghost effect below 0 by using a negative number.  You can also set the ghost effect directly to a number out of the range with the "Set ghost effect to [ ]" block

### Steps to Reproduce

Create a new project
On the cat build these stacks:
screen shot 2018-12-27 at 9 55 49 am
Press space
Note: the cat fades away
Press b
note: the cat returns
press space twice
Note: the cat fades away
Press b
Note: the cat does not return
press b again
Note: the cat returns.  The value for the ghost effect was 200 so you had to reduce it by 100 twice to get it to appear again.
Drag out and run the block "Set ghost effect to [300]"
Note: the cat disappears
Press space
Note: the cat does not return
Press space twice more
Note: the cat returns.  You successfully set the ghost effect to 300. 

Here's a useful project to test it with: https://scratch.mit.edu/projects/274813682/

Mac Chrome

towerofnix:
[Cherry-picked](https://stackoverflow.com/questions/1526044/partly-cherry-picking-a-commit-with-git) changes from 35c88904253f7f8ef494a32c2644157fbe4d45cb (#1877), /cc @apple502j.

Pasted from #1877:

---

### Resolves

Resolves #1875

### Proposed Changes

Clamps ghost effect using MathUtil

### Reason for Changes

It goes more than 100, or less than 0, which is not compatible

### Test Coverage

Use the script in #1875

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -4,6 +4,7 @@ const RenderedTarget = require('../sprites/rendered-target');
 const uid = require('../util/uid');
 const StageLayering = require('../engine/stage-layering');
 const getMonitorIdForBlockWithArgs = require('../util/get-monitor-id');
+const MathUtil = require('../util/math-util');
 
 /**
  * @typedef {object} BubbleState - the bubble state associated with a particular target.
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM! Thanks @towerofnix and @apple502j!

PR #1881, opened 2018-12-29
Change legacy pen color block's behavior

apple502j:
### Resolves
Resolves LLK/scratch-render#382

### Proposed Changes
The old `set pen color to ()` block makes the transparency to 0.

### Reason for Changes
Compatibility issues.

### Test Coverage
Project 272543304 should make two different circles

### DIFF
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -675,7 +675,6 @@ class Scratch3PenBlocks {
         const hueValue = Cast.toNumber(args.HUE);
         const colorValue = hueValue / 2;
         this._setOrChangeColorParam(ColorParam.COLOR, colorValue, penState, false);
-        penState = this._getPenState(util.target);
         this._setOrChangeColorParam(ColorParam.TRANSPARENCY, 0, penState, false);
         this._legacyUpdatePenColor(penState);
     }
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_pen/index.js
```diff
@@ -671,11 +671,12 @@ class Scratch3PenBlocks {
      * @param {object} util - utility object provided by the runtime.
      */
     setPenHueToNumber (args, util) {
-        const penState = this._getPenState(util.target);
+        let penState = this._getPenState(util.target);
```
cwillisf:
```suggestion
        const penState = this._getPenState(util.target);
```

src/extensions/scratch3_pen/index.js
```diff
@@ -671,11 +671,12 @@ class Scratch3PenBlocks {
...
         const hueValue = Cast.toNumber(args.HUE);
         const colorValue = hueValue / 2;
         this._setOrChangeColorParam(ColorParam.COLOR, colorValue, penState, false);
-
+        penState = this._getPenState(util.target);
```
cwillisf:
```suggestion

```

cwillisf (changes requested):
Good catch, @apple502j! Thanks for the contribution!

One small detail: there's no need to call `_getPenState` a second time. It should be safe to modify the state multiple times before "committing" the changes with `_legacyUpdatePenColor`.

cwillisf (approved):

Issue #1873, opened 2018-12-26
Unicode 'BACKSPACE' stop to read project


thisandagain:
### Expected Behavior

Should be open and run project in v3 as v2.

### Actual Behavior

If a project contains Unicode character 'BACKSPACE' (\u0008), it will not run on Scratch 3.0 beta. The scripts are disappeared after the character and not editable in v3.0 any more. 

### Steps to Reproduce

[Unicode Character 'BACKSPACE' (\u0008) ](https://scratch.mit.edu/projects/273360075/)

[Unicode Character 'BACKSPACE' (\u0008)  on Scratch 3.0 beta](https://beta.scratch.mit.eduLLK/scratch-gui#273360075)

### Operating System and Browser

MacOS 10.13.6
Chrome 70.0.3538.110

---
Moved from LLK/scratch-gui#4190
/cc @yokobond

kchadha:
This is directly related to #1077. The problem with both of these is that xml does not allow control characters.

One of the problems I was dealing with when I was trying to resolve this issue a while back was -- what is/are the place(s) where we should catch and fix this issue since these illegal characters can sneak into the block XML from a variety of different places (listing just a few below):

- imported project (e.g. from 2.0)
- variable name via the variable modal -- how likely is it that this character can be typed unintentionally...
- a different text field in the 3.0 editor whose contents can appear on a block (e.g. sprite name, costume/sound name, etc.)

yokobond:
The cause of this problem is that the error 'PCData invalid Char value' was returned by DOMParser when making svg text from string in the project at SvgRenderer.loadString(svgString, fromVersion2). 

Invalid char like that is not only \u0008 but also \u001c \u001d etc. (ref: https://www.w3.org/TR/xml/#charsets ). These invalid characters can be easily and unconsiously input by the default Japanese IME on MacOS, then so many project was destructed in Japanese and will be happen in the future continuously.

kchadha:
Removing the critical label from this as most issues created because of this character are now more recoverable (by saving and reloading the project) as a result of the changes in LLK/scratch-parser#53.

Issue #1882, opened 2018-12-30
Refuse to open projects with hacked arguments until a fix is implemented


towerofnix:
It's already [a known issue](https://github.com/LLK/scratch-vm/issues/1030) that projects with hacked arguments successfully open, but do so in a very bad way -- each field that contained a block in 2.0 instead contains the stringified array representing the block.

![In 2.0, the blocks "when green flag clicked: say ((join 'vol' 'ume') of 'stage')](https://user-images.githubusercontent.com/9948030/50550346-3c922c00-0c45-11e9-8832-4bdd42fa4850.png)

![In 3.0, the blocks "when green flag clicked: say ('concatenate:with:,vol,ume' of 'stage')](https://user-images.githubusercontent.com/9948030/50550351-4f0c6580-0c45-11e9-9cc7-915046546749.png)

While the long-term fix to this is to implement support for these hacked arguments, the code for that might not be implemented in the neartime future.

With the release of Scratch 3.0, a problem arises: if somebody opens a project they made which used hacked blocks, they will (sooner or later) see that the project doesn't work properly. So they'll likely make changes to the project blocks in order to debug the problem. In doing so, **the project will autosave, overwriting the existing 2.0 project, causing the user to lose the ability to recover their project's hacked blocks (by opening it in 2.0).**

I think there's a tool that the Scratch Team can use for recovering old versions of projects, but this isn't public, so there might be considerable manual work for Scratch Team moderators who get requests to restore old versions of projects.

As a temporary fix, I'm suggesting that these projects should fail to open in Scratch 3.0 until there is an implementation for hacked arguments.

(Not sure if this issue should go in scratch-parser?)

thisandagain:
Thanks for raising this @towerofnix. I'd like to discuss this with the team and we'll post a response here.

Alzter:
Please, please do this. Some great projects have been made with hacked blocks and they flat out don't work in 3.0. If they are able to be viewed and edited, they could be saved by the creator and that would permanently break the project even if a patch is later released.

My proposal for how we should go about fixing hacked blocks is to just make most of them work without hacking. Straight up removing hacked blocks limits the creative potential of projects and lowers the roof of how much can be achieved with the program.

Alzter:
[Also here's some ideas I had for making hacked blocks usable normally, with mockups included](https://github.com/LLK/scratch-vm/issues/1030#issuecomment-450810120)

apple502j:
[...]

Issue #1077, opened 2018-04-21
Variable name with special character causes all stacks to be hidden and program will not work


Mr-Dave2:
### Expected Behavior

Program stacks should be visible and the program should execute as it does on Scratch 2.0

_Please describe what should happen_

Variable names, even ones with Unicode characters, should have no effect on displaying program stacks or how the program executes.

### Actual Behavior

Stacks and the variable are not visible and the program will not work.

### Steps to Reproduce

I don't know where this character came from.  I have tried other Unicode characters and I have not found another one that causes any problems.  Credit to @shu279 for creating variable names that start with .  I was playing "Cat's Step Adventure -Part1-" https://scratch.mit.edu/projects/194603038 by shu279 and wanted to try it on Scratch 3.0 but it would not work and none of the stacks could be seen.  

I  did a cut-n-paste to create a variable with the special-character name in a new Scratch 2.0 project (https://scratch.mit.edu/projects/217845144/) and then created a small stack that used that variable.  When loaded into Scratch 3.0, the stack is not visible and the program will not work.

I also tried a cut-n-paste into vim and the character showed up as a control-H.  Vim does not support Unicode so this probably does not mean anything.

### Operating System and Browser

Windows 7 (64 bit), Chrome 65.0.3325.181

Mr-Dave2:
I found another program that will not run or display its stacks and it’s apparently also because of characters used in a name (of a list this time).  (Actually, it does something; it will play a sound for about a second but there is no motion).

The program is https://scratch.mit.edu/projects/215321405/ and for Scratch 3.0 (https://llk.github.io/scratch-gui/develop/#215321405).  I remixed the program and the only change I made was to delete the list named “>> Highscores <<”.  After saving the project I could load it into Scratch 3.0 (https://llk.github.io/scratch-gui/develop/#218249134) and it would run with stacks now visible.  (The sound still stops after one second but that is a different issue: “Deleting clones while playing notes causes music extension to stop working #1018”.

Mr-Dave2:
More people have reported this as the likely problem with their projects.

All projects that use the PTE projects to draw text (https://scratch.mit.edu/projects/172111618/).
This user mentions the problem in this post https://scratch.mit.edu/discuss/topic/299791/?page=254#post-3190127 and another mentions it here: https://scratch.mit.edu/discuss/topic/299791/?page=254#post-3190125

I have three more test cases:
[...]

PR #1883, opened 2018-12-31
Remove Unicode Backspace from XML to prevent broken SVG

yokobond:
### Resolves

Fix of #1873 Unicode 'BACKSPACE' stop to read project

### Proposed Changes

Remove the problematic character at the process to make XML source that build SVG graphics.

### Reason for Changes

The character '\0008' breaks SVG so that it should not be used but it will input in any text field accidentally by users. V2.0 accept the char and no probrem to read and convert to V3.0 which project including it. Removing the char at the last process to make SVG will fix the both problems.  The escape XML will good place to do this. 

### Test Coverage

New unit-test was added for xmlEscape().

### DIFF
--- a/src/util/xml-escape.js
+++ b/src/util/xml-escape.js
@@ -7,13 +7,15 @@
  * @return {string} XML-escaped string, for use within an XML tag.
  */
 const xmlEscape = function (unsafe) {
-    return unsafe.replace(/[<>&'"]/g, c => {
+    // eslint-disable-next-line no-control-regex
+    return unsafe.replace(/[<>&'"\u0008]/g, c => {
         switch (c) {
         case '<': return '&lt;';
         case '>': return '&gt;';
         case '&': return '&amp;';
         case '\'': return '&apos;';
         case '"': return '&quot;';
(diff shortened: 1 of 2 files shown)


kchadha (changes requested):
@yokobond, unfortunately this fix does not work. Pulling in these changes and testing with the project in the original issue (273360075) shows errors in the console.

You are right about the fix being related to what can and cannot be displayed in XML/SVG, but unfortunately fixing the problem fully will need a lot more changes. I am currently working on a fix for a related issue (#1077), and so hopefully this issue will get resolved with that fix as well.

towerofnix:
Would this change prevent projects that intentionally detect the backspace character from functioning correctly? We want to be able to load this projects - but not at the expense of them running as they're intended to, right? (Same type of reasoning as in #1882.) Or does this just change how the block displays, but not how it saves in the sb3 (i.e. is the backspace character still present in the sb3 json)?

kchadha:
@towerofnix, with this change, the backspace control character is still saved in the sb3 json.

PR #1897, opened 2019-01-07
Fix cloud variable limit

kchadha:
### Resolves

Resolves LLK/scratch-gui#4250

### Proposed Changes

Update the cloud variable limit from 8 to 10.

### Reason for Changes

Incompatibility with Scratch 2.0.

### Test Coverage

Updated existing tests that check for the limit.


paulkaplan (approved):
LGTM

PR #1898, opened 2019-01-07
Re enable tests that were unavailable during maintenance

paulkaplan:
/cc @kchadha @rschamp

### DIFF
--- a/test/unit/util_task-queue.js
+++ b/test/unit/util_task-queue.js
@@ -1,4 +1,4 @@
-const test = require('tap').test;
+const test = require('tap').skip;
 
 const TaskQueue = require('../../src/util/task-queue');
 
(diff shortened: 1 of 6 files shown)


kchadha (changes requested):
This also enables a test that was being skipped from before maintenance -- the util_task_queue test. That one was being skipped because it's a flaky test.

test/unit/util_task-queue.js
```diff
@@ -1,4 +1,4 @@
-const test = require('tap').skip;
+const test = require('tap').test;
```
kchadha:
```suggestion
const test = require('tap').skip;
```

paulkaplan: Gotchya thanks

kchadha (approved):
LGTM

PR #1899, opened 2019-01-07
Do not deep clone before serializing.

paulkaplan:
This was ok when we did not attach assets, but it is not cool now. With large assets, this made saving extremely slow.

After consulting @kchadha and looking through the history, I believe that the deep cloning was there as a relic of past save/load work, rather than a defensive measure, so I feel confident about making this change. I looked through all the block/asset serializing code and didn't see anywhere that modifications were being made to the data structure.

Fixes https://github.com/LLK/scratch-gui/issues/4256

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -14,7 +14,6 @@ const StageLayering = require('../engine/stage-layering');
 const log = require('../util/log');
 const uid = require('../util/uid');
 const MathUtil = require('../util/math-util');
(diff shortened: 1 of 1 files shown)


kchadha (approved):
LGTM

thisandagain:
This looks reasonable to me, though there is a little bit of risk here. @BryceLTaylor Let's make sure this gets lots of test time. Thanks @paulkaplan !

kchadha:
@thisandagain, agreed!

PR #1900, opened 2019-01-08
Clamp the currentCostume number when loading a project

kchadha:
### Resolves

Resolves an issue reported by sentry `cannot find property bitmapResolution of undefined`. This was causing a BSOD in at least the following two projects:

This sb2 project:
65811674

This sb3 project: 
277128060

### Proposed Changes

When loading a project, clamp the currentCostume number to an appropriate value (an index available in the actual list of the target's costumes).

### Reason for Changes

Fixes the projects so that they can load.

Filed a separate issue (#1901) for trying to figure out how the project got saved in this state in the first place...

### Test Coverage

Tested with existing projects reporting this error on sentry.

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -12,6 +12,7 @@ const Color = require('../util/color');
 const log = require('../util/log');
 const uid = require('../util/uid');
 const StringUtil = require('../util/string-util');
+const MathUtil = require('../util/math-util');
 const specMap = require('./sb2_specmap');
 const Comment = require('../engine/comment');
 const Variable = require('../engine/variable');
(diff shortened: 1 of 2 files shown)


src/serialization/sb2.js
```diff
@@ -812,7 +816,9 @@ const specMapBlock = function (block) {
     const opcode = block[0];
     const mapped = opcode && specMap[opcode];
     if (!mapped) {
-        log.warn(`Couldn't find SB2 block: ${opcode}`);
+        log.warn(`Couldn't find SB2 block: ${opcode}, trying as custom procedure call`);
+        // block = ['call', block];
```
thisandagain: Looks like this may be some cruft that was left in.

thisandagain (changes requested):
Looks good! One minor nit.

src/serialization/sb2.js
kchadha: Done.

kchadha:
@thisandagain, I removed the cruft and also added some new code: a log warning and a patch for *serialization* so that we don't continue to contribute to saving broken projects. There is a new TODO comment that references #1901.

thisandagain (approved):

PR #1907, opened 2019-01-10
Add blocks to GDX-FOR extension

spectranaut:
### Proposed Changes

This PR adds three more blocks to the extension and filters acceleration from gravity on the "When acceleration" hat.

### How to test

You can play with the extension here: https://vernierst.github.io/scratch-gui/goforce/

Make sure you have the most recent Scratch-Link (1.1.28) and a vernier GDX-FOR device. To connect to the device, add the extension (it will have a "microbit" image) and click the power button the GDX-FOR device.

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -367,6 +367,20 @@ class Scratch3GdxForBlocks {
         ];
     }
 
+    get FACE_MENU () {
+        return [
+            {
+                text: 'up',
+                value: 'up'
+            },
+            {
+                text: 'down',
+                value: 'down'
+            }
+        ];
+    }
+
+
     get COMPARE_MENU () {
         return [
             {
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_gdx_for/index.js
```diff
@@ -562,6 +616,14 @@ class Scratch3GdxForBlocks {
             return false;
         }
     }
+    whenJumped () {
```
ericrosenbaum, line 619: I think this could just be `return this.isFreeFalling()`?

src/extensions/scratch3_gdx_for/index.js
```diff
@@ -562,6 +616,9 @@ class Scratch3GdxForBlocks {
...
+        return this.isFreeFalling;
```
ericrosenbaum: Probably needs the `()`?

ericrosenbaum (approved):
LGTM! I have a bunch of details, cleanup and questions but I'll file those as follow up issues

PR #1912, opened 2019-01-11
Populate executable targets when runtime.targets is populated 

kchadha:
### Resolves

Resolves LLK/scratch-gui#4257. Also possibly LLK/scratch-gui#4302...

### Proposed Changes

Don't populate executable targets before targets are finished installing. This ensures that scripts don't run before a project is done loading.

### Reason for Changes

Without these changes it was possible to start running blocks  in a project before targets were fully installed. This was causing some uncaught exceptions like the one described in LLK/scratch-gui#4257, but also behavioral changes in the projects like turning global variables into local variables.

### Test Coverage

Added to a unit test which ensures that the layer order produced by deserialization is correct. This layer order is then used when targets are getting installed to populate executableTargets in the correct order. Previously executableTargets were getting populated when the target was created (which was pre-sorted by this same property).

Manually tested with some example projects that use the 'when timer > 0; broadcast message' construct as well as with some test cases that were previously turning global variables into local ones.

Thanks @picklesrus for working through this with me!

### DIFF
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -165,8 +165,7 @@ class Scratch3ControlBlocks {
         // Create clone
         const newClone = cloneTarget.makeClone();
         if (newClone) {
-            this.runtime.targets.push(newClone);
-            this.runtime.addExecutable(newClone);
+            this.runtime.addTarget(newClone);
 
             // Place behind the original target.
             newClone.goBehindOther(cloneTarget);
--- a/src/blocks/scratch3_control.js
+++ b/src/blocks/scratch3_control.js
@@ -166,6 +166,10 @@ class Scratch3ControlBlocks {
         const newClone = cloneTarget.makeClone();
         if (newClone) {
             this.runtime.targets.push(newClone);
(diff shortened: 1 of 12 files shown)


paulkaplan (approved):
The code looks good, but I'm a bit concerned about maintainability. If in the future we are changing this code and forget that we need to add the executable target when adding a target to the target list... bad things happen and there are no run time checks for the consistency of executable targets vs. targets (i.e the original bug this fixes). Is there a performant way to make the executable order completely internal to the runtime object? Exposing it feels error prone. What do you think?

paulkaplan:
Marked as LG since the code is fine, but i'd love to get your thoughts on the review comment above^

kchadha:
@paulkaplan, just submitted a change to address your suggestion. There is a new `addTarget` function on the runtime which adds to the executableTargets and targets lists at the same time. In installTargets, we take an extra step of sorting executableTargets after it has been populated, and the comment on the addTarget function indicates that needs to happen.

paulkaplan (approved):
LG, thanks for thinking through that

Issue #1908, opened 2019-01-10
Running Vernier sensor block during connection process causes an error


ericrosenbaum:
### Expected Behavior

Using a Vernier extension block should not cause an error in the console.

### Actual Behavior

During the connection process, after you have clicked "connect" on your device, but before it has connected, if any block that accesses a sensor value runs, you'll get an error in the console:

```
Uncaught TypeError: Cannot read property 'getSensor' of null
    at GdxFor.getAccelerationX (index.js:270)
```

### Steps to Reproduce

- Add the GDX-FOR extension, and dismiss the connection modal
- Check the monitor checkbox for force (or add hat block to the workspace, or run a forever loop containing a sensor reporter)
- Connect to the device, and the error appears during the connection process

cc @spectranaut

spectranaut:
Oh wow I wanted to test this but thought I'd never been able to click through the modal fast enough. Thanks for the steps, @ericrosenbaum :)

ericrosenbaum:
Fixed by https://github.com/LLK/scratch-vm/pull/1914

PR #1914, opened 2019-01-11
Prevent console errors when disconnecting/reconnect to Vernier extension

spectranaut:
### Resolves

https://github.com/LLK/scratch-vm/issues/1908

### Proposed Changes

This PR does two things: (1) it keeps track of when the measurements have actually been enabled (before, it only checked when Scratch Link had first connect to the device) and (2) it prevents sending 'close' to a device which has already been closed (by clearing the internal `this._device` after closing.

### Reason for Changes

The extension works find in the gui without these changes, however, it produces errors in the console.

### Test Coverage
- Add the GDX-FOR extension, and dismiss the connection modal
- Check the monitor checkbox for force (or add hat block to the workspace, or run a forever loop containing a sensor reporter)
- Connect to the device, and the error appears during the connection process


ericrosenbaum (approved):
LGTM!

PR #1916, opened 2019-01-14
Prevent excess event emitting in Vernier extension

ericrosenbaum:
Previously we were using a workaround to clear the stored sensor values inside the Vernier module, but this was causing lots of events to be emitted, and possibly causing performance issues.

Instead, set an internal flag to prevent the values from being stored in the first place.


spectranaut (approved):
looks good!

PR #1920, opened 2019-01-15
Properly disconnect BLE socket in GDX-FOR scan/disconnect

### LINKED ISSUE #1919: GDX-FOR extension doesn't disconnect properly on scan
1. Scan for GDX-For: see device appear in scan list named 'GDX-FOR ###'
2. Dismiss connection dialog
3. Scan for micro:bit
4. **Expected**: see device appear in scan list named 'micro-bit device name'
4. **Actual**: see device appear in scan list named 'GDX-FOR ###'

### LINKED ISSUE #1921: GDX-FOR extension doesn't properly disconnect when user disconnects
Report from Ian Honohan:
"The disconnection of the GDX-FOR can fail. This is intermittent and I don't have a reproducible workflow for this, yet. It appears that when an initial connection fails or when the user chooses to disconnect from the GDX-FOR, the GDX-FOR continues to flash its green LED, indicating it is still connected to the system. At this point, you cannot reconnect until you cycle the power on the GDX-FOR device, or if you reload the page."

evhan55:
### Resolves

- Resolves #1919: GDX-FOR extension doesn't disconnect properly on scan
- Resolves #1921: GDX-FOR extension doesn't properly disconnect when user disconnects

### Proposed Changes

- Changes from `_device.close` to `scratchLinkSocket.disconnect` since _device is null in the discovery/scan case

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -75,8 +75,8 @@ class GdxFor {
      * Called by the runtime when user wants to scan for a peripheral.
      */
     scan () {
-        if (this._device) {
-            this._device.close();
+        if (this._scratchLinkSocket) {
+            this._scratchLinkSocket.disconnect();
         }
 
         this._scratchLinkSocket = new BLE(this._runtime, this._extensionId, {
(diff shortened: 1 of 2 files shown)


ericrosenbaum (approved):
LGTM!

PR #1924, opened 2019-01-15
Removing withResponse flag in GDX-FOR extension write method

evhan55:
### Resolves

- Resolves https://github.com/LLK/scratch-link/issues/111

### Testing

- Should be tested with latest Scratch Link from `develop`, on both Mac and Windows


cwillisf (approved):
This makes sense to me, as long as it works reliably on both macOS and Windows.

PR #1925, opened 2019-01-15
Do not load Vernier extension yet

ericrosenbaum:
Due to compatibility issues with the Vernier Go Direct module which need investigation, do not load the Vernier extension yet.


paulkaplan (approved):
LG thanks!

PR #1928, opened 2019-01-16
Improve the performance of the timer utilitiy

kchadha:
### Proposed Changes

Timer utility was creating a new object every time it was trying to calculate the amount of time that hat elapsed when running as part of a wait block (multiple times per frame). This was causing a minor GC.

Save the nowObj in an instance property in the block-utility so that it does not return a new object every time the getter is called.

### Reason for Changes

Performance improvement for the wait block.

### Test Coverage

Benchmark suite comparison on Mac Chrome (numbers on the left are after improvements, numbers on the right are develop):

image

### DIFF
--- a/src/engine/block-utility.js
+++ b/src/engine/block-utility.js
@@ -23,7 +23,7 @@ class BlockUtility {
         this.thread = thread;
 
         this._nowObj = {
-            now: () => this.runtime.currentMSecs
+            now: () => this.sequencer.runtime.currentMSecs
         };
     }
 
--- a/src/engine/block-utility.js
+++ b/src/engine/block-utility.js
@@ -21,6 +21,10 @@ class BlockUtility {
          * @type {?Thread}
          */
         this.thread = thread;
+
+        this._nowObj = {
+            now: () => this.runtime.currentMSecs
+        };
     }
 
     /**
@@ -46,9 +50,7 @@ class BlockUtility {
      */
     get nowObj () {
         if (this.runtime) {
-            return {
-                now: () => this.runtime.currentMSecs
-            };
+            return this._nowObj;
         }
         return null;
     }


src/engine/block-utility.js
```diff
@@ -21,6 +21,10 @@ class BlockUtility {
          * @type {?Thread}
          */
         this.thread = thread;
+
+        this._nowObj = {
+            now: () => this.runtime.currentMSecs
```
mzgoddard: A small performance optional detail: do `this.sequencer.runtime.currentMSecs` instead. Directly dereferencing and avoiding the property call will be faster.

mzgoddard (approved):
I was noticing some performance differences in two branches I was comparing and noticed the (pretty significant) increase in garbage collection. I almost remade your work here before I remembered you mentioned this PR to. I'll be including it in my perf branch.

src/engine/block-utility.js
kchadha: @mzgoddard, I'm not sure I understand the change suggested..

Shouldn't `this.runtime` and `this.sequencer.runtime` be the same object?

kchadha: Ah, nvm, just saw the `this.runtime` getter.

kchadha: @mzgoddard, since `this.sequencer` could be null in the constructor, how about the following:

```suggestion
            now: () => this.sequencer && this.sequencer.runtime.currentMSecs
```

mzgoddard: We don't need worry about being safe with `this.sequencer`. If you call *any* BlockUtility method without setting `sequencer` and `thread` I think you should expect an error. (I want to get rid of the `this.runtime` check here in `nowObj`. I'll open a separate PR for that. Don't need to add more stuff here.)

Anywhere BlockUtility may be used that doesn't set `sequencer` and `thread` during normal vm execution would fair us better to throw an error instead of silently failing.

As well, none of the existing methods check for sequencer or thread.

I guess testing could benefit. But I kind of feel like writing tests that involve blocks without setting up more of the vm they would need to operate and confirm their results would be a disservice to us. Then thrown errors there would also help us.

paulkaplan (approved):
Change LGTM. I agree with @mzgoddard about deciding on whether runtime/sequencer is required for the block utility. It would also allow us to get away from magic getters. 

Aside: I'd love to talk with the group about the use of JS getters/setters. I don't totally understand when they should be used.

kchadha:
@paulkaplan, agreed!

src/engine/block-utility.js
kchadha:
```suggestion
            now: () => this.sequencer.runtime.currentMSecs
```

kchadha:
@mzgoddard, @paulkaplan, made the change for referencing `this.sequencer.runtime` directly and created a new tech debt issue for removing the `this.runtime` check in the `nowObj` getter (https://github.com/LLK/scratch-vm/issues/1936)

PR #1930, opened 2019-01-18
Cache hat block information for the runtime

### LINKED ISSUE #1707: Cache script blocks for allScriptsDo
### Expected Behavior

Quickly loop over script blocks filtered by uses of allScriptsDo, like filtering by opcode.

### Actual Behavior

Every frame, Runtime starts threads for all edge hats. This currently loops over all script blocks, the top block of a command block list, for all targets. Doing this the inner function passed to allScriptsDo filters by opcode so only edge hats are started. The opcode filter can be cached in Blocks like other caches which are reset on any block change. This way a pre-filtered set of script blocks by opcode and possibly other filters can be saved and reused.

### Possible Optimizations

- Cache a filtered set of script blocks by opcode
- Cache the hatFields for each script block
- Cache a filtered set of script blocks by opcode and hatFields
- Filter thread existence check by target first. Prefer the object identity equality comparison over a string comparison.

mzgoddard:
### Resolves

#1707 

### Proposed Changes

- Add BlocksRuntimeCache to cache data about block scripts, their hats, and their data (like broadcast option)
- <del>Group threads on their target in addition to runtime.targets and loop those target to determine if a thread should be started or restarted for a hat</del>
- Remove Object.assign use in _pushThreads

### Reason for Changes

Edge activated hats are started every frame and broadcasts are a feature that may be used a lot in a project. Starting hats currently loops all scripts in all targets in most cases. With that scripts are checked that are not the right hat/opcode. This information can easily be grouped and cached so every second iteration for a hat/opcode only needs to loop over the desired hats and test if the fields match any optional fields.

### Benchmark Results

In general all platforms normally experience an improvement with this change on projects.

Here are some exceptions:
- Safari (2017 MBP 15") has a general issue with this change for 155128646 (stacky towers). It experiences a small decrease in performance.
- Chrome (2017 MBP 15") starting 89811578 (solar system) seems to be the only one negatively effected in a large way.
- Firefox (2017 MBP 15") experiences a minor degradation in 130041250 (floating blocks).

Raspberry PI data could not be collected for 89811578 (solar system). The PI stalled on this project likely due to needing too much memory or too much memory thrashing.

<details>
<summary>Benchmark Results Table</summary>

Platform | Project ID | Warm Up | Recording | develop | after change | difference
-------- | ---------- | ------- | --------- | ------: | -----------: | ---------:
raspberrypi | 130041250 | 0 | 2000 | 3187 | 3321 | +4.2
book | 130041250 | 0 | 2000 | 169252 | 169119 | -0.1
chrome | 130041250 | 0 | 2000 | 1093599 | 1140890 | +4.3
firefox | 130041250 | 0 | 2000 | 773011 | 754854 | -2.3
safari | 130041250 | 0 | 2000 | 1797192 | 2052938 | +14.2
raspberrypi | 130041250 | 4000 | 6000 | 41998 | 39521 | -5.9
book | 130041250 | 4000 | 6000 | 278924 | 285923 | +2.5
chrome | 130041250 | 4000 | 6000 | 1338833 | 1352135 | +1.0
firefox | 130041250 | 4000 | 6000 | 961047 | 933778 | -2.8
safari | 130041250 | 4000 | 6000 | 2067681 | 2100278 | +1.6
raspberrypi | 14844969 | 0 | 2000 | 8528 | 14097 | +65.3
book | 14844969 | 0 | 2000 | 255686 | 329475 | +28.9
chrome | 14844969 | 0 | 2000 | 1093788 | 1142420 | +4.4
firefox | 14844969 | 0 | 2000 | 830292 | 898024 | +8.2
safari | 14844969 | 0 | 2000 | 2425291 | 2864760 | +18.1
raspberrypi | 14844969 | 1000 | 6000 | 61521 | 72142 | +17.3
book | 14844969 | 1000 | 6000 | 409130 | 455599 | +11.4
chrome | 14844969 | 1000 | 6000 | 1147805 | 1219053 | +6.2
firefox | 14844969 | 1000 | 6000 | 937069 | 1077716 | +15.0
safari | 14844969 | 1000 | 6000 | 2789899 | 3138463 | +12.5
raspberrypi | 173918262 | 0 | 5000 | 25746 | 32022 | +24.4
book | 173918262 | 0 | 5000 | 313615 | 333111 | +6.2
chrome | 173918262 | 0 | 5000 | 1071776 | 1073457 | +0.2
firefox | 173918262 | 0 | 5000 | 717066 | 756411 | +5.5
safari | 173918262 | 0 | 5000 | 2117538 | 2363462 | +11.6
raspberrypi | 173918262 | 5000 | 5000 | 57276 | 61210 | +6.9
book | 173918262 | 5000 | 5000 | 409647 | 419373 | +2.4
chrome | 173918262 | 5000 | 5000 | 1107828 | 1131176 | +2.1
firefox | 173918262 | 5000 | 5000 | 699914 | 818517 | +16.9
safari | 173918262 | 5000 | 5000 | 1609219 | 1870705 | +16.2
raspberrypi | 155128646 | 0 | 5000 | 18827 | 21581 | +14.6
book | 155128646 | 0 | 5000 | 195730 | 205951 | +5.2
chrome | 155128646 | 0 | 5000 | 1059996 | 1082332 | +2.1
firefox | 155128646 | 0 | 5000 | 563900 | 584788 | +3.7
safari | 155128646 | 0 | 5000 | 2414075 | 2287622 | -5.2
raspberrypi | 155128646 | 5000 | 5000 | 30147 | 34159 | +13.3
book | 155128646 | 5000 | 5000 | 260194 | 265460 | +2.0
chrome | 155128646 | 5000 | 5000 | 841821 | 909526 | +8.0
firefox | 155128646 | 5000 | 5000 | 455500 | 479676 | +5.3
safari | 155128646 | 5000 | 5000 | 600826 | 567534 | -5.5
chrome | 89811578 | 0 | 5000 | 1082200 | 817461 | -24.5
firefox | 89811578 | 0 | 5000 | 911832 | 1009328 | +10.7
book | 89811578 | 0 | 5000 | 498503 | 501233 | +0.5
safari | 89811578 | 0 | 5000 | 2857443 | 2871101 | +0.5
chrome | 89811578 | 5000 | 5000 | 1091999 | 1116591 | +2.3
firefox | 89811578 | 5000 | 5000 | 1030964 | 1112638 | +7.9
book | 89811578 | 5000 | 5000 | 553632 | 562060 | +1.5
safari | 89811578 | 5000 | 5000 | 3251637 | 3173566 | -2.4

</details>

### DIFF
--- a/src/engine/blocks-runtime-cache.js
+++ b/src/engine/blocks-runtime-cache.js
@@ -1,0 +1,78 @@
+/**
+ * @fileoverview
+ * The BlocksRuntimeCache caches data about the top block of scripts so that
+ * Runtime can iterate a targeted opcode and iterate the returned set faster.
+ * Many top blocks need to match fields as well as opcode, since that matching
+ * compares strings in uppercase we can go ahead and uppercase the cached value
+ * so we don't need to in the future.
+ */
+
+/**
+ * A set of cached data about the top block of a script.
+ * @param {Blocks} container - Container holding the block and related data
+ * @param {string} blockId - Id for whose block data is cached in this instance
+ */
+class RuntimeScriptCache {
+    constructor (container, blockId) {
+        /**
+         * Container with block data for blockId.
+         * @type {Blocks}
+         */
+        this.container = container;
+
+        /**
+         * ID for block this instance caches.
+         * @type {string}
+         */
+        this.blockId = blockId;
+
+        const block = container.getBlock(blockId);
+        const fields = container.getFields(block);
+
+        /**
+         * Formatted fields or fields of input blocks ready for comparison in
+         * runtime.
+         *
+         * This is a clone of parts of the targeted blocks. Changes to these
+         * clones are limited to copies under RuntimeScriptCache and will not
+         * appear in the original blocks in their container. This copy is
+         * modified changing the case of strings to uppercase. These uppercase
+         * values will be compared later by the VM.
+         * @type {object}
+         */
+        this.fieldsOfInputs = Object.assign({}, fields);
+        if (Object.keys(fields).length === 0) {
+            const inputs = container.getInputs(block);
+            for (const input in inputs) {
+                if (!inputs.hasOwnProperty(input)) continue;
+                const id = inputs[input].block;
+                const inputBlock = container.getBlock(id);
+                const inputFields = container.getFields(inputBlock);
+                Object.assign(this.fieldsOfInputs, inputFields);
+            }
+        }
+        for (const key in this.fieldsOfInputs) {
+            const field = this.fieldsOfInputs[key] = Object.assign({}, this.fieldsOfInputs[key]);
+            if (field.value.toUpperCase) {
+                field.value = field.value.toUpperCase();
+            }
+        }
+    }
+}
+
+/**
+ * Get an array of scripts from a block container prefiltered to match opcode.
+ * @param {Blocks} container - Container of blocks
+ * @param {string} opcode - Opcode to filter top blocks by
+ */
+exports.getScripts = function () {
+    throw new Error('blocks.js has not initialized BlocksRuntimeCache');
+};
+
+/**
+ * Exposed RuntimeScriptCache class used by integration in blocks.js.
+ * @private
+ */
+exports._RuntimeScriptCache = RuntimeScriptCache;
+
+require('./blocks');
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -5,6 +5,7 @@ const MonitorRecord = require('./monitor-record');
 const Clone = require('../util/clone');
 const {Map} = require('immutable');
 const BlocksExecuteCache = require('./blocks-execute-cache');
+const BlocksRuntimeCache = require('./blocks-runtime-cache');
 const log = require('../util/log');
 const Variable = require('./variable');
 const getMonitorIdForBlockWithArgs = require('../util/get-monitor-id');
@@ -71,7 +72,13 @@ class Blocks {
              * actively monitored.
              * @type {Array<{blockId: string, target: Target}>}
              */
-            _monitored: null
+            _monitored: null,
+
+            /**
+             * A cache of hat opcodes to collection of theads to execute.
+             * @type {object.<string, object>}
+             */
+            scripts: {}
         };
 
         /**
@@ -504,6 +511,7 @@ class Blocks {
         this._cache.procedureDefinitions = {};
         this._cache._executeCached = {};
         this._cache._monitored = null;
+        this._cache.scripts = {};
     }
 
     /**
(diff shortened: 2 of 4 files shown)


mzgoddard:
- [x] I'll be reworking the commits a bit to pull out a few things and giving them more descriptive comments than wip.

mzgoddard:
I removed the thread grouping by target. There a lot of ways threads are killed to account for. Hard to make sure we get all of them so I dedicate that work later. This may mean the posted data above is less accurate now.

kchadha:
@mzgoddard, it looks like integration tests are failing...

mzgoddard:
@kchadha Yeah I've looked into them a few times but haven't dug deep enough to find whats off. Focusing on the loading stuff at the moment.

mzgoddard:
- Fix the boolean conversions in `_pushThread`

kchadha:
@mzgoddard these changes don't seem to work with extension hat blocks when the extension is loaded during construction of the project (instead of loading a project that already uses an extension). E.g. If I add the microbit extension to the default project and add the stack `when A button pressed; turn 15 degrees`, pressing the A button on the connected micro:bit does not work. I can confirm that I'm connected to the micro:bit because I can run command blocks (e.g. display [heart]). Saving and loading this project however fixes the issue.

kchadha (changes requested):
@mzgoddard these changes don't seem to work with extension hat blocks when the extension is loaded during construction of the project (instead of loading a project that already uses an extension). E.g. If I add the micro:bit extension to the default project and add the stack `when A button pressed; turn 15 degrees`, pressing the A button on the connected micro:bit does not work. I can confirm that I'm connected to the micro:bit because I can run command blocks (e.g. display [heart]). Saving and loading this project however fixes the issue.

Edit: it seems like only the blocks that were already in the loaded project run. If you add more micro:bit hat block stacks to the project that already has the extension loaded, those blocks don't run.

mzgoddard:
Ah. The cache busting is missing for scripts. On it.

mzgoddard:
- Fix the hat block cache issue by resetting it with the other block caches.
- Add a test that added hats are executed too.

src/engine/blocks-runtime-cache.js
```diff
@@ -0,0 +1,76 @@
...
+        const block = container.getBlock(blockId);
+        const fields = container.getFields(block);
+
+        /**
+         * Formatted fields or fields of input blocks ready for comparison in
+         * runtime.
+         *
+         * This copy of the fields or the fields of all inputs is created and
+         * modified changing the case of strings to uppercase. Part of this is
+         * to make sure we do not overwrite the fields stored in the raw block.
```
mzgoddard: Start this comment with the fact that this should not change the original block.

src/engine/blocks-runtime-cache.js
```diff
@@ -0,0 +1,76 @@
...
+
+        const block = container.getBlock(blockId);
+        const fields = container.getFields(block);
+
+        /**
+         * Formatted fields or fields of input blocks ready for comparison in
+         * runtime.
+         *
+         * This copy of the fields or the fields of all inputs is created and
+         * modified changing the case of strings to uppercase. Part of this is
```
kchadha: This comment is a little confusing, it makes it seem like the uppercase is related to copying the fields. =]

kchadha (changes requested):
Thanks for fixing the issue with adding new hat blocks. It seems to be fixed. Could you add the comment we talked about on Thursday? After that LGTM.

mzgoddard:
@kchadha I fixed the comment.

cwillisf (commented):
Hmm... the general idea seems sound enough, though I don't understand why it would cause a performance hit in some cases. In the worst case, for example if a project contains at most one hat of any opcode*, shouldn't this code run at the same speed as the old code? Any guesses where the performance hit comes from in those cases where this change hurts, and what the worst-case performance is?

Separately: I'm not sure I understand what's going on with the mutual `require` between `blocks-runtime-cache.js` and `blocks.js`. If I'm reading this correctly it's set up to enforce tight coupling between the two JS files, but I'm not sure why that's necessary. If `getScripts` were moved into `blocks-runtime-cache.js` and the `require('./blocks.js')` removed, wouldn't everything work the same? The `getScripts` method doesn't look like it tries to access anything "private" within `blocks.js`. If there's an interdependency that makes the separation necessary then I think a comment would help, but if not I'd prefer a more straightforward implementation for the sake of maintenance.

mzgoddard:
@cwillisf 

> Hmm... the general idea seems sound enough, though I don't understand why it would cause a performance hit in some cases.

My current guess is jit heuristic. We are decreasing the number of calls to the  inner startHats function.  That inner function has two potentially expensive for loops over all threads. The projects with consistent performance hit for firefox and safari are to my knowledge the ones with less hat usage. So if this is the case the looping over threads is slower not be optimized to the next jit level.

The removal of the `if(optMatchFields)` may be a factor where the if is helping to skip a expensive case where optMatchFields is falsey and the if operates on that faster than the for-in.

I'm not sure what to make of the Chrome exception in solar system. That kind of difference makes me think its a major garbage collection or function deoptimization going on.

> Separately: I'm not sure I understand what's going on with the mutual require between blocks-runtime-cache.js and blocks.js.

It is accessing something private. It is accessing `_cache`. I think we should have some interface between Blocks and its targeted caches where only blocks.js accesses `blocks._cache`. Thats the current bidirectional interface between runtime cache, execute cache, and blocks.

cwillisf (approved):
👍

kchadha (approved):

PR #1933, opened 2019-01-19
Feature/field type support

### LINKED ISSUE #1647: Custom fields in extensions
Got a report that Blockly.Field.register isn't working for field types that aren't yet defined. Seems to work okay to override existing types, but not new ones.

Needs more investigation.

---
Moved from LLK/scratch-blocks#1725
/cc @picklesrus

ErikMejerHansen:
### Resolves
#1647

### Proposed Changes
Add support extensions to define custom field types.

This is done by adding a new element "customFieldTypes" to the extension info structure. Ex:
```
customFieldTypes: {
   angleField: {
      implementation: { fromJson: options => new AngleField(options) },
      output: 'number',
      outputShape: 2,
   }
}
```
Field types are defined by an implementation that has to match what is expected by ScratchBlocks.Field.register and its output and shape.

src/engine/runtime.js has been updated to handle the new "customFieldTypes"-field:
- Existing (global) field types cannot be overridden
- New fields are "namespaced" to the extension in the same way as opcodes are.

Once the custom field type has been picked up by scratch-vm a "EXTENSION_FIELD_ADDED" event is emitted.
It is then up to the hosting app to call ScratchBlocks.Field.register to register the field type with ScratchBlocks. Ex:
```
vm.addListener('EXTENSION_FIELD_ADDED', fieldInfo => {
   this.ScratchBlocks.Field.register(fieldInfo.name, fieldInfo.implementation);
});
```

Note: There are a lot of extra whitespace changes in order to fulfill the linting rules.

### Reason for Changes
It'd be nice for extensions to be able to add new field_types without having to issue a pull request to ScratchBlocks. 

### Test Coverage

No tests added :(

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1161,9 +1161,8 @@ class Runtime extends EventEmitter {
         }
 
         const defaultValue =
-            typeof argInfo.defaultValue === 'undefined'
-                ? ''
-                : escapeHtml(maybeFormatMessage(argInfo.defaultValue, this.makeMessageContextForTarget()).toString());
+            typeof argInfo.defaultValue === 'undefined' ? '' :
+                escapeHtml(maybeFormatMessage(argInfo.defaultValue, this.makeMessageContextForTarget()).toString());
 
         if (argTypeInfo.check) {
             argJSON.check = argTypeInfo.check;
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -557,6 +557,14 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Event name for reporting that an extension as asked for a custom field to be added
+     * @const {string}
+     */
+    static get EXTENSION_FIELD_ADDED () {
+        return 'EXTENSION_FIELD_ADDED';
+    }
+
+    /**
      * Event name for updating the available set of peripheral devices.
      * This causes the peripheral connection modal to update a list of
      * available peripherals.
@@ -779,6 +787,7 @@ class Runtime extends EventEmitter {
             color1: extensionInfo.colour || '#0FBD8C',
             color2: extensionInfo.colourSecondary || '#0DA57A',
             color3: extensionInfo.colourTertiary || '#0B8E69',
+            customFieldTypes: {},
             blocks: [],
             menus: []
         };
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
This is great! Thanks, @ErikMejerHansen!

The one concern I have is that this won't easily work across a worker boundary. We plan to isolate unofficial extensions using [Web Workers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) in the future, and communication between extension workers and the VM/Runtime will be very restricted for both [technical](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) and safety reasons. That said, I think this is still a great feature for internal and official extensions, as well as for prototyping.

ErikMejerHansen:
Thanks for looking at it @cwillisf !
Was a bit worried about the web workers issue. Is that something something that would block this PR from being merged?

griffpatch:
*sad face* - I'm guessing my physics extension is not going to work once
the Web Worker framework is in place. I've been accessing the convex hulls,
and sprites internal positions, etc with complete freedom up till now - It
was very fun while it lasted :(

cwillisf:
@griffpatch I'm planning to expose quite a bit more information to extensions in Scratch 3.0... we may be able to get you (and everyone else) "official" access to everything your physics extension needs even across the Web Worker boundary. There are very few things that we're already certain that we don't want to allow, mostly for reasons related to security or privacy.

kchadha (approved):
LGTM

PTegmark:
> Field types are defined by an implementation that has to match what is expected by ScratchBlocks.Field.register and its output and shape. 

Could you elaborate on this? How do I use the implementation element to create a new field type? 

I'm trying to make a new field type that is a variant of the matrix field type (the matrix field type is the one used by the Micro:Bit extension, which creates a dropdown that has a 5x5 grid on it). So far I have edited the file scratch-blocks/core/field_matrix.js to make the matrix field type suit my needs, but the problem with this is that the Micro:Bit extension has a block that also uses the matrix field type, so my changes have impacted the Micro:Bit extension as well. In order to solve this problem, I am trying to make a new field type which acts like the matrix field type, but references a different file instead of referencing field_matrix.js. 

Can I accomplish this by adding a customFieldTypes element to my extension's index.js file, and if so, how? I can't seem to find any documentation anywhere on how to make new field types in Scratch 3.

Issue #1847, opened 2018-12-14
Makey Makey blocks are not translated


apple502j:
The menu item is translated but the block itself is not.

PR #1937, opened 2019-01-23
Localize Makey Makey extension blocks

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1847

### Proposed Changes

Use localized strings on the Makey Makey extension blocks.


chrisgarrity (approved):
Looks good, and I checked that they're extracted by the scripts.

Issue #1733, opened 2018-11-06
Default input to the "speak" block should be localized


ericrosenbaum:
### Expected Behavior

When you have the editor set to (e.g.) japanese, and then you add the text to speech extension, the default input to the "speak" block should be in japanese. This should work for any language currently supported by the text to speech extension, and other languages should default to the English "hello". 

The code shown here should accomplish this (and did work, when I was testing locally): https://github.com/LLK/scratch-vm/blob/3bd079781002d339c77850a74d44b0f1ee785081/src/extensions/scratch3_text2speech/index.js#L262-L271

### Actual Behavior

Currently, the speak block always shows the default input "hello"

### Steps to Reproduce

- Switch to Japanese (or another language supported by the Text to Speech extension, i.e. included in the list of languages in the "set language" block in the extension).
- Add the Text to Speech extension
- Look at the default value of the input to the "speak" block

PR #1938, opened 2019-01-23
Fix Text to Speech extension default input localization

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1733

### Proposed Changes

Don’t use a variable in formatMessage! 

@chrisgarrity I'm not certain this is enough to make this work- lets look at this together.


chrisgarrity (approved):
This change does ensure that the default 'hello' gets extracted for upload to transifex. I'm pretty sure it'll also work to get translated when it's a supported language, but that will be easier to test once there are some translations. I think we may get some 'bug' reports from translators of 'unsupported' languages that they've translated it, but that it doesn't appear.

ericrosenbaum:
Thanks @chrisgarrity! I'll check in with you again before we test this one to confirm that we've deployed translations.

PR #1941, opened 2019-01-24
Add reference for `runtime` to blocks container

### LINKED ISSUE #1942: Pass in reference to runtime when creating `new Blocks` in the tests
#1941 adds a reference to the `runtime` to the blocks container constructor. Currently we have a lot of tests that create a new blocks container via `new Blocks()`. These references are used in ways that don't actually use the runtime, so they are technically okay, but the tests should be updated to reflect the new API.

e.g. Blocks.changeBlock used to have an optRuntime argument which was technically being used before being checked that it was falsey, but it happened to be used in the tests using a control path that didn't end up de-refrencing the non-existant optRuntime argument. This could easily lead to bugs.

cc/ @chrisgarrity

kchadha:
#### Update
**Closes #1942.**

### Proposed Changes

Add a reference to the runtime in each block container instance.

Thank you to @chrisgarrity for pairing on this!

### Reason for Changes

This is work towards fixing erroneous project saving. We want to eventually be able to track actual project changes introduced by changes to the blocks (e.g. creating a new block, changing a block field) vs. always emitting a project change whenever we receive a blockly event (https://github.com/LLK/scratch-vm/blob/develop/src/engine/blocks.js#L493). We will eventually remove that line and make that `emitProjectChanged` call within each of the `___Block` functions (moveBlock, changeBlock, createBlock, deleteBlock) when a change actually takes place.

### Test Coverage

This is not a functionality change so we did not add any new tests.

We did add a warning for calling `changeBlock` without the blocks container having a reference to the runtime. Previously this was an optional argument and none of the tests are testing any runtime needing functionality. ~We have filed a tech debt issue (#1942) to update all the tests to pass in a runtime when creating a new blocks container instance.~ We added a reference to the `runtime` to the existing calls to the Blocks constructor in the tests.

We also removed the optRuntime parameter that was being passed into `blocklyListen` in one of the tests since it is no longer needed. That test already follows a code path (e.g. deserializing a project) that will construct the blocks containers in the right way.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -17,11 +17,14 @@ const getMonitorIdForBlockWithArgs = require('../util/get-monitor-id');
 
 /**
  * Create a block container.
+ * @param {Runtime} runtime The runtime this block container operates within
  * @param {boolean} optNoGlow Optional flag to indicate that blocks in this container
  * should not request glows. This does not affect glows when clicking on a block to execute it.
  */
 class Blocks {
-    constructor (optNoGlow) {
+    constructor (runtime, optNoGlow) {
+        this.runtime = runtime;
+
         /**
          * All blocks in the workspace.
          * Keys are block IDs, values are metadata about the block.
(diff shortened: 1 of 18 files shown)


paulkaplan (approved):
LGTM, thanks for including those test updates

PR #1943, opened 2019-01-24
Start executing hats before other threads change values

### LINKED ISSUE #906: "when timer > 0" hat + "reset timer" block doesn't make a loop
In 2.0, this runs forever after starting:
image

But in 3.0, it stops after a finite number of times.

ktbee:
### Resolves

Fixes #906 

### Proposed Changes

Use the runtime's `currentMSecs` timestamp for clock's now value and start executing hat threads in the runtime's `startHats` before all other threads execute.

### Reason for Changes

There is a timer and a thread execution order incompatibility for edge activated hats between Scratch 2 and 3. In Scratch 2, it's easier to have a timer's time elapsed value equal zero. The timer uses `currentMSecs` as its current time value, and that timestamp is mostly only updated [once per step of the sequencer](https://github.com/LLK/scratch-flash/blob/72e4729b8189d11bbe51b6d94144b0a3c392ac9a/src/interpreter/Interpreter.as#L245) (though it can get updated in other circumstances as well). Since the timer's elasped value equals zero, the "greater than hat" in this example project can be triggered when the timer goes from zero to a value greater than zero. This PR switches the clock's timer to use the runtime's `currentMSecs` as its now value, which will update less frequently than the previous `Date.now()` values.

In addition to this timestamp not being updated as frequently as in Scratch 3, we found a race condition in Scratch 2 where [`currentMSecs` is updated _after_ edge triggered hats are processed.](https://github.com/LLK/scratch-flash/blob/develop/src/scratch/ScratchRuntime.as#L179-L180) This means these hats are effectively using the timestamp from the last step. This PR starts hat threads in `startHats` so that the hats are processed before `currentMSecs` is updated in the sequencer's `stepThreads`.  Processing hats before other threads will be helpful as well for any other edge activated hats that need to process some data before other threads start changing the project's state. This'll be particularly useful for extension hat blocks. 

### Test Coverage

These changes break a number of tests. Here's my plan:

- [x] Investigate broken tests and confirm they're broken in a way expect
- [x] Fix broken tests
- [x] Add new tests for updated hat thread timing


ktbee:
I've finished updating tests and fixing bugs other tests caught (🎉 ), and added a new test to make sure a "timer greater than 0" hat is activated once every other step when the timer is reset. Here's a summary/explanation of the test changes:

- In io_clock.js, I changed the assertions to confirm that the `projectTimer` doesn't increase during a step and also confirmed that it increases after then next step has finished. @thisandagain, I believe you wrote the bulk of this test originally, so let me know if I'm missing anything in these changes
- In hat-threads-run-every-frame.js: 
     - I switched out the loudness hat in the test fixture with the timer hat since we want it to always be processed as true and we don't have an audio engine available in this test.
     - In the test "edge activated hat should trigger for both sprites when sprite is duplicated", I removed the extra `execute` functions calls since this behavior now happens in `startHats` and is no longer completely prevented by [`currentStepTime being zero`](https://github.com/LLK/scratch-vm/blob/develop/test/integration/hat-threads-run-every-frame.js#L125).
     - @kchadha, I believe you wrote the bulk of this test originally, so let me know if I'm missing anything in these changes

thisandagain:
The test updates look good to me. Thanks @ktbee !

kchadha:
@ktbee, the changes look okay. I am a little concerned about the change and what the implications are for project execution (e.g. the change affects all hats and not just edge-triggered hats). It would be great to get this on a separate testing branch to test out the changes.

cc/ @BryceLTaylor

test/integration/hat-threads-run-every-frame.js
```diff
@@ -136,7 +136,6 @@ test('edge activated hat should trigger for both sprites when sprite is duplicat
             t.assert(vm.runtime.threads[0].status === Thread.STATUS_RUNNING);
             // Run execute on the thread to populate the runtime's
             // _edgeActivatedHatValues object
-            execute(vm.runtime.sequencer, vm.runtime.threads[0]);
```
kchadha, line 139: The comment above this should also be deleted if this line is getting deleted

kchadha:
@ktbee, regarding swapping out the loudness hat with the timer hat in the test fixture, I'm not sure what you mean by:

> since we want it to always be processed as true

ktbee:
@kchadha in terms of why I switched to a timer "greater than hat" rather than a loudness one, we want the hat to always be activated or have its activation condition be `true` in the test. That's what I meant to say with "processed to be true", but let me know if I can clarify anything more. 

Also here is a version of scratch-gui with this PR's branch deployed to it. Let me know if this is enough for you testing needs @BryceLTaylor @kchadha: https://ktbee.github.io/scratch-gui/ !

kchadha:
> @kchadha in terms of why I switched to a timer "greater than hat" rather than a loudness one, we want the hat to always be activated or have its activation condition be `true` in the test. That's what I meant to say with "processed to be true", but let me know if I can clarify anything more.

@ktbee, I guess I don't understand why that is a necessary condition for the test now when it wasn't before these changes. To me, that indicates that this code isn't working with the previous test, but I wonder if I'm missing something...

ktbee:
@kchadha good point, previously the test worked. It fails with this PR's changes because of the earlier execution of hat threads. Because tests in hat-threads-run-every-frame.js just check whether a thread is a hat thread and its status, it doesn't matter if the hat ever returns true (which it wouldn't with the `loudness` hat) since no other blocks on that thread are tested to see if they are executed. With these changes, our hat thread gets executed earlier, so it will be destroyed by the time we check its status in the test. We need the hat to return true so that its thread sticks around long enough for the test to check it.

src/engine/sequencer.js
```diff
@@ -138,7 +138,6 @@ class Sequencer {
                     activeThread.status === Thread.STATUS_DONE) {
                     // Finished with this thread.
                     stoppedThread = true;
-                    this.runtime.updateCurrentMSecs();
```
kchadha, line 141: As discussed offline, I'm not sure I understand this change or whether it's necessary? It seems like it might be analogous to this code in Scratch 2, but I'm not sure:

https://github.com/LLK/scratch-flash/blob/develop/src/interpreter/Interpreter.as#L277

kchadha:
@ktbee, marking as needs work because I have a question in the comments.

src/engine/sequencer.js
ktbee: Yup, I think this update of `currentMSecs` is analogous to that line in Scratch 2, or at least that's how I understood it when I added it initially. However after taking another look at it and some feedback from @mzgoddard, I've realized that we don't need this `currentMSecs` update because in Scratch 2 it was used to eventually end this while loop: https://github.com/LLK/scratch-flash/blob/develop/src/interpreter/Interpreter.as#L251 

And in Scratch 3, the analogous while loop uses the Timer util, which is updated by Date.now:

https://github.com/LLK/scratch-vm/blob/develop/src/engine/sequencer.js#L86-L89

I don't think this `currentMSecs` update is causing any bugs and I don't think removing it fixes any bugs, so I'll take it out of this PR and make a follow-up PR to research whether or not we want to take out that line.

ktbee:
@kchadha as discussed in the comments, I re-added the extra update to `currentMSecs`. When I did that, it broke this sb2 test that is run by execute.js: [hat-thread-execution.sb2.zip](https://github.com/LLK/scratch-vm/files/3019809/hat-thread-execution.sb2.zip)

So it turns out there is a difference between whether [we update `currentMSecs` here](https://github.com/LLK/scratch-vm/blob/develop/src/engine/sequencer.js#L141) or not, and it looks like it makes Scratch 3 a little less compatible if we have it. I've taken this test out for now since we have other tests covering this change in hat-threads-run-every-frame.js and io_clock.js, and will add it to the follow up issue to investigate removing this line.

ktbee:
And here's the follow-up issue! https://github.com/LLK/scratch-vm/issues/2074

ktbee:
It turns out that removing that extra update to `currentMSecs` is more crucial than I had thought, which @kchadha and I discovered after digging into this a little more offline. With this extra `currentMSecs` update, the original bug in #906 isn't fixed. So I've re-removed that line and added my test for it.

src/engine/sequencer.js
kchadha: @ktbee and I discussed this change offline and it looks like the removal of this line is crucial to the bug this PR is fixing. Since we're now executing hats earlier (in order to catch the timer value at 0 instead of at some value > 0), we need to remove this update to currentMSecs right after the thread is killed.

kchadha (approved):
@ktbee thanks for making the updates and sorry for the back and forth on this PR! LGTM! =D

Issue #1939, opened 2019-01-24
Message and variable name including '&' destructs toolbox and blocks


yokobond:
### Expected Behavior

'&' should be able to use in message or variable names. 

### Actual Behavior

When a message or variable contains '&', the message or variable blocks in the Toolbox has no name and blocks in the Code area is empty. 

This problem is reported in discussion forum. 
[When "&" is included in the message, the existing block disappears. - Discuss Scratch](https://scratch.mit.edu/discuss/topic/337375/?fbclid=IwAR2q-siYCFqzo62IeUhIcKTrNRM8_lhZG4EdGVeXRQEFZYPRNVO386myakM)

### Steps to Reproduce

1. Name a new message or variable with '&'.
2. Open 'Costume' tab and re-open 'Code' tab. 

### Operating System and Browser

MacOS 10.13.6
Chrome 71.0.3578.98

apple502j:
Kind of #1077 and regression of #178

kchadha:
Tracking this in #1077.

PR #1945, opened 2019-01-25
Escape variable name in XML

yokobond:
### Resolves

Fix for #1939 and partially #1077 .
Special characters '<>&\'"' can be used in variable and message name. 

### Proposed Changes

aVariable.toXML() escapes special characters in the name. 

### Reason for Changes

Blocks in pallet and code pane build SVG string from XML presentation of variables. Variable name is used in blocks, so that the name of variables should be escaped special character in its XML. 

Broadcast messages is a kind of variable in block presentation. Then this code fix the problem of broadcast message including special characters too. 

### Test Coverage

Added new unit test cases for engine/variable. It validates that toXML() returns escaped special characters '<>&\'"'. 
It contains tests for default specification too. 

Integration test was not included, because these special characters are no harm in scratch-vm. This fix can be checked only on scratch-gui.

### DIFF
--- a/test/unit/engine_variable.js
+++ b/test/unit/engine_variable.js
@@ -1,0 +1,110 @@
+const test = require('tap').test;
+const Variable = require('../../src/engine/variable');
+const htmlparser = require('htmlparser2');
+
+test('spec', t => {
+    t.type(typeof Variable.SCALAR_TYPE, typeof Variable.LIST_TYPE);
+    t.type(typeof Variable.SCALAR_TYPE, typeof Variable.BROADCAST_MESSAGE_TYPE);
+
+    const varId = 'varId';
+    const varName = 'varName';
+    const varIsCloud = false;
+    let v = new Variable(
+        varId,
+        varName,
+        Variable.SCALAR_TYPE,
+        varIsCloud
+    );
+
+    t.type(Variable, 'function');
+    t.type(v, 'object');
+    t.ok(v instanceof Variable);
+
+    t.equal(v.id, varId);
+    t.equal(v.name, varName);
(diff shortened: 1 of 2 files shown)


kchadha:
@yokobond, thank you for submitting this pull request. We are actively working on the fix for #1077, and while your fix works for part of the issue, we are working on a larger change that might incorporate these changes as part of it.

Since we are still working on the fix for the larger issue, we will hold off on merging this pull request until we have figured out all the pieces of the larger issue to make sure these changes don't conflict with that work.

test/unit/engine_variable.js
```diff
@@ -0,0 +1,110 @@
+const test = require('tap').test;
+const Variable = require('../../src/engine/variable');
+const htmlparser = require('htmlparser2');
+
+test('spec', t => {
+    t.type(typeof Variable.SCALAR_TYPE, typeof Variable.LIST_TYPE);
```
kchadha, line 6: I don't think this is necessary to fix now, but this test (and the one below) would probably be better written as the following:

```suggestion
    t.type(typeof Variable.SCALAR_TYPE, 'string');
```

We usually use the 'spec' test to verify the existence and types of properties on the class we are testing.

kchadha (approved):
@yokobond thank you for making this pull request and for adding the tests! I left a minor suggestion in the comments, but I will merge this pull request as I would like to incorporate it into other changes I am making to resolve #1077.

test/unit/engine_variable.js
yokobond: I agree this. It must be string not only same type. 
I overlooked that the return value is used by scratch-blocks as it is.

PR #1946, opened 2019-01-25
Fix saving and loading of slider min max values.

kchadha:
### Proposed Changes

Slider min and max values weren't getting saved for monitors correctly. Fixes this issue.
Makes the userAgent property in the meta section of sb3 files more consistent (e.g. always a string type like the sb3 parser expects).

### Test Coverage

 Add test for sb2 to sb3 conversion that tests all monitor properties remain the same after save and load. The new test is basically the same as monitor_sb2.js but adds saving and loading before checking the monitor properties.

This is our first sb2 -> sb3 test, so there were some hitches. One unresolved issue to investigate is why running vm.start before the 2 -> 3 conversion doesn't find the stage properly when running monitor blocks...

Thanks to @chrisgarrity for helping with this!!! =]

### DIFF
--- a/test/integration/monitors_sb2_to_sb3.js
+++ b/test/integration/monitors_sb2_to_sb3.js
@@ -1,0 +1,155 @@
+const path = require('path');
+const test = require('tap').test;
+const makeTestStorage = require('../fixtures/make-test-storage');
+const readFileToBuffer = require('../fixtures/readProjectFile').readFileToBuffer;
+const VirtualMachine = require('../../src/index');
+
+const projectUri = path.resolve(__dirname, '../fixtures/monitors.sb2');
+const project = readFileToBuffer(projectUri);
+
+const vm = new VirtualMachine();
+vm.attachStorage(makeTestStorage());
+
+// TODO figure out why running threads doesn't work in this test
+// vm.start();
+vm.clear();
+vm.setCompatibilityMode(false);
+vm.setTurboMode(false);
+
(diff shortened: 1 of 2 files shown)


test/integration/monitors_sb2_to_sb3.js
```diff
@@ -0,0 +1,155 @@
...
+const vm = new VirtualMachine();
+vm.attachStorage(makeTestStorage());
+
+// TODO figure out why running threads doesn't work in this test
+// vm.start();
+vm.clear();
+vm.setCompatibilityMode(false);
+vm.setTurboMode(false);
+
+test('Set Up - load the sb2 project', t => {
```
rschamp: I think for this kind of thing you want to use `beforeEach` in the suite for clarity (since it's not a test) and possibly to ensure the right order of operations if we start parallelizing things (although AFAICT tap currently doesn't parallelize tests within a suite).

paulkaplan (approved):
LGTM!

rschamp (approved):
Looks great! Just a suggestion about test set up — maybe it might help with the stuff you have commented out?

test/integration/monitors_sb2_to_sb3.js
kchadha: Done! Thanks for finding it. =]

kchadha:
@rschamp Thanks for finding the beforeEach! Changed the test to use it.

PR #1944, opened 2019-01-24
Benchmark load time

mzgoddard:
### Resolves

Benchmark loading Scratch and Scratch Projects so we can make that faster.

### Proposed Changes

- Build benchmark.js
- Store global time values and use the performance API to record some load timing data
- Render the overall load duration in the RunningView area of the benchmark

### Reason for Changes

- Make benchmark loading and execution closer to how other uses of vm operate, as a large javascript file.
- Record and render loading duration so we can compare between changes to the VM and dependencies.
- Changing vm benchmark before gui to get an experimental area up without necessarily negatively impacting the primary users of Scratch.

### Benchmark Data

Instead of being comparison data, so here are some baselines.

project id | platform | run 1 | run 2 | run 3 | run 4 | run 5 | avg
--------- | ------ | ---- | ---- | ---- | ---- | ---- | ----
169401431 | chrome | 9820 | 9715 | 9675 | 10270 | 10323 | 9960.6
169401431 | firefox | 6568 | 5919 | 5348 | 5478 | 5307 | 5724
169401431 | safari | 9048 | 8531 | 7954 | 8368 | 7786 | 8337.4
169401431 | book | 25781 | 19570 | 20712 | 19658 | 19547 | 21053.6
173918262 | chrome | 2086 | 2446 | 1753 | 1796 | 1809 | 1978
173918262 | firefox | 2073 | 1371 | 1558 | 1358 | 1435 | 1559
173918262 | safari | 1943 | 1306 | 1361 | 1246 | 1306 | 1432.4
173918262 | book | 12337 | 6408 | 7018 | 6283 | 6338 | 7676.8
173918262 | pi | 22557 | 18754 | 17990 | 18212 | 19023 | 19307.2

PR #1947, opened 2019-01-25
Deserialize sb2 assets before other sprite data and blocks

mzgoddard:
### Resolves

Decrease average time spent loading a project.

### Proposed Changes

Depends on #1944.

Pass over the sb2 data, and find all image and sound assets and start loading them immediately.

### Reason for Changes

Currently asset loading starts when it's sprite is being created and its blocks. It takes a noticeable amount of time to get through all of sprites and by extension loading all of the assets. Instead we can make a pass to just handle the asset data and then a second pass over the other sprite data. This lets us start loading all assets a lot earlier making the project appear to load faster by removing the delay.

### Benchmark Data

You can compare against the baseline data here: https://github.com/LLK/scratch-vm/pull/1944.

project id | platform | run 1 | run 2 | run 3 | run 4 | run 5 | avg
--------- | ------ | ---- | ---- | ---- | ---- | ---- | ----
169401431 | chrome | 8664 | 7805 | 7487 | 7642 | 7504 | 7820.4
169401431 | firefox | 5431 | 4254 | 4424 | 4136 | 3970 | 4443
169401431 | safari | 8510 | 7343 | 7695 | 7626 | 7632 | 7761.2
169401431 | book | 22997 | 18725 | 18057 | 18274 | 18548 | 19320.2
173918262 | chrome | 1262 | 1331 | 1068 | 1054 | 1038 | 1150.6
173918262 | firefox | 1756 | 1181 | 1057 | 1168 | 1163 | 1265
173918262 | safari | 2000 | 1226 | 1135 | 1226 | 1226 | 1362.6
173918262 | book | 11492 | 6567 | 6688 | 7039 | 6379 | 7633
173918262 | book | 4099 | 3081 | 3359 | 3152 | 3252 | 3388.6
173918262 | pi | 22537 | 19767 | 18516 | 14893 | 15078 | 18158.2

### DIFF
--- a/src/serialization/sb2.js
+++ b/src/serialization/sb2.js
@@ -391,47 +391,28 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
 };
 
 /**
- * Parse a single "Scratch object" and create all its in-memory VM objects.
- * TODO: parse the "info" section, especially "savedExtensions"
+ * Parse the assets of a single "Scratch object" and load them. This
+ * preprocesses objects to support loading the data for those assets over a
+ * network while the objects are further processed into Blocks, Sprites, and a
+ * list of needed Extensions.
  * @param {!object} object - From-JSON "Scratch object:" sprite, stage, watcher.
  * @param {!Runtime} runtime - Runtime object to load all structures into.
- * @param {ImportedExtensionsInfo} extensions - (in/out) parsed extension information will be stored here.
  * @param {boolean} topLevel - Whether this is the top-level object (stage).
  * @param {?object} zip - Optional zipped assets for local file import
- * @return {!Promise.<Array.<Target>>} Promise for the loaded targets when ready, or null for unsupported objects.
+ * @return {?{costumePromises:Array.<Promise>,soundPromises:Array.<Promise>,children:object}}
+ *   Object of arrays of promises and child objects for asset objects used in
+ *   Sprites. null for unsupported objects.
  */
-const parseScratchObject = function (object, runtime, extensions, topLevel, zip) {
+const parseScratchAssets = function (object, runtime, topLevel, zip) {
     if (!object.hasOwnProperty('objName')) {
-        if (object.hasOwnProperty('listName')) {
-            // Shim these objects so they can be processed as monitors
-            object.cmd = 'contentsOfList:';
-            object.param = object.listName;
-            object.mode = 'list';
-        }
-        // Defer parsing monitors until targets are all parsed
-        object.deferredMonitor = true;
-        return Promise.resolve(object);
+        // Skip parsing monitors. Or any other objects missing objName.
+        return null;
     }
 
-    // Blocks container for this object.
-    const blocks = new Blocks(runtime);
-    // @todo: For now, load all Scratch objects (stage/sprites) as a Sprite.
-    const sprite = new Sprite(blocks, runtime);
-    // Sprite/stage name from JSON.
-    if (object.hasOwnProperty('objName')) {
-        if (topLevel && object.objName !== 'Stage') {
-            for (const child of object.children) {
-                if (!child.hasOwnProperty('objName') && child.target === object.objName) {
-                    child.target = 'Stage';
-                }
-            }
-            object.objName = 'Stage';
-        }
+    const assets = {costumePromises: [], soundPromises: [], children: []};
 
-        sprite.name = object.objName;
-    }
     // Costumes from JSON.
-    const costumePromises = [];
+    const costumePromises = assets.costumePromises;
     if (object.hasOwnProperty('costumes')) {
         for (let i = 0; i < object.costumes.length; i++) {
             const costumeSource = object.costumes[i];
@@ -476,7 +457,7 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
         }
     }
     // Sounds from JSON
-    const soundPromises = [];
+    const soundPromises = assets.soundPromises;
     if (object.hasOwnProperty('sounds')) {
         for (let s = 0; s < object.sounds.length; s++) {
             const soundSource = object.sounds[s];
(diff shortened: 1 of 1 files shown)


cwillisf (dismissed):
The idea seems sound but it looks like this breaks quite a few of the tests. Do you know what's going on there?

mzgoddard:
@cwillisf There was a change I made that was in a critical part for deserializing sb2. Forgot that `Promise.resolve` doesn't like being called without its `Promise` context. Fixing that solves all of the failures.

And I pushed the change on Friday but Travis didn't rerun.

mzgoddard:
- Re-recorded chrombook and raspberry pi numbers. Building baseline numbers for them I realized the numbers here were previously baseline.

src/serialization/sb2.js
```diff
@@ -809,7 +844,13 @@ const sb2import = function (json, runtime, optForceSprite, zip) {
         extensionIDs: new Set(),
         extensionURLs: new Map()
     };
-    return parseScratchObject(json, runtime, extensions, !optForceSprite, zip)
+    return Promise.resolve(parseScratchAssets(json, runtime, extensions, !optForceSprite, zip))
+        // Force this promise to wait for the next loop in the js tick. Let
+        // storage have some time to send off asset requests.
+        .then(assets => Promise.resolve(assets))
```
cwillisf, line 859: I missed this (in its other form) the first time around... could you explain a bit more why this is needed? I get that it delays this `then` chain by one tick but why do we need that? It seems like somebody (storage? `parseScratchAssets`?) should return something that resolves at the right time without the need for an arbitrary one-tick delay. Is there a `Promise.all(...)` missing from somewhere maybe?

mzgoddard: This is a bit of an onion.

ScratchStorage uses promises that delay making the actual requests until later. Some of those promises can return promises themselves, further delaying the actual request. As is we can't know when a request is actually made.

Additionally defining when a request is made I think is tricky. Is it when a fetch or xhr call is made? What if that fails and you try the next tool or source. So then is it when the first tool receives a 200/300 status code? I think may be better to leave this as a vague idea and instead to make part of scratch that are synchronous into green threads.

Activities that take a lot of time like deserializing a project or the initial construction of the gui's dom should be made into green threads where they yield to other work by delaying further work with a promise. For example deserializing an sb2's could yield a promise at the start of each object. That way the stage and each of the sprites would let some other work happen before they do work. Using promises to green thread this (unless those promises used setTimeout or another event) would continue to block the main thread so the render would appear to happen all at once as it does now.

The solution above is a minimal achievement of this idea so that another Issue or PR could have a better discussion of the concept.

src/serialization/sb2.js
```diff
@@ -390,47 +390,20 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
...
- * TODO: parse the "info" section, especially "savedExtensions"
+ * Parse the assets of a single "Scratch object" and load them.
  * @param {!object} object - From-JSON "Scratch object:" sprite, stage, watcher.
  * @param {!Runtime} runtime - Runtime object to load all structures into.
  * @param {ImportedExtensionsInfo} extensions - (in/out) parsed extension information will be stored here.
  * @param {boolean} topLevel - Whether this is the top-level object (stage).
  * @param {?object} zip - Optional zipped assets for local file import
- * @return {!Promise.<Array.<Target>>} Promise for the loaded targets when ready, or null for unsupported objects.
+ * @return {{costumePromises:Array.<Promise>,soundPromises:Array.<Promise>,children:object}}
+ *   Promise for the loaded targets when ready, or null for unsupported objects.
```
kchadha: This return description needs to be updated...

src/serialization/sb2.js
```diff
@@ -390,47 +390,20 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
 };
 
 /**
- * Parse a single "Scratch object" and create all its in-memory VM objects.
- * TODO: parse the "info" section, especially "savedExtensions"
+ * Parse the assets of a single "Scratch object" and load them.
  * @param {!object} object - From-JSON "Scratch object:" sprite, stage, watcher.
```
kchadha, line 398: This comment should reflect that this is a pre-processing step to start retrieving and loading assets and that watchers will be ignored... (because they don't have costumes and sounds...)

A side effect of these changes is that we could load extra assets from malformed project jsons (e.g. there's nothing preventing a watcher from having costumes and sounds). I would prefer to have an early exit in here similar to the one at the top of parseScratchObject...

src/serialization/sb2.js
```diff
@@ -390,47 +390,20 @@ const parseMonitorObject = (object, runtime, targets, extensions) => {
...
-    if (object.hasOwnProperty('objName')) {
-        if (topLevel && object.objName !== 'Stage') {
-            for (const child of object.children) {
-                if (!child.hasOwnProperty('objName') && child.target === object.objName) {
-                    child.target = 'Stage';
-                }
-            }
-            object.objName = 'Stage';
-        }
+const parseScratchAssets = function (object, runtime, extensions, topLevel, zip) {
```
kchadha: Are all of these arguments necessary? I think extensions and topLevel shouldn't be needed.

src/serialization/sb2.js
```diff
@@ -504,10 +477,70 @@ const parseScratchObject = function (object, runtime, extensions, topLevel, zip)
             // the file name of the sound should be the soundID (provided from the project.json)
             // followed by the file ext
             const assetFileName = `${soundSource.soundID}.${ext}`;
-            soundPromises.push(deserializeSound(sound, runtime, zip, assetFileName)
-                .then(() => loadSound(sound, runtime, sprite))
```
kchadha, line 509: What's the reasoning for handling loadSound later (when parsing other target info) and not here?

kchadha (changes requested):
I left some suggestions/questions in the code review

src/serialization/sb2.js
mzgoddard: Sprites don't exist yet. `loadSound` needs a soundBank to store the loaded SoundPlayers on. Sprites have one, but I don't think we should create a Sprite yet at this point. Instead I have a follow up for this https://github.com/LLK/scratch-vm/pull/1972

src/serialization/sb2.js
mzgoddard: `extensions` isn't needed.

`topLevel` is needed for costume rotationCenter values.

src/serialization/sb2.js
mzgoddard: Hmm. Good point.

mzgoddard:
- Made changes following @kchadha's comments

kchadha (approved):
LGTM, thanks for making those changes!

PR #1949, opened 2019-01-27
Round ghost effect

### LINKED ISSUE #1661: Should use math.round() for ghost effect
### Expected Behavior

If you get the ghost effect to 99.9999 it should really use 100.

### Actual Behavior

Currently if your ghost effect value is 99.999999 it doesn't treat it as 100 in 3.0, but it does in 2.0.  This means you can click a sprite that is ghosted to close to 100 but not quite 100.  In 2.0 you click through that sprite.

This can be seen in this project:https://github.com/LLK/scratch-gui/issues/2820
It changes the ghost effect by 3.333333 on the Thumbnail sprite 30 times (for a total of 99.99999)

Here's the relevant stack from the "Thumbnail" sprite:
screen shot 2018-10-16 at 2 37 32 pm

This came up in our discussion of this issue: https://github.com/LLK/scratch-gui/issues/2820

### Steps to Reproduce

Run the project linked above
Note: you can't click through the thumbnail in 3.0 after it "disappears"

apple502j:
### Resolves
Resolves #1661 
### Proposed Changes
Rounds ghost effect.
### Reason for Changes
Compatibility

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -501,7 +501,6 @@ class Scratch3LooksBlocks {
         const effect = Cast.toString(args.EFFECT).toLowerCase();
         let value = Cast.toNumber(args.VALUE);
         if (effect === 'ghost') {
-            value = 
             value = MathUtil.clamp(Math.round(value),
                 Scratch3LooksBlocks.EFFECT_GHOST_LIMIT.min,
                 Scratch3LooksBlocks.EFFECT_GHOST_LIMIT.max);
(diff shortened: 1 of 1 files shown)


kchadha (commented):
@apple502j, thank you for the pull request! Upon further investigation, it seems that the scratch2 code is not actually rounding the ghost effect value and there's something else going on here, so the information in #1661 is not accurate. I will update that issue.

I am closing this PR as #1661 needs more investigation about what's actually causing the compatibility issue in this project: 236788917

PR #1956, opened 2019-01-28
Lazy eval big dependencies

mzgoddard:
### Resolves

Decrease average time evaluating javascript and data embedded in javascript.

### Proposed Changes

Depends on https://github.com/LLK/scratch-vm/pull/1944.

- Evaluate the following code on demand
  - extensions
  - scratch-parser
  - sb1, sb2, and sb3 (de)serialization
  - text-encoding

### Reason for Changes

Built Scratch javascript files are multiple megabytes in size. To load a Scratch project, that code must be downloaded or fetched from a cache, compiled, and evaluated before a project can starting loading by downloading its data file. If we can start that later download earlier Scratch will seem to load faster. One step we can make do that is to evaluate parts of Scratch later.

The selected elements of Scratch whose evaluation is delayed in this pull request either are not needed by all systems (text-encoding is a polyfill) or not needed by all projects (extensions, sb1, sb2, and sb3 (de)serialization) or not needed to download the data.

The following is the amount of (unminified) code whose parsing is delayed here:
- music extension data: ~1.9MB
- extensions: 355KB
- scratch-parser: 410KB
- text-encoding: 616KB
- sb1, sb2, sb3 (de)serialization: 224KB

Total: ~3.5MB

This work could be taken further by putting the extensions, parser, and serialization code in their own javascript chunk and loading it in a script tag following the main js tag. This would reduce the amount of code that must be downloaded and compiled before the project can initialize loading.

### Benchmark Data

You can compare with baseline numbers from https://github.com/LLK/scratch-vm/pull/1944.

project id | platform | run 1 | run 2 | run 3 | run 4 | run 5 | avg
--------- | ------ | ---- | ---- | ---- | ---- | ---- | ----
169401431 | chrome | 7051 | 6714 | 7551 | 7214 | 7448 | 7195.6
169401431 | firefox | 5041 | 4180 | 4030 | 3783 | 3878 | 4182.4
169401431 | safari | 7687 | 7049 | 7153 | 7498 | 7166 | 7310.6
169401431 | book | 20750 | 19229 | 19645 | 18672 | 18504 | 19360
173918262 | chrome | 1190 | 1032 | 1022 | 967 | 1098 | 1061.8
173918262 | firefox | 1328 | 1092 | 1150 | 1159 | 1199 | 1185.6
173918262 | safari | 1295 | 1199 | 1120 | 1051 | 1111 | 1155.2
173918262 | book | 3291 | 2621 | 2694 | 2610 | 3081 | 2859.4
173918262 | pi | 24826 | 18603 | 18489 | 14952 | 14560 | 18286

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -7,32 +7,21 @@ const BlockType = require('./block-type');
 // These extensions are currently built into the VM repository but should not be loaded at startup.
 // TODO: move these out into a separate repository?
 // TODO: change extension spec so that library info, including extension ID, can be collected through static methods
-const Scratch3PenBlocks = require('../extensions/scratch3_pen');
-const Scratch3WeDo2Blocks = require('../extensions/scratch3_wedo2');
-const Scratch3MusicBlocks = require('../extensions/scratch3_music');
-const Scratch3MicroBitBlocks = require('../extensions/scratch3_microbit');
-const Scratch3Text2SpeechBlocks = require('../extensions/scratch3_text2speech');
-const Scratch3TranslateBlocks = require('../extensions/scratch3_translate');
-const Scratch3VideoSensingBlocks = require('../extensions/scratch3_video_sensing');
-const Scratch3Speech2TextBlocks = require('../extensions/scratch3_speech2text');
-const Scratch3Ev3Blocks = require('../extensions/scratch3_ev3');
-const Scratch3MakeyMakeyBlocks = require('../extensions/scratch3_makeymakey');
-// todo: only load this extension once we have a compatible way to load its
-// Vernier module dependency.
-// const Scratch3GdxForBlocks = require('../extensions/scratch3_gdx_for');
 
 const builtinExtensions = {
-    pen: Scratch3PenBlocks,
-    wedo2: Scratch3WeDo2Blocks,
-    music: Scratch3MusicBlocks,
-    microbit: Scratch3MicroBitBlocks,
-    text2speech: Scratch3Text2SpeechBlocks,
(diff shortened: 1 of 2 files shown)


mzgoddard:
- Add benchmark data

rschamp (approved):
It seems like this could be brittle, since by default people will add new dependencies to the top of the file (since that's easiest to read/parse). For the future, is it possible to configure Webpack to do this kind of optimization automatically? It seems like it should be able to see when a dependency is only used in a certain scope.

mzgoddard:
> It seems like this could be brittle, since by default people will add new dependencies to the top of the file (since that's easiest to read/parse).

We could move the function wrapped requires back to about the original location of the requires.

> For the future, is it possible to configure Webpack to do this kind of optimization automatically? It seems like it should be able to see when a dependency is only used in a certain scope.

Hmm. The easy way to configure this would be with a loader. It would make a intermediate module in the webpack output wrapping the original require like the function currently does.

Webpack could be configured with a plugin to detect the uses of a variable are in functions and  hoist the require into those functions.

I don't know of a loader or plugin off the top of my head that does this. The first approximate one I can think of is bundle-loader that creates webpack chunks for each entry. That'd turn the extensions into a chunk for each (or as modified by the webpack config, for example they could be grouped into a larger chunk if they don't reach a minimum size.

Otherwise this would be proabably be a new loader or plugin. The loader would be very easy to make. 3 lines minimum. (So it might exist on npm but is probably not popular and so hard to find.) The plugin would be a lot more involved (mostly by needing to write plugins for webpack's acorn parser plugin wrapper).

PR #1958, opened 2019-01-28
Update, re-order and clean up Vernier Force & Accel extension blocks

ericrosenbaum:
- Use constants for gesture thresholds and gravity
- Swap tilt x and y calculations
- Re-factor logic in the peripheral for scaling accel and spin speed values 
- Report spin speed in degrees per frame
- Use enums for all string menu values
- Localize menu text where appropriate
- Re-order blocks

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -44,6 +44,12 @@ const MOVED_THRESHOLD = 3;
 const SHAKEN_THRESHOLD = 30;
 
 /**
+ * Threshold for acceleration magnitude, to check if we are facing up.
+ * @type {number}
+ */
+const FACING_THRESHOLD = 9;
+
+/**
  * Acceleration due to gravity, in m/s^2.
  * @type {number}
  */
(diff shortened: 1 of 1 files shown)


evhan55:
Code re-org looks okay so far to me.

Blocks:
- `forever <facing up>?` feels flaky to me (reports false/true intermittently when it's facing up
screen shot 2019-01-29 at 2 08 29 pm

- is `forever <falling>?` supposed to be true just momentarily or steady while falling?
screen shot 2019-01-29 at 2 08 23 pm

ericrosenbaum:
@evhan55 good catch on the `<facing up>` block- I had inadvertently changed the threshold, but that is now fixed and feels stable to me.

The `<falling?>` block should be true while the sensor is in free fall. One problem is that if the sensor is rotating in the air, the acceleration from that prevents free fall detection. The Vernier team is working on a solution to that using gyro measurements, so the fix will come in a future PR. For now, it's easiest to test that block by trying to drop the sensor in such a way that it doesn't rotate as it falls.

evhan55:
lgtm

PR #1959, opened 2019-01-29
Declare const once to be reused in for loop

ktbee:
### Resolves

No issues exists for this, but @mzgoddard and I noticed a small potential performance improvement while working on other issues. 

### Proposed Changes

Move declaring `hatMeta` out of a for-loop since its value doesn't need to be set or updated within those loops. 

### Reason for Changes

Performance benefit by avoiding redeclaring a variable more than once.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1461,6 +1461,8 @@ class Runtime extends EventEmitter {
         }
         const instance = this;
         const newThreads = [];
+        // Look up metadata for the relevant hat.
+        const hatMeta = instance._hats[requestedHatOpcode];
 
         for (const opts in optMatchFields) {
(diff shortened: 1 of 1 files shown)


kchadha (commented):
This seems like a safe change, but I'm concerned about instance._hats being updated while the loop is running (e.g. loading a new extension or uploading a new sprite), and it seems like this change has more of an opportunity to result in `hatMeta` being undefined which would result in an error below...

Do you think this a big enough performance boost to warrant the risk associated with the change?

ktbee:
@kchadha good question! Since allScriptsDo is synchronous and doesn't call any other functions that could install extensions, I don't think instance._hats can be updated mid-loop. That makes moving the variable declaration worth it from my perspective, but let me know if I'm missing anything.

kchadha (approved):

PR #1962, opened 2019-01-29
Project dirty state fixes

kchadha:
### Proposed Changes

Depends on #1941 (and currently the diff includes changes from that PR).

@chrisgarrity and I worked on all of these changes together.

Fix erroneous project changed state reporting.

This PR starts to decouple emitting a targets update from emitting a project changed.
Prior to this PR, we were erroneously emitting a project changed state in the following cases:
1) Loading a project
2) Running a project (scratch 2.0 did not consider running a project to change the dirty state of a project, this also includes clicking a block in the toolbox or on a sprite's workspace).
3) Switching from project page to editor mode
4) Switching sprites

The above scenarios were getting triggered by emitting project changes in the following scenarios:
1) during `vm.clear()` (which happens during load project),
2) during renaming a sprite whether or not such a change occurred (also happens during project load), 3) when receiving any scratch blocks event in a blocks container that can affect the project (e.g. blocks containers for actual sprites but not blocks containers for the flyout or monitors).
4) during `runtime._step()` which is always running (before/during/after project load)

After investigating and comparing the cases in which Scratch 2.0 and Scratch 3.0 track dirty state, we decided to resolve these issues by:

1) No longer emitting project state changes from `runtime._step()`
2) No longer emitting project state changes from `vm.clear()`
3) Being more specific about when blocks events emit project changes (e.g. we check if an actual change took place in the VM rather than just upon receiving an event). This doesn't resolve all of our issues since we have some scenarios in which switching sprites still causes a project change because we're effectively deleting a block and recreating that block in the VM. This is a known issue we hope to address in a future PR.
4) As a side effect of removing emitting project changes from `runtime._step`, we now specifically track project changes that were previously being emitted as a side effect (e.g. adding a sprite, adding a costume, etc. -- all the changes in `src/virtual-machine.js`)
5) changing `renameSprite()` to only emit a project change/targets update if an actual rename occurred.

### Test Coverage

Added lots of unit tests and an integration test. Also did lots of manual testing!
We have documented our investigation and manual testing in the following doc. We have also documented some known issues in the doc.

https://docs.google.com/document/d/1DKM4rf-ajioOrAeipsgtuIluJNMIo71L3MfgB84y3JE/edit

Changes can be tested at https://llk.github.io/scratch-gui/project-state-changes


src/engine/blocks.js
```diff
@@ -506,6 +511,16 @@ class Blocks {
         this._cache._monitored = null;
     }
 
+    /**
+     * Emit a project changed event if this is a block container
+     * that can affect the project state.
+     */
+    emitProjectChanged () {
+        if (!this.forceNoGlow) {
```
rschamp, line 519: I'm wondering if we want to take this time to rename that option, or create a separate, explicit option related to project state here?

src/engine/blocks.js
```diff
@@ -564,7 +582,8 @@ class Blocks {
                 // Changing the value in a dropdown
                 block.fields[args.name].value = args.value;
 
-                if (!optRuntime){
+                if (!this.runtime){
+                    log.warn('Runtime is not optional, it should get passed in when the block container is created.');
```
rschamp: I think this check should be farther up (and fatal) or can just be removed altogether. We assume it exists above... and all over the place elsewhere.

src/engine/blocks.js
```diff
@@ -592,7 +611,8 @@ class Blocks {
             block.mutation = mutationAdapter(args.value);
             break;
         case 'checkbox': {
-            if (!optRuntime) {
+            if (!this.runtime) {
+                log.warn('Runtime is not optional, it should get passed in when the block container is created.');
```
rschamp: Remove?

src/engine/blocks.js
```diff
@@ -663,6 +683,10 @@ class Blocks {
         }
         }
 
+        // TODO maybe track actual changes,
```
rschamp: Is this super hard to figure out in this pass?

src/virtual-machine.js
```diff
@@ -559,10 +559,12 @@ class VirtualMachine extends EventEmitter {
...
                 const projectVersion = validatedInput[0].projectVersion;
                 if (projectVersion === 2) {
-                    return this._addSprite2(validatedInput[0], validatedInput[1]);
+                    return this._addSprite2(validatedInput[0], validatedInput[1])
+                        .then(() => this.runtime.emitProjectChanged());
                 }
                 if (projectVersion === 3) {
-                    return this._addSprite3(validatedInput[0], validatedInput[1]);
+                    return this._addSprite3(validatedInput[0], validatedInput[1])
+                        .then(() => this.runtime.emitProjectChanged());
```
rschamp: Since we reject in case of neither of these, can't we put this in the next `then()`? I.e. chained to `.then(validatedInput =>...).then(/* here? */)`?

src/virtual-machine.js
```diff
@@ -94,8 +94,8 @@ class VirtualMachine extends EventEmitter {
         this.runtime.on(Runtime.VISUAL_REPORT, visualReport => {
             this.emit(Runtime.VISUAL_REPORT, visualReport);
         });
-        this.runtime.on(Runtime.TARGETS_UPDATE, () => {
-            this.emitTargetsUpdate();
+        this.runtime.on(Runtime.TARGETS_UPDATE, emitProjectChanged => {
+            this.emitTargetsUpdate(emitProjectChanged);
```
rschamp, line 98: Is this ever true?

rschamp (changes requested):
Overall this looks great, thanks for all the tests!  I don't think this necessarily has to change before merging, there were just more than a few questions/suggestions.

kchadha:
Towards resolving LLK/scratch-gui#4275

src/engine/blocks.js
kchadha: @rschamp, I think you're right. We had this check from before updating tests to resolve #1942 and this can probably be updated now.

src/virtual-machine.js
kchadha: Right now it isn't, but I don't know whether it's better to just assume always false here...

In a future pass we plan to completely separate out targets update and project changed, so we can fix this there then too. I'll file an actual issue for it.

kchadha: Filed in https://github.com/LLK/scratch-vm/issues/1964

src/engine/blocks.js
kchadha: Done!

src/engine/blocks.js
kchadha: Done!

src/engine/blocks.js
kchadha: Ah I had actually put this in here as a note to myself to see whether we needed to track actual changes or if this was good enough. I think in general it is usually rare that we would receive a `change` event from scratch-blocks where a change did not occur. This is unlike the other cases (create, move, delete) where we get the respective events from scratch-blocks but effectively no change occurs in the VM.

I don't *think* it's possible to get a change event for a block without some user interaction with the block. @picklesrus, I would be interested to know, can you think of scenarios where this would happen?

I think I'll probably remove this TODO comment.

src/virtual-machine.js
kchadha: Yep, that makes sense! Good catch!

kchadha: Done!

rschamp (approved):

PR #1967, opened 2019-02-03
GDXFOR sensor array refactor

### LINKED ISSUE #1953: Refactor sensor state in Vernier Force and Acceleration extension
In our other hardware extensions,  the `_peripheral._sensors` object is initialized with default sensor values at extension load. Data from the hardware directly updates these values. The block opcodes can then simply use `_peripheral`'s getters, which are always safe (i.e. whether or not the hardware is connected). 

In the vernier extension, we currently have a more complex setup: `_peripheral` has a `_device` created by the Vernier godirect module, which has internal state tracking the sensor values. The `_device` object only exists after the hardware has been connected. This situation necessitated adding both a boolean _sensorsEnabled and a function _canReadSensors(), which is used in every sensor value getter and requires the “default” state of sensor values to be distributed throughout the code. This setup seems complex and error prone.

I think we should change it to be more like our other extensions. The `_peripheral` should have a `_sensors` object maintaining the sensor values, which listens for sensor update events from the hardware. The `_sensors` object just needs to be reset to defaults on load and on disconnect.

evhan55:
Resolves #1953: Refactor sensor state in Vernier Force and Acceleration extension

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -19,7 +19,7 @@ const Scratch3Ev3Blocks = require('../extensions/scratch3_ev3');
 const Scratch3MakeyMakeyBlocks = require('../extensions/scratch3_makeymakey');
 // todo: only load this extension once we have a compatible way to load its
 // Vernier module dependency.
-// const Scratch3GdxForBlocks = require('../extensions/scratch3_gdx_for');
+const Scratch3GdxForBlocks = require('../extensions/scratch3_gdx_for');
 
 const builtinExtensions = {
     pen: Scratch3PenBlocks,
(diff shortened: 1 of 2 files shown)


evhan55:
@ericrosenbaum: Submitting this PR as a first pass / WIP.  One of the files has the extension manager change needed for this to launch, which won't be in the final version of the PR.

In the meantime, can you check this branch and see if the values all look like they are coming out ok for you in the blocks?

If so, then I can continue the refactor and clean up the spin speed calculations out into a single helper function.

ericrosenbaum:
Looks great so far! I tested all the sensor values and they're working.

evhan55:
Note: I also need to clear the values on disconnect, which might add a small complication to this PR, maybe making it a bigger change than intended:  The GDXFOR extension should follow the same, recently-added 'disconnect' behavior in the other extensions

evhan55:
@ericrosenbaum I made a few more changes:

- separated out an `_onSensorValueChanged`
- separated out an `_spinSpeedFromGyro` helper function
- cleared the `this._sensors` array on disconnect, I hope this is working correctly

I am not 100% confident I didn't lose something important along the way, so would be good to check all the sensor values again.  Also, all the different changes might be more easily seen directly in the file.

ericrosenbaum (approved):
Looks great! Tested everything and it's still working well.

PR #1969, opened 2019-02-04
Benchmark more data

mzgoddard:
### Resolves

More data to take loading performance further.

### Proposed Changes

Depends on #1944 

- Display time to load just the (json) data
- Display progress and time for content
- Display progress and time hydrating the downloaded assets
- Start loading the target project immediately instead of waiting for the onload event
- Display memory used

### Reason for Changes

- Display time to load just the (json) data
  Knowing how long it takes to fetch the data also tells us about when the content begins to download. Getting these steps to occur earlier and with smaller gaps inbetween work done will mean a faster load time.
- Display progress and time for content
  Once the raw data (content) has loaded we know when turning that data into assets will not be blocked by not having all the project's data yet.
- Display progress and time hydrating the downloaded assets
  Turning data into assets is the most time consuming step of loading a project. It is also the hardest step to optimize. This final time is the previous time value shown and our final measure of how well loading performance is doing.
- Start loading the target project immediately instead of waiting for the onload event
  We want scratch and any project being loaded to be as fast as possible, so we want to make loading projects in the benchmark as fast as possible to help show what we can carry over to scratch-gui if it isn't in another repo's respective code.
- Display memory used
  This is a little related to loading performance. Using less memory also means less garbage collection. As well this will help reduce the memory used to put less strain on systems with less resources.

### Benchmark data

There is no new data with this change. I might revisit and break down the times with the new measures.

### DIFF
--- a/src/playground/benchmark.js
+++ b/src/playground/benchmark.js
@@ -725,7 +725,7 @@ const renderBenchmarkData = function (json) {
     setShareLink(json);
 };
 
-window.onload = function () {
+const onload = function () {
     if (location.hash.substring(1).startsWith('view')) {
         document.body.className = 'render';
         const data = location.hash.substring(6);
@@ -747,3 +747,5 @@ if (window.performance) {
 }
 
 window.ScratchVMEvalEnd = Date.now();
+
+onload();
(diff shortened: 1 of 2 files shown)


cwillisf (approved):
LGTM!

kchadha (approved):
LGTM though there are some merge conflicts that need to be resolved

mzgoddard:
- Rebase on develop

PR #1970, opened 2019-02-04
Use Promise.all, createImageBitmap, and one canvas in `load-costume.js`

mzgoddard:
### Resolves

Project loading performance.

### Proposed Changes

Depends on #1944 

- Use one canvas (if possible) for merging bitmaps and their text layer
- Use Promise.all to load the bitmap and text layer
- Use createImageBitmap if available

### Reason for Changes

- Use one canvas (if possible) for merging bitmaps and their text layer
  Reduce memory allocations when possible. Creating multiple canvas elements and their underlying bitmap memory region uses a lot of memory. Reducing these allocations will reduce the memory used as well as time spent both allocating them and garbage collecting them.

  Conveniently how `load-costume.js` currently uses Promises means this new code will only ever need one Canvas. Promises when executing their callbacks will call the next promise in the chain immediately if possible. As fetchBitmapCanvas_ and loadBitmap_ do no async work after loading the image the canvas will be drawn to, resized, and turned into a BitmapSkin. Then the Canvas is free for the next bitmap costume.
- Use Promise.all to load the bitmap and text layer
  Loading the bitmap and the text layer here was using some common code plus some work that we can rely on a common helper for. We can turn the boolean that waits for both images into a Promise.all call that will wait for the bitmap and text layer.
- Use createImageBitmap if available
  createImageBitmap creates an ImageBitmap which, as we need, can be drawn to a canvas. createImageBitmap takes Blobs as one of its possible inputs. We can use these facts to use createImageBitmap on systems that support it, and call it with a blob of the Asset's data. It takes less time to create a Blob from an arraybuffer than to encode that arraybuffer as base64. This lets us save time by creating a blob instead or a datauri.

### Benchmark Data

Pending ...

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -74,86 +74,70 @@ const fetchBitmapCanvas_ = function (costume, runtime, rotationCenter) {
         return Promise.reject('No V2 Bitmap adapter present.');
     }
 
-    return new Promise((resolve, reject) => {
-        const baseImageElement = new Image();
-        let textImageElement;
+    return Promise.all([costume.asset, costume.textLayerAsset].map(asset => {
+        if (!asset) {
+            return null;
+        }
 
-        // We need to wait for 2 images total to load. loadedOne will be true when one
-        // is done, and we are just waiting for one more.
-        let loadedOne = false;
+        if (typeof createImageBitmap !== 'undefined') {
+            return createImageBitmap(
+                new Blob([asset.data], {type: asset.assetType.contentType})
+            );
+        }
 
-        const onError = function () {
-            // eslint-disable-next-line no-use-before-define
-            removeEventListeners();
-            reject('Costume load failed. Asset could not be read.');
-        };
-        const onLoad = function () {
-            if (loadedOne) {
-                // eslint-disable-next-line no-use-before-define
-                removeEventListeners();
-                resolve([baseImageElement, textImageElement]);
-            } else {
-                loadedOne = true;
-            }
-        };
+        return new Promise((resolve, reject) => {
+            const image = new Image();
+            image.onload = function () {
+                resolve(image);
+                image.onload = null;
+                image.onerror = null;
+            };
+            image.onerror = function () {
+                reject('Costume load failed. Asset could not be read.');
+                image.onload = null;
+                image.onerror = null;
+            };
+            image.src = asset.encodeDataURI();
+        });
+    }))
+        .then(([baseImageElement, textImageElement]) => {
+            const mergeCanvas = getCanvas();
 
-        const removeEventListeners = function () {
-            baseImageElement.removeEventListener('error', onError);
-            baseImageElement.removeEventListener('load', onLoad);
+            const scale = costume.bitmapResolution === 1 ? 2 : 1;
+            mergeCanvas.width = baseImageElement.width;
+            mergeCanvas.height = baseImageElement.height;
+
+            const ctx = mergeCanvas.getContext('2d');
+            ctx.drawImage(baseImageElement, 0, 0);
             if (textImageElement) {
-                textImageElement.removeEventListener('error', onError);
-                textImageElement.removeEventListener('load', onLoad);
+                ctx.drawImage(textImageElement, 0, 0);
             }
-        };
+            let canvas = mergeCanvas;
+            if (scale !== 1) {
+                canvas = runtime.v2BitmapAdapter.resize(mergeCanvas, canvas.width * scale, canvas.height * scale);
+            }
 
-        baseImageElement.addEventListener('load', onLoad);
-        baseImageElement.addEventListener('error', onError);
-        if (costume.textLayerAsset) {
-            textImageElement = new Image();
-            textImageElement.addEventListener('load', onLoad);
-            textImageElement.addEventListener('error', onError);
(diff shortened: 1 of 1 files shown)


src/import/load-costume.js
```diff
@@ -32,6 +32,53 @@ const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
...
+        create () {
+            return this.pool.pop() || document.createElement('canvas');
+        }
+
+        /**
+         * Release the canvas to be reused.
+         * @param {HTMLCanvasElement} A canvas element.
+         */
+        release (canvas) {
+            this.clear();
```
cwillisf, line 74: Do you mean to call `clear()` here? This means that the canvases that enter the pool will last at most one JS tick. I guess that would help with thrashing during one tick but it seems like letting them last for several ticks would be even better...? Or is the idea that you're just trying to avoid one JS tick creating many canvas elements that will all be eligible for GC by the end of the tick?

mzgoddard: I agree holding canvases for several ticks may be better for mixing loading performance and memory used as a simple strategy.

I decided to simplify this Pool class where I could since I don't intend this specific implementation to be used outside of this file. Really I only wanted this class to handle the general improvement of reusing canvases and some safety if I'm wrong about one fact of this file.

As is, load-costume will only have 0 or 1 canvases in this pool. As the promise chain is written it will "create", use, and release the canvas synchronously every time without leaving the javascript context. That is to say the promise chain all resolves after the load event of the images and browsers will exhaust those resolve jobs before returning to the native context to possibly emit another load event. (Of course this could change from a change to the file or other change, or if I'm just wrong. In those cases we have the safety of the list of canvases hear.)

-----

I haven't spent enough time thinking on architecture to better solve this. It's clear to me that we need a large behaviour managing this. But there is a bunch of needs. I think I want to propose two things a TaskLoop and ReferenceManager on VM or something to manage large processing jobs or large memory resources.

mzgoddard: *Facepalm*

Crud pretty sure what I just wrote out means this version of the canvas pool stand alone doesn't help. I've done testing on a bunch of changes standalone and others grouped with other changes and I can't keep track of what things depend on each other anymore clearly.

...

Oh wait. No thats not right. I already thought of this.

https://github.com/LLK/scratch-vm/pull/1970/files#diff-a63caa2e62f9dd4327204d2dd61b4b3fR53

The pool is cleaned up after 1 second after a canvas is released when no work to clear the pool is queued.

src/import/load-costume.js
```diff
@@ -54,123 +101,114 @@ const fetchBitmapCanvas_ = function (costume, runtime, rotationCenter) {
...
-            baseImageElement.removeEventListener('error', onError);
-            baseImageElement.removeEventListener('load', onLoad);
+            const ctx = mergeCanvas.getContext('2d');
+            ctx.drawImage(baseImageElement, 0, 0);
             if (textImageElement) {
-                textImageElement.removeEventListener('error', onError);
-                textImageElement.removeEventListener('load', onLoad);
+                ctx.drawImage(textImageElement, 0, 0);
+            }
+            let canvas = mergeCanvas;
```
kchadha, line 148: Why do we need this assignment (also why are both `canvas` and `mergeCanvas` being returned below)?

mzgoddard: Because of the line, two lines down.

```js
canvas = runtime.v2BitmapAdapter.resize(mergeCanvas, canvas.width * scale, canvas.height * scale);
```

v2BitmapAdapter returns a different canvas than the one passed in. So we cannot know the `mergeCanvas` value once `canvas` is overwritten if we only passed `canvas` to the next section.

```js
canvasPool.release(mergeCanvas);
```

In other words `canvas` is not guaranteed to be `mergeCanvas`. `mergeCanvas` is the one we want to return to the resource pool.

mzgoddard:
- Rebase on develop

src/import/load-costume.js
kchadha: Why does it matter which canvas we return to the resource pool?

If we never had this line, and then instead in the conditional two lines down, we had:

```mergeCanvas = runtime.v2BitmapAdapter.resize(mergeCanvas, mergeCanvas.width * scale, mergeCanvas.height * scale);```

Wouldn't the old canvas get garbage collected?

mzgoddard: I figured I'd use ownership semantics here. I am making the assumptions mergeCanvas is owned by this file and functions and that the canvas returned by resize is owned by resize. Since resize borrows mergeCanvas and this file borrows resize's canvas, we should only hold on to mergeCanvas. Since we don't own resize's canvas, we shouldn't retain it and later transform it by merging bitmaps on it.

The point is to keep canvases from being garbage collected. Under this formality we can only keep the mergeCanvas.

kchadha: That makes sense to me, thanks for explaining. Do you mind adding a comment to the code so that this is called out explicitly as I imagine this question would come up again and the code could get rewritten in a way that isn't cohesive with this thought process.

mzgoddard: Sure thing.

mzgoddard:
- Rebase on develop
- Add comment for `let canvas = mergeCanvas;`

kchadha (approved):
LGTM! Thanks for the update.

PR #1972, opened 2019-02-04
Sound bank over sprite

mzgoddard:
### Resolves

Load sounds as early as possible.

### Proposed Changes

Depends on #1947 

- Kick off sound loading like costumes

### Reason for Changes

Like costumes, starting the requests for sounds as early as possibly by making those requests before we load data can allow that data to download while other work is done.

### Benchmark Data

Pending ...


kchadha (approved):
LGTM

PR #1973, opened 2019-02-05
Fix variable characters

kchadha:
### Resolves

Work towards #1077.

### Proposed Changes

Thanks to @picklesrus for pairing on this!

Fix issues with importing projects from scratch 2 (or even scratch 3 projects that were originally from 2.0) that have variables or lists with special characters in their names. When importing a 2.0 project, we were adding the user defined variable name (which could include special xml-unsafe chars) to the variable id. This was preventing blocks referencing variables with these special characters from showing up in the project (or in the toolbox).

Known Issues:
- broadcast messages with special characters (imported from 2.0 or from 2 -> 3)
- custom procedures ""
- Backspace control character (and other control characters)

### Test Coverage

Added tests!


src/util/string-util.js
```diff
@@ -57,6 +57,25 @@ class StringUtil {
             return value;
         });
     }
+    /**
+     * A function to replace unsafe characters (not allowed in XML) with safe ones. This is used
+     * in cases where we're replacing non-user facing strings (e.g. variable IDs).
+     * When replacing user facing strings, the xmlEscpe utility function should be used
```
towerofnix: This should be `xmlEscape`, right?

kchadha: good catch!

```suggestion
     * When replacing user facing strings, the xmlEscape utility function should be used
```

PR #1978, opened 2019-02-05
Vernier gdxfor extension: add spacers and rename "spin"

ericrosenbaum:


ericrosenbaum:
@evhan55 I realized that because this extension is on develop, translators have already started working on it. If we want to make sure they use the new string on a block, we have to update the id (which I just did).

evhan55:
@ericrosenbaum Sounds good, thanks.

evhan55:
This looks really nice to me!  Simple and powerful at the same time.

PR #1981, opened 2019-02-06
update text-encoding to latest version

mzgoddard:
### Resolves

Faster project loading and less memory usage.

### Proposed Changes

Update to latest text-encoding.

### Reason for Changes

scratch-gui depends on text-encoding@0.7.0. If scratch-vm depends on 0.6.4, gui has to build an extra copy of text-encoding, one for gui (and other modules using 0.7.0, like scratch-storage) and another for scratch-vm, and one for scratch-sb1-converter. This increases the size of the JS build and if text-encoding is used on a system that needs it, it will be evaluated twice and split up the work decreasing the change the JS VM optimizes the text-encoding implementation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12222,13 +12222,6 @@
         "js-md5": "0.7.3",
         "minilog": "3.1.0",
         "text-encoding": "^0.7.0"
-      },
-      "dependencies": {
-        "text-encoding": {
-          "version": "0.7.0",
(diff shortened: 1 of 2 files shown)


paulkaplan (approved):
LG! good catch on the different versions

kchadha:
@mzgoddard, do you mind rebasing this PR off of develop and updating the package-lock.json?

PR #1982, opened 2019-02-06
GDX-FOR better freefall algorithm and different tilt options

stocktonkincade:
### Proposed Changes

For the Go Direct Force (GDX-FOR) extension:
* New freefal algorithm
* Simpler options for tilt.

### Reason for Changes

* A freefall algorithm that better accounts for rotation while the device is in freefall
* Tilt for x/y was unstable when attempting to use the +/-180 range. Changed from x/y to front/back and left/right and clamp the range to +/-90. Similar to micro:bit implementation

### Test Coverage

* Check that the freefall "started falling" hatblock and falling boolean behave
* Check that the tilt reporter shows and reports new options and ranges

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -275,29 +275,15 @@ class GdxFor {
         return this._sensors.force;
     }
 
-    getTiltX () {
-        let x = this.getAccelerationX();
-        let y = this.getAccelerationY();
-        let z = this.getAccelerationZ();
-
-        let xSign = 1;
-        let ySign = 1;
-        let zSign = 1;
-
-        if (x < 0.0) {
-            x *= -1.0; xSign = -1;
-        }
-        if (y < 0.0) {
-            y *= -1.0; ySign = -1;
-        }
-        if (z < 0.0) {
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
This is looking great so far- thanks @stocktonkincade!

Free fall detection worked nicely for me when I tossed the sensor, even when the sensor was spinning. 

@khanning @thisandagain I'm curious if you have any thoughts on this approach to removing the effect of spin from free fall detection. 

One small thing to change: TILT_MENU now needs localization (we don't localize x and y, but we do localize front/back/left/right).

evhan55:
The changes in code all look good to me so far!

I haven't tested the new behavior on the device yet but I can test that as soon as I am near my device again.

stocktonkincade:
Good catch, @ericrosenbaum. I added the localization to the tilt menu items.

evhan55:
I just tested with the device:

- `falling?` now feels smoother and more predictable as I tossed the device around
- `tilt front/back/right/left` feels much easier to understand for me, now, and the numbers coming back felt predictable as well

evhan55:
All looks good to me 👍

ericrosenbaum (approved):
LGTM!

PR #1984, opened 2019-02-07
Vernier gdxfor input rate change

### LINKED ISSUE #1976: Vernier gdxfor sensor should update every 100ms
- Sensor update interval should be set to 100ms (not faster)
- Move that value into a constant

evhan55:
### Resolves

#1976: Vernier gdxfor sensor should update every 100ms


ericrosenbaum (approved):
LG! I tested all the sensors with my "test all the sensors" project, and it has about the same responsiveness. The one real change I've noticed is that the graphs are very slightly steppier, which makes sense.

PR #1986, opened 2019-02-07
Vernier gdxfor "falling?" report false when not connected

### LINKED ISSUE #1977: Vernier gdxfor extension "falling?" reports true while hardware disconnected
### Expected Behavior

While the Vernier Force and Accel sensor is not connected, the reporter blocks should all report reasonable default values.

### Actual Behavior

The `<falling?>` reporter reports true, while the sensor is not connected.

### Steps to Reproduce

- Load the extension, without connecting to the sensor.
- Clicking the falling? reporter, and it reports true.

evhan55:
### Resolves

Resolves #1977: Vernier gdxfor extension "falling?" reports true while hardware disconnected.

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -822,6 +822,13 @@ class Scratch3GdxForBlocks {
     }
 
     isFreeFalling () {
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM!

PR #1990, opened 2019-02-07
Vernier gdxfor tilt values 0 when disconnected

### LINKED ISSUE #1988: Vernier gdxfor tilt values should be 0 when disconnected
### Expected Behavior

Tilt reporter should report 0 when the device is disconnected.

### Actual Behavior

While disconnected, the tilt reporter reports -90 for "front", 90 for "back", etc.

### Steps to Reproduce

- Add the extension but do not connect to a device
- Click the "tilt" reporter

evhan55:
### Resolves

Resolves #1988: Vernier gdxfor tilt values should be 0 when disconnected

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -775,6 +775,13 @@ class Scratch3GdxForBlocks {
     }
 
     getTilt (args) {
+        // When the peripheral is not connected, the acceleration magnitude
+        // is 0 instead of ~9.8, which ends up registering a tilt value; so
+        // we need to return 0 here to prevent returning a tilt value.
+        if (!this._peripheral.isConnected()) {
+            return 0;
+        }
+
         switch (args.TILT) {
         case TiltAxisValues.FRONT:
             return Math.round(this._peripheral.getTiltFrontBack(false));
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
Thanks! Looks like the comment needs fixing

src/extensions/scratch3_gdx_for/index.js
```diff
@@ -745,6 +745,13 @@ class Scratch3GdxForBlocks {
     }
 
     getTilt (args) {
+        // When the peripheral is not connected, the acceleration magnitude
+        // is 0 instead of ~9.8, which ends up registering a tilt value; so
```
evhan55: @ericrosenbaum oh I thought you said the acceleration being 0 ends up affecting the tilt, right? Maybe I misunderstood

ericrosenbaum: How about "Tilt values are calculated using acceleration due to gravity, so we need to return 0 when the peripheral is not connected."

ericrosenbaum: Good to merge with that fix to the comment- thanks!

ericrosenbaum (approved):
LGTM!

PR #1993, opened 2019-02-08
handle sb2 files zipped in a folder

chrisgarrity:
NOTE: depends on https://github.com/LLK/scratch-parser/pull/51
### Resolves

This is the VM half of fixing https://github.com/LLK/scratch-parser/issues/42

### Proposed Changes
If the asset can’t be found by name, look for the same asset in a folder within the zipfile.

### Reason for Changes

sb2 supported two formats of zip archive:
- a flat list of files
- a folder containing the files

### Test Coverage
- added test for zip file that has project files within nested folder `default`


test/fixtures/readProjectFile.js
```diff
@@ -8,5 +8,9 @@ module.exports = {
     extractProjectJson: function (path) {
         const zip = new AdmZip(path);
         return JSON.parse(zip.readAsText('project.json', 'utf8'));
+    },
+    extractNestedProjectJson: function (path, folder) {
```
kchadha: I think it would be better for this logic to go inside `extractProjectJson` so that we don't have to think about how the fixture was created when writing a test.

I think instead of taking the folder name, it should do something similar to the code changes in this PR and in scratch-parser where it looks for the `project.json` in a folder if it fails to find it at the top level.

There's an API function listed here (`getEntryChildren`) that might be helpful:

https://github.com/cthackers/adm-zip/wiki/ZipFile

kchadha (dismissed):
Requesting a change to the test helper functions so that there is only one helper API for extracting the project.json

test/fixtures/readProjectFile.js
chrisgarrity: I used `getEntries` instead so that it didn't need to know the folder name.

kchadha (approved):
LGTM! Thanks for updating that test helper!

kchadha:
@chrisgarrity, could you update the scratch-parser version in this PR to pull in the related changes?

PR #1994, opened 2019-02-08
Fix WeDo2 'motor on for' duration

### LINKED ISSUE #1616: WeDo2: 'turn motor for 1 sec block' acts strange in forever loop
### Expected Behavior

`forever > turn motor for 1 sec` should consistently turn the motor on and off every second

### Actual Behavior

`forever > turn motor for 1 sec` sometimes turns the motor for more than a second (e.g. 3 seconds before it stops and restarts)

evhan55:
### Resolves

Resolves #1616: WeDo2: 'turn motor for 1 sec block' acts strange in forever loop

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1290,8 +1290,9 @@ class Scratch3WeDo2Blocks {
                 }
             });
 
-            // Run for some time even when no motor is connected
-            setTimeout(resolve, durationMS);
+            // Run for some time even when no motor is connected, include
+            // the time needed to brake the motor
+            setTimeout(resolve, durationMS + WeDo2Motor.BRAKE_TIME_MS);
         });
     }
 


ericrosenbaum:
I believe this is not an appropriate fix to the problem, because it adds one second of yield time to the block. So for example, if you run "turn motor on for 1 sec", the block yields for 2 seconds (or if you turn it for 0.1 sec, the block yields for 1.1 sec, etc.). So it would break the timing behavior of existing projects. 

The bug itself as reported seems pretty marginal- the only symptom of the bug is that the motor does not (very) briefly pause between executions of the block. If you put pretty much anything else into the forever loop after it, it works fine. I'm not convinced that it is something that we need to fix, because it doesn't really break any functionality. Also, I would guess the resolution of the problem is a bit more subtle, because it has to do with a race condition. We could talk more about exactly what causes it (which would take some investigation- maybe the brake command that is starting to be sent at the end of a "turn motor on for" opcode" is over-ridden by the new motor on command that is sent immediately after, when we're in that forever loop? in which cause what we should do is actually subtract a very small amount of time from the duration of `motor.turnOnFor`, to account for the BLE communication timing, but leave the duration of the `motorOnFor` opcode as is...).

evhan55:
@ericrosenbaum Ooh I see, thanks for investigating and explaining! I guess that makes sense and I am not worried either way. We can close as 'won't fix' ?

ericrosenbaum:
I think that makes sense- thanks!

evhan55:
Not fixing

PR #1995, opened 2019-02-11
Support isDiscrete flag on monitors

### LINKED ISSUE #1971: Support `isDiscrete` flag on slider monitors
Support `isDiscrete` flag on slider monitors
- import `isDiscrete` from sb2 projects
- serialize and import `isDiscrete` flag from sb3 projects
- add `isDiscrete` flag to the [monitor record](https://github.com/LLK/scratch-vm/blob/3dd0d0a657917ff6479c258213a201dbe757754b/src/engine/monitor-record.js) and have it sent to the GUI

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1971

### Proposed Changes

Adds the `isDiscrete` flag on the MonitorRecord object, and adds support for serializing/deserializing from both sb2 and sb3.

### Test Coverage

Updated the tests, including the following cases:
- No `isDiscrete` property listed in sb3 file => results in `True`, which is the implicit value right now.
- isDiscrete set to false in sb2 import
- isDiscrete set to false in sb3 import, saving to sb3 and then re-importing it

Note that the first commit of this PR just updates a couple width/height values in the tests. This is because I couldn't open the `monitors.sb2` file and re-save it without those changes, and I wasn't able to modify the JSON directly (when I re-zipped / re-named to sb2 it would fail to parse in the tests?) I wasn't sure how to handle this problem so I just made the first commit what I needed to change in order to just open/save that file. The next commit is real changes. @kchadha is there a better way to manage that?


kchadha:
@paulkaplan, you may have run into LLK/scratch-parser#42 when you were trying to re-zip the sb2.

It should be resolved now (thanks @chrisgarrity)!

kchadha (approved):
LGTM!

PR #1998, opened 2019-02-13
Task queue enhancements

### LINKED ISSUE #1807: Task queue tests are flaky

cwillisf:
### Resolves

Resolves #1807

### Proposed Changes

Several improvements to `TaskQueue`, as well as several supporting changes:

- Fixed an issue where `TaskQueue`'s `cancelAll()` could cause a "cannot read property of undefined" exception.
- Reduced the number of `.then(...)` and `setTimeout(...)` steps involved in running a task through `TaskQueue`, simplifying the code and possibly improving performance.
- `TaskQueue` now supports specifying a maximum total task cost. Attempting to add a task exceeding this cost will return a rejected promise.
- `TaskQueue` now bases all timing and scheduling on a replaceable `Timer` instance.
- New `MockTimer` acts like a `Timer` but allows programmatic control over its time value.
- `TaskQueue` tests are now based on `MockTimer`, resolving #1807.

### Reason for Changes

The `TaskQueue` should now be ready to use for extension throttling.

### Test Coverage

- `TaskQueue` tests are much more extensive
- A few tests have been added for `Timer`
- A new set of tests have been added for `MockTimer`


src/util/task-queue.js
```diff
@@ -12,18 +12,33 @@ class TaskQueue {
...
+     * @property {number} maxTotalCost - reject a task if total queue cost would pass this limit (default: no limit).
      * @memberof TaskQueue
      */
-    constructor (maxTokens, refillRate, startingTokens = maxTokens) {
+    constructor (maxTokens, refillRate, options = null) {
         this._maxTokens = maxTokens;
         this._refillRate = refillRate;
         this._pendingTaskRecords = [];
-        this._tokenCount = startingTokens;
+        this._tokenCount = options.hasOwnProperty('startingTokens') ? options.startingTokens : maxTokens;
```
evhan55, line 24: This throws an exception if `options` is null:

```
task-queue.js:24 Uncaught TypeError: Cannot read property 'hasOwnProperty' of null
    at new TaskQueue (task-queue.js:24)
```

src/util/task-queue.js
```diff
@@ -12,18 +12,33 @@ class TaskQueue {
...
+        this._maxTotalCost = options.hasOwnProperty('maxTotalCost') ? options.maxTotalCost : Infinity;
```
evhan55, line 25: This throws an exception if `options` is null, similar as above.

src/util/task-queue.js
```diff
@@ -35,38 +50,57 @@ class TaskQueue {
...
+            if (currentTotalCost + cost > this._maxTotalCost) {
+                return Promise.reject('Maximum total cost exceeded');
+            }
+        }
+        const newRecord = {
+            cost
+        };
+        newRecord.promise = new Promise((resolve, reject) => {
+            newRecord.cancel = () => {
+                reject(new Error('Task canceled'));
```
evhan55, line 64: typo 'cancelled'

cwillisf: Heh - I was worried about this. It turns out that "canceled" is the preferred US spelling whereas "cancelled" is the preferred British spelling.
See also: https://www.grammarly.com/blog/canceled-vs-cancelled/

evhan55: LOL, lies!

src/util/timer.js
```diff
@@ -90,6 +90,25 @@ class Timer {
...
     }
+
+    /**
+     * Call a handler function after a specified amount of time has elapsed.
+     * @param {function} handler - function to call after the timeout
+     * @param {number} timeout - number of milliseconds to delay before calling the handler
+     * @returns {number} - the ID of the new timeout
+     */
+    setTimeout (handler, timeout) {
+        return global.setTimeout(handler, timeout);
```
mzgoddard, line 101: `global` only exists in Node. Browser environments have `window` and `self`.

You could `typeof setTimeout !== 'undefined'` for it's existence. That'll work in any JS environment.

mzgoddard:
```js
if (typeof setTimeout !== 'undefined') {
    return setTimeout(handler, timeout);
} else {
    // something else
}
```

mzgoddard: I should clarify `global` will work in browsers but only because webpack is polyfilling global as window for us.

cwillisf: Webpack automatically converts `global` into `window` when targeting browsers -- or, rather, Webpack always provides a value for `global` and in the browser case it's generally `window`.

Anyway, `global` is only really here to make it explicit that I'm not making a recursive call. The function would work just as well without `global` or `self` or `window`, it would just look weird(er) at first glance:
```js
setTimeout (handler, timeout) {
        return setTimeout(handler, timeout);
}
```

ericrosenbaum (approved):
Tested with Evelyn's changes to the microbit extension to use this and it's working great!

src/util/timer.js
mzgoddard: > it would just look weird(er) at first glance:

I like this point.

> converts `global` into `window` when targeting browsers

webpack wraps your code in a second closure that depends on [webpack/buildins/global.js](https://github.com/webpack/webpack/blob/master/buildin/global.js). It notices the keyword and adds that wrapper. It also does that for `Buffer`, `process`, and I think some others.

```js
// the wrapping function stored in webpack's module array
function (module, exports, require) {
  (function (global) {
    // your module code here
  }.call(this, require('webpack/buildins/global')));
}
```

webpack is smart enough to not do this if `global` or any of the others are defined in a scope the variable name can reach in the module.

evhan55:
Thumbs up to the PR from me!

Issue #706, opened 2017-10-11
Extensions should receive stop button event


ericrosenbaum:
### Expected Behavior

If a Lego motor is running, you should be able to stop it by pressing the stop button.

Generally, extensions should have a function that gets called on the stop event. 

### Actual Behavior

The stop button is not yet hooked up to extensions.

thisandagain:
@ericrosenbaum My instinct is to not label this as `critical` (blocking release) because the play note blocks now have a limit on them. Do you agree?

ELing9527:
@thisandagain 2025年希望我能知道这个功能是否已经开发完成了，缺少发布这最后一步吗？希望能有幸获悉，它的发布

PR #2002, opened 2019-02-15
Allow primitives to have "cancel" behavior; cancel "play sound until done" on thread retire

### LINKED ISSUE #7: Clicking a “play sound until done” block while it's highlighted should stop the sound

### LINKED ISSUE #40: Behavior on tap running stack
This is something that I think we need to resolve eventually. Right now in Scratch, if you click a running stack, that stack stops. But what happens if it was running a long-running block, like "motor on for 5 secs"? In our current model, the motor would stay on, I think... Perhaps tapping a running stack shouldn't stop it, or we need to think about how to implement stopping it.

towerofnix:
### Resolves

Resolves LLK/scratch-audio#7. Closes #40 (as this PR provides an implementation).

### Proposed Changes

At a high level, this PR allows classes such as `Scratch3SoundBlocks` to specify behavior that "cancels" an actively executing block. These functions are called when you click a script that is already running, and its active block is one of that opcode. The "play sound until done" block makes use of this; if you retire a thread where it is the active block, the sound it started will be stopped.

Internally, instances of `BlockCached` now keep track of their "cancel block" function alongside the main block function. This is referred to when a thread is retired; if present, it's called just like the normal block function.

### Reason for Changes

When a script is executing, it is highlighted with a yellow "glow" bordering its outline. If you click a script while it's executing, it will lose this glow and stop running further blocks in the script. The implication, to the user, is that the script is stopped. Internally, this is what is happening. But, before this PR, "stopping" a thread did not also mean stopping active side-effects. Intuitively, side-effects should be stopped when a script is stopped; otherwise it does not seem so much like you are *stopping* the the script.

This PR doesn't (currently) implement "cancel" blocks into extensions, but it provides the basic code. Once it's implemented, blocks like "play note for beats" and "turn motor on for (N) seconds" can be changed so that their side-effects are cancelled as soon as the thread is retired.

### Test Coverage

I tested this manually. I'm not certain if unit tests are wanted here? I saw that there don't seem to be any for `execute.js` so far, only the higher-level classes (e.g. Thread, Sequencer); this PR doesn't add much behavior to those classes.

### DIFF
--- a/src/blocks/scratch3_sound.js
+++ b/src/blocks/scratch3_sound.js
@@ -143,6 +143,16 @@ class Scratch3SoundBlocks {
         };
     }
 
+    /**
+     * Retrieve the cancel block primitives implemented by this package.
+     * @return {object.<string, FUnction>} Mapping of opcode to Function.
+     */
+    getCancelPrimitives () {
+        return {
+            sound_playuntildone: this.cancelPlaySoundAndWait
+        };
+    }
+
     getMonitored () {
         return {
             sound_volume: {
@@ -162,11 +172,10 @@ class Scratch3SoundBlocks {
     }
 
     _playSound (args, util, storeWaiting) {
-        const index = this._getSoundIndex(args.SOUND_MENU, util);
-        if (index >= 0) {
+        const soundId = this._getSoundId(args.SOUND_MENU, util);
+        if (soundId) {
             const {target} = util;
             const {sprite} = target;
-            const {soundId} = sprite.sounds[index];
             if (sprite.soundBank) {
                 if (storeWaiting === STORE_WAITING) {
                     this._addWaitingSound(target.id, soundId);
@@ -176,6 +185,26 @@ class Scratch3SoundBlocks {
                 return sprite.soundBank.playSound(target, soundId);
             }
         }
+    }
+
+    cancelPlaySoundAndWait (args, util) {
+        const soundId = this._getSoundId(args.SOUND_MENU, util);
+        const {target} = util;
+        const {sprite} = target;
+        if (sprite.soundBank) {
+            sprite.soundBank.stop(target, soundId);
+        }
+    }
+
+    _getSoundId (indexArg, util) {
+        const index = this._getSoundIndex(indexArg, util);
+        if (index >= 0) {
+            const {target} = util;
+            const {sprite} = target;
+            const {soundId} = sprite.sounds[index];
+            return soundId;
+        }
+        return null;
     }
 
     _addWaitingSound (targetId, soundId) {
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -593,6 +593,10 @@ const cancelExecution = function (sequencer, thread) {
     const currentBlockId = thread.peekStack();
     const {blockCached} = getBlockCached(currentBlockId, thread, runtime);
 
+    if (!blockCached) {
+        return;
+    }
+
     // Only the last op is actually a block and not an input; it is the one
     // that may have a cancel implementation.
     const ops = blockCached._ops;
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -210,6 +210,18 @@ class BlockCached {
         this._definedBlockFunction = false;
 
         /**
+         * The block opcode's "cancel" implementation function.
+         * @type {?function}
+         */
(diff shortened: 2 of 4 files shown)


thisandagain:
@kchadha @ericrosenbaum This may be a good candidate for "co-review". :-)

towerofnix:
This also closes #706, since `retireThread` and so also the cancel functions of blocks, get called when the stop sign is clicked.

Fwiw, this doesn't *implement* stopping the motor as discussed in #706 and #40; it only adds the support for that feature. It's probably worth implementing stopping the motor in this PR, but I don't have a lego motor or any external devices to do that.

Implementing motor-stopping into this would mean adding support for this to extensions (since the motor blocks are all part of extensions), which is a good idea anyway since that looks to be the direction Scratch 3.0 is going now.

thisandagain:
@ericrosenbaum @kchadha Bump

ericrosenbaum:
@towerofnix sorry it took us a long time to look at this! @kchadha and I were just looking into it, and we realized that Scratch 1 and 2 also did not stop the sound for "play sound until done" when you click the stack to stop it, so that specific thing feels less important than I originally thought when I filed that issue (https://github.com/LLK/scratch-audio/issues/7). This would be a nice feature though, especially, as you point out, for the extensions that make sound or control motors. We'll investigate the implications of the change a bit further.

ericrosenbaum:
After some more discussion, we decided to close this PR, and revisit in the future if needed for internal design and implementation.

towerofnix:
@ericrosenbaum alright. Since I spent a while on this, even if it's not help-wanted and I didn't *particularly* expect it to get merged, I'd appreciate just a summary of thoughts on this PR shared here?

towerofnix:
Ping @ericrosenbaum?

ericrosenbaum:
Sorry I didn't provide a more detailed response. We appreciate your work on this. Our main concern is the risk of introducing bugs, due to this being a fairly complex feature. We realized that because previous versions of scratch do not have this feature, it's not itself a bug fix but a a new feature. So we need to decide as a group whether we want to add it (which we have decided to hold off on for now). If we do decide to add it in the future, due to the complexity, we'll want to do our own design and implementation of the feature within the team. Your implementation in that case would serve as a useful reference. Hope that helps!

towerofnix:
Alright, I understand! Thanks for the explanation.

ELing9527:
@towerofnix 您好，想请问这个pr当时是在scratch-vm 的那个版本基础上使用的，我可以知道吗？

towerofnix:
@ELing-Lonely Please see my email reply, or read it below:

> Hello,
>
> I created this pull request in February 2019. That is very shortly after Scratch 3.0 was first released to the public (January 2019) and since then, it has been 6 years. I do believe the pull request was based on the version of scratch-vm which, at the time, was current. I have not tested the code in any more recent version of scratch-vm. Unfortunately that means I cannot tell you if it would require only minor adaptations, or require more serious ones.
>
> However, during those six years, Scratch has not changed very much. It looks like the VM file “src/engine/execute.js” is still basically the same as it used to be. Therefore, I expect changes would be pretty minimal. You may even be able to use  [“git cherry-pick”](https://git-scm.com/docs/git-cherry-pick) or [“git rebase”](https://git-scm.com/docs/git-rebase) without much trouble. I have not tried this out myself.
>
> If you are interested, I do recommend trying it yourself. May I ask why you were wondering? Are you creating a modified version of Scratch which you are able to share about?
>
> And it's no bother at all! Thank you for reaching out.

ELing-Lonely:
@towerofnix 
您好

     我并没有创建一个修改版本的Scratch的打算，只是在学习使用Scratch-vm 时发现，停止功能没有停止一切的能力，查看了issues ，发现您在2019年提出的pr,可以解决此问题，但社区开发者好像没有采用，但我觉得可以解决我的问题所以有此一问，在此在次表示感谢。谢谢您的回信。

ELing9527:
@towerofnix
您好
     我并没有创建一个修改版本的Scratch的打算，只是在学习使用Scratch-vm 时发现，停止功能没有停止一切的能力，查看了issues ，发现您在2019年提出的pr,可以解决此问题，但社区开发者好像没有采用，但我觉得可以解决我的问题所以有此一问，在此在次表示感谢。谢谢您的回信。

PR #2008, opened 2019-02-20
Vernier Force & Accel extension updates

ericrosenbaum:
Changes based on design discussions and playtest observations:

- Remove the “moved” gesture
- Add “when tilted” hat and "tilted?" boolean, with "any" option
- Swap tilt front/back and left/right (they were reversed)
- Rename reporters to “tilt angle" and “spin speed”
- Re-order blocks
- Update block and menu icons
- Set sensor update rate to 80ms


evhan55:
The blocks look good to me!
I tested all the blocks individually, they all seemed to work fine on my end (macOS) except for one tiny nit and another tiny observation:
- the force monitor reports a value of `-1` sometimes when the force sensor is not being touched at all but the device is being shaken/moved/tilted around
- the monitor label `Force and Acceleration: force (val)` took me a few seconds to parse as well

evhan55 (approved):

PR #2009, opened 2019-02-20
Deserialize sb3 assets before loading data

mzgoddard:
### Resolves

Load faster.

### Proposed Changes

Depends on #1972

Parse and load costumes and sounds before parsing the rest of the data.

### Reason for Changes

Parsing a project's data and creating the relevant Sprites, Blocks, and other objects takes a fair bit of time. We can load the overall project faster by making the network requests for the necessary data and then parsing the whole data set.

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -823,46 +823,32 @@ const deserializeBlocks = function (blocks) {
 
 
 /**
- * Parse a single "Scratch object" and create all its in-memory VM objects.
+ * Parse the assets of a single "Scratch object" and load them. This
+ * preprocesses objects to support loading the data for those assets over a
+ * network while the objects are further processed into Blocks, Sprites, and a
+ * list of needed Extensions.
  * @param {!object} object From-JSON "Scratch object:" sprite, stage, watcher.
  * @param {!Runtime} runtime Runtime object to load all structures into.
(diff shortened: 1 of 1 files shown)


test/unit/project_changed_state.js
```diff
@@ -48,7 +48,11 @@ test('Adding a sprite (from sprite3) should emit a project changed event', t =>
     vm.addSprite(sprite3).then(() => {
         t.equal(projectChanged, true);
         t.end();
-    });
+    })
+        .catch(e => {
```
kchadha: Why is this change necessary?

kchadha (approved):
Small question in the comments, but otherwise LG!

test/unit/project_changed_state.js
mzgoddard: Hmm. This isn't necessary. I added it to clarify why the test was failing because an error in the chain would keep the above then clause from executing. The lack of an error report would have made it hard to fix. But now that I fixed that issue we don't need this.

Should I remove it?

mzgoddard: Wrote this last week. But its posting was stucking a review queue. I hit the wrong button.

mzgoddard: Gonna remove this catch statement.

kchadha: LG! Thanks!

kchadha (approved):
LGTM!

PR #2011, opened 2019-02-20
Give all backdrops a more equal chance of being randomly picked

### LINKED ISSUE #2010: Uneven probability distribution in switching to a random backdrop
### Expected Behavior

`switch backdrop to [random backdrop v]` has an equal chance of picking all backdrops, except for the current backdrop which should never be picked.

### Actual Behavior
Unless the currently selected backdrop is the penultimate one,
- the last backdrop will never be picked; and
- the backdrop directly after the current backdrop has a higher chance of being picked

This is because the current algorithm picks a random number between 0 and (n - 2) (inclusive). This means that the index of the last costume (n - 1) will never be picked.
https://github.com/LLK/scratch-vm/blob/a86dc5bb58ae66e8234896dfe99f548f59fdb92a/src/blocks/scratch3_looks.js#L410
Next, the algorithm checks if the current costume has the same index as the one that has been picked, and if it has, it increments the selected costume by 1.
https://github.com/LLK/scratch-vm/blob/a86dc5bb58ae66e8234896dfe99f548f59fdb92a/src/blocks/scratch3_looks.js#L411

So if the randomly selected number is the index of the next costume, *or* the index of the current costume, then the next costume will be picked. This means that the probability for that next costume is double the probability for any other costume. Of course, the last costume can never be picked, unless it's being nudged by the incrementation (i.e. the currently selected costume is the penultimate one).

There are two ways to fix this:
- Change the `===` to a `>=`
- Change the `selectedIndex += 1` to a `selectedIndex = numCostumes - 1`

I don't know which one is more readable (they should both be quite clear in my view), but the latter is probably marginally more optimised because it is rarer that the computational work of adding 1 is needed.

### Steps to Reproduce

This test project has the following backdrops (in this order):

1. Blue
2. Yellow
3. Purple

Every tenth of a second, it switches to the Blue backdrop and then switches to a random backdrop. The expectation is that it will flicker between Yellow and Purple. But because we are not in the penultimate position (this would be Yellow), the last backdrop is never picked so we always get Yellow and there is no flickering.

The test project: https://scratch.mit.edu/projects/288178914/

Firefox on Debian Linux.

joker314:
### Resolves

Resolves #2010 

### Proposed Changes

If you attempt to pick the current backdrop in the "set backdrop to [random backdrop]", don't use the next backdrop as a fallback, instead, use the last backdrop.

This removes the unfair advantage given to the next backdrop; and the unfair disadvantage given to the last backdrop.

### Reason for Changes

To eradicate any confusion / make the project behave as users would expect.

This method is chosen over the other one I also brought up in that issue because it is marginally more efficient.

Edit: would a randInt or randFloat utility function be useful, since it could be shared with the random number Scratch block?
### Test Coverage

It's hard to deterministically unit test random phenomenon; though if asked I would be happy to either mock `Math.random`, or run a test 100 times and check if results are at least a little close to what would be expected.

I have instead used **manual testing**: I loaded my test project (https://scratch.mit.edu/projects/288178914/) into this code and it now works (flickering between the two other colours). I changed what the base backdrop was and it always flickered between the two other backdrops.


paulkaplan (dismissed):
Seems like if we are going to make this change we should make it guaranteed to pick a random one instead of trading one bias for another. Could you come up with a utility function for this, something like picking a random number in a range (in this case [0, (numCostumes-1)] inclusive) and excluding a single index (in this case stage.currentCostume)?

joker314:
Hi @paulkaplan ! Thanks for leaving your review. I'm happy to add the changes but I'm not sure I understand your comment

> instead trading one bias for another

Let's say we have 6 costumes, and costume number 3 is selected. This means that there are 5 costumes we are allowed to pick: 1, 2, 4, 5, 6.

We then generate one of 5 things. It doesn't matter what these things are, and commonly this is `Math.floor(Math.random() * n) + 1)`. In actuality, without the `+ 1`, but it's easier to talk about a 1-indexed world. (This PR's particular code is 0-indexed).

If we did nothing, we would have a bias:

The 6th costume doesn't get anything that maps to it, and the 3rd costume is mapped to when it shouldn't be


This is *different* to the bias that this PR is addressing, but I'm just saying that to map {0..5} to {0..6 except 3} requires taking the bias from 3 and moving it to 6, somehow.

My PR does it in this way:

The number 3 is mapped to the 6th costume

Do you think that this way would make it clearer as to what the code is doing/make sure it works?

Each number greater than or equal to 3 is mapped to its neighbour one above

So I suppose my question is: is this purpose of using a utility function in this case to make the code more readable by naming all the components using parameters, or for another purpose? (I just want to make sure I understand correctly).

Thanks again for the review!

joker314:
Awesome @paulkaplan, I've added utility function into `MathUtil`. I've tested with the testing project I made.

I was wondering if you think the massive comment I added is unneeded? Do you think changing from mapping just the banned number to the last number, to instead adding 1 for every number that's greater than or equal to the banned number:

Diagram of adding 1 to every number that is greater than or equal to the banned number

would make everything more intuitive?

paulkaplan:
@joker314 thanks for putting in this work! This is a fun function to write. Your "pick the upper if you picked the excluded" works fine, but yeah I agree it would be more intuitive to just increment the choice by 1 if you picked the excluded (it took me a while to figure out why you can always use the upper, the shift is pretty understandable). I would get rid of the big comment block, the jsdoc above the function is enough. 

Could you add tests for this function? We have good test coverage over utilities like that. You can see how to write these based on the existing tests in `test/unit/util_math.js`.

joker314:
@paulkaplan Thanks for your comments! I've incorporated them into this pull request, as noted below.

> I agree it would be more intuitive to just increment the choice by 1

I just pushed a commit which uses this algorithm instead, as I agree it's easier to understand.

> Could you add tests for this function?

Sure! I overrode `Math.random` so as to allow deterministic tests. I hope this will do.

> I would get rid of the big comment block

Done!

joker314:
@paulkaplan, I think this issue is mislabelled. Could you change it back to "needs review" since I've made some changes since the last review? Thanks!

adroitwhiz:
It looks like the unit tests you added are failing.

adroitwhiz (approved):

adroitwhiz:
@joker314 Thanks for your patience on this issue :smile:

PR #2014, opened 2019-02-22
Fix micro:bit auto-disconnect from setInterval to setTimeout

### LINKED ISSUE #1537: Microbit: `setInterval` should be `setTimeout`
The `setInterval` used for auto-disconnecting should probably be a `setTimeout` instead

evhan55:
### Resolves

- Resolves #1537: Microbit: `setInterval` should be `setTimeout`

### Testing

- Test by removing power from micro:bit and checking that disconnect alert pops up after **4.5 secs**

### DIFF
--- a/src/extensions/scratch3_microbit/index.js
+++ b/src/extensions/scratch3_microbit/index.js
@@ -299,7 +299,7 @@ class MicroBit {
      */
     _onConnect () {
         this._ble.read(BLEUUID.service, BLEUUID.rxChar, true, this._onMessage);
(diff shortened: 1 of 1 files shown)


ericrosenbaum (changes requested):
Looks like this should use `clearTimeout()` in each case, instead of `clearInterval()`

evhan55:
@ericrosenbaum Ah yes, thank you! makes sense

ericrosenbaum (approved):
LGTM!

PR #2018, opened 2019-02-25
Load Vernier force and accel extensions

ericrosenbaum:
Previously, we had commented out the extension manager lines that loaded the Vernier gdxfor extension, due to a problem building GUI into WWW. This was due to an incompatibility with the Vernier GoDirect node module required by the extension which has now been resolved.

@rschamp and I tested this change by deploying to staging and confirming that the extension loads.

We verified that staging loads the editor and can load the extension without problems on:

- [x] Mac Chrome 
- [x] Mac Safari
- [x] Mac Firefox
- [x] Windows Chrome
- [ ] Windows Edge
- [x] Windows Firefox
- [x] iOS safari
- [x] Android Chrome

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -19,7 +19,7 @@ const Scratch3Ev3Blocks = require('../extensions/scratch3_ev3');
 const Scratch3MakeyMakeyBlocks = require('../extensions/scratch3_makeymakey');
 // todo: only load this extension once we have a compatible way to load its
 // Vernier module dependency.
-// const Scratch3GdxForBlocks = require('../extensions/scratch3_gdx_for');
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
The extension does not work on Microsoft Edge. Connection to the devices immediately fails, with an error about `TextDecoder`.  It turns out that TextDecoder is not available on Edge:

https://caniuse.com/#search=TextDecoder

ericrosenbaum:
Blocked on updating the Vernier module version, after this PR lands: https://github.com/VernierST/godirect-js/pull/36

ericrosenbaum:
superseded by https://github.com/LLK/scratch-vm/pull/2069

PR #2024, opened 2019-03-01
Rate limiting for micro:bit using TaskQueue

### LINKED ISSUE #1508: Hardware Extensions: Add TaskQueue to microbit
microbit

evhan55:
### Resolves

- Resolves #1508: Hardware Extensions: Add TaskQueue to microbit

### Proposed Changes

Use the new TaskQueue utility to rate limit the hardware command sent over the Scratch Link socket to the microbit, to prevent trying to send too many commands at once and potentially lock up the peripheral or the browser, etc.

### Test Coverage

Should test stacks that send many commands in a short period of time (e.g. broadcast > when I receive > forever display symbol)


src/extensions/scratch3_microbit/index.js
```diff
@@ -138,6 +134,16 @@ class MicroBit {
          */
         this._busy = false;
 
+        /**
+         * A task queue to track communication tasks sent over the socket.
+         * The bucket in this task queue holds 1 task at a time, and refills
+         * at a rate of 10 tasks per second, from a queue that holds at maximum
+         * 30 tasks.  If more than 30 tasks are added to the task queue in a short
+         * period, some tasks may be rejected (ignored) by the task queue.
+         * @type {TaskQueue}
```
cwillisf: Maybe worth adding something about how the TaskQueue is being used to limit the rate of Bluetooth messages by limiting the rate of execution of blocks which send Bluetooth messages.

src/extensions/scratch3_microbit/index.js
```diff
@@ -138,6 +134,16 @@ class MicroBit {
          */
         this._busy = false;
 
+        /**
+         * A task queue to track communication tasks sent over the socket.
+         * The bucket in this task queue holds 1 task at a time, and refills
+         * at a rate of 10 tasks per second, from a queue that holds at maximum
+         * 30 tasks.  If more than 30 tasks are added to the task queue in a short
```
cwillisf: "from a queue that holds tasks with a maximum total cost of 30. Since most tasks have a cost of 1 this means the queue will generally have at most 30 tasks."

src/extensions/scratch3_microbit/index.js
```diff
@@ -844,14 +840,19 @@ class Scratch3MicroBitBlocks {
...
     displayText (args) {
-        const text = String(args.TEXT).substring(0, 19);
-        if (text.length > 0) this._peripheral.displayText(text);
-        const yieldDelay = 120 * ((6 * text.length) + 6);
-
-        return new Promise(resolve => {
-            setTimeout(() => {
-                resolve();
-            }, yieldDelay);
+        return this._peripheral._queue.do(() => {
```
cwillisf:
```suggestion
        const text = String(args.TEXT).substring(0, 19);
        return this._peripheral._queue.do(() => {
```

evhan55: Note: might be good to check all the block op code routines and push all argument preparation outside of the task queue `do` call.

ericrosenbaum (approved):
LGTM! tested it in a few different ways (forever broadcast triggering a display random, multiple parallel forever stacks, doing that simultaneous with reading sensor values) and everything behaves as expected.

The max total cost of 30 makes it so you can get into a situation where queued messages will send for three seconds, but that seems fine- it's a reasonable compromise, and definitely an edge case (only occurs if you use the broadcast trick to send messages very fast and stop the stack by clicking it rather than using the stop button).

Issue #1743, opened 2018-11-08
Dropped sprite names that are numbers do not work


ericrosenbaum:
### Expected Behavior

It should be possible to drop a variable that reports a number onto an input that expects a sprite name in order to select a sprite with a name that is a number. 

This bug affects any block that accepts a sprite name as an input:
- Go to
- Glide to
- Point towards
- Create clone of
- Distance to
- X position of

For example, here the bananas are a sprite named "3" and dropping blocks reporting the number 3 onto the "distance to" block works correctly in Scratch 2.0:

![screen shot 2018-11-08 at 11 38 05 am](https://user-images.githubusercontent.com/567844/48213008-d545b300-e34a-11e8-83ac-4c495b60e116.png)

I saw this issue while trying out the "chaos game" project, which currently does not work at all in 3.0, because it uses both "point toward" and "distance to" with dropped numeric values to refer to sprites:
https://scratch.mit.edu/projects/243588792/
https://llk.github.io/scratch-gui/develop/#243588792

### Actual Behavior

In 3.0, the same blocks report 10,000:

![screen shot 2018-11-08 at 11 38 15 am](https://user-images.githubusercontent.com/567844/48213050-eee6fa80-e34a-11e8-8015-023df0545330.png)

### Steps to Reproduce

- Try stacks as in screenshots above
- Note that the "chaos game" project does not work: https://llk.github.io/scratch-gui/develop/#243588792

ericrosenbaum:
I'm not sure whether the runtime's getSpriteTargetByName function should do a cast to string itself, or if each block implementation should do the cast. We generally do casting like this in the block implementations, but obviously that is error prone already, where we miss cases (e.g. the current bug). cc @cwillisf

ktbee:
Thought I'd note something interesting I found while repro-ing: when you don't use the "round" block, the Scratch Cat sprite is able to say the correct distance to the sprite with the numeric name. The "round" block might be part of the culprit here. 

![screen shot 2019-02-14 at 12 44 20 pm](https://user-images.githubusercontent.com/7991694/52806168-50a7b680-3056-11e9-9aed-4bd4bd0aa38f.png)

joker314:
@kchadha Yeah, the "round" block returns a JavaScript Number, whereas the raw variable block returns a JavaScript String. We still need to have the distinction because things like the "switch to costume" blocks take the type of the data into account if there is a costume with a name that is a number.

[...]

Issue #2013, opened 2019-02-21
Cast sprite names in one place before passing to getSpriteTargetByName 


ktbee:
### Expected Behavior

All blocks should be able to get a sprite target when passing a name if the sprite exists. 

### Actual Behavior

If the sprite name is a number, it might not be able to find the sprite. `getSpriteTargetByName` expects a string, and if a block doesn't cast a name to a string, the sprite target can't be found. We should cast the name to a string in one place instead of multiple times in individual blocks. 

### Steps to Reproduce

See https://github.com/LLK/scratch-vm/issues/1743 for repro steps and more details

ktbee:
@mzgoddard I assigned you this issue since I know this is something you wanted to tackle in your optimization work.

PR #2038, opened 2019-03-06
Cast sprite name to string before getSpriteTargetByName

ktbee:
### Resolves

https://github.com/LLK/scratch-vm/issues/1743

### Proposed Changes

For blocks that use `getSpriteTargetByName`, cast the sprite name to a string before passing to `getSpriteTargetByName`. These are the blocks that are fixed by this change:

- Go to
- Glide to
- Point towards
- Create clone of
- Distance to
- X position of
- Touching sprite

### Reason for Changes

Currently `getSpriteTargetByName` doesn't cast sprite names to a string, so any sprites with a number for a name might not be found if a number is the type of the value passed to `getSpriteTargetByName`. As discussed in https://github.com/LLK/scratch-vm/issues/1743, it would generally be better to have this string casting done in one place instead of having the same logic repeated in multiple blocks. I am doing this non-DRY solution for now however because @mzgoddard has plans as part of his performance optimizations to make the string casting happen elsewhere in the VM in one place. That work is being tracked in https://github.com/LLK/scratch-vm/issues/2013

### Test Coverage

I added sprite-number-name.sb2 as one of the tests run by execute.js.


src/sprites/rendered-target.js
```diff
@@ -840,6 +840,7 @@ class RenderedTarget extends Target {
      * @return {boolean} True iff touching a clone of the sprite.
      */
     isTouchingSprite (spriteName) {
+        spriteName = StringUtil.stringify(spriteName);
```
kchadha: Was there a particular reason you chose `StringUtil.stringify` here? I think this one should also be `Cast.toString`

kchadha (changes requested):
Change requested in the code comments.

src/sprites/rendered-target.js
ktbee: I was thinking I could use a util that already was required in the VM since it's such a small change, but in retrospect I was probably overthinking it. I'll add `Cast` and use `toString`.

kchadha (approved):
LGTM! Thanks!

Issue #1283, opened 2018-06-29
Improper clamping behavior on graphic effects


thisandagain:
### Steps to Reproduce
- https://scratch.mit.edu/projects/228123708/
- https://preview.scratch.mit.edu/#228123708

### Reference via Mr-Dave
```
I looked at your ice cream maker project and I believe that the problem you are having is that Scratch 3.0 does not limit the min and max values for what we can set for the brightness and ghosting settings.  You can see the difference in this small project.  Under Scratch 2.0 the cat and the backdrop recover and look the same no matter that both have had their values set to more that the max or min of 100 and -100.  However, run this same project under Scratch 3.0 and all you'll see is solid black.

Other people have mentioned having the same problem.
```

LiFaytheGoblin:
Related to this is how the effect should behave if it is at the 'edge', like in this case:
<img width="976" alt="screen shot 2018-08-29 at 3 45 16 pm" src="https://user-images.githubusercontent.com/5471273/44811407-a53ba000-aba2-11e8-9d56-d14b122bda72.png">

Project 242576633

ktbee:
closed via #2042

PR #2042, opened 2019-03-08
add clampEffect for limiting range of brightness and ghost effects

ktbee:
### Resolves

https://github.com/LLK/scratch-vm/issues/1283

### Proposed Changes

Add `clampEffect` to limit the range of ghost and brightness effects

### Reason for Changes

[In Scratch 2](https://github.com/LLK/scratch-flash/blob/646523e6846ad0dd993213a38b46fe2ea511d026/src/filters/FilterPack.as#L86-L88) we limited the values for the brightness graphic effect to between -100 and 100. These changes add that limit in a separate function that handles the previously added ghost effect limits as well. This allows us to use the same clamping logic in two blocks (`setEffect` and `clampEffect`). 

### Test Coverage

Added `clamp graphic effects` test in unit/blocks_looks.js

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -77,6 +77,14 @@ class Scratch3LooksBlocks {
     }
 
     /**
+     * Limit for brightness effect
+     * @const {object}
+     */
+    static get EFFECT_BRIGHTNESS_LIMIT (){
+        return {min: -100, max: 100};
+    }
+
+    /**
      * @param {Target} target - collect bubble state for this target. Probably, but not necessarily, a RenderedTarget.
      * @returns {BubbleState} the mutable bubble state associated with that target. This will be created if necessary.
      * @private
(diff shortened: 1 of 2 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -484,27 +492,36 @@ class Scratch3LooksBlocks {
         );
     }
 
+    clampEffect (effect, value) {
+        let clampedValue;
```
joker314: For effects other than "ghost" and "brightness", this code should return the value as passed instead of returning `undefined`. Shouldn't this be

```suggestion
        let clampedValue = value;
```

joker314 (changes requested):
Thanks for making the PR! Just a question about how this would affect effects other than "brightness" and "ghost" in the comment above.

thisandagain:
@picklesrus Mind taking a look at this PR?

src/blocks/scratch3_looks.js
ktbee: Oh right, that's a good catch. Thank you @joker314!

ktbee: I'm surprised forgetting to return any value for `clampedValue` didn't break any tests. I'll add that scenario to my test for this PR as well. Thanks again!

picklesrus (approved):
Changes look good to me. Thanks! Will merge!

PR #2045, opened 2019-03-11
Allow for situation where we get a move event to attach a shadow.

### LINKED ISSUE #1691: Some custom block inputs are "invisible"
### Expected Behavior

All inputs in calls of the custom block look ordinary, i.e. white and can be typed in.

### Actual Behavior

There are a *lot* of weird issues with calls of this custom block.

Here is what they look like when you open the project:

Screenshot of three calls of the same custom block which has three inputs; First: "behavior" is empty - Second: "behavior" and "spacing" are empty - Third: "spacing" is empty

Here is what they look like when you drag out the "pick random" block from the "velocity" input:

Screenshot of the three calls again; in the first, there is an ordinary empty input, but in the other two, the input is "missing", just a darker red empty input-shaped field

**Note that, during the development of the project that used this custom block, I added new inputs to the "Spawn in a Circle" custom block multiple times.** I know I originally created the custom block with just the first input, "behavior:". Later I created the "velocity" and "spacing" inputs, but I'm not sure what order. I think I may have created "spacing" first, but I'm 90% not sure.

### Steps to Reproduce

Download this project and open it in the Scratch 3.0 GUI, and observe the ensuing weirdness as described in "actual behavior": [untitled-project-8_17_2018-11_38_37 AM.zip](https://github.com/LLK/scratch-blocks/files/2297636/untitled-project-8_17_2018-11_38_37.AM.zip)

Firefox 62.0b17, Debian.

### LINKED ISSUE #4610: Editing custom blocks cause its "inputs" (?) to be bigger when reloading page
### Expected Behavior

This shouldn't happen.

### Actual Behavior

customblock

### Steps to Reproduce

1. Make a custom block
2. Edit it by adding a new input
3. Reload the page

Ubuntu Linux, Firefox 65.0

paulkaplan:
This happens after adding a custom procedure input to an existing custom procedure call block.

### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/4610
Fixes https://github.com/LLK/scratch-blocks/issues/1691

### Proposed Changes

Allows the `move` block event to attach a shadow to an input. This never happens in the UI directly from the user,  but it happens after changing a custom procedure by adding an input. In that situation, the call block is first mutated to add an input, then a shadow is created, then it is moved to be attached to the call block. That last step was only assigning the block to the `block` attribute, leaving the `shadow` attribute null. But the block is itself a shadow, so should be assigned to both shadow/block attributes. 

### Reason for Changes

Fix the above bugs.

### Test Coverage

Manually tested with the repro cases on the linked issues, and added a unit test to make sure the code works.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -735,6 +735,12 @@ class Blocks {
                 if (this._blocks[e.newParent].inputs.hasOwnProperty(e.newInput)) {
                     oldShadow = this._blocks[e.newParent].inputs[e.newInput].shadow;
                 }
+
+                // If the block being attached is itself a shadow, make sure to set
+                // both block and shadow to that blocks ID. This happens when adding
+                // inputs to a custom procedure.
+                if (this._blocks[e.id].shadow) oldShadow = e.id;
+
                 this._blocks[e.newParent].inputs[e.newInput] = {
                     name: e.newInput,
                     block: e.id,
--- a/test/unit/engine_blocks.js
+++ b/test/unit/engine_blocks.js
@@ -358,6 +358,43 @@ test('move no obscure shadow', t => {
     t.end();
 });
 
+test('move - attaching new shadow', t => {
+    const b = new Blocks(new Runtime());
+    // Block/shadow are null to mimic state right after a procedure_call block
+    // is mutated by adding an input. The "move" will attach the new shadow.
+    b.createBlock({
+        id: 'foo',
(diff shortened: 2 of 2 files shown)


kchadha:
LGTM! Thanks for investigating and fixing this!

kchadha (approved):
LGTM! Thanks for investigating and fixing this!

PR #2046, opened 2019-03-12
Add eight new languages to Text to Speech extension

ericrosenbaum:
**Eight new languages**
This PR adds support for these languages to the Text to Speech extension:
Chinese (Mandarin), Korean, Hindi, Swedish, Norwegian, Romanian, Turkish, Welsh

**Support for voice options for languages that provide only one voice gender**
We did not initially include these languages because the speech synthesis service we use, Amazon Polly, provides only a female voice for each (details [here](https://docs.aws.amazon.com/polly/latest/dg/voicelist.html)). In the other languages we support, they provide both a female and a male voice, which we use to create the four distinct voice options in the extension ("alto" is female, "tenor" is male, "squeak" is the female voice pitched up, "giant" is the male voice pitched down). To support these additional languages, I've added new pitch adjustments, so that a single voice can be used in four versions ("alto" is unmodified, "tenor" is pitched down, "squeak" is pitched up, and "giant" is pitched way down). 

**New data structure**
In order to cleanly add the flag to indicate which voices needed these new pitch adjustments, I used a new data structure that holds this flag (`singleGender`).

**Four types of locale code**
In order to clarify the difference between them, I included a comment in the code differentiating between Scratch locales, supported locales, extension locales, and speech synth locales.

**Loading the extension in the correct language for additional languages**
Previously, we had assumed a one-to-one relationship between languages that you could set the editor to ("Scratch locales") and text to speech languages ("extension locales"). There are actually three cases in which we support two written languages that can use a single spoken language (Chinese, Japanese, and Norwegian). In other words, those are cases where there are two supported locales for one extension locale. In these cases, when the editor is set to any of those languages, we load the extension with the correct language set. 

**Future Work**
- We plan to localize the language names in the "set language" block's menu.
- We plan to support language droppability, i.e. when you drop the "language" reporter from the translate extension into the "set language" block's input, it should correctly set the extension language if it is supported.
- If the speech synthesis service adds them, we could provide support for multiple spoken languages for each scratch locale.


src/extensions/scratch3_text2speech/index.js
```diff
@@ -376,13 +510,49 @@ class Scratch3Text2SpeechBlocks {
...
+     * @param {string} locale a locale code.
+     * @returns {?string} a locale supported by the extension.
+     */
+    _getExtensionLocaleForSupportedLocale (locale) {
+        for (const lang in this.LANGUAGE_INFO) {
+            if (this.LANGUAGE_INFO[lang].locales.includes(locale)) {
+                return lang;
+            }
+        }
+        log.error(`cannot find extension locale for locale ${locale}`);
```
picklesrus, line 523: Should this return some default value or null in the case where it isn't found? Or no because it is only used for saving?  I can't quite tell whether state.textToSpeechLanguage is being set in the same cases as it was before and if not, if that is intentional.

src/extensions/scratch3_text2speech/index.js
```diff
@@ -457,16 +627,29 @@ class Scratch3Text2SpeechBlocks {
...
+            }
+            if (state.voiceId === GIANT_ID) {
+                playbackRate = FEMALE_GIANT_RATE;
+            }
+        }
 
         if (state.voiceId === KITTEN_ID) {
             words = words.replace(/\S+/g, 'meow');
-            locale = 'en-US';
+            locale = this.LANGUAGE_INFO[this.DEFAULT_LANGUAGE].speechSynthLocale;
```
picklesrus, line 652: Just confirming that I'm reading this right and the kitten voice is still always english.

picklesrus:
Just one question but generally looks good!

src/extensions/scratch3_text2speech/index.js
ericrosenbaum: The logic for setting `stage.textToSpeechLanguage` is the same as before- the `log.error` should only happen in the case that the `LANGUAGE_INFO` data structure somehow gets out of sync (in which case it's a useful error).

src/extensions/scratch3_text2speech/index.js
ericrosenbaum: meow (yup)!

PR #2048, opened 2019-03-14
Temporarily remove Vernier Force and Accel extension

ericrosenbaum:
We need to temporarily remove the Vernier Force and Acceleration extension until an issue with the Vernier GoDirect module that it depends on is resolved.


thisandagain (approved):

ericrosenbaum:
no longer needed

Issue #1997, opened 2019-02-12
Localize name of Text to Speech extension


ericrosenbaum:
Looks like we inadvertently failed to localize the name of the Text to Speech extension

PR #2056, opened 2019-03-19
Localize name of Text to Speech extension

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1997

### Proposed Changes

Use a formatMessage to localize the name of the Text to Speech extension 

### Reason for Changes

Generally, strings should be localized. Some extension names that are brand names are not localized (such as "micro:bit"), but this is not one of those.


chrisgarrity (approved):
LGTM

PR #2058, opened 2019-03-21
Remove Hindi from text2speech extension

ericrosenbaum:
Amazon's Hindi language speech synthesis has very low quality- needs investigation, so we will remove it for now.


paulkaplan (approved):
LGTM (and @picklesrus)

PR #2060, opened 2019-03-22
Load core extensions synchronously

kchadha:
### Resolves

First step towards converting core blocks into extensions.

### Proposed Changes

Allows loading internal (non-worker) extensions synchronously. This should work for any existing extension, but will be used for loading core block extensions in the future. An example core block category is also provided and used in the tests.

To manually test, you can pull in these changes and add coreExample to the `CORE_EXTENSIONS` const in src/virtual-machine.js and it should load with the other core block categories.

Known issue:
- currently the core extension loads slightly after the other categories in the toolbox, and this will eventually need to be fixed in the gui when we end up getting rid of the default toolbox xml.

### Test Coverage

Added unit tests and integration test for loading an extension (or a dispatch service) synchronously.

### DIFF
--- a/src/dispatch/central-dispatch.js
+++ b/src/dispatch/central-dispatch.js
@@ -47,12 +47,12 @@ class CentralDispatch extends SharedDispatch {
         const {provider, isRemote} = this._getServiceProvider(service);
         if (provider) {
             if (isRemote) {
-                log.error("Cannot use 'callSync' on a remote service provider.");
-                return;
+                throw new Error(`Cannot use 'callSync' on remote provider for service ${service}.`);
             }
 
             return provider[method].apply(provider, args);
         }
+        throw new Error(`Provider not found for service: ${service}`);
     }
 
     /**
--- a/src/dispatch/central-dispatch.js
+++ b/src/dispatch/central-dispatch.js
@@ -36,6 +36,39 @@ class CentralDispatch extends SharedDispatch {
     }
 
     /**
+     * Synchronously call a particular method on a particular service provided locally.
+     * Calling this function on a remote service will fail.
+     * @param {string} service - the name of the service.
+     * @param {string} method - the name of the method.
+     * @param {*} [args] - the arguments to be copied to the method, if any.
+     * @returns {*} - the return value of the service method.
+     */
+    callSync (service, method, ...args) {
+        const {provider, isRemote} = this._getServiceProvider(service);
+        if (provider) {
+            if (isRemote) {
+                log.error("Cannot use 'callSync' on a remote service provider.");
+                return;
+            }
+
+            return provider[method].apply(provider, args);
+        }
+    }
+
+    /**
+     * Synchronously set a local object as the global provider of the specified service.
+     * WARNING: Any method on the provider can be called from any worker within the dispatch system.
+     * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui', 'extension9'.
+     * @param {object} provider - a local object which provides this service.
+     */
+    setServiceSync (service, provider) {
+        if (this.services.hasOwnProperty(service)) {
+            log.warn(`Central dispatch replacing existing service provider for ${service}`);
+        }
+        this.services[service] = provider;
+    }
+
+    /**
      * Set a local object as the global provider of the specified service.
      * WARNING: Any method on the provider can be called from any worker within the dispatch system.
      * @param {string} service - a globally unique string identifying this service. Examples: 'vm', 'gui', 'extension9'.
@@ -45,10 +78,7 @@ class CentralDispatch extends SharedDispatch {
     setService (service, provider) {
         /** Return a promise for consistency with {@link WorkerDispatch#setService} */
         try {
-            if (this.services.hasOwnProperty(service)) {
-                log.warn(`Central dispatch replacing existing service provider for ${service}`);
-            }
-            this.services[service] = provider;
+            this.setServiceSync(service, provider);
             return Promise.resolve();
         } catch (e) {
             return Promise.reject(e);
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -11,7 +11,7 @@ const BlockType = require('./block-type');
 const builtinExtensions = {
     // This is an example that isn't loaded with the other core blocks,
     // but serves as a reference for loading core blocks as extensions.
-    coreExample: require('../blocks/scratch3_core_example'),
+    coreExample: () => require('../blocks/scratch3_core_example'),
     // These are the non-core built-in extensions.
     pen: () => require('../extensions/scratch3_pen'),
     wedo2: () => require('../extensions/scratch3_wedo2'),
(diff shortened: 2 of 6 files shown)


src/dispatch/central-dispatch.js
```diff
@@ -35,6 +35,39 @@ class CentralDispatch extends SharedDispatch {
...
+    callSync (service, method, ...args) {
+        const {provider, isRemote} = this._getServiceProvider(service);
+        if (provider) {
+            if (isRemote) {
+                log.error("Cannot use 'callSync' on a remote service provider.");
+                return;
+            }
+
+            return provider[method].apply(provider, args);
+        }
```
cwillisf, line 54:
```suggestion
        }
        throw new Error(`Provider not found for service: ${service}`);
```

src/dispatch/central-dispatch.js
```diff
@@ -35,6 +35,39 @@ class CentralDispatch extends SharedDispatch {
...
+     * @param {string} method - the name of the method.
+     * @param {*} [args] - the arguments to be copied to the method, if any.
+     * @returns {*} - the return value of the service method.
+     */
+    callSync (service, method, ...args) {
+        const {provider, isRemote} = this._getServiceProvider(service);
+        if (provider) {
+            if (isRemote) {
+                log.error("Cannot use 'callSync' on a remote service provider.");
+                return;
```
cwillisf:
```suggestion
                throw new Error(`Cannot use 'callSync' on remote provider for service ${service}.`);
```

src/extension-support/extension-manager.js
```diff
@@ -108,6 +112,27 @@ class ExtensionManager {
...
+     * fail if the provided id is not does not match an internal extension.
+     * @param {string} extensionId - the ID of an internal extension
+     */
+    loadExtensionIdSync (extensionId) {
+        if (builtinExtensions.hasOwnProperty(extensionId)) {
+            /** @TODO dupe handling for non-builtin extensions. See commit 670e51d33580e8a2e852b3b038bb3afc282f81b9 */
+            if (this.isExtensionLoaded(extensionId)) {
+                const message = `Rejecting attempt to load a second extension with ID ${extensionId}`;
+                log.warn(message);
+                return; // TODO Do we want to throw an error here?
```
cwillisf: Good question... my gut feeling is that trying to load a second copy of an extension isn't a "fatal" error since the goal of loading an extension is to gain access to the blocks, and after this path is done the blocks should be available in the dupe case because they already were available before the call. I'm open to further discussion though :)

src/extension-support/extension-manager.js
```diff
@@ -124,7 +149,7 @@ class ExtensionManager {
 
             const extension = builtinExtensions[extensionURL]();
             const extensionInstance = new extension(this.runtime);
-            return this._registerInternalExtension(extensionInstance).then(serviceName => {
+            return Promise.resolve(this._registerInternalExtension(extensionInstance)).then(serviceName => {
                 this._loadedExtensions.set(extensionURL, serviceName);
```
cwillisf: All this is synchronous now, right? In that case it might be easier to read & understand if it's written this way:
```js
this._registerInternalExtension(...);
this._loadedExtensions.set(...);
return Promise.resolve();
```

src/extension-support/extension-manager.js
```diff
@@ -150,7 +175,7 @@ class ExtensionManager {
                     dispatch.call('runtime', '_refreshExtensionPrimitives', info);
                 })
                 .catch(e => {
-                    log.error(`Failed to refresh buildtin extension primitives: ${JSON.stringify(e)}`);
+                    log.error(`Failed to refresh built-in extension primitives: ${JSON.stringify(e)}`);
```
cwillisf, line 181: No, I meant that the extension primitives were built from tin!
:iwasneverhere:

cwillisf (changes requested):
Looks great! I added a few nit-picky requests but I don't see any major issues.

src/dispatch/central-dispatch.js
```diff
@@ -35,6 +35,39 @@ class CentralDispatch extends SharedDispatch {
...
+     * @param {string} service - the name of the service.
+     * @param {string} method - the name of the method.
+     * @param {*} [args] - the arguments to be copied to the method, if any.
+     * @returns {*} - the return value of the service method.
+     */
+    callSync (service, method, ...args) {
+        const {provider, isRemote} = this._getServiceProvider(service);
+        if (provider) {
+            if (isRemote) {
+                log.error("Cannot use 'callSync' on a remote service provider.");
```
kchadha:
```suggestion
```

src/extension-support/extension-manager.js
kchadha: I think that sounds reasonable!

```suggestion
                return;
```

kchadha: Should we make it so that the analogous code in loadExtensionURL doesn't reject the promise here?

kchadha:
@cwillisf, I think I addressed all of your comments/suggestions and also followed up with a question. =]

src/extension-support/extension-manager.js
```diff
@@ -9,6 +9,10 @@ const BlockType = require('./block-type');
 // TODO: change extension spec so that library info, including extension ID, can be collected through static methods
 
 const builtinExtensions = {
+    // This is an example that isn't loaded with the other core blocks,
+    // but serves as a reference for loading core blocks as extensions.
+    coreExample: require('../blocks/scratch3_core_example'),
```
rschamp: Recently there was a performance pass to do these requires asynchronously. Is this require intentionally synchronous?

src/extension-support/extension-manager.js
```diff
@@ -108,6 +112,27 @@ class ExtensionManager {
         return this._loadedExtensions.has(extensionID);
     }
 
+    /**
+     * Synchronously load an internal extension (core or non-core) by ID. This call will
+     * fail if the provided id is not does not match an internal extension.
+     * @param {string} extensionId - the ID of an internal extension
+     */
+    loadExtensionIdSync (extensionId) {
+        if (builtinExtensions.hasOwnProperty(extensionId)) {
```
rschamp: Should we not throw or warn if it's not built in?

test/unit/dispatch.js
```diff
@@ -62,3 +62,24 @@ test('remote', t => {
...
+    t.equal(a, 42);
+
+    const b = dispatch.callSync('SyncDispatchTest', 'doubleArgument', 9);
+    t.equal(b, 18);
+
+    const c = dispatch.callSync('SyncDispatchTest', 'doubleArgument', 123);
+    t.equal(c, 246);
+
+    try {
+        dispatch.callSync('SyncDispatchTest', 'throwException');
```
rschamp: My understanding is that you should be able to use, e.g. 
```suggestion
        t.throws(() => dispatch.callSync('SyncDispatchTest', 'throwException'), new Error('This is a test exception thrown by DispatchTest'));
```

rschamp (changes requested):
The only comment here I think is consequential is the one about the async require, since we don't want to become inconsistent and lose the benefit of the perf work.

src/extension-support/extension-manager.js
kchadha: Good catch! This was not intentional.

src/extension-support/extension-manager.js
kchadha: I'll add a warning.

test/unit/dispatch.js
kchadha: Done.

kchadha:
@rschamp addressed all your comments! Thanks!

src/extension-support/extension-manager.js
cwillisf: Oh. Hmm... I always find it awkward when I disagree with my past self... ;)

Yeah, I guess I would recommend just returning an empty `Promise.resolve()` from `loadExtensionURL` in the equivalent branch. I guess maybe I made that reject before I realized that rejecting a promise is just as "serious" as throwing an exception.

rschamp (approved):

src/extension-support/extension-manager.js
kchadha: I think this plan makes sense.

cwillisf (approved):

src/extension-support/extension-manager.js
cwillisf: I don't think the change in `loadExtensionURL` needs to block this PR; we can improve consistency in a later change if you'd like. I have this vague sense that we'll be actively working in this area for a while ;)

PR #2061, opened 2019-03-22
Add LEGO BOOST Scratch 3.0 extension

knandersen:
### Proposed Changes

Adds the LEGO BOOST extension to the scratch-vm

### Reason for Changes

Centralize development, review, and launch of the LEGO BOOST extension


knandersen:
Worth noting that the BLE advertisement data filter is currently commented out as it depends on that feature existing in Scratch Link on Windows computers (https://github.com/LLK/scratch-link/issues/112).

evhan55:
Hi @knandersen !  I am finally getting to this PR and starting by testing basic functionality of all the blocks, first.  Most of the blocks work really well and as expected for me!

I only found two tiny issues in basic block functionality:
- `when [color] color seen` doesn't seem to trigger for almost anything for me, and similarly the `color` reporter block mostly reports `'none'`.  Is there a trick to using the color sensor?
- setting motor power gave me a tiny bug: when setting from `100%` power to `10%` power, the motor seems to jitter when first running it, and then it is consistently at `10%`.  the same is not true for going in the reverse order (setting to `10%` power first, then `100%`).  I am not sure how consistent this is.

Everything else worked really well for me (I even tried all motor option combos), amazing! Up next, I will look at the code itself.

package-lock.json
```diff
@@ -10,16 +10,16 @@
       "integrity": "sha512-zbO/DtTnaDappBflIU3zYEgATLToRDmW5uN/EGH1GXaes7ydfjqmAoK++xmJIA+8HfDw7UyPZNdM8fhGhfmMhw==",
```
evhan55: You can remove the `package.json` from the PR by checking out the latest `package.json` from develop and pushing that as a commit to this branch.

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+ * @type {string}
+ */
+// eslint-disable-next-line max-len
+const iconURI = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAFAAAABQCAMAAAC5zwKfAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAACpQTFRF////fIel5ufolZ62/2YavsPS+YZOkJmy9/j53+Hk6+zs6N/b6dfO////tDhMHAAAAA50Uk5T/////////////////wBFwNzIAAAA6ElEQVR42uzX2w6DIBAEUGDVtlr//3dLaLwgiwUd2z7MJPJg5EQWiGhGcAxBggQJEiT436CIfqXJPTn3MKNYYMSDFpoAmp24OaYgvwKnFgL2zvVTCwHrMoMi+nUQLFthaNCCa0iwclLkDgYVsQp0mzxuqXgK1MRzoCLWgkPXNN2wI/q6Kvt7u/cX0HtejN8x2sXpnpb8J8D3b0Keuhh3X975M+i0xNVbg3s1TIasgK21bQyGO+s2PykaGMYbge8KrNrssvkOWDXkErB8UuBHETjoYLkKBA8ZfuDkbwVBggQJEiR4MC8BBgDTtMZLx2nFCQAAAABJRU5ErkJggg==';
+
+/**
+ * Boost BLE UUIDs.
+ * @enum {string}
+ */
+
```
evhan55: nit: extra line

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+    characteristic: '00001624-1212-efde-1623-785feabcd123',
+    sendInterval: 100,
+    sendRateMax: 20
+};
+
+/**
+ * Boost Motor Max Power.
+ * @constant {number}
+ */
+
```
evhan55: nit: extra line

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+    MOTOR_SENSOR: 2, // Set motors to report their position
+    UNKNOWN: 0 // Anything else will use the default mode (mode 0)
+};
+
+/**
+ * Helper function for converting a JavaScript number to an INT32-number
+ * @param {number} number - a number
+ * @return {array} - a 4-byte array of Int8-values representing an INT32-number
+ */
+const numberToInt32Array = function (number) {
```
evhan55, line 224: These two helper functions look good to me up here for now.  I do wonder if they are used among other LEGO devices moving forward, if they can't move into a util class somewhere (e.g. `MathUtil`) if they'll be useful in a few other extensions.

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+        dataview.getInt8(0)
+    ];
+};
+
+/**
+ * Helper function for converting a regular array to a Little Endian INT32-value
+ * @param {Array} array - an array containing UInt8-values
+ * @return {number} - a number
+ */
+const int32ArrayToNumber = function (array) {
```
evhan55, line 241: Same comment as other helper function.

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+         * @type {Object.<string, number>}
+         * @private
+         */
+        this._sensors = {
+            tiltX: 0,
+            tiltY: 0,
+            color: BoostColor.NONE
+        };
+
+        this._colorBucket = [];
```
evhan55, line 601: Does this need a JSDoc comment?

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+
+/**
+ * Boost BLE UUIDs.
+ * @enum {string}
+ */
+
+const BoostBLE = {
+    service: '00001623-1212-efde-1623-785feabcd123',
+    characteristic: '00001624-1212-efde-1623-785feabcd123',
+    sendInterval: 100,
```
evhan55, line 31: I like the idea of putting these rate limiter vars in this enum, we should do it in the other extensions!

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+        return this._motors[index];
+    }
+
+    /**
+     * Stop all the motors that are currently running.
+     */
+    stopAllMotors () {
+        this._motors.forEach(motor => {
+            if (motor) {
+                // Send the motor off command without using the rate limiter.
```
evhan55, line 668: Good comment! 👍

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+     * Called by the runtime when user wants to scan for a Boost peripheral.
+     */
+    scan () {
+        if (this._ble) {
+            this._ble.disconnect();
+        }
+        this._ble = new BLE(this._runtime, this._extensionId, {
+            filters: [{
+                services: [BoostBLE.service]/* ,
+                manufacturerData: {
```
evhan55, line 732: Good idea to leave this comment in while we try to include it in Scratch Link

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+     *
+     * Payload is accepted as an array since these vary across different subcommands.
+     *
+     * @param  {number} portID - the port (Connect ID) to send a command to.
+     * @param  {number} execution - Byte containing startup/completion information
+     * @param  {number} subCommand - the id of the subcommand byte.
+     * @param  {array}  payload    - the list of bytes to send as subcommand payload
+     * @return {array}            - a generated output command.
+     */
+    generateOutputCommand (portID, execution, subCommand, payload) {
```
evhan55, line 822: I'm glad these` generateOutputCommand` and `generateInputCommand` were a useful pattern for the Boost, too!

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+            enableNotifications
+        ]);
+
+        command.unshift(command.length + 1);
+
+        return command; // Prepend payload with length byte;
+    }
+
+    /**
+     * Sets LED mode and initial color and starts reading data from peripheral after BLE has connected.
```
evhan55: Is this comment accurate?

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+        );
+        this._pingDeviceId = window.setInterval(this._pingDevice, BoostPingInterval);
+    }
+
+    /**
+     * Process the sensor data from the incoming BLE characteristic.
+     * @param {object} base64 - the incoming BLE data.
+     * @private
+     */
+    _onMessage (base64) {
```
evhan55, line 876: I absolutely love how you have formatted this function ❤️ 👍  .  I'd love to see about trying to do something similar for the EV3 source code for `_onMessage`!

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+        default:
+        }
+    }
+
+    /**
+     * Ping the Boost hub. If the Boost hub has disconnected
+     * for some reason, the BLE socket will get an error back and automatically
+     * close the socket.
+     */
+
```
evhan55: nit: extra line

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+     * sensor or motor internally, and then register for notifications on input
+     * values if it is a sensor.
+     * @param {number} portID - the port to register a sensor or motor on.
+     * @param {number} type - the type ID of the sensor or motor
+     * @private
+     */
+    _registerSensorOrMotor (portID, type) {
+        // Record which port is connected to what type of device
+        this._ports[portID] = type;
+
```
evhan55, line 985: nit: extra line?

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+    /**
+     * Turn specified motor(s) on for a specified duration.
+     * @param {object} args - the block's arguments.
+     * @property {MotorID} MOTOR_ID - the motor(s) to activate.
+     * @property {int} DURATION - the amount of time to run the motors.
+     * @return {Promise} - a promise which will resolve at the end of the duration.
+     */
+    motorOnFor (args) {
+        // TODO: cast args.MOTOR_ID?
```
evhan55, line 1602: This comment is referring to the fact that `args.MOTOR_ID` is sent to a Promise later, and perhaps is better if it is handled/cleaned/verified before that Promise.  It might be good to fix all these casting `TODOs` in all the opcodes for the Boost, but probably in a later PR, where maybe we can address them in both the WeDo and Boost extensions at once.

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+            if (motor) {
+                return new Promise(resolve => {
+                    motor.turnOnForDegrees(degrees, sign);
+                    motor.pendingPromiseFunction = resolve;
+                });
+            }
+            return null;
+        });
+        // To prevent the block from returning a value, an empty function is added to the .then
+        return Promise.all(promises).then(() => {});
```
evhan55, line 1655: This isn't 100% clear to me, but since testing the motors went well for me, I am assuming it's something necessary to handle multiple promises from multiple motors.

src/extensions/scratch3_boost/index.js
```diff
@@ -0,0 +1,1963 @@
...
+    setMotorPower (args) {
+        // TODO: cast args.MOTOR_ID?
+        this._forEachMotor(args.MOTOR_ID, motorIndex => {
+            const motor = this._peripheral.motor(motorIndex);
+            if (motor) {
+                motor.power = MathUtil.clamp(Cast.toNumber(args.POWER), 0, 100);
+                if (motor.isOn) {
+                    if (motor.pendingTimeoutDelay) {
+                        motor.turnOnFor(motor.pendingTimeoutStartTime + motor.pendingTimeoutDelay - Date.now());
+                    } else if (motor.pendingPositionDestination) {
```
evhan55, line 1717: Is the glitch I found in going from `10%` to `100%` maybe found in here somewhere?  I'm not studying the math carefully, so just guessing.

evhan55:
Just did a first pass at reviewing the code, it reads really gorgeous and cleanly to me!  I would be happy with landing this in the state that it's in.  I commented on some minor nits and questions here and there, but overall it feels like it's a really great state to me, including the use of the RateLimiter.  Any further changes, etc. could go in future PR's.

I especially love the `_onMessage` and would love your help trying to do something similar for the EV3 and generally getting the EV3 extension code to be as clean as this one is.

evhan55 (approved):
Pending removal of the `package.json`, everything else looks good to me and further major changes could go in future PR's.

src/extensions/scratch3_boost/index.js
knandersen: Good point! If we want to do that though, I think we'd have to make them more generic, since this one is hardcoded to return "little endian"-format numbers.

src/extensions/scratch3_boost/index.js
knandersen: Yes it does. Thanks for catching that :)

src/extensions/scratch3_boost/index.js
knandersen: Thanks!

src/extensions/scratch3_boost/index.js
knandersen: Thanks!

src/extensions/scratch3_boost/index.js
knandersen: @evhan55 Yeah I inherited the function from the WeDo 2.0 extension, so that makes sense. In the case of Boost, I can probably just check if it exists in the `BoostPort`-enum. Thoughts?

src/extensions/scratch3_boost/index.js
knandersen: It was added because without it, when running motor-commands on all ports without motors plugged into those ports, the block commands for the non-motor ports would execute immediately. (hope that made sense, it's hard to describe 😆
By mapping only the ports that have motors on them, and checking the promises against `Promise.all`, the blocks yield correctly. 

I do agree that it doesn't read very well. so I'll try to add some comments to the code!

src/extensions/scratch3_boost/index.js
knandersen: Removed, thanks!

src/extensions/scratch3_boost/index.js
knandersen: Removed, thanks!

src/extensions/scratch3_boost/index.js
knandersen: Awesome, thanks! 😃

src/extensions/scratch3_boost/index.js
knandersen: No it's not, I've moved the LED part to `_registerSensorOrMotor()`. Thanks!

ericrosenbaum:
I found some problems, but I will file those as issues to follow up on so we can more easily track them. I agree with @evhan55 that once `package-lock.json` is removed from this PR we can merge it. Thanks @knandersen!

ericrosenbaum (approved):
LGTM!

knandersen:
> I found some problems, but I will file those as issues to follow up on so we can more easily track them. I agree with @evhan55 that once `package-lock.json` is removed from this PR we can merge it. Thanks @knandersen!

@ericrosenbaum @evhan55 Great! I've updated the package-lock.json, so we should be good! 🎉

PR #2063, opened 2019-03-22
Move facing options into gesture menu

ericrosenbaum:
Remove the <facing [up, down]?> boolean block, and add two new options into the existing hat block "when [shaken, started falling]", for "turned face up" and "turned face down".

Screen Shot 2019-03-22 at 5 20 34 PM

cc @carljbowman

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -417,7 +417,9 @@ const PushPullValues = {
  */
 const GestureValues = {
     SHAKEN: 'shaken',
-    STARTED_FALLING: 'started falling'
+    STARTED_FALLING: 'started falling',
+    TURNED_FACE_UP: 'turned face up',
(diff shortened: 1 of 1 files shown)


evhan55:
I tested the new hat:

- `when shaken`: still works as expected
- `when started falling`: still works as expected
- `when turned face up`: sometimes triggers a couple of times for me when I turn it, the sensitivity feels a bit high?
- `when turned face down`: similar as above

But if we think that sensitivity is okay, then it looks good to me!
Will mark as good to merge.

carljbowman:
Not sure if there were other considerations for the phrasing on these new menu items. It might not be a problem, but we us "turn" quite often on blocks and sometimes it means different things: e.g. "turn motor on", "turn 15 degrees". This may be a reason to consider an another term.

That said, the alternatives may be too obscure:
"when (flipped face up)" & "when (flipped face down)"

PR #2065, opened 2019-03-25
Fix #4573 by using an existing utitlity that re-ids blocks when we du…

picklesrus:
…plicate a sprite.

### Resolves

Fixes https://github.com/LLK/scratch-gui/issues/4573

### Proposed Changes

Use the util that share the love does to give blocks new ids when duplicating a sprite.  

### Test Coverage

Added a unit test for duplicate that checks that a duplicated block has a different id from the original.

### DIFF
--- a/src/sprites/sprite.js
+++ b/src/sprites/sprite.js
@@ -2,6 +2,7 @@ const RenderedTarget = require('./rendered-target');
 const Blocks = require('../engine/blocks');
 const {loadSoundFromAsset} = require('../import/load-sound');
 const {loadCostumeFromAsset} = require('../import/load-costume');
+const newBlockIds = require('../util/new-block-ids');
 const StringUtil = require('../util/string-util');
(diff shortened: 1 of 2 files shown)


kchadha (approved):
LGTM!

src/sprites/sprite.js
```diff
@@ -136,8 +137,12 @@ class Sprite {
 
     duplicate () {
         const newSprite = new Sprite(null, this.runtime);
+        const copiedBlocks = JSON.parse(JSON.stringify(Object.values(this.blocks._blocks)));
```
picklesrus: this object.values needs to go

kchadha:
```suggestion
        const blocksContainer = this.blocks._blocks;
        const originalBlocks = Object.keys(blocksContainer).map(key => blocksContainer[key]);
        const copiedBlocks = JSON.parse(JSON.stringify(originalBlocks));
```

Issue #2067, opened 2019-03-25
Vernier "when turned face up/down" hat often triggers twice


ericrosenbaum:
### Expected Behavior

The "when turned face [up, down]" hat often triggers twice when you turn the device over.

### Actual Behavior

The hat block should fire once when you turn the device over.

Smoothing the Z axis acceleration didn't help much. I think I'll try using a simple hysteresis approach (e.g. like a Schmitt trigger).

PR #2068, opened 2019-03-25
Stabilize Vernier "turned faced up/down" hat events

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/2067

### Proposed Changes

On the threshold used for detecting that the sensor is facing up or facing down, add an offset if the sensor is already facing up or down. 

### Reason for Changes

Having a different threshold in each direction ("hysteresis") helps to prevent rapid fluctuations. 

For example, when you turn the sensor over so it is facing down, there is often a brief spike in z axis acceleration, then it drops a bit, then it stabilizes at the acceleration due to gravity. Previously, this caused the acceleration value to cross the threshold twice, triggering the hat block twice. 

With this change, as soon as the acceleration value crosses the threshold, the threshold is offset downward. The small decrease after that is not enough to drop below the offset threshold, so it doesn't trigger again.

### DIFF
--- a/src/extensions/scratch3_gdx_for/index.js
+++ b/src/extensions/scratch3_gdx_for/index.js
@@ -81,6 +81,12 @@ const SHAKEN_THRESHOLD = 30;
 const FACING_THRESHOLD = 9;
 
 /**
+ * An offset for the facing threshold, used to check that we are no longer facing up.
+ * @type {number}
+ */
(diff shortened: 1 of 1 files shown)


evhan55:
This new behavior for `when turned face down` and `when turned face up` feels much better to me!  Things only trigger once as I would expect.  Using a threshold-offset in the code looks good to me, too.

👍

PR #2070, opened 2019-03-26
Add RateLimiter to EV3

### LINKED ISSUE #2015: Hardware Extensions: Add RateLimiter to EV3
When filing PR, include test case

evhan55:
### Resolves

- Resolves #2015: Hardware Extensions: Add RateLimiter to EV3


evhan55:
@ericrosenbaum This is a barebones implementation that I think covers all the same bases the implementation in the WeDo2 covers.  You are welcome to test it now as I sit on it for a day and think about whether it's missing something else!  It's possible that this is ready to go...

evhan55:
Hm `BTSendInterval` still needs to be used, maybe?

ericrosenbaum:
@evhan55 re: `BTSendInterval`, I think that interval is useful for any opcode that:
- sends a command
- does not already have a timeout that resolves it (e.g. motor on for seconds)

As far as I can tell there are no opcodes like that in the extension (motor turn and beep blocks have timeouts, and the set motor power block does not send a command, just sets extension state).

So, looks like it can be removed.

evhan55:
@ericrosenbaum Sweet, thanks! I will remove it.

ericrosenbaum:
@evhan55 in my testing so far, I found something kinda weird. This code sends a broadcast on a key event, turning on two motors. I think this is a typical attempt to use a keyboard-controlled driving robot, and it has some unexpected results:

Screen Shot 2019-03-26 at 11 11 30 AM

Here's what happens if you hold down the key continuously:

- At first it looks fine: you hold down the key, it triggers once turning on the motors briefly, then they turn off
- Then, the keyboard autorepeat (roughly 10 presses per second, on my mac) kicks in. Now, the broadcasts are arriving faster (~100ms) than the duration of the motor commands (250ms), so they are getting interrupted and restarted repeatedly.
- It still looks fine for several seconds: both motors are running continuously
- Motor B starts to stutter, pausing and restarting repeatedly
- Eventually, both motors completely stop
- Then (still holding down the key), both motors restart, but for some reason they are now at a lower power level!

Maybe we are hitting an EV3 hub internal limit, like it's queuing up commands, and has to pause? needs more investigation... maybe our rate limit should be lower?

evhan55:
@ericrosenbaum Whoa that sounds pretty intense! Yes maybe trying a slower rate would help with that?  Would it slow other operations down too much though?

ericrosenbaum:
Something I noticed:

`_pollValues` also calls the `send` function, which is rate limited. We should think about whether the polling should be excluded from the rate limiting (probably not, but it changes how I was imagining what is happening, and what the effective rate is that you can actually send commands using blocks!)

evhan55:
@ericrosenbaum oh I thought I had checked them all but somehow missed that one, good point! how would it change the effective rate? because there's only so much the EV3 can handle in terms of incoming messages?

ericrosenbaum:
@evhan55 no I just meant, basically, if you make a forever loop in scratch that runs 10 motor blocks per second, I would have thought we were sending 10 messages per second. But there are also always 10 polling requests being sent every second. So in fact, in that scenario, we are sending 20 messages per second.

evhan55:
@ericrosenbaum ah yep yep, makes sense.

evhan55:
@ericrosenbaum I am testing this PR again with the following set of stacks

Screen Shot 2019-05-17 at 11 40 48 AM

I updated the `BTSendRateMax` from `20` to `40` and that seems to actually fix the `when down arrow key pressed` issue you found on March 26th .  Is that fix the opposite of what we were guessing, that slowing down the rate limiter would help?  I tried changing `BTSendRateMax` to 10 and that made the symptom worse, but `40` seems to work well?

I am assigning back to you for review, or please let me know if I am missing other testing scenarios to improve my coverage of rate limiting tests. Thanks!

ericrosenbaum (approved):
LGTM! I tested on mac and windows with a similar set of stacks to what @evhan55 used above. In the edge case where "motor turn for seconds" blocks are being run at very high speed using "forever [broadcast]", on production, the hub locks up with the motors stuck on, but this PR prevents that problem. Overall the behavior seems fine.

PR #2083, opened 2019-04-02
Allow extensions to make buttons

cwillisf:
This PR depends on LLK/scratch-gui#4712 for the `MAKE_A_*` callback registrations.

### Proposed Changes

This change adds support for a "button" (pseudo-) block type to the extension interface. These entries are expected to have a `func` field but no `opcode`. For now the `func` must be one of a few existing callback keys handled by `scratch-blocks` but in the future we could open this up to allow connecting an extension method to a button.

### Reason for Changes

Converting variables, lists, and "My Blocks" into extensions means supporting the buttons used to create those items. This approach also allows us to experiment with other extension buttons in the future, if we'd like.

### Test Coverage

A unit test has been added to ensure that `BlockType.BUTTON` gets interpreted correctly.

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -375,16 +375,24 @@ class ExtensionManager {
             blockAllThreads: false,
             arguments: {}
         }, blockInfo);
-        blockInfo.opcode = this._sanitizeID(blockInfo.opcode);
+        blockInfo.opcode = blockInfo.opcode && this._sanitizeID(blockInfo.opcode);
         blockInfo.text = blockInfo.text || blockInfo.opcode;
 
-        if (blockInfo.blockType !== BlockType.EVENT) {
+        switch (blockInfo.blockType) {
+        case BlockType.EVENT:
+            if (blockInfo.func) {
+                log.warn(`Ignoring function "${blockInfo.func}" for event block ${blockInfo.opcode}`);
+            }
+            break;
+        case BlockType.BUTTON:
+            if (blockInfo.opcode) {
+                log.warn(`Ignoring opcode "${blockInfo.opcode}" for button with text: ${blockInfo.text}`);
+            }
+            break;
+        default:
             blockInfo.func = blockInfo.func ? this._sanitizeID(blockInfo.func) : blockInfo.opcode;
 
-            /**
-             * This is only here because the VM performs poorly when blocks return promises.
-             * @TODO make it possible for the VM to resolve a promise and continue during the same Scratch "tick"
-             */
+            // Avoid promise overhead if possible
             if (dispatch._isRemoteService(serviceName)) {
                 blockInfo.func = dispatch.call.bind(dispatch, serviceName, blockInfo.func);
             } else {
@@ -392,12 +400,11 @@ class ExtensionManager {
                 const func = serviceObject[blockInfo.func];
                 if (func) {
                     blockInfo.func = func.bind(serviceObject);
-                } else if (blockInfo.blockType !== BlockType.EVENT) {
+                } else {
                     throw new Error(`Could not find extension block function called ${blockInfo.func}`);
                 }
             }
-        } else if (blockInfo.func) {
-            log.warn(`Ignoring function "${blockInfo.func}" for event block ${blockInfo.opcode}`);
+            break;
         }
 
         return blockInfo;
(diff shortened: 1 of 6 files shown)


towerofnix:
>Converting ... "My Blocks" into extensions

:eyes: This shows that extensions definitely have potential in this direction..!

test/unit/extension_conversion.js
```diff
@@ -12,6 +12,11 @@ const testExtensionInfo = {
     id: 'test',
     name: 'fake test extension',
     blocks: [
+        {
+            func: 'CREATE_VARIABLE',
```
kchadha: could you also add this to the `coreExample` extension for easy testing? =D

src/extension-support/extension-manager.js
```diff
@@ -375,29 +375,36 @@ class ExtensionManager {
...
+                log.warn(`Ignoring function "${blockInfo.func}" for event block ${blockInfo.opcode}`);
+            }
+            break;
+        case BlockType.BUTTON:
+            if (blockInfo.opcode) {
+                log.warn(`Ignoring opcode "${blockInfo.opcode}" for button with text: ${blockInfo.text}`);
+            }
+            break;
+        default:
             blockInfo.func = blockInfo.func ? this._sanitizeID(blockInfo.func) : blockInfo.opcode;
```
kchadha, line 397: Should we log a warning or an error if opcode is undefined here? Or if both opcode and func are undefined? Alternatively, if they are undefined, should we default to a built in no-op function? Maybe we do that somewhere else...

kchadha: Ah, never mind, I see we throw an error below. Nevertheless, I'll leave the question about defaulting to a no-op function here.

test/integration/internal-extension.js
```diff
@@ -83,21 +83,24 @@ test('load sync', t => {
     t.ok(vm.extensionManager.isExtensionLoaded('coreExample'));
 
     t.equal(vm.runtime._blockInfo.length, 1);
-    t.equal(vm.runtime._blockInfo[0].blocks.length, 1);
+    t.equal(vm.runtime._blockInfo[0].blocks.length, 2);
```
kchadha, line 88: Maybe add a comment here that there are 2 items but only one actual block. The other one is a button.

kchadha (approved):
LGTM! Just some minor questions/comments.

src/extension-support/extension-manager.js
cwillisf: Good idea -- done. Note that the exception thrown here will be caught and reported by some code in `_prepareExtensionInfo` which dumps the whole `blockInfo` so there's no need to add any `blockInfo` fields in the exception message.

test/unit/extension_conversion.js
```diff
@@ -153,9 +163,15 @@ test('registerExtensionPrimitives', t => {
     runtime.on(Runtime.EXTENSION_ADDED, blocksInfo => {
         t.equal(blocksInfo.length, testExtensionInfo.blocks.length);
 
+        blocksInfo.forEach(blockInfo => {
+            // `true` here means "either an object or a non-empty string but definitely not null or undefined"
+            t.true(blockInfo.info, 'Every block and pseudo-block must have a non-empty "info" field');
```
kchadha, line 168: I think in other VM unit tests we use `t.ok` for the truthy test. I believe that's just a synonym for t.true... ¯\_(ツ)_/¯

cwillisf: Yeah -- I felt like `t.true` was semantically more accurate in this case since usually we do `t.ok(something == desired)`, and in this case I just want to check that it's "truthy". Really I kind of want `t.typeIsOneOf(blockInfo.info, ['object','string']);` but that doesn't exist.

I guess it could be `t.ok(-1 != ['object', 'string'].indexOf(typeof blockInfo.info))` but that seems even harder to read than the other options :(

Also, obligatory "you dropped this: \\"

towerofnix: `t.ok(['object', 'string'].includes(typeof blockInfo.info))` perhaps?

kchadha: > Really I kind of want `t.typeIsOneOf(blockInfo.info, ['object','string']);` but that doesn't exist.

Agreed.

> I guess it could be `t.ok(-1 != ['object', 'string'].indexOf(typeof blockInfo.info))` but that seems even harder to read than the other options :(

Also agreed.

> Also, obligatory "you dropped this: \"

I think GitHub ate it! It's there if I try to edit the raw comment text but idk what markdown feature GitHub thinks I'm trying to use.

PR #2095, opened 2019-04-05
Fix #2085 by handling discarded hub commands

### LINKED ISSUE #2085: BOOST extension: stalled motor causes rotation block to hang
### Expected Behavior

When you run the LEGO BOOST "turn motor A for 1 rotations" block, if the motor is physically prevented from rotating (e.g. if it is stalled by too much load), the block should eventually stop yielding, so the stack can continue to the next block. We will need to discuss the exact desired behavior here.

### Actual Behavior

If the motor is stalled, the block appears to yield indefinitely.

### Steps to Reproduce

Use a "turn motor A for 1 rotations" followed by another block in the stack. Stop the motor from completing the rotation, and note that the stack never continues.

knandersen:
This was caused because the case for BoostMessage.PORT_FEEDBACK didn't handle the BoostPortFeedback.DISCARDED type, which corresponds to a command failing on the Boost hub.

### Resolves

https://github.com/LLK/scratch-vm/issues/2085

### Proposed Changes

the case now declares a `commandCompleted` constant that has a truthy value if a command has either been completed or discarded.

### Reason for Changes

Stalling motor-commands were causing blocks to hang.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -943,8 +943,9 @@ class Boost {
             const motor = this.motor(portID);
             if (motor) {
                 motor._status = feedback;
-                if (feedback === (BoostPortFeedback.COMPLETED ^ BoostPortFeedback.IDLE) &&
-                    motor.pendingPromiseFunction) {
+                // Makes sure that commands resolve both when they actually complete and when they fail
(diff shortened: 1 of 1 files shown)


evhan55:
This looks fixed to me!  Tested on macOS.

evhan55 (approved):
Seems good!

knandersen:
This pull request was filed incorrectly. Filed correctly in: #2099

PR #2096, opened 2019-04-05
Fix 2087, 2088, and 2089 by properly setting parameters in rotation-commands and not using command feedback when relying on scratch-vm for motor timing

### LINKED ISSUE #2087: BOOST extension: can't change motor direction while it's running
### Expected Behavior

While a motor is running, having been started in any of the ways available (motor on, on for seconds, or on for rotations), it should be possible while it is running to change the direction of the motor using the "set motor direction" block. In the case of "on for seconds", the motor should continue running for the remainder of the time. In the case of "on for rotations", the motor should (at least approximately) complete the rotations.
 
### Actual Behavior

Not currently working.

### LINKED ISSUE #2088: BOOST extension: can't change motor speed while it's running
### Expected Behavior

While a motor is running, having been started in any of the ways available (motor on, on for seconds, or on for rotations), it should be possible while it is running to change the speed of the motor using the "set motor speed" block. In the case of "on for seconds", the motor should continue running for the remainder of the time. In the case of "on for rotations", the motor should (at least approximately) complete the rotations.

### Actual Behavior

Not currently working.

### LINKED ISSUE #2089: BOOST extension: default motor speed should be 50%?
In the EV3 extension, we set the initial value for the power of the motors to 50%, instead of 100% (i.e. not the default value for the input of the block, but the actual value set internally when the extension loads). We should consider doing the same for the BOOST. 

This would make the initial experience of the motor moving a bit more gentle. It also gives you room for an increase, so that setting it to 100% actually causes it to change.

knandersen:
Resolves #2087 - Can't change motor direction while it's running
Resolves #2088 - Can't change motor power while it's running
Resolves #2089 - Default motor speed should be 50%?

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -33,10 +33,13 @@ const BoostBLE = {
 };
 
 /**
- * Boost Motor Max Power.
+ * Boost Motor Max Power Add. Defines how much more power than the target speed
+ * the motors may supply to reach the target speed faster.
+ * Lower number == softer, slower reached target speed.
+ * Higher number == harder, faster reached target speed.
  * @constant {number}
  */
-const BoostMotorMaxPower = 100;
+const BoostMotorMaxPowerAdd = 10;
 
 /**
  * A time interval to wait (in milliseconds) in between battery check calls.
@@ -280,7 +283,7 @@ class BoostMotor {
          * @type {number}
          * @private
          */
-        this._power = 100;
+        this._power = 50;
 
         /**
          * This motor's current relative position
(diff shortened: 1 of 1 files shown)


evhan55:
As part of testing this PR, I should test the issue I found in the original Boost PR for the jittering found when switching from power setting `10%` to `100%`

evhan55:
I am testing this branch and have the following results so far:

✅ 
- Changing motor direction works for `turn motor [A] for [#] seconds`
- Changing motor direction works for `turn motor [A] for [#] rotations`
- Motor default is set to 50%

❌ 
- Changing motor direction during a `turn motor [A] on` successfully reverses the motor, but then stops after some time
- Changing motor power during all three kinds of motor blocks seems to set the motor to a low power and then stop (I was testing 20% -> start motor -> set motor to 100% -> motor slows down and/or stops, instead of becoming faster)

knandersen:
This pull request was filed incorrectly. Filed correctly in: #2100

PR #2097, opened 2019-04-05
Fix #2086 by resolving pending promise even when power set to 0%

### LINKED ISSUE #2086: BOOST extension: behavior of "rotations" block for 0% power?
We should consider what the desired behavior is for the "turn motor A for 1 rotations" block, in the case where the motor power has been set at or close to 0%, which will prevent the motor from completing the rotation. 

This may be solved by our solution to https://github.com/LLK/scratch-vm/issues/2085

knandersen:
Resolves #2086: rotation-based motor-commands hang when power set to 0%

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -471,7 +471,11 @@ class BoostMotor {
      * @param {number} direction - rotate in this direction
      */
     turnOnForDegrees (degrees, direction) {
-        if (this._power === 0) return;
+        if (this._power === 0) {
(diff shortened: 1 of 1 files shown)


evhan55:
This seems good to me now!  The `turn motor [A] for [#] rotations` doesn't seem to hang anymore at power level `0%`, and it seems okay for power levels close to `0%`.

evhan55 (approved):
LGTM!

knandersen:
This pull request was filed incorrectly. Filed correctly in: #2101

Issue #2091, opened 2019-04-02
BOOST extension: "when color seen" hat not working


ericrosenbaum:
The "when color seen" hat does not seem to trigger at all. I know this was working at some point! maybe the code just got out of sync.  

In general, it looks like some cleanup may be needed in how color is represented in the code, to help distinguish between the numeric values reported by the sensor, the values used internally in the menu, the localized names used on the menu, and the localized names reported by the reporter.

I would also suggest renaming the `_colorBucket` data structure, since we are using a "token bucket" in the rate limiter which is quite different- this approach seems more like a sliding window? but it could just be called `_colorSamples` or `_colorArray`.

evhan55:
oooh `colorBucket` rename idea sounds good

knandersen:
Yeah I agree about the renaming (addressed in https://github.com/LLK/scratch-vm/pull/2096), and that a cleanup between numeric and string representation could be useful.

The hat-block works for me though! @ericrosenbaum can you give me steps to repro? If it still doesn't work, can you check and see if the reporter works? just to see if it's the entire sensor not working for you.

I do get some false positives for `black` (kinda considering if we should remove that color all together, thoughts?).

evhan55:
Here is a video of me testing it yesterday showing some instability:
https://photos.app.goo.gl/qzp2RYTW2RJmDdjz8

And here is a photo of a red candle next to a red LEGO brick, the red candle never reported `red`:
![img_20190405_164606_1024](https://user-images.githubusercontent.com/699840/55671406-b72ea100-585d-11e9-853a-64bdc16366ca.jpg)

Talking with @knandersen, he believes this may be best case scenario but I will let him explain more.

evhan55:
@knandersen Said he will take this on so I will assign to him.  Please feel free to reassign to me if you'd like me to take it on instead.

knandersen:
@ericrosenbaum @evhan55 

Just a few notes about the sensor:

- It is designed to detect certain LEGO bricks (shiny plastic): Bright Red, Blue, Dark Green, Yellow, White, and Black. 
- Its range is somewhere between 5-15mm.

After having tested the sensor this morning I feel like it's actually performing as well as I would ever expect it to. I do however feel like I'm biased when testing the sensor, because I know how to use it 'properly', but others naturally may expect it to perform differently.

I'd like to have a discussion (including some hands-on demo) around this, because I do see some things we could do. Examples:
[...]

Issue #1929, opened 2019-01-16
Hardware Extensions: separate extension cleanup and disconnect 


ericrosenbaum:
Per discussion with @evhan55, we decided to make each hardware extension have a "clean up" function, which is separate from the disconnect function. This will allow us to pass the clean up function as the disconnect callback that we pass to the socket object. Then, in `_triggerDisconnectError` (formerly `_sendDisconnectError`) we can always call `disconnect`, and remove the branching.

PR #2100, opened 2019-04-08
Fix 2087, 2088, and 2089 by properly setting parameters in rotation-commands and not using command feedback when relying on scratch-vm for motor timing

### LINKED ISSUE #2087: BOOST extension: can't change motor direction while it's running

### LINKED ISSUE #2088: BOOST extension: can't change motor speed while it's running

### LINKED ISSUE #2089: BOOST extension: default motor speed should be 50%?

knandersen:
Resolves #2087 - Can't change motor direction while it's running
Resolves #2088 - Can't change motor power while it's running
Resolves #2089 - Default motor speed should be 50%?

Clarifies the renaming aspect mentioned in #2091


evhan55:
This tests ok for me after the latest commit!

- Can change direction on motor for all three motor on blocks (`seconds/rotations/on`)
- Can change power on motor for all three motor blocks (`seconds/rotations/on`)
- Motor default is at 50%

src/extensions/scratch3_boost/index.js
```diff
@@ -590,15 +594,16 @@ class Boost {
...
+            previousColor: BoostColor.NONE
         };
 
         /**
          * An array of values from the Boost Vision Sensor.
          * @type {Array}
          * @private
          */
-        this._colorBucket = [];
+        this._colorSamples = [];
```
evhan55, line 606: Does this change belong in another PR, or should the issue this resolves be mentioned in the comment for this PR?

evhan55:
There is some code in this PR related to color sensing, should it be mentioned in the PR comment and reference an issue so I can test it?

src/extensions/scratch3_boost/index.js
knandersen: Yeah thanks for pointing that out @evhan55. It's from #2091!

src/extensions/scratch3_boost/index.js
```diff
@@ -1926,10 +1934,10 @@ class Scratch3BoostBlocks {
         case BoostColorLabel.ANY:
             if (Object.keys(BoostColor).find(key => BoostColor[key])
                 .toLowerCase() !== this.getColor()) {
-                if (this.getColor() === this._peripheral.oldColor) {
```
evhan55, line 1930: Was this a bug before and should have been `this._peripheral._sensors.oldColor`? (Just trying to understand the change completely)

evhan55:
The color name change should be okay, but I am not testing it since the color behavior is flaky at the moment in general for me, especially without LEGO bricks for testing.  I would say the color name change in this PR should be fine except for line #1929 which actually changes behavior a tiny bit.  If that is just a small bug fix, then it could land with this PR.  However, if it's important for me to test the change at line #1929, then I would maybe put all the color name change plus that bug fix in line #1929 in another PR later that addresses the color sensor changes if any will get made.  What do you think @knandersen ?

src/extensions/scratch3_boost/index.js
knandersen: yes, you're right, it should be `this._peripheral._sensors.previousColor`(also updated the naming). It was addressed in commit e3cdbff

evhan55:
This feels mostly good to me now! 🎉 

- Can change motor direction for all motor on/seconds/rotations
- Can change motor power for all motor on/seconds/rotations
- Default was 50%
- Color bucket name change seems fine

Note:
- There is a tiny issue with rotations maybe not completing 100%, but I would rather find a repro for it and to resolve in a future PR, what do you think?

evhan55 (approved):
Looks mostly fine to me!  I think it might be good to merge at this step for testing at tomorrow's workshop.

knandersen:
Thanks @evhan55!
I had a really good chat with @ericrosenbaum earlier today about cleaning up handling what state the motor is in, since the code right now is error-prone and very hard to read.
I've tried to put our thinking in writing:

Since the motors can either really be off, on forever, on for an amount of time, or on for an amount of rotation, the idea would be to create a BoostMotorStatus-enum, with these values:

```
BoostMotorStatus {
  OFF: 0
  ON_FOREVER: 1
  ON_FOR_TIME: 2
  ON_FOR_ROTATION: 3
}
```

`BoostMotor.isOn()` would be replaced by `BoostMotor.getStatus()` which would get `BoostMotor._status` which can have a value from the `BoostMotorStatus`-enum.
A new function called `BoostMotor.setStatus(boostMotorStatus)` would be created, which would be responsible for not only giving the motor a status, but depending on the value supplied to it will clear the relevant motor-states, i.e.:

```
setStatus(state) {
  switch(state){
    case BoostMotorState.OFF:
      // Clear timed- and rotation-based motor-states
    case BoostMotorState.ON_FOREVER:
      // Clear rotation-based and timed motor-states
    case BoostMotorState.ON_FOR_TIME:
      // Clear rotation-based motor-states
    case BoostMotorState.ON_FOR_ROTATION:
      // Clear timed motor-states
  }
}
```

Thoughts? cc @cwillisf

evhan55:
@knandersen 

Clean up sounds good as long as it keeps the functionality.  It would be easier for me to see if it makes the code more legible once this proposed enum is actually implemented.

One extra thought: my opinion would be to land the functionality as is, to keep a snapshot of "working code" which was tricky to get right. Then, make a new PR with the clean up steps, hopefully feeling safer to do that refactor knowing that you can always revert to working code.  Just my two cents!

PR #2101, opened 2019-04-08
Fix #2086 by resolving pending promise even when power set to 0%

### LINKED ISSUE #2086: BOOST extension: behavior of "rotations" block for 0% power?

knandersen:
Resolves #2086: rotation-based motor-commands hang when power set to 0%

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -471,7 +471,7 @@ class BoostMotor {
      * @param {number} direction - rotate in this direction
      */
     turnOnForDegrees (degrees, direction) {
-        if (this._power === 0) {
+        if (this.power === 0) {
             this.pendingPromiseFunction();
             return;
         }
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -471,7 +471,11 @@ class BoostMotor {
      * @param {number} direction - rotate in this direction
      */
     turnOnForDegrees (degrees, direction) {
-        if (this._power === 0) return;
+        if (this._power === 0) {
+            this.pendingPromiseFunction();
+            return;
+        }
+        
         degrees = Math.max(0, degrees);
 
         const cmd = this._parent.generateOutputCommand(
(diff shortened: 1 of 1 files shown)


evhan55:
Hi, I am testing this again in the new PR and finding a small issue I had seen before in the old PR but wasn't able to reproduce.  I am able to reproduce it now:

1. `set motor A power to 100%`
2. `turn motor A for 1 rotations`
3. `set motor A power to 1%`
4. `turn motor A for 1 rotations`

**Expected**: motor slow moves one rotation
**Actual**: motor moves slowly, stalls for a second, and then finishes the rotation slowly

I get the stalling to happen pretty reliably with those steps above

knandersen:
> Hi, I am testing this again in the new PR and finding a small issue I had seen before in the old PR but wasn't able to reproduce. I am able to reproduce it now:
> 
> 1. `set motor A power to 100%`
> 2. `turn motor A for 1 rotations`
> 3. `set motor A power to 1%`
> 4. `turn motor A for 1 rotations`
> 
> **Expected**: motor slow moves one rotation
> **Actual**: motor moves slowly, stalls for a second, and then finishes the rotation slowly
> 
> I get the stalling to happen pretty reliably with those steps above

Thanks so much for catching this @evhan55. It seems like I keep moving back and forth between two problems. I've described it in a commit for a related PR/issue: https://github.com/LLK/scratch-vm/pull/2100/commits/ecbbacd4c099a4c018b50b510e34cbf53a07b9cd

evhan55:
@knandersen Ah, interesting!  Seems like a tricky dilemma!
Might be worth showing to Eric and Andrew, et. al.

evhan55:
Looks like the performance issue I am seeing will be addressed in a future issue: https://github.com/LLK/scratch-vm/issues/2104.  So, this seems good to land for now!

evhan55:
Okay just tested with the very latest commits again, all looks good and power 0% doesn't make rotation block hang.  Possible performance issues to be addressed later: https://github.com/LLK/scratch-vm/issues/2104.

evhan55 (approved):
LGTM!

PR #2102, opened 2019-04-09
Pass block info to extension methods

cwillisf:
### Resolves

This is a supporting change for the extensionification project.

### Proposed Changes

Extension methods now receive a third argument which contains the section of the `getInfo` results which the extension provided for the block being executed.

### Reason for Changes

Sometimes a single extension method needs to service several different instances of the same opcode. This can happen with variables or custom procedures, for example. This change allows the extension method to inspect the `blockInfo` for instance data, including arbitrary extension-specific properties if necessary.

### Test Coverage

There is now a check check for correct `blockInfo` being passed into an extension method in `test/integration/internal-extension.js`.

### DIFF
--- a/src/extension-support/extension-manager.js
+++ b/src/extension-support/extension-manager.js
@@ -390,25 +390,39 @@ class ExtensionManager {
                 log.warn(`Ignoring opcode "${blockInfo.opcode}" for button with text: ${blockInfo.text}`);
             }
             break;
-        default:
-            if (blockInfo.opcode) {
-                const funcName = blockInfo.func ? this._sanitizeID(blockInfo.func) : blockInfo.opcode;
-
-                // Avoid promise latency unless necessary
-                if (dispatch._isRemoteService(serviceName)) {
-                    blockInfo.func = (args, util) => dispatch.call(serviceName, funcName, args, util, blockInfo);
-                } else {
-                    const serviceObject = dispatch.services[serviceName];
-                    if (!serviceObject[funcName]) {
-                        // The function might show up later as a dynamic property of the service object
-                        log.warn(`Could not find extension block function called ${funcName}`);
-                    }
-                    blockInfo.func = (args, util) => serviceObject[funcName](args, util, blockInfo);
-                }
-            } else {
+        default: {
+            if (!blockInfo.opcode) {
                 throw new Error('Missing opcode for block');
             }
+
+            const funcName = blockInfo.func ? this._sanitizeID(blockInfo.func) : blockInfo.opcode;
+
+            const getBlockInfo = blockInfo.isDynamic ?
+                args => args && args.mutation && args.mutation.blockInfo :
+                () => blockInfo;
+            const callBlockFunc = (() => {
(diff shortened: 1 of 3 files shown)


src/extension-support/extension-manager.js
```diff
@@ -391,23 +391,22 @@ class ExtensionManager {
             }
             break;
         default:
-            if (!blockInfo.opcode) {
-                throw new Error('Missing opcode for block');
-            }
-
-            blockInfo.func = blockInfo.func ? this._sanitizeID(blockInfo.func) : blockInfo.opcode;
+            if (blockInfo.opcode) {
```
kchadha: Any reason for rearranging this conditional (from the previous structure of falsey check and early exit)?

src/extension-support/extension-manager.js
```diff
@@ -391,23 +391,22 @@ class ExtensionManager {
...
-                    blockInfo.func = func.bind(serviceObject);
+                // Avoid promise latency unless necessary
+                if (dispatch._isRemoteService(serviceName)) {
+                    blockInfo.func = (args, util) => dispatch.call(serviceName, funcName, args, util, blockInfo);
                 } else {
-                    throw new Error(`Could not find extension block function called ${blockInfo.func}`);
+                    const serviceObject = dispatch.services[serviceName];
+                    if (!serviceObject[funcName]) {
+                        // The function might show up later as a dynamic property of the service object
+                        log.warn(`Could not find extension block function called ${funcName}`);
```
kchadha: Should this function exit after warning here? Below, the function RHS of the arrow function is just going to get bound to `undefined(args, util, blockInfo)` right?

kchadha (commented):
Some minor comments and questions, otherwise LG!

src/extension-support/extension-manager.js
cwillisf: It's kind of silly... but this isn't legal JavaScript:
```js
default:
    if (!blockInfo.opcode) {
        throw new Error('Missing opcode for block');
    }
    const funcName = "stuff"; // it's not legal to make a new `const` here
```

You can do it if you introduce a new scope with `{}` which sort of naturally led me to the arrangement proposed here...

src/extension-support/extension-manager.js
cwillisf: Since it's looking up `serviceObject[funcName]` in the body of the (anonymous) function, it's possible that `serviceObject[funcName]` will become a valid function call between now and the first time it gets called. This could happen if the extension dynamically generates methods and monkey-patches itself, for example, especially if it needs to download some data first.

I don't expect this to be common but technically the way we make "remote" service calls doesn't even do this check at all, so I figured that if it's allowed there it should be allowed here. 🤷‍♂️

src/extension-support/extension-manager.js
kchadha: Hmmmm okay. I think I'd prefer adding the new block of scope only because it's easier to read an early exit vs. keeping track of which block of code corresponds to which part of the if/else etc. Esp. since there are nested ifs involved here.

PR #2107, opened 2019-04-09
Fix BOOST whenColor hat

### LINKED ISSUE #2091: BOOST extension: "when color seen" hat not working

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/2091

### Proposed Changes

This is a quick fix to make the BOOST "when color seen" hat work for the options other than "any".

### Reason for Changes

This line looks like it was a typo: `return this.getColor() === color;` That definitely never worked, because `color` is the color utility object. 

The "any" option is handled separately, and seems to work.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1925,7 +1925,7 @@ class Scratch3BoostBlocks {
      * @return {boolean} - true if the tilt sensor is tilted past a threshold in the specified direction.
      */
     whenColor (args) {
-        return this._isColor(args.COLOR);
+        return this._isColor(args);
     }
 
     /**
@@ -1944,7 +1944,7 @@ class Scratch3BoostBlocks {
      * @private
      */
     _isColor (args) {
-        switch (args) {
+        switch (args.COLOR) {
         case BoostColorLabel.ANY:
             if (Object.keys(BoostColor).find(key => BoostColor[key])
                 .toLowerCase() !== this.getColor()) {
@@ -1956,7 +1956,7 @@ class Scratch3BoostBlocks {
             }
             break;
         default:
-            return this.getColor() === color;
+            return this.getColor() === args.COLOR;
         }
     }
 


evhan55:
I tested this with the following code:

Screen Shot 2019-04-10 at 9 29 32 AM

And the following brick pattern:
IMG_20190410_093003

And the sprite would turn when seeing both `'red'` and `'yellow'` if I held the sensor about 1cm away from the bricks.  I also confirmed that those colors were being seen by monitoring the reporter, and it works!

Seems like a good fix to me!

src/extensions/scratch3_boost/index.js
```diff
@@ -1944,7 +1944,7 @@ class Scratch3BoostBlocks {
      * @private
      */
     _isColor (args) {
-        switch (args) {
```
evhan55, line 1947: It was switching on an array of arguments before?

evhan55 (approved):
lgtm!

evhan55:
Thanks @ericrosenbaum for clarification!  Also testing the non-any options seen here, with the same brick pattern as above:

Screen Shot 2019-04-10 at 9 40 07 AM
Screen Shot 2019-04-10 at 9 40 11 AM
Screen Shot 2019-04-10 at 9 40 25 AM

The first two code snippets correctly trigger when the sensor is near the red and yellow bricks, respectively!
Also, the third code snippet does not trigger at all.

Seems like the non-any options are also working well!

evhan55:
lgtm again!

knandersen:
Looks great! I also really like using `args` and `args.COLOR` throughout to avoid ambiguity, which is what I suspect happened when using `color`!

As @ericrosenbaum pointed out however, there's still a matter of fixing the issue mentioned here: https://github.com/LLK/scratch-vm/issues/2091#issuecomment-480844119

Should we file that as a separate issue or change this PR to only partly resolve #2091?

src/extensions/scratch3_boost/index.js
ericrosenbaum: yup, that was also breaking the "any" condition I think, which I didn't notice

ericrosenbaum:
Thanks! I will file an issue about the problem with multiple "when color any" hats

PR #2111, opened 2019-04-11
Fix #2103 and #2110 by clearly defining motor state and how it's modified

### LINKED ISSUE #2103: BOOST extension: Motor state handling can be improved
Based on the work on #2100, it seems like the new possible motor functionality that Boost introduces requires a better way of handling the change of motor states, e.g. changing between time-based and rotation-based commands.

I had a really good chat with @ericrosenbaum earlier today about cleaning up handling what state the motor is in, since the code right now is error-prone and very hard to read.
I've tried to put our thinking in writing:

Since the motors can either really be off, on forever, on for an amount of time, or on for an amount of rotation, the idea would be to create a BoostMotorStatus-enum, with these values:

```
BoostMotorStatus {
  OFF: 0
  ON_FOREVER: 1
  ON_FOR_TIME: 2
  ON_FOR_ROTATION: 3
}
```

`BoostMotor.isOn()` would be replaced by `BoostMotor.getStatus()` which would get `BoostMotor._status` which can have a value from the `BoostMotorStatus`-enum.
A new function called `BoostMotor.setStatus(boostMotorStatus)` would be created, which would be responsible for not only giving the motor a status, but depending on the value supplied to it will clear the relevant motor-states, i.e.:

```
setStatus(state) {
  switch(state){
    case BoostMotorState.OFF:
      // Clear timed- and rotation-based motor-states
    case BoostMotorState.ON_FOREVER:
      // Clear rotation-based and timed motor-states
    case BoostMotorState.ON_FOR_TIME:
      // Clear rotation-based motor-states
    case BoostMotorState.ON_FOR_ROTATION:
      // Clear timed motor-states
  }
}
```

### LINKED ISSUE #2110: BOOST extension: Changing power/direction immediately after timed motor commands makes it run forever
### Expected Behavior

Changing motor speed/direction immediately after a timed motor command should only change the state of the motor.
Screenshot 2019-04-11 at 09 02 53

### Actual Behavior

Currently it will make the motor run forever. It seems to be related to the issues described in #2103, where motor-state can be ambiguous and isn't being reset properly between commands.

### Steps to Reproduce

See blocks above

knandersen:
Resolves #2103: Motor state handling can be improved
Resolves #2110 : Changing motor speed/direction after running for timed motor commands makes it run forever

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -415,20 +415,9 @@ class BoostMotor {
      * @param {BoostMotorState} value - set this motor's state.
      */
     set status (value) {
-        switch (value) {
-        case BoostMotorState.OFF:
-        case BoostMotorState.ON_FOREVER:
-            this._clearRotationState();
-            this._clearTimeout();
-            break;
-        case BoostMotorState.ON_FOR_TIME:
-            this._clearRotationState();
-            break;
-        case BoostMotorState.ON_FOR_ROTATION:
-            this._clearRotationState();
-            this._clearTimeout();
-            break;
-        }
+        // Clear any time- or rotation-related state from motor and set new status.
+        this._clearRotationState();
+        this._clearTimeout();
         this._status = value;
     }
 
@@ -472,7 +461,7 @@ class BoostMotor {
      * @private
      */
     _turnOn () {
-        if (this._power === 0) return;
+        if (this.power === 0) return;
         const cmd = this._parent.generateOutputCommand(
             this._index,
             BoostOutputExecution.EXECUTE_IMMEDIATELY,
@@ -1672,9 +1661,7 @@ class Scratch3BoostBlocks {
         return new Promise(resolve => {
             this._forEachMotor(args.MOTOR_ID, motorIndex => {
                 const motor = this._peripheral.motor(motorIndex);
-                if (motor) {
-                    motor.turnOnFor(durationMS);
-                }
+                if (motor) motor.turnOnFor(durationMS);
             });
 
             // Run for some time even when no motor is connected
@@ -1732,9 +1719,7 @@ class Scratch3BoostBlocks {
         // TODO: cast args.MOTOR_ID?
         this._forEachMotor(args.MOTOR_ID, motorIndex => {
             const motor = this._peripheral.motor(motorIndex);
-            if (motor) {
-                motor.turnOnForever();
-            }
+            if (motor) motor.turnOnForever();
         });
 
         return new Promise(resolve => {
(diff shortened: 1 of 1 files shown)


knandersen:
Two things I'm wondering:

1. Ideally the motor state should be set in the `BoostMotor`-functions rather than in the opcodes, since the former are "closer" to the actual BLE-commands getting sent. This, however, is hard to do, because of the interaction between `turnOnFor()` and `turnOn()`, since the latter doesn't know that it's called by the former, and thus can't know what motor-state to change to.

2. Are there actually scenarios where we're interested in retaining motor-state relating to either timed- or rotation-based behavior? With this PR, `BoostMotor.status = value` has a `switch/case` that determines whether to clear motor state relating to `time` or `rotation`, but perhaps this is unnecessary, and there should just be a `_clearMotorState()`-function that resets both, since a motor can only be doing one thing at a time.

Thoughts? cc @ericrosenbaum @evhan55

ericrosenbaum:
Great questions! From our discussion:

1. I like the idea of the BoostMotor object managing the motor state. One solution for the issue with `turnOnFor()` would be to make a new function like `turnOnForever`, and call `turnOn` internally from the motor object only (i.e. make `turnOn` "private"). Then, `turnOnFor` and `turnOnForever` would set the motor state, and both would call `turnOn` (which would not).

2. Yeah! seems like a good idea have a function that clears the motor state.

evhan55:
Tested this branch and:

- Fixes the issue in #2110 for me!  I was able to reproduce it using the STR stack from #2110 without this change, and then I saw it fixed with this code.
- The new motor state handling feels very readable to me.  I only have one question from reading through it a couple of times, which I commented in the code directly.

Looks good to me!

src/extensions/scratch3_boost/index.js
```diff
@@ -439,10 +457,11 @@ class BoostMotor {
...
 
     /**
-     * Turn this motor on indefinitely.
+     * Turn this motor on indefinitely
+     * @private
      */
-    turnOn () {
-        if (this._power === 0) return;
+    _turnOn () {
+        if (this.power === 0) return;
```
evhan55, line 464: `this.power === 0;` is distinct from `BoostMotorState.OFF` ?

src/extensions/scratch3_boost/index.js
```diff
@@ -1729,15 +1763,18 @@ class Scratch3BoostBlocks {
...
-                if (motor.isOn) {
-                    if (motor.pendingTimeoutDelay) {
-                        motor.turnOnFor(motor.pendingTimeoutStartTime + motor.pendingTimeoutDelay - Date.now());
-                    } else if (motor.pendingPromiseFunction) {
-                        const p = Math.abs(motor.pendingPositionDestination - motor.position);
-                        motor.turnOnForDegrees(p, Math.sign(p));
-                    } else {
-                        motor.turnOn();
-                    }
+                switch (motor.status) {
```
evhan55, line 1766: This refactor feels really good!

evhan55 (approved):
Other than wondering how `this.power === 0` is different from `BoostMotorState.OFF`, everything lgtm!

src/extensions/scratch3_boost/index.js
knandersen: Yeah, `this.power` describes the motor's power between 0-100, where as `BoostMotorState.OFF` is an enumerated value.

PR #2115, opened 2019-04-12
Fix #2108 by making setMotorPower() and setMotorDirection() yield for a tick

### LINKED ISSUE #2108: BOOST extension: should "set motor speed" block yield for a tick?
### Expected Behavior

I expected to be able to use a stack like this to smoothly ramp the motor speed up and down:

Screen Shot 2019-04-09 at 5 56 48 PM

### Actual Behavior

The stack above produces an erratic behavior, because none of the blocks yield, so they run at thousands of blocks per sec, instead of the expected 30/sec.

I was able to get the desired buttery-smooth speed ramping behavior using this stack (the "set y" block yields for a tick because of the need for redraw, but I could have used e.g. "wait 0", or any other yielding block): 

Screen Shot 2019-04-09 at 5 55 03 PM

This suggests to me that it would be helpful if this opcode yielded for a tick (e.g. using `return Promise.resolve();`). There may be some scenario where this is not desirable though? We should discuss.

knandersen:
### Resolves

#2108

### Proposed Changes

Adds a `return Promise.resolve()` to the `setMotorPower()` and setMotorDirection()`-opcodes

### Reason for Changes

Before, as described in #2108, changing motor power or direction very fast would make something like ramping a motor speed cause the motors to behave very erratically, and could potentially result in spamming BLE commands to the hardware.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1722,6 +1722,7 @@ class Scratch3BoostBlocks {
      * @param {object} args - the block's arguments.
      * @property {MotorID} MOTOR_ID - the motor(s) to be affected.
      * @property {int} POWER - the new power level for the motor(s).
+     * @return {Promise} - returns a promise to make sure the block yields.
      */
     setMotorPower (args) {
         // TODO: cast args.MOTOR_ID?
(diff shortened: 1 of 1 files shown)


evhan55:
I have tested this with the STR stack from #2108 and it tests great for me!
Without this change, I saw the erratic behavior.  With the change in this PR, the motor continuously ramped up and down in speed.

evhan55 (approved):
lgtm!

PR #2119, opened 2019-04-16
BOOST color sensing fixes

### LINKED ISSUE #2090: BOOST extension: color names should be localized
### Expected Behavior

All text presented by an extension to the user should be localized.

### Actual Behavior

Looks like the color names used by the color reporter are not localized.

### LINKED ISSUE #2112: BOOST extension: using multiple "when color any" hats does not work
### Expected Behavior

It should be possible to use the "when color seen" hat with the "any" option selected in two different stacks at the same time.

### Actual Behavior

If you make two stacks started by "when color any", only one of them will start when the color changes:

Screen Shot 2019-04-11 at 1 29 01 PM

### LINKED ISSUE #2113: BOOST extension: "when color any" hat does not trigger twice in a row for the same color
### Expected Behavior

It should be possible to show the color sensor any sequence of colors, and have the "when color any" hat trigger on each transition between two colors, or between "none" and a color.

### Actual Behavior

If you show the sensor something red, then take it away, then show it again, it triggers the first time, but not the second time.

I noticed that the hat opcode is returning "undefined" most of the time, and only occasionally `true` or `false`, which is probably related to this bug.  Edge-triggered hat opcodes should always return either true or false.

### LINKED ISSUE #2114: BOOST extension: color reporter design?
Currently the BOOST extension includes a "color" reporter, which reports a string which is the name of the color the sensor currently sees, such as "red", "yellow", or "none".

If we keep this approach, we'll need to localize these color names. Then the problem would be that if I create code that checks value of the reporter, it will work for me but not for people using the editor in other languages:

Screen Shot 2019-04-11 at 1 43 22 PM

For example, if I load this project with the editor set to Spanish, the reporter will report "rojo" instead of "red", and the condition will never be true.

So, we might consider instead creating a boolean reporter block with a menu listing the color names- something like:

<seeing color [red]?>

This would allow you to do the same thing as above, but more simply, in a way that works across languages. 

If we replace the color string reporter with the boolean, though, we will lose the ability to live monitor the color sensor's value, which is very useful.

ericrosenbaum:
### Resolves

resolves https://github.com/LLK/scratch-vm/issues/2114
resolves https://github.com/LLK/scratch-vm/issues/2113
resolves https://github.com/LLK/scratch-vm/issues/2112
resolves https://github.com/LLK/scratch-vm/issues/2090

### Proposed Changes

This is a set of related fixes and changes to the BOOST extension color sensing blocks:

- Remove the `color` string reporter (which also removes the need to localize that reporter's output)
- Add a boolean reporter: `<seeing [any color] brick?>`
- Change the wording on the hat block to: `when [any color] brick seen`
- Clean up the hat block functionality by using a new system to refer to colors internally. Previously we were using a combination of ids (which were also used as UI strings) and numeric indices (which are sent from the hub). This change uses a single set of IDs to represent color: as menu values, where sensor values are stored, and in opcodes when comparing colors. The conversion from index to ID is done in just one place, when the BLE message arrives. 
- Fix the ability for multiple "when color any" hats to trigger, by moving where `previousColor` is set out of the opcode function   

### Questions

The `BoostColorIdByIndex` object may not be the cleanest solution for looking up a color id by index. Maybe this should be some other data structure? Or a function with a switch statement?

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -101,30 +101,34 @@ const BoostPort = {
 };
 
 /**
- * Ids for each color sensed by the Boost Vision Sensor
+ * Ids for each color sensor value used by the extension.
+ * @readonly
+ * @enum {string}
  */
-const COLOR_ID_ANY = 'any';
-const COLOR_ID_NONE = 'none';
-const COLOR_ID_RED = 'red';
-const COLOR_ID_BLUE = 'blue';
-const COLOR_ID_GREEN = 'green';
-const COLOR_ID_YELLOW = 'yellow';
-const COLOR_ID_WHITE = 'white';
-const COLOR_ID_BLACK = 'black';
+const BoostColor = {
+    ANY: 'any',
+    NONE: 'none',
+    RED: 'red',
+    BLUE: 'blue',
+    GREEN: 'green',
+    YELLOW: 'yellow',
+    WHITE: 'white',
+    BLACK: 'black'
+};
 
 /**
- * Object to look up a color ID by color index
+ * Enum for indices for each color sensed by the Boost vision sensor.
  * @readonly
  * @enum {number}
  */
-const BoostColorIdByIndex = {
-    255: COLOR_ID_NONE,
-    9: COLOR_ID_RED,
-    3: COLOR_ID_BLUE,
-    5: COLOR_ID_GREEN,
-    7: COLOR_ID_YELLOW,
-    10: COLOR_ID_WHITE,
-    0: COLOR_ID_BLACK
+const BoostColorIndex = {
+    [BoostColor.NONE]: 255,
+    [BoostColor.RED]: 9,
+    [BoostColor.BLUE]: 3,
+    [BoostColor.GREEN]: 5,
+    [BoostColor.YELLOW]: 7,
+    [BoostColor.WHITE]: 10,
+    [BoostColor.BLACK]: 0
 };
 
 /**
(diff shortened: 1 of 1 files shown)


evhan55:
Testing this now for functionality:

- #2114: the new `seeing [color] brick?` boolean reporter works well for me!
- #2113: `when [any color] brick seen` hat triggers twice in a row for the same color
- #2112: two `when [any color] brick seen` hats trigger at the same time
- #2090: color names in menus seem localized in the code, now

The tester project sent to me by @ericrosenbaum works really great!  This PR seems to significantly improve the color sensing functionality in the Boost extension, and makes it feel very usable.

I will look at the code and give comments separately.

src/extensions/scratch3_boost/index.js
```diff
@@ -101,18 +101,30 @@ const BoostPort = {
 };
 
 /**
- * Enum for indexed colors read by the Boost Vision Sensor
+ * Ids for each color sensed by the Boost Vision Sensor
+ */
+const COLOR_ID_ANY = 'any';
```
evhan55: I tried to simplify this by not needing the extra color ID values and wasn't successful at finding a better alternative.  However, I wonder if these could get renamed to `COLOR_ANY` and `COLOR_RED` etc, for a bit of simplification?

src/extensions/scratch3_boost/index.js
```diff
@@ -101,18 +101,30 @@ const BoostPort = {
...
  */
-const BoostColor = {
-    NONE: 255,
-    RED: 9,
-    BLUE: 3,
-    GREEN: 5,
-    YELLOW: 7,
-    WHITE: 10,
-    BLACK: 0
+const BoostColorIdByIndex = {
```
evhan55: This could probably stay named `BoostColor`, also for the sake of a simpler name, just a thought!

evhan55 (approved):
lgtm!

evhan55:
The new data structures are still working for me and read more clearly!

evhan55 (approved):
lgtm!

PR #2123, opened 2019-04-17
use xmlEscape instead of escape-html for extensions

cwillisf:
### Proposed Changes

Remove `escape-html` from our dependency list in favor of the `xmlEscape` function that we have in `util/xml-scape.js`.

### Reason for Changes

The data being escaped is XML intended for scratch-blocks, so escaping all HTML entities is excessive. Since this is the only place where we use `escape-html` the presence of the dependency is unnecessary.

I'm about to add another place where we need to escape some text in XML intended for scratch-blocks, so this seems like a good time to make the switch proposed in this PR.

Note that `escape-html` is still listed in our `package-lock.json` since it's a dependency of `webpack-dev-server`. That said, `webpack-dev-server` is only a development dependency, which means that `escape-html` is also now only a development dependency and is no longer part of our build output.

### Test Coverage

Existing extension tests cover this change.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -72,7 +72,6 @@ module.exports = [
         },
         externals: {
             'decode-html': true,
-            'escape-html': true,
             'format-message': true,
             'htmlparser2': true,
             'immutable': true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -3865,7 +3865,8 @@
     "escape-html": {
       "version": "1.0.3",
       "resolved": "https://registry.npmjs.org/escape-html/-/escape-html-1.0.3.tgz",
-      "integrity": "sha1-Aljq5NPQwJdN4cFpGI7wBR0dGYg="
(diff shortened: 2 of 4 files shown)


webpack.config.js
```diff
@@ -72,7 +72,6 @@ module.exports = [
         },
         externals: {
             'decode-html': true,
-            'escape-html': true,
```
kchadha, line 75: Could you explain what's happening here?

kchadha (commented):
Changes look good! Just had a minor question about the change in the webpack config...

webpack.config.js
cwillisf: This line effectively tells Webpack to exclude `escape-html` from the output bundle based on the promise that it will be provided externally. Since it's now gone from our (non-dev) dependency list there's no need to exclude it from the output any more.

kchadha (approved):

PR #2126, opened 2019-04-19
Runtime script cache fix

### LINKED ISSUE #1707: Cache script blocks for allScriptsDo

mzgoddard:
### Resolves

#1707 

### Proposed Changes

- https://github.com/LLK/scratch-vm/pull/1930
- Normalize DONE thread status with the hat execution order change

### Reason for Changes

After 1930 was merged failed to pass tests. The test it failed is introduced in 1930. The hat execution order change changed the expected thread status (DONE). A small addition (a5c44732) with 1930 is in this PR to help normalize the thread status in this scenario.

### DIFF
--- a/src/engine/blocks-runtime-cache.js
+++ b/src/engine/blocks-runtime-cache.js
@@ -1,0 +1,78 @@
+/**
+ * @fileoverview
+ * The BlocksRuntimeCache caches data about the top block of scripts so that
+ * Runtime can iterate a targeted opcode and iterate the returned set faster.
+ * Many top blocks need to match fields as well as opcode, since that matching
+ * compares strings in uppercase we can go ahead and uppercase the cached value
+ * so we don't need to in the future.
+ */
+
+/**
+ * A set of cached data about the top block of a script.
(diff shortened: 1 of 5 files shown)


mzgoddard:
- Move the thread status fix into sequencer. The fix in startHats broke broadcast tests.

kchadha (approved):
LGTM, thanks for fixing this!

PR #2132, opened 2019-04-22
Make Boost set motor speed opcode yield for send interval duration

### LINKED ISSUE #2127: BOOST extension: setting speed continuously can result in motor being "stuck on"
### Expected Behavior

When you use the BOOST "turn motor on for seconds" block, the motor should always stop after the specified number of seconds.

### Actual Behavior

If you run a stack that sets the motor speed in a loop, and then click the "turn motor for 1 rotation" block, about a quarter of the time the motor does not stop, and gets "stuck on".

### Steps to Reproduce

Click the top stack to start the loop, then click the "turn motor for seconds" block:

Screen Shot 2019-04-19 at 4 19 57 PM

If you try this a few times, sometimes the motor will just stay on instead of stopping after a second.

evhan55:
### Resolves

- Resolves #2127: BOOST extension: setting speed continuously can result in motor being "stuck on"
- Resolves a similar issue to #2127 for setting direction continuously

### Proposed Changes

Have 'set motor speed' and 'set motor direction' opcodes yield for send interval duration instead of a single tick.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1804,7 +1804,11 @@ class Scratch3BoostBlocks {
                 }
             }
         });
-        return Promise.resolve();
+        return new Promise(resolve => {
+            window.setTimeout(() => {
+                resolve();
+            }, BoostBLE.sendInterval);
+        });
     }
 
     /**
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1856,7 +1856,11 @@ class Scratch3BoostBlocks {
                 }
             }
         });
-        return Promise.resolve();
+        return new Promise(resolve => {
+            window.setTimeout(() => {
+                resolve();
+            }, BoostBLE.sendInterval);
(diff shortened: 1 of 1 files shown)


evhan55:
@ericrosenbaum 

This seems to fix the problem.  Without the fix, I got the STR to happen after trying to turn the motor on after 4 or 5 times.  With the fix, I tried to turn the motor on over 20 times and didn't get it to reproduce.

I noticed the `setMotorDirection` opcode also doesn't yield correctly, should this PR handle that?

ericrosenbaum:
re: `setMotorDirection` that sounds great yes, please add!

evhan55:
@ericrosenbaum 

Okay, I added the send interval delay to `setMotorDirection` as well.

It seems to fix the same kind of problem in your STR, replacing`'set motor A power to 100%` with `set motor A direction to this way`.  I got the same symptom (on forever after 4 or 5 tries), and the same fix (doesn't reproduce even after 20 tries).

ericrosenbaum (approved):
LGTM!

PR #2133, opened 2019-04-22
Prevent Boost turnOnForDegrees jittering on power set

### LINKED ISSUE #2125: BOOST extension: setting speed continuously interferes with "turn for rotations"
### Expected Behavior

Setting the BOOST motor speed continuously in a loop should not interfere with the functionality of other blocks.

### Actual Behavior

If you run a stack that sets the motor speed in a loop, and then click the "turn motor for 1 rotation" block, the motor turns one rotation and then freaks out, rapidly jittering for up to several seconds. 

### Steps to Reproduce

Click the top stack to start the loop, then click the "turn motor for rotations" block:

Screen Shot 2019-04-19 at 3 56 24 PM

A more typical case, where the speed is being changed:

Screen Shot 2019-04-22 at 4 11 50 PM

evhan55:
### Resolves

- Resolves #2125 BOOST extension: setting speed continuously interferes with "turn for rotations"
- Also resolves potential issues with setting direction during "turn for rotations", as per comment: https://github.com/LLK/scratch-vm/issues/2125#issuecomment-487614200

### Proposed Changes

Prevent setting motor power and direction if it's already in rotation state.

### Testing

Trying to change direction or power during motor rotations now has no effect:

Screen Shot 2019-04-29 at 12 00 25 PM

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1800,8 +1800,8 @@ class Scratch3BoostBlocks {
                         motor.turnOnFor(motor.pendingTimeoutStartTime + motor.pendingTimeoutDelay - Date.now(), false);
                         break;
                     case BoostMotorState.ON_FOR_ROTATION: {
-                        const p = Math.abs(motor.pendingPositionDestination - motor.position);
-                        motor.turnOnForDegrees(p, Math.sign(p), false);
+                        const p = Math.abs(motor.pendingPositionDestination - motor.position, false);
+                        motor.turnOnForDegrees(p, Math.sign(p));
                         break;
                     }
                     }
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1788,23 +1788,14 @@ class Scratch3BoostBlocks {
         this._forEachMotor(args.MOTOR_ID, motorIndex => {
             const motor = this._peripheral.motor(motorIndex);
             if (motor) {
-                const power = MathUtil.clamp(Cast.toNumber(args.POWER), 0, 100);
-                // if motor power is already equal to the scaled input power, don't do anything
-                if (motor.power !== MathUtil.scale(power, 1, 100, 10, 100)) {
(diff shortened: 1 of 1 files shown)


src/extensions/scratch3_boost/index.js
```diff
@@ -1788,19 +1788,23 @@ class Scratch3BoostBlocks {
...
-                    motor.turnOnFor(motor.pendingTimeoutStartTime + motor.pendingTimeoutDelay - Date.now(), false);
-                    break;
-                case BoostMotorState.ON_FOR_ROTATION: {
-                    const p = Math.abs(motor.pendingPositionDestination - motor.position, false);
-                    motor.turnOnForDegrees(p, Math.sign(p));
-                    break;
-                }
+                const power = MathUtil.clamp(Cast.toNumber(args.POWER), 0, 100);
+                // if motor power is already equal to the scaled input power, don't do anything
+                if (motor.power !== MathUtil.scale(power, 1, 100, 10, 100)) {
```
evhan55: `MathUtil.scale(power, 1, 100, 10, 100)` is what the setter for power for the motor uses.  If this overall fix is what we want to use to fix the bug in #2125, then we could maybe move that scaling to a helper function.

evhan55:
With this quick fix, this scenario below still jitters, because `turnOnForDegrees` is called for every power setting change:

Screen Shot 2019-04-22 at 4 11 50 PM

ericrosenbaum:
I haven't had time to review this yet- but I like the idea of bailing out from updating the motor command in progress if the set speed command does not actually change the state (we could apply the same logic for setting direction as well I think?). 

I think this should not be considered to resolve the underlying bug though.

ericrosenbaum:
I just removed the "resolves" line from the description so we don't accidentally close the bug

evhan55:
Sounds good to remove the resolves tag! Makes sense that if this kind of bailing out is used, to use it for set direction, too.

ericrosenbaum (approved):
LGTM!

PR #2139, opened 2019-04-23
Reverse the sign on Boost Motor A position reporter

### LINKED ISSUE #2135: BOOST extension: motor A position reporter seems reversed
### Expected Behavior

Turning the motor clockwise should result in the reported motor position increasing.

### Actual Behavior

For motor A (but not for motor B or an external motor), turning the motor clockwise results in the motor position decreasing.

### Steps to Reproduce

Run code that continuously sets a sprite's direction to the motor position:

Screen Shot 2019-04-23 at 12 13 02 PM

Note that when you turn something plugged into motor A, if you turn it clockwise, the sprite turns counter-clockwise.

evhan55:
- Resolves #2135: BOOST extension: motor A position reporter seems reversed

### Proposed Fix

The motor in Boost port A is reversed by design, so this PR proposes to reverse its value in the reporter so it matches polarity with the other motor port reported values.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1016,10 +1016,7 @@ class Boost {
                 break;
             case BoostIO.MOTOREXT:
             case BoostIO.MOTORINT:
-                // The motor position in port A is reversed by design, so we need
-                // to reverse it here so that all motors match
-                this.motor(portID).position = ((portID === BoostPort.A) ? -1 : 1) *
-                    int32ArrayToNumber(data.slice(4, 8));
+                this.motor(portID).position = int32ArrayToNumber(data.slice(4, 8));
                 break;
             case BoostIO.CURRENT:
             case BoostIO.VOLTAGE:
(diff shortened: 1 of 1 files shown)


evhan55:
@knandersen 

I am looking at #2135 and can't find a bug in the Boost extension code that could be causing this issue.  I tested a hack-fix (in the code in this PR) by just reversing the sign on port A's motor position, and it seems to fix the bug.

Is it possible that one side of the Boost has reversed motor position values? (but not for the external motors C and D for some reason?)

ericrosenbaum:
@evhan55 you're correct that this isn't caused by a bug- in fact, the two internal motors report their positions with opposite handedness, by design. I'm guessing there's a logic to why this is the case, having to do with simplifying the control for driving two wheeled vehicles, but I can't quite wrap my head around it. So the place you put the change makes sense to me. 

The external motor behaves as expected, with no inversion, so nothing needs to be changed for that.

I just tested this change and noticed that it fixes the bug in the narrowest sense possible- "forever set rotation to motor A position" now has the correct clockwise behavior. But the motor reporter always reports a negative value, which is obviously not what we want. It should behave the same as the other motor reporters, and report a positive value. I believe all that's necessary is to multiply the value by -1 before wrap clamping, instead of after.

evhan55:
@ericrosenbaum Ooh thanks! Before the wrap clamping makes sense of course, sorry! I will fix and test assign for review.

ericrosenbaum:
@evhan55 this change is quite different, because instead of modifying the output of the reporter block, it modifies the internally stored position, which is used in other places. So this change could have other side effects or introduce bugs. Could you please verify?

evhan55:
@ericrosenbaum Thank you for catching that!  Per our discussion today, I moved the position reversal back to the reporter opcode function, so that it doesn't affect any internal state.  Please let me know what you think, and also if you think the comment isn't clear enough.

ericrosenbaum (approved):
LGTM!

PR #2140, opened 2019-04-23
Change Boost motor turning behavior for '0 speed' case

### LINKED ISSUE #2136: BOOST extension: setting speed to 0 after motor is on results in it being stuck on
### Expected Behavior

A motor should never get stuck on: it should always be possible to stop all motors using the stop button, or to stop motors using the "turn motor off" block. 

### Actual Behavior

If you turn a motor on, then set its speed to 0, it will become stuck on. The stop all button does not stop it, and the "turn motor off" block does not stop it.

Screen Shot 2019-04-23 at 12 20 22 PM

Or, click the "turn motor on for 3 seconds" block, and then click the "set speed to 0" block. The motor will continue indefinitely, instead of stopping after three seconds (to stop it, you can change the value in the set speed block to something above 0, then click on it):

Screen Shot 2019-04-24 at 5 36 46 PM

and a third case ?:

Screenshot 2019-04-26 at 17 27 12

evhan55:
### Resolves

- Resolves #2136: BOOST extension: setting speed to 0 after motor is on results in it being stuck on

### Proposed Changes

- Remove all the `this.power === 0` checks in the motor turning functions, which were brought over from the WeDo2 extension, and were incompatible with the `BoostMotorState` paradigm.
- Remove all use of `resetState` flag since it didn't seem necessary after removing the `this.power===0` checks.
- Fix the motor `status` setter to not clear rotation promise if in `ON_FOR_ROTATION` and not clear timeout promise if in `ON_FOR_TIME`

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -493,44 +493,33 @@ class BoostMotor {
 
     /**
      * Turn this motor on indefinitely
-     * @param {boolean} [resetState=true] - whether to reset the state of the motor when running this command.
      */
-    turnOnForever (resetState = true){
-        // if (this.power === 0) return;
-        if (resetState) this.status = BoostMotorState.ON_FOREVER;
+    turnOnForever (){
+        this.status = BoostMotorState.ON_FOREVER;
         this._turnOn();
 
-        console.log('ON_FOREVER');
+        console.log('this.status', this.status);
     }
 
     /**
      * Turn this motor on for a specific duration.
      * @param {number} milliseconds - run the motor for this long.
-     * @param {boolean} [resetState=true] - whether to reset the state of the motor when running this command.
      */
-    turnOnFor (milliseconds, resetState = true) {
-        // if (this.power === 0) return;
-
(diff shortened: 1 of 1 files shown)


evhan55:
@knandersen 

This PR has code that fixes the problem seen in #2136. However, the fixes I put in may break other cases that I am not thinking about. Do you have a sense for how to test the motor blocks to see that behavior remains robust?

I noticed that `setMotorPower` actually calls `turnOnForever` when the status is `ON_FOREVER`, so then I just had `turnOnForever` call `turnOff` if power is 0, and `turnOff` to ignore the `power === 0` case, because it should be safe to turn off the motor if the power was already 0 or if it has become 0 in a new motor status/state.

evhan55:
A couple of notes to myself at this moment in time for this PR:

- behavior and highlighting seems pretty stable
- setting power to 0 during `ON_FOR_ROTATION` causes `ON_FOR_ROTATION` to stop altogether
- `turn motor A on` loses highlight as soon as motor is turned on, which might be desired

ericrosenbaum (approved):
Tested again- looks good to me!

Issue #2120, opened 2019-04-16
Extensionification: All blocks should be implemented using extension system


cwillisf:
# Extensionification!

## Goals

- All blocks in Scratch 3.0 should be entirely defined within `scratch-vm`. Adding a new block should not require modifications in `scratch-gui` or `scratch-blocks`.
- All blocks in Scratch 3.0 should be implemented in the same way. We should not maintain two (or more) different ways to define blocks.
- The extension system should be enhanced such that it becomes powerful enough to implement every existing Scratch 3.0 block.

## Tasks

* Supporting features
  * [x] Extensions should be able to place buttons in the toolbox
  * [ ] Extension blocks should refresh when changing targets
  * [ ] Extension blocks should refresh on certain actions (create/update/delete variable, list, procedure, message, etc.)
  * [ ] Extensionified categories can be interleaved with categories defined in `scratch-gui`'s `makeToolboxXML`
* Serialization
  * [ ] Existing SB3 files, including those containing variables, lists, or custom procedures, can be loaded correctly when those categories are extensionified.
  * [ ] An SB3 file saved with extensionified variables, lists, or custom procedures can be loaded correctly in an older editor (esp. Scratch Desktop).
* Categories
  * [ ] Motion
  * [ ] Looks
  * [ ] Sound
  * [ ] Events (including `broadcast _` & friends)
  * [ ] Control (including `stop _`)
  * [ ] Sensing (including `_ of _`)
  * [ ] Operators
  * [ ] Variables
  * [ ] My Blocks

harbaum:
One of the most basic features for this is the support for translations. Currently these are in scratch-blocks/msg/messages.js and the translation system supposed to be there doesn't seem to be functional at all.

cwillisf:
@harbaum we actually have a separate translation system for Scratch extensions. We're still figuring out the details but documentation on the current iteration is available here and will be updated as we progress:
https://github.com/LLK/scratch-vm/blob/develop/docs/extensions.md#translation

harbaum:
Yes, i know. I tried to use it but it had no effect. I then searched the code and found no code that would care for these translations.

apple502j:
@BryceLTaylor hm, adroitwhiz said "RIP extensionification" https://github.com/LLK/scratch-gui/issues/4400#issuecomment-627797933 - is it true?

CommanderGL:
Will this finally be implemented because if not I'm putting it in my fork

PR #2141, opened 2019-04-24
Add a runtime event for requesting a toolbox update

kchadha:
Adds a new runtime event which the vm listens for and issues a toolbox update for extension blocks. This event is triggered on switching sprites so that extensions can contain target-dependent blocks. This is necessary for the extensionification of variables and custom procedures (towards #2120).

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -510,7 +510,7 @@ class Runtime extends EventEmitter {
     }
 
     /**
-     * Event name for report that a change was made that can be saved
+     * Event name for report that a change was made to an extension in the toolbox.
      * @const {string}
      */
     static get TOOLBOX_EXTENSIONS_NEED_UPDATE () {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -510,6 +510,14 @@ class Runtime extends EventEmitter {
     }
 
     /**
+     * Event name for report that a change was made that can be saved
+     * @const {string}
+     */
+    static get TOOLBOX_EXTENSIONS_NEED_UPDATE () {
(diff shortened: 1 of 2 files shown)


src/engine/runtime.js
```diff
@@ -1963,10 +1971,15 @@ class Runtime extends EventEmitter {
      * @param {!Target} editingTarget New editing target.
      */
     setEditingTarget (editingTarget) {
+        const oldEditingTarget = this._editingTarget;
```
cwillisf, line 1974: It might make sense to just skip everything and early-out if the editing target isn't really changing

src/engine/runtime.js
```diff
@@ -509,6 +509,14 @@ class Runtime extends EventEmitter {
         return 'PROJECT_CHANGED';
     }
 
+    /**
+     * Event name for report that a change was made that can be saved
```
cwillisf: comment needs update :)

cwillisf (approved):
I made a couple minor suggestions but honestly I'd be OK with merging this to e16n as-is...

src/engine/runtime.js
kchadha: Hmmm that's a good point but I'm not sure I want to make that change here, since it seems like more of a refactor that could have unintentional side effects rather than necessary for this PR.

src/engine/runtime.js
kchadha: Good catch!

kchadha:
```suggestion
     * Event name for report that a change was made to an extension in the toolbox.
```

PR #2143, opened 2019-04-24
support non-droppable menus in extensions

### LINKED ISSUE #709: Let extensions make non-droppable menus
### Expected Behavior

Any Scratch 3.0 block should be possible to build in an extension.

### Actual Behavior

It's currently not possible to make a drop-down menu which cannot be overridden by inserting a reporter. That is, all extension menus currently accept reporters.

Note that another way to fix this would be to make all Scratch 3.0 menus accept reporters...

cwillisf:
### Resolves

Resolves #709

### Proposed Changes

When an extension defines a menu, the two existing definition forms are still allowed:
```js
menuWithStaticItems: ['a', 'b'],
menuWithDynamicItems: 'getDynamicItems'
```

This pull request adds support for defining a menu with an object, and makes the above forms shorthand for an object which specifies only an `items` property. These definitions are equivalent to those above:
```js
menuWithStaticItems: { items: ['a', 'b'] },
menuWithDynamicItems: { items: 'getDynamicItems' }
```

Doing so means we can also add more settings to the object. This pull request adds support for a `rejectsReporters` flag which makes the menu "non-droppable" -- it won't accept reporters:
```js
nonDroppableMenu: { items: 'getVariableNames', rejectReporters: true }
```

**I am open to suggestions for a better name than `rejectReporters`.** I would prefer to keep it a "positive" flag such that omitting the flag or setting it to a falsey value both result in a droppable menu by default.

### Reason for Changes

There are a few menus in Scratch 3.0 which don't accept reporters, such as the variables and lists menus and the key selection menu for the "when key pressed" hat block. Extensionifying those blocks means supporting non-droppable menus.

CC @ericrosenbaum @evhan55


docs/extensions.md
```diff
@@ -311,7 +311,16 @@ class SomeBlocks {
 
                 // Dynamic menu: returns an array as above.
                 // Called each time the menu is opened.
-                menuB: 'getItemsForMenuB'
+                menuB: 'getItemsForMenuB',
+
+                // The examples above are shorthand for setting only the `items` property in this full form:
+                menuC: {
+                    // This flag makes a "non-droppable" menu: the menu will not accept reporters.
+                    rejectReporters: true,
```
towerofnix: Frankly, save for someone hypothetically confusing the term with "reject" in Promise lingo, I think `rejectReporters` isn't too bad.. :P

Another option would be something along the lines of "only" - `onlyDropdown`, `onlyField`, etc. But I think `rejectReporters` is more clear than those. There's also `disallowReporters`, but that sort of screams, "It *should* be `allowReporters: false`!" - negation in the value rather than the name, that is (and `: false` is already explicitly out of the question).

PS a fun thing with making menus be in object form is that we can add entries like `changeCallback`, for implementing things like the "stop (all/this script/other scripts in sprite)" block!

cwillisf:
After discussion in a meeting earlier today, we decided that defaulting to non-droppable menus would be best.

While ideally most or all drop-down menus would accept reporters, it's easy for that to get confusing or even frustrating unless the extension author thinks through a lot of details. To avoid bad experiences being the norm for dropdown menus in extensions, we decided to make this an opt-in feature.

I added more documentation detail on this topic in 88da71474a25adca9864ca1b10933e3a386ea5ec, including a few of the specific considerations to think through before making a menu accept reporters.

CC @thisandagain @ericrosenbaum @knandersen @kchadha @evhan55 @cwillisf

cwillisf:
Since 642a8c09b09b1997be16382d2ea623b95d1f3322 involves a lot of indentation changes I encourage reviewers to check this box:
image

docs/extensions.md
```diff
@@ -120,6 +120,123 @@ class SomeBlocks {
...
+encourage extension authors to make their menus accept reporters when possible, doing so requires careful
+consideration to avoid confusion and frustration on the part of those using the extension.
+
+A few of these considerations include:
+
+* The valid values for the menu should not change when the user changes the Scratch language setting.
+  * In particular, changing languages should never break a working project.
+* The average Scratch user should be able to figure out the valid values for this input without referring to extension
+  documentation.
+  * One way to ensure this is to make an item's text match or include the item's value.
```
kchadha: Maybe it makes sense to include an example of this (e.g. the Music Extension lists the extension number as part of the menu item text)?

src/engine/runtime.js
```diff
@@ -830,18 +825,24 @@ class Runtime extends EventEmitter {
...
-     *  Read extension information, convert menus, blocks and custom field types
+     * Read extension information, convert menus, blocks and custom field types
      * and store the results in the provided category object.
      * @param {CategoryInfo} categoryInfo - the category to be filled
      * @param {ExtensionMetadata} extensionInfo - the extension metadata to read
      * @private
      */
     _fillExtensionCategory (categoryInfo, extensionInfo) {
+        categoryInfo.blocks = [];
+        categoryInfo.customFieldTypes = {};
```
kchadha, line 845: Any particular reason for moving this out of `_registerExtensionPrimitives` and moving it here instead?

src/engine/runtime.js
```diff
@@ -830,18 +825,24 @@ class Runtime extends EventEmitter {
...
+        categoryInfo.menus = [];
+        categoryInfo.menuInfo = {};
+
         for (const menuName in extensionInfo.menus) {
             if (extensionInfo.menus.hasOwnProperty(menuName)) {
-                const menuItems = extensionInfo.menus[menuName];
-                const convertedMenu = this._buildMenuForScratchBlocks(menuName, menuItems, categoryInfo);
+                const menuValue = extensionInfo.menus[menuName];
```
kchadha: I think `menuValue` is a weird name. Maybe we can just go w/`menuInfo` or in cases like this I use 'Desc' as short for descriptor, e.g. `menuDesc` ¯\\_(ツ)_/¯

src/engine/runtime.js
```diff
@@ -883,19 +884,21 @@ class Runtime extends EventEmitter {
...
      * @private
      */
-    _buildMenuForScratchBlocks (menuName, menuItems, categoryInfo) {
+    _buildMenuForScratchBlocks (menuName, menuInfo, categoryInfo) {
         const menuId = this._makeExtensionMenuId(menuName, categoryInfo.id);
-        let options = null;
-        if (typeof menuItems === 'function') {
-            options = menuItems;
+        let menuItems = null;
+        if (typeof menuInfo.items === 'function') {
```
kchadha: Why would this be a function? O.o

src/engine/runtime.js
```diff
@@ -1218,29 +1222,52 @@ class Runtime extends EventEmitter {
...
 
-        const shadowType = (argInfo.menu ?
-            this._makeExtensionMenuId(argInfo.menu, context.categoryInfo.id) :
-            argTypeInfo.shadowType);
-        const fieldType = argInfo.menu || argTypeInfo.fieldType;
+        let valueName;
+        let shadowType;
+        let fieldName;
+        if (argInfo.menu) {
+            const menuInfo = argInfo.menu && context.categoryInfo.menuInfo[argInfo.menu];
```
kchadha: We're already in an if for `argInfo.menu` so we don't need the additional check here, right?
```suggestion
            const menuInfo = context.categoryInfo.menuInfo[argInfo.menu];
```

src/engine/runtime.js
```diff
@@ -1218,29 +1222,52 @@ class Runtime extends EventEmitter {
...
+            if (menuInfo.acceptReporters) {
+                valueName = placeholder;
+                shadowType = this._makeExtensionMenuId(argInfo.menu, context.categoryInfo.id);
+                fieldName = argInfo.menu;
+            } else {
+                argJSON.type = 'field_dropdown';
+                argJSON.options = menuInfo.items;
+                valueName = null;
```
kchadha, line 1255: why not just initialize these to `null` when declaring them above?

src/engine/runtime.js
```diff
@@ -1218,29 +1222,52 @@ class Runtime extends EventEmitter {
...
         if (shadowType) {
             context.inputList.push(`<shadow type="${shadowType}">`);
+        }
 
-            // <field> is a text field that the user can type into. Some shadows, like the color picker, don't allow
-            // text input and therefore don't need a field element.
-            if (fieldType) {
-                context.inputList.push(`<field name="${fieldType}">${defaultValue}</field>`);
-            }
+        // <field> is a text field that the user can type into. Some shadows, like the color picker, don't allow
```
kchadha: This comment is actually not strictly correct... especially now that we have non-droppable menus...

src/extension-support/extension-manager.js
```diff
@@ -310,15 +310,22 @@ class ExtensionManager {
     _prepareMenuInfo (serviceName, menus) {
         const menuNames = Object.getOwnPropertyNames(menus);
         for (let i = 0; i < menuNames.length; i++) {
-            const item = menuNames[i];
+            const menuName = menuNames[i];
```
kchadha, line 312: Maybe add a comment here saying we're normalizing the menu definitions to have the structure `{items: [...]}`.

kchadha (changes requested):
Some minor comments/questions inline.

aeons:
Hey, what's the status on this?

kchadha:
@aeons, @cwillisf and I are actively working on this as part of a larger project to convert core blocks in scratch into extensions (aka using the extension spec to define them). See #2120 for more info. 

This is a pull request to get these changes into the `e16n` branch, and there will be a subsequent pull request to get the `e16n` branch onto `develop`.

aeons:
Alright, thanks.

ErikMejerHansen:
@kchadha: Is the plan to finish up all the e16n work and then merge the `e16n`-branch or will there be multiple merges as the work progresses?
This feature would be very useful to have without having to wait for all of the e16n work to finish.

kchadha:
@ErikMejerHansen, we will be merging e16n as work progresses. There are currently two PRs out for merging e16n into develop LLK/scratch-gui#4790 and #2161. Our hope is that we will be getting those two PRs merged in soon.

src/engine/runtime.js
cwillisf: `_fillExtensionCategory` is also called by `_refreshExtensionPrimitives` so we need to clear these out to prevent duplicating things on refresh. Previously `_refreshExtensionPrimitives` was clearing blocks and menus but not custom fields, so it was an easy problem to miss. By putting these here it's harder to forget to clear a new `categoryInfo` field in `_registerExtensionPrimitives` before filling it.

src/engine/runtime.js
cwillisf: this is a possibly-controversial habit of mine: if I have a variable that gets initialized in one of several ways I generally leave it uninitialized and then make sure to initialize it in each path. That way if something happens to the control flow (like a `break` gets added or something) and results in the possibility of skipping initialization of the variable, an IDE or linter can report that the variable might be uninitialized below. Initializing to `null` makes sure that the variable is definitely initialized but doesn't guarantee that it's initialized to a correct value...

TL;DR: it's a habit of mine that I like for maintenance reasons but if other people don't like it then I'm happy to switch.

src/engine/runtime.js
cwillisf: Ah, yeah... I was going for "name/value" but `menuInfo` is probably better

PR #2145, opened 2019-04-25
Raise params to the next frame when pushing

mzgoddard:
### Resolves

Runtime performance related to stack operations including procedure parameters and util.stackFrame (thread.stackFrame.executionContext).

### Proposed Changes

Related to https://github.com/LLK/scratch-vm/pull/2147 and https://github.com/LLK/scratch-vm/pull/2148.

- Remove procedures_definition block function
- Remove use of utils.stackFrame in procedures_call
- _StackFrame.params is always set
- Set params after starting a procedure
- Carry params into deeper stack frames
- Set params to null prototype objects
- executionContext is always set
- Reset executionContext when needed instead of setting it on demand
- Move reported from stackFrame to thread
- Clean up other reporting members
- Add Thread.pointer
- Add Thread.stackFrame

### Reason for Changes

> - Remove procedures_definition block function

The VM supports blocks without functions. An empty block function takes time to work through finding the function, calling it, and figuring out what to do with the result. So remove it and get on to other work.

> - Remove use of utils.stackFrame in procedures_call

The VM handles the stack correctly and will execute procedures_call twice. (I'm guessing this was a workaround for a bug we incidentally solved at some point.)

> - _StackFrame.params is always set
> - Set params after starting a procedure
> - Carry params into deeper stack frames

Always set to some object we can look up and set parameter values faster. Carry params forward when pushing the stack lets us immediately access the current procedure's parameters. We don't need to look at params in any other stack frames.

> - Set params to null prototype objects

Since only Thread interacts with params we can use the benefits of a null prototype Object without the concerns of the same.

> - executionContext is always set
> - Reset executionContext when needed instead of setting it on demand

Resetting when we need to instead of always resetting can save us some time. Having executionContext already set lets us remove a branch statement when retrieving executionContext.

> - Move reported from stackFrame to thread
> - Clean up other reporting members

`reported` members are used for promise returned blocks. As a thread is paused while the promise resolves the thread only needs to store one set of `reported` members. Moving that to the thread we can remove that from StackFrame and not worry about needing to unset it.

> - Add Thread.pointer
> - Add Thread.stackFrame

Keep the "top" of the stack as direct members of the Thread. This way we can directly access them without needing to find out the length of the stack and get the last item.


src/blocks/scratch3_procedures.js
```diff
@@ -13,46 +13,39 @@ class Scratch3ProcedureBlocks {
      */
     getPrimitives () {
         return {
-            procedures_definition: this.definition,
```
kchadha, line 16: Maybe add a comment in place of this line for why procedures_definition isn't mentioned as a one of the 'primitives'

src/blocks/scratch3_procedures.js
```diff
@@ -13,46 +13,39 @@ class Scratch3ProcedureBlocks {
...
             procedures_call: this.call,
             argument_reporter_string_number: this.argumentReporterStringNumber,
             argument_reporter_boolean: this.argumentReporterBoolean
         };
     }
 
-    definition () {
-        // No-op: execute the blocks.
```
kchadha, line 24: @cwillisf Do you have any concerns for extensionification of custom procs that the procedure definition function will not exist?

In general we expect either `opcode` or both `opcode` and `func` to be defined for a given block right? We should make sure we start to handle these non-existent functions well.

src/blocks/scratch3_procedures.js
kchadha: On second thought, what if we made `procedures_definition: null` an entry here instead of removing it? In general, I don't like having `undefined` be something meaningful...

src/blocks/scratch3_procedures.js
```diff
@@ -13,46 +13,39 @@ class Scratch3ProcedureBlocks {
...
-    }
-
     call (args, util) {
-        if (!util.stackFrame.executed) {
```
kchadha, line 28: @cwillisf, @thisandagain, any idea why we would have had this? It seems it's back from when custom procedure calls were first introduced to the VM so I'm hesitant about removing it without fully understanding why it was there... (see @mzgoddard's corresponding comment in the PR description about executing the call function twice).

src/engine/thread.js
```diff
@@ -4,6 +4,12 @@
  */
 const _stackFrameFreeList = [];
 
+/**
+ * Default params object to for stack frames not inside a procedure.
```
kchadha: Nit: typo in this comment

kchadha: Maybe also explicitly call out the `Object.create(null)` here and add a note about why we're using it and that look-ups of params will not be able to use `hasOwnProperty`

src/engine/thread.js
```diff
@@ -27,92 +33,70 @@ class _StackFrame {
...
+        this.executionContext = {};
 
         /**
-         * A context passed to block implementations.
-         * @type {Object}
+         * Has this frame changed and need a reset?
+         * @type {boolean}
          */
-        this.executionContext = null;
+        this.needReset = false;
```
kchadha: Nit about naming: -->`this.needsReset`

src/engine/thread.js
```diff
@@ -27,92 +33,70 @@ class _StackFrame {
...
     /**
      * Reuse an active stack frame in the stack.
-     * @param {?boolean} warpMode defaults to current warpMode
-     * @returns {_StackFrame} this
      */
-    reuse (warpMode = this.warpMode) {
-        this.reset();
-        this.warpMode = Boolean(warpMode);
-        return this;
+    reuse () {
```
kchadha: why not just call reset here?

src/engine/thread.js
```diff
@@ -27,92 +33,70 @@ class _StackFrame {
...
      * @returns {_StackFrame} The clean stack frame with correct warpMode setting.
      */
-    static create (warpMode) {
-        const stackFrame = _stackFrameFreeList.pop();
-        if (typeof stackFrame !== 'undefined') {
-            stackFrame.warpMode = Boolean(warpMode);
+    static create (warpMode, params) {
+        const stackFrame = _stackFrameFreeList.pop() || null;
+        if (stackFrame !== null) {
+            stackFrame.warpMode = warpMode;
```
kchadha: should `warpMode` still be cast to Boolean in the RHS here? I think it can sometimes be a string...

src/engine/thread.js
```diff
@@ -186,7 +183,25 @@ class Thread {
...
+         * @type {*}
+         */
         this.justReported = null;
+
+        /**
+         * The id of the block that we will report the promise resolved value
+         * for.
+         * @type {string}
+         */
+        this.reporting = null;
```
kchadha: Nit: Can we rename this to `this.reportingBlock` for clarity? As is, the name could be mistaken for a boolean.

src/engine/thread.js
```diff
@@ -239,13 +254,14 @@ class Thread {
...
-            const parent = this.stackFrames[this.stackFrames.length - 1];
-            this.stackFrames.push(_StackFrame.create(typeof parent !== 'undefined' && parent.warpMode));
-        }
+        this.stack.push(this.pointer);
+        this.pointer = blockId;
+
+        const parent = this.stackFrame;
+        const warpMode = parent !== null ? parent.warpMode : false;
+        const params = parent !== null ? parent.params : defaultParams;
+        this.stackFrames.push(this.stackFrame);
```
kchadha:
```suggestion
        this.stackFrames.push(parent);
```

src/engine/thread.js
```diff
@@ -254,17 +270,20 @@ class Thread {
      * @param {string} blockId Block ID to push to stack.
      */
     reuseStackForNextBlock (blockId) {
-        this.stack[this.stack.length - 1] = blockId;
-        this.stackFrames[this.stackFrames.length - 1].reuse();
+        this.pointer = blockId;
+        if (this.stackFrame.needReset) this.stackFrame.reuse();
```
kchadha: It looks like this is the only place in the vm that `reuse` is being used. Should we just consolidate `reuse` and `reset` now given that they have the same functionality?

src/engine/thread.js
```diff
@@ -254,17 +270,20 @@ class Thread {
...
     }
 
     /**
      * Pop last block on the stack and its stack frame.
      * @return {string} Block ID popped from the stack.
      */
     popStack () {
-        _StackFrame.release(this.stackFrames.pop());
-        return this.stack.pop();
+        const popped = this.pointer;
```
kchadha: `popped` is a weird name for this given the code. Maybe `lastPointer` or `lastBlock` would be better?
I think we should also update the comment on this function so that it doesn't necessarily have to reference stack functionality directly (even though the function name itself does). Maybe something like:

'Remove the last block from the stack and its stack frame. Update the current instruction pointer.' or something...

src/engine/thread.js
```diff
@@ -321,14 +340,22 @@ class Thread {
         this.justReported = typeof value === 'undefined' ? null : value;
     }
 
+    /**
+     * Return a execution context for a block to use.
+     * @returns {object} the execution context
+     */
+    getExecutionContext () {
+        const frame = this.stackFrame;
+        frame.needReset = true;
```
kchadha: Why does looking at the executionContext indicate that the frame needs a reset?

src/engine/thread.js
```diff
@@ -321,14 +340,22 @@ class Thread {
...
     /**
      * Initialize procedure parameters on this stack frame.
      */
     initParams () {
-        const stackFrame = this.peekStackFrame();
-        if (stackFrame.params === null) {
-            stackFrame.params = {};
-        }
+        const stackFrame = this.stackFrame;
+        stackFrame.params = Object.create(null);
```
kchadha, line 378: does it make sense to just use `defaultParams` here?

kchadha: Do we want to guard against overwriting already existing params here?

src/engine/thread.js
```diff
@@ -348,15 +375,9 @@ class Thread {
...
-            const frame = this.stackFrames[i];
-            if (frame.params === null) {
-                continue;
-            }
-            if (frame.params.hasOwnProperty(paramName)) {
-                return frame.params[paramName];
-            }
-            return null;
+        const stackFrame = this.stackFrame;
+        if (typeof stackFrame.params[paramName] !== 'undefined') {
```
kchadha, line 399: what about `return stackFrame.params[paramName] || null`?

src/engine/thread.js
```diff
@@ -388,8 +409,8 @@ class Thread {
      */
     isRecursiveCall (procedureCode) {
         let callCount = 5; // Max number of enclosing procedure calls to examine.
-        const sp = this.stack.length - 1;
-        for (let i = sp - 1; i >= 0; i--) {
+        const sp = this.stack.length;
+        for (let i = sp - 1; i >= 1; i--) {
```
kchadha: Why don't we look at `stack[0]`?

kchadha (changes requested):
Left some questions and comments inline. It also looks like tests are failing.

src/blocks/scratch3_procedures.js
mzgoddard: We already handle non-existent functions (hats like event_whenflagclicked and shadows). Do we instead need to handle confirming that non-existence is expected?

src/blocks/scratch3_procedures.js
mzgoddard: My guess is dirty stack frame behaviour, with a pop out of a procedure after a loop and with isLoop having not been unset. And so sequencer have treated exiting the procedure like it was a loop, left the stack at the procedure call id, and yielded.

src/engine/thread.js
mzgoddard: Performance. I think reset and reuse need to be fast. So I inlined the new reset here. I think it would be better for us to get rid of one of reset or reuse then call one from the other.

src/engine/thread.js
mzgoddard: I was wondering about this too. Also because I wanted to know what ways warpMode may be set.

https://github.com/LLK/scratch-vm/blob/develop/src/engine/thread.js#L247

If this warpMode was a string that means its parent warpMode would be a string. That is true in this PR and in develop. The thread creating a stack frame sets the warpMode to its parent warpMode. If the parent stack frame's warpMode is a string ... then we are already running with a warpMode that is a string.

Outside of Thread I found one line that changes warpMode.

https://github.com/LLK/scratch-vm/blob/develop/src/engine/sequencer.js#L339

Sequencer is handling the warp value in case its a string. And pretty explicitly setting warpMode to true if the warp in mutation is true.

I don't think I was previously aware, but this means we only need to check and set a warpTimer in the beginning of Sequnecer.stepThread and in Sequencer.stepToProcedure after setting warpMode true. Once warpMode is true, its true until we pop to a stack frame where its false.

src/engine/thread.js
mzgoddard: I think so yes.

src/engine/thread.js
mzgoddard: Ooop. That is what I meant to do. :smile:

src/engine/thread.js
mzgoddard: That'll be null for params that are 0, false, or empty string. We could do a ternary.

```
stackFrame.params[paramName] = 0;
return stackFrame.params[paramName] || null;

>>> null
```

src/engine/thread.js
mzgoddard: The operations to the stack in this PR leave it with the same number of stack members by making the first member null. So its not a real stack value, but gives this PR the length side effect, and restores pointer to null on the last popStack.

Basically its done this way to fulfill existing tests that look at stack length.

src/engine/thread.js
kchadha: Right sorry forgot that we're dealing with an actual param value and not an object/null/undefined

src/engine/thread.js
kchadha: Hmmm that seems a little too subtle/scary from a maintenance perspective. We should probably remove the null from the beginning of the stack and update the tests...

src/engine/thread.js
mzgoddard: With this PR most StackFrame properties remain do not need to change from when the frame is initialized to when the frame is released into the free list. isLoop and executionContext (and needReset) are the remaining properties that may change.

We need to reset isLoop and executionContext when we reuseStackForNextBlock or popStack on that stack frame. Or that we need to reset executionContext if it is used. I thought this may be a way we capture that info. If a block doesn't get the executionContext then it cannot have used it and we can skip a bit of work resetting it.

There is an alternative, similar to what was in block utility and with previous thread code. When getExecutionContext is called it'll assign an object to executionContext if needed. Then when we reuse or release we can set executionContext to null.

Those are kind of the same and not. One assumes we need to reset after use, and the other assumes we need to create once used.

src/engine/thread.js
mzgoddard: Yeah. I could respond to this quickly because I had recent experience with this logic. 😉

src/engine/thread.js
mzgoddard: Sounds good to me. That'll also help with tests that explicitly check for popStack returning undefined.

src/engine/thread.js
```diff
@@ -321,14 +338,22 @@ class Thread {
         this.justReported = typeof value === 'undefined' ? null : value;
     }
 
+    /**
+     * Return a execution context for a block to use.
```
towerofnix: "an" execution context? :shipit:

src/blocks/scratch3_procedures.js
cwillisf: Looks like that `if` dates back to the initial version of the code from @tmickel back in October of 2016. I can't say for sure why it was written this way but I suspect @mzgoddard is right that it was meant to avoid running the same procedure again after it exits without advancing the thread.

src/blocks/scratch3_procedures.js
mzgoddard: Ah. Hmm. Setting this to null, we need to change a test in the runtime method that register's primitives. It tries to bind null like its a function. That is to say the register primitives algorithm only accepts functions currently.

I'd guess we'll need to change other things in there.

src/engine/thread.js
mzgoddard: I changed it to `reportingBlockId`.

src/engine/thread.js
mzgoddard: Hmm. Aw no, I think I misunderstood when I last responded.

This is compatible with Scratch 2 procedure behaviour to my knowledge. This PR swaps the order the stack is changed. In this PR the stack is pushed and then parameters are set. The parent procedure's stack frame is left alone and the new stack frame will start a new set of parameters.

src/engine/thread.js
mzgoddard: Removing the null from the stack turned out kind of weird. Primarily code that tested `stack.length === 0` could not switch to `stack.pointer === null` but instead `thread.stackFrame === null`. The pointer/stack data can be null to mean the end of a stack and not just the end of the thread. The pointer-less/all-stack version uses `.length` of 1 or more while `stack[0] === null`. Representing its in a stack but at the end.

So this makes `pointer === null` a little ambiguous but really points to using `stackFrame === null` instead. Since it is only null when the thread is not working.

src/engine/thread.js
mzgoddard: Fixed. Thanks. :smile:

mzgoddard:
I should have all my ducks in a row now with what we have discussed so far.

src/blocks/scratch3_procedures.js
mzgoddard: I'm talking about https://github.com/LLK/scratch-vm/pull/2145/files#diff-38d875e1d50b3158dc96404b65275a54R715.

src/engine/thread.js
kchadha: Ah okay.

That still leaves the question of whether we should use `defaultParams` here instead of explicitly using `Object.create(null)` again.

src/engine/thread.js
kchadha: Ah, okay. Where did we end up landing with this?

kchadha:
There seems to be a compatibility issue with this PR. I tested out project 312525055 on this PR vs production/develop and I'm seeing a behavior only on this PR where the cat switches to the black rectangle costume and slides to the left (w/o keyboard input):

incompatibility

If you get to the bottom of this issue, it would be nice to have an integration test for it.

kchadha (changes requested):
Marking as needs work b/c of incompatibility found w/existing develop/production.

src/engine/thread.js
mzgoddard: I think we landed on `stack[0]` cannot be null. Or at least that is currently the implementation in the PR. In this PR currently, values in stack will always be a blockId. `pointer` can be null, but none of the values in stack will be null.

mzgoddard:
I fixed the bug you found @kchadha. It was the dirty isLoop problem. I've added some tests to `test/unit/engine_sequencer.js` to protect against future regression.

kchadha:
@mzgoddard just noticed what seems like another bug. I was testing project 304890904 from production locally w/the changes from this PR.

Things seem to work great until I try to click on the original big red circle to delete it, and then it seems like the whole browser tab freezes.

mzgoddard:
@kchadha Here's a minimal repro.

Screen Shot 0031-06-25 at 12 07 16 PM

I found, tested, and fixed two bugs in `Thread.stopThisScript` from that. One I added in the PR and a pre-existing one. The executionContext use by procedures_call was occluding the bug.

The pre-existing bug is caused by a technical incompatibility with scratch 2. scratch-flash does one more thing in stopThisScript, it sets the block pointer to null [scratch-flash/src/interpreter/Thread.as#L115](https://github.com/LLK/scratch-flash/blob/develop/src/interpreter/Thread.as#L115). That would let Sequencer pop the procedure. Instead develop currently pops to the procedure call block, Sequencer leaves it alone and reruns the procedure call. Necessitating the executionContext check work around.

There is a relevant unit test on stopThisScript but it looks like it depends on the technical incompatibility by looking at the stack and looking for the procedure call block id. It should instead check that stack is null.

src/engine/thread.js
```diff
@@ -254,34 +280,41 @@ class Thread {
...
+     * Move the instruction pointer to the last procedure call block and resume
+     * execution there or to the end of this thread and stop executing this
+     * thread.
      */
     stopThisScript () {
         let blockID = this.peekStack();
         while (blockID !== null) {
             const block = this.target.blocks.getBlock(blockID);
             if (typeof block !== 'undefined' && block.opcode === 'procedures_call') {
+                this.pushStack(null);
```
kchadha, line 316: Could you add a comment here about why this is significant?

kchadha:
@mzgoddard I re-tested this PR after the bugfix you added, and I think it looks good to go. Could you add a code comment about the line you added in stopThisScript about why pushing null onto the stack is significant (and/or what bug it was fixing)? After that LGTM.

src/engine/thread.js
mzgoddard: I added:

```js
// If we do not push this null, Sequencer will not step to the
// next block and will erroneously call the procedure a second
// time.
//
// By pushing null, Sequencer will treat it as the end of a
// substack, pop the stack and step to the next block.
```

mzgoddard:
- Add a comment inside stopThisScript on pushing null

kchadha (approved):
LGTM

BryceLTaylor:
Testing Plan:

Description: This pr changes how Custom procedures store variables so they are more easily accessible to the vm while running the block.  

To test:
Make a custom procedure with a number input, a text input (same as number but we'll use it for text) and a boolean input
Put the block into a forever loop
Add code to the procedure definition while it's running
Add an if statement that uses the boolean input
Add an if / then block using the variables
add a statement that uses the number input
add a statement that uses the text input
Change the values of the inputs while the stack is running
Use a "say 'word' for [time]" block at the end of the stack
Use a "say 'word' for [time]" block in the beginning/middle of the stack

Make a custom procedure that runs without screen refresh and repeat the above steps
Edit the block and add another variable while running
edit the block and delete a variable while running

call the block recursively

Make sure to add the variables from the custom procedure everywhere

PR #2148, opened 2019-04-25
Thread always done

mzgoddard:
### Resolves

General performance around iterating threads during a step.

### Proposed Changes

<del>Depends on https://github.com/LLK/scratch-vm/pull/2145 and https://github.com/LLK/scratch-vm/pull/2147.</del>
<ins>This PR is now standalone and does not depend on other PRs.</ins>

- Remove Thread.isKilled
- Test if a thread is STATUS_YIELD_TICK after STATUS_RUNNING
- Remove stack.length === 0 tests

### Reason for Changes

> - Remove Thread.isKilled

The current code base doesn't actually remove the thread so I'm not sure we need this additional data point.

Removing isKilled lets us remove the isKilled loop at the beginning of _step. Which currently, must loop over all threads.

The one way isKilled threads are configured that way is through stop thread, which also sets their status to DONE.

> - Test if a thread is STATUS_YIELD_TICK after STATUS_RUNNING

Generally most threads will be RUNNING and not YIELD_TICK. We can have the special case after the primary case in the same if/else branch chain and if it is YIELD_TICK, i-- to run it again and have the RUNNING branch activate. This removes a branch test from most thread execution.

> - Remove stack.length === 0 tests

The work in the depending PRs get us to a point where status === DONE when stack is length 0. So they should be synamous and we can use the one test.

### DIFF
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -221,13 +221,10 @@ class Sequencer {
             thread.warpTimer = new Timer();
             thread.warpTimer.start();
         }
-        // Store the current block ID outside of the loop to later set it to
-        // thread.blockGlowInFrame.
-        let currentBlockId = thread.peekStack();
         // The thread entered a null block while outside of stepThreads during a
         // stepHat or promise resolution. We can't unwrap during stepHat or
         // promise resolution if we need to change the thread status.
-        if (currentBlockId === null) {
+        if (thread.peekStack() === null) {
             // Unwrap the stack until we find a loop block, a non-null
             // block, or the top of the stack.
             this.unwrapStack(thread);
@@ -238,7 +235,7 @@ class Sequencer {
         }
         while (thread.status === Thread.STATUS_RUNNING) {
             // Save the current block ID to notice if we did control flow.
-            currentBlockId = thread.peekStack();
+            const currentBlockId = thread.peekStack();
 
             if (this.runtime.profiler === null) {
                 // Execute the current block.
@@ -284,8 +281,6 @@ class Sequencer {
                 thread.status = Thread.STATUS_RUNNING;
             }
         }
-        // Indicate the block that just executed.
-        thread.blockGlowInFrame = currentBlockId;
         // If we yielded out of the thread, set status to RUNNING so stepThreads
         // can count it as an activeThread and possibly step all threads an
         // extra time.
--- a/src/engine/sequencer.js
+++ b/src/engine/sequencer.js
@@ -77,7 +77,7 @@ class Sequencer {
         // Start counting toward WORK_TIME.
         this.timer.start();
         // Count of active threads.
-        let numActiveThreads = Infinity;
+        let activeThreads = true;
         // Whether `stepThreads` has run through a full single tick.
         let ranFirstTick = false;
         const doneThreads = [];
@@ -86,7 +86,7 @@ class Sequencer {
         // 2. Time elapsed must be less than WORK_TIME.
         // 3. Either turbo mode, or no redraw has been requested by a primitive.
         while (this.runtime.threads.length > 0 &&
-               numActiveThreads > 0 &&
+               activeThreads &&
                this.timer.timeElapsed() < WORK_TIME &&
                (this.runtime.turboMode || !this.runtime.redrawRequested)) {
             if (this.runtime.profiler !== null) {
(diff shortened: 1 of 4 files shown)


mzgoddard:
@kchadha I refactored this PR so its now standalone.

This still achieves the goal that a thread with an empty stack array or stackFrame array has a status of DONE. This may leave the thread's peekStack at null and a status that is not DONE as long as the stack/stackFrames arrays are not length 0.

To achieve that it does a few added things that the prior dependee branches did.

> - Perform all sequencer and thread state logic in sequencer

Some logic in runtime.js and execute.js perform sequencer logic for hat and promise support. That logic is now either an existing thread method, a new sequencer method, or handled by Sequencer.stepThread.

> - add unwrapStack

Add a method to pop a thread's stack until it is exhausted, hits a loop or non-null block. The method's logic is reduced from its prior form to loop until the thread's status changes or peekStack is not null.

> - treat popping into a loop block literally as STATUS_YIELD

Part of unwrapStack replaces some warpMode consideration logic with setting the thread status to yield. Other code in stepThread can then consider if the yielded loop block should execute anyways due to warpMode or other cases.

> - remove `continue;` and `return;` statements from stepThread

Reorder the logic so cases using continue and return and be removed and allow the loop to execute normally and not exit prematurely.

src/engine/sequencer.js
```diff
@@ -75,7 +75,7 @@ class Sequencer {
         // Start counting toward WORK_TIME.
         this.timer.start();
         // Count of active threads.
-        let numActiveThreads = Infinity;
+        let activeThreads = true;
```
kchadha, line 80: Could you update the comment above this line?

src/engine/sequencer.js
```diff
@@ -94,48 +94,48 @@ class Sequencer {
...
+                    // behaviour.)
+                    i--;
+                    // Skip ahead to running this thread again. We don't want
+                    // this to count towards activeThreads until after
+                    // execution.
+                    continue;
                 }
+                // Check if the thread is running and step threads again of if
+                // it completed while it just stepped to make sure remove it
+                // before the next iteration of all threads.
```
kchadha: I think this comment has some typos...

src/engine/sequencer.js
```diff
@@ -168,109 +167,128 @@ class Sequencer {
...
+
+            if (thread.status === Thread.STATUS_YIELD &&
+                thread.peekStackFrame().warpMode &&
+                thread.warpTimer.timeElapsed() <= Sequencer.WARP_TIME) {
+                // In warp mode, yielded blocks are re-executed immediately.
+                thread.status = Thread.STATUS_RUNNING;
+            }
+        }
+        // Indicate the block that just executed.
+        thread.blockGlowInFrame = currentBlockId;
```
kchadha: Shouldn't this happen inside the while loop?

src/engine/sequencer.js
```diff
@@ -168,109 +167,128 @@ class Sequencer {
...
+        ) {
+            thread.popStack();
+
+            const stackFrame = thread.peekStackFrame();
+            if (stackFrame === null) {
+                // No more stack to run!
+                this.retireThread(thread);
+            } else if (stackFrame.isLoop) {
+                // The current level of the stack is marked as a loop.
+                // Return to yield for the frame/tick in general.
```
kchadha: This comment is a little out of date since we're not returning out of a function anymore.

src/engine/sequencer.js
```diff
@@ -168,109 +167,128 @@ class Sequencer {
...
+
     /**
      * Step the requested thread for as long as necessary.
      * @param {!Thread} thread Thread object to step.
      */
     stepThread (thread) {
-        let currentBlockId = thread.peekStack();
-        if (!currentBlockId) {
-            // A "null block" - empty branch.
-            thread.popStack();
```
kchadha: I think these changes need to be rebased off of current develop since this PR is missing the fix from #2126

kchadha (changes requested):
Some minor comments/changes requested, and it seems this branch needs to be updated against current develop to pull in changes from #2126

src/engine/sequencer.js
mzgoddard: Hmm. `blockGlowInFrame` is one of those thread parts that I thought was odd to be in sequencer. I think it'd make the most sense being next to wherever requestScriptGlowInFrame is set.

Looking back at this, I've done just that. I moved blockGlowInFrame into execute alongside requestScript before the loops in execute(). That'll update the blockGlowInFrame every time we run a block.

kchadha:
@mzgoddard, these changes are almost good to go, but I did notice that some stacks aren't glowing (they are in production). E.g. in this project 288115558, Sprite1 has a `when [any] key pressed, next costume` stack, and if I hold down a key (e.g. space), the stack will repeatedly glow in production but doesn't glow at all locally w/these changes pulled in.

PR #2156, opened 2019-04-29
Add standalone player example to playground

spectranaut:
This PR adds a experimental "standalone player" to the GUI's playground. The standalone-player.html  is example of how to create minimalist scratch project player without the editor view or the GUI.

I originally made the PR on GUI here, but it seems like (since the point is to not use GUI) we could put it in the VM playground instead. Original PR with feedback: https://github.com/LLK/scratch-gui/pull/4729


spectranaut:
heya @paulkaplan -- could you look over this PR? It's the same player as the one you reviewed for scratch-gui (https://github.com/LLK/scratch-gui/pull/4729), so should be easy to look over :)

PR #2160, opened 2019-04-30
BOOST extension: Always clear motor promises when setting state

ericrosenbaum:
We saw a bug (no issue filed) where if you ran a "turn motor A for rotations" block, and before it completes, run another "turn motor A for rotations", the first block will hang (i.e. stay highlighted).

This change fixes that problem, and a similar problem with the "on for seconds" block.


evhan55 (commented):
lgtm!

PR #2161, opened 2019-05-01
Supporting VM changes for extensionification

cwillisf:
### Dependency

This PR and LLK/scratch-gui#4790 should be merged at about the same time, as they make a small-but-important change to the way extension block information is communicated between the VM and the GUI.

### Proposed Changes

This is a grab-bag of supporting changes for extensionification, including:
* Support for interpreting extension block info through a Blockly mutation (late / "dynamic") instead of in the VM (immediate / "static")
  * This makes it possible to have two different blocks with the same opcode but different attributes in some other way, such as two variable blocks or custom procedure call blocks.
  * This also opens the door to removing some of the `scratch-blocks` / Blockly knowledge from `scratch-vm` if we decide to remove the "static" path.
* Loading a "core" extension will now place the extension blocks in the correct location in the toolbox instead of at the bottom. Non-"core" extensions will still load at the bottom.
* It's now possible to make core extension blocks look the same as the non-extension core blocks:
  * Extension blocks which don't have an extension icon will no longer be "inflated" using the `scratch_extension` Blockly extension.
  * Colors specified by an extension are now honored.
* Extension metadata will now be updated when changing the current editing target by selecting a different sprite/stage.

### Reason for Changes

It might make sense to review and/or merge this as one PR, or it might make sense to break it up into more than one. Either way, this PR can serve as a way to keep tabs on the current state of the `e16n` branch.

### Test Coverage

* Added `test/unit/engine_mutation-adapter.js` to test the new mutation adapter features.
* Modified `test/integration/internal-extension.js`, `test/unit/engine_blocks.js`, and `test/unit/extension_conversion.js` to cover new or modified block and category information.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -152,6 +152,10 @@ class SomeBlocks {
             // Will be used as the extension's namespace.
             id: 'someBlocks',
 
+            // Core extensions only: override the default extension block colors.
+            color1: '#FF8C1A',
+            color2: '#DB6E00',
+
             // Optional: the human-readable name of this extension as string.
             // This and any other string to be displayed in the Scratch UI may either be
             // a string or a call to `formatMessage`; a plain string will not be
--- a/src/engine/mutation-adapter.js
+++ b/src/engine/mutation-adapter.js
@@ -13,6 +13,10 @@ const mutatorTagToObject = function (dom) {
     for (const prop in dom.attribs) {
         if (prop === 'xmlns') continue;
         obj[prop] = decodeHtml(dom.attribs[prop]);
+        if (prop === 'blockinfo') {
+            obj.blockInfo = JSON.parse(obj.blockinfo);
+            delete obj.blockinfo;
+        }
     }
     for (let i = 0; i < dom.children.length; i++) {
         obj.children.push(
--- a/src/engine/mutation-adapter.js
+++ b/src/engine/mutation-adapter.js
@@ -13,6 +13,8 @@ const mutatorTagToObject = function (dom) {
     for (const prop in dom.attribs) {
         if (prop === 'xmlns') continue;
         obj[prop] = decodeHtml(dom.attribs[prop]);
+        // Note: the capitalization of block info in the following lines is important.
+        // The lowercase is read in from xml which normalizes case. The VM uses camel case everywhere else.
         if (prop === 'blockinfo') {
             obj.blockInfo = JSON.parse(obj.blockinfo);
             delete obj.blockinfo;
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -787,11 +787,9 @@ class Runtime extends EventEmitter {
 
         this._fillExtensionCategory(categoryInfo, extensionInfo);
 
-        const fieldTypeDefinitionsForScratch = [];
         for (const fieldTypeName in categoryInfo.customFieldTypes) {
             if (extensionInfo.customFieldTypes.hasOwnProperty(fieldTypeName)) {
                 const fieldTypeInfo = categoryInfo.customFieldTypes[fieldTypeName];
-                fieldTypeDefinitionsForScratch.push(fieldTypeInfo.scratchBlocksDefinition);
 
                 // Emit events for custom field types from extension
                 this.emit(Runtime.EXTENSION_FIELD_ADDED, {
@@ -801,9 +799,7 @@ class Runtime extends EventEmitter {
             }
         }
 
-        const allBlocks = fieldTypeDefinitionsForScratch.concat(categoryInfo.blocks).concat(categoryInfo.menus);
-
-        this.emit(Runtime.EXTENSION_ADDED, allBlocks);
+        this.emit(Runtime.EXTENSION_ADDED, categoryInfo);
     }
 
     /**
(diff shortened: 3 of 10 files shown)


docs/extensions.md
```diff
@@ -152,6 +152,10 @@ class SomeBlocks {
             // Will be used as the extension's namespace.
             id: 'someBlocks',
 
+            // Core extensions only: override the default extension block colors.
+            color1: '#FF8C1A',
+            color2: '#DB6E00',
```
towerofnix, line 157: Should `color3` be here?

Also, when this is called `color1`, `color2`, `color3`, to me it sounds like it's just yelling to be in the format `colors: [first, second, third]`. But that takes an additional comment making it explicit that this is an *ordered* array, not just a set of colors. So, instead, I think the names `colorPrimary`, `colorSecondary`, and `colorTertiary` would be better, along with a description of what those mean (which will be easier to remember than color1/2/3!) and maybe a link to [the color specification](https://github.com/LLK/scratch-blocks/wiki/Colors).

Also, yeah, internally I think these [*are*](https://github.com/LLK/scratch-vm/pull/2161/commits/91f0d59be078f8c120b5429352f284ce87bc8c6b#diff-38d875e1d50b3158dc96404b65275a54R785) given the names color1/color2/color3. But (assuming that can't easily be changed immediately) it's still future proofing to give more specific names to the colors in extensions. Making a change to better names later would mean either all future extension code requiring updates, or including code in the VM to support both color1/2/3 and primary/secondary/tertiary formats (which makes any future code use *either* of those formats, which is sure to be confusing).

cwillisf: Yeah - I've gone back and forth with how I feel about which names to use and whether to use an array. I feel like `colorPrimary`, `colorSecondary`, and `colorTertiary` don't offer much advantage since they basically mean the same as `color1`, `color2`, and `color3` and are harder to spell.

Still, I agree that these names aren't great either. Ideally I'd like more semantically-meaningful names. I'll think on this a bit more...

Also: I agree that a link to the color spec would be a good idea. Thanks for the suggestion!

kchadha: Hmmm... semantically meaningful names... maybe we can talk to @carljbowman about this?

I *think* the meanings of the colors are as follows:
color1/colorPrimary: the base color of the block (usually the most prevalent color on the block)
color2/colorSecondary: The color of shadows/droppable menus on the block (e.g. the dark green on the boolean blocks)
color3/colorTertiary: The color of the outline of the blocks

If these definitions are actually correct, maybe we could have something like:

baseColor, shadowColor (wouldn't actually want to use the word shadow here necessarily), and outlineColor...?

image

cwillisf: I talked to @carljbowman about this a while ago but forgot to put notes here -- sorry!

Anyway, he suggested that we stick with non-semantic names. I probably won't capture his reasoning accurately but the way I understood it is that there are a bunch of semantically-disconnected uses of each color that are hard to capture in just a few words.

Also, apparently [that color spec](https://github.com/LLK/scratch-blocks/wiki/Colors) is very out-of-date to the point that @carljbowman wasn't sure we should be linking to it since it might point people in the wrong direction. To me that suggests we should either take it down (and not link to it here) or update it as a separate project (and add a link here after that).

src/engine/mutation-adapter.js
```diff
@@ -13,6 +13,10 @@ const mutatorTagToObject = function (dom) {
     for (const prop in dom.attribs) {
         if (prop === 'xmlns') continue;
         obj[prop] = decodeHtml(dom.attribs[prop]);
+        if (prop === 'blockinfo') {
```
kchadha, line 18:
```suggestion
        // Note: the capitalization of block info in the following lines is important. 
        // The lowercase is read in from xml which normalizes case. The VM uses camel case everywhere else.
        if (prop === 'blockinfo') {
```

src/engine/runtime.js
```diff
@@ -801,9 +816,7 @@ class Runtime extends EventEmitter {
             }
         }
 
-        const allBlocks = fieldTypeDefinitionsForScratch.concat(categoryInfo.blocks).concat(categoryInfo.menus);
-
-        this.emit(Runtime.EXTENSION_ADDED, allBlocks);
+        this.emit(Runtime.EXTENSION_ADDED, categoryInfo);
```
kchadha, line 819: 🎉

src/engine/runtime.js
```diff
@@ -812,18 +825,16 @@ class Runtime extends EventEmitter {
...
-        let extensionBlocks = [];
         for (const categoryInfo of this._blockInfo) {
             if (extensionInfo.id === categoryInfo.id) {
                 categoryInfo.name = maybeFormatMessage(extensionInfo.name);
                 categoryInfo.blocks = [];
                 categoryInfo.menus = [];
                 this._fillExtensionCategory(categoryInfo, extensionInfo);
-                extensionBlocks = extensionBlocks.concat(categoryInfo.blocks, categoryInfo.menus);
+
+                this.emit(Runtime.BLOCKSINFO_UPDATE, categoryInfo);
```
kchadha: Documenting a conversation from offline...
It looks like this change would make this line run multiple times per call to `_refreshExtensionPrimitives` but in practice, this change is the same as before, this emit runs only once per call to `_refreshExtensionPrimitives` because there should only ever be one category in `this._blockInfo` which matches the given extension id.

We should refactor this in the future to prevent confusion.

src/engine/runtime.js
```diff
@@ -1272,12 +1286,12 @@ class Runtime extends EventEmitter {
...
-            xmlParts.push(`<category name="${name}" id="${categoryInfo.id}"
-                ${statusButtonXML} ${colorXML} ${menuIconXML}>`);
-            xmlParts.push.apply(xmlParts, paletteBlocks.map(block => block.xml));
-            xmlParts.push('</category>');
-        }
-        return xmlParts.join('\n');
+            return {
+                id: categoryInfo.id,
+                xml: `<category name="${name}" id="${categoryInfo.id}" ${statusButtonXML} ${colorXML} ${menuIconXML}>${
+                    paletteBlocks.map(block => block.xml).join('')}</category>`
```
kchadha: Is the `.join('')` here necessary or just for ease of reading?

src/engine/runtime.js
cwillisf: Refactored in fd776025e5fe2cfd2b2d5f7fa004ea94985b1447

kchadha (approved):
LGTM!

PR #2163, opened 2019-05-02
Add support for custom context menu items to the extension spec.

kchadha:
### Resolves

Towards #2120 

Related to LLK/scratch-gui#4794

### Proposed Changes

Allow extensions to specify custom context menu items for dynamic blocks.

### Reason for Changes

Needed for extensionifying myBlocks and variables which have custom context menu items for editing a custom procedure or renaming/deleting a variable.

**Note:** this feature is only active for dynamic extension blocks.

### Test Coverage

Added a custom context menu to a new block in the core example extension.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -575,7 +575,9 @@ should appear in.
 The options for the context menu item context are as follows:
 
 `ContextMenuContext.ALL` - The context menu item should always appear in the block's context menu. This is the default option.
+
 `ContextMenuContext.TOOLBOX_ONLY` - The context menu item should only appear in the block's context menu when the block is in the toolbox.
+
 `ContextMenuContext.WORKSPACE_ONLY` - The context menu item should only appear in the block's context menu when the block is on the main workspace.
 
 By default, a context menu item will appear on the block in both the toolbox as well as the
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -487,3 +487,111 @@ class SomeBlocks {
     };
 }
 ```
+
+## Experimental / In Development Features
+
+The following are experimental features that are under active development and subject to change during the process
+of converting the core Scratch blocks into using the extension spec.
+
+### Dynamic Blocks
+The content above describes defining static extension blocks (e.g. blocks that will always keep the same shape). While most
+blocks in the Scratch language fall into this category, there are others which may dynamically change their shape based
+state information. An example of this is the `control_stop` block which changes whether or not it can have a command block
+attached after it based on which menu item is selected. Blocks that may dynamically change their shape are referred to below as
+"dynamic blocks".
+
+In order to support Scratch blocks like variables or custom procedure call blocks, we have added the support for dynamic blocks.
+A block is specified to be dynamic by using the `isDynamic` flag in the block specification.
+
+```js
+class SomeBlocks {
+    // ...
+    getInfo () {
+        return {
+            // ...
+            blocks: [
+                {
+                    isDynamic: true
+                    opcode: 'dynamicReporter',
+                    blockType: BlockType.REPORTER,
+                    text: 'my dynamic reporter block',
+                }
+            ]
+        };
+    }
+    // ...
+}
+```
+
+#### Adding Custom Context Menu Options
+Dynamic blocks can have custom context menu options in addition to the default options for adding
+a block comment, deleting the block, and duplicating the block.
+
+In order to specify custom context menu options, you can provide a list of context menu item descriptors
+which contain the text label for the menu item as well as the name of the function in the extension that should be run when
+the block is selected.
+
+```js
+class SomeBlocks {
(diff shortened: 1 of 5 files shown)


test/integration/internal-extension.js
```diff
@@ -100,14 +100,17 @@ test('load sync', t => {
 
     t.equal(vm.runtime._blockInfo.length, 1);
 
-    // blocks should be an array of two items: a button pseudo-block and a reporter block.
-    t.equal(vm.runtime._blockInfo[0].blocks.length, 2);
+    // blocks should be an array of two items: a button pseudo-block, a reporter block, and a command block.
```
cwillisf: Rephrasing the comment like this might make it harder to accidentally miss updating the number in the comment
```suggestion
    // blocks should be an array of [button pseudo-block, reporter block, command block].
```

test/integration/internal-extension.js
```diff
@@ -100,14 +100,17 @@ test('load sync', t => {
...
+    t.equal(vm.runtime._blockInfo[0].blocks.length, 3);
     t.type(vm.runtime._blockInfo[0].blocks[0].info, 'object');
     t.type(vm.runtime._blockInfo[0].blocks[0].info.func, 'MAKE_A_VARIABLE');
     t.equal(vm.runtime._blockInfo[0].blocks[0].info.blockType, 'button');
     t.type(vm.runtime._blockInfo[0].blocks[1].info, 'object');
     t.equal(vm.runtime._blockInfo[0].blocks[1].info.opcode, 'exampleOpcode');
     t.equal(vm.runtime._blockInfo[0].blocks[1].info.blockType, 'reporter');
+    t.type(vm.runtime._blockInfo[0].blocks[2].info, 'object');
+    t.equal(vm.runtime._blockInfo[0].blocks[2].info.opcode, 'exampleDynamicOpcode');
+    t.equal(vm.runtime._blockInfo[0].blocks[2].info.blockType, 'command');
```
cwillisf, line 120: Shouldn't this include a test for the context menu itself? Maybe something like this? (WARNING: not tested...)
```suggestion
    t.equal(vm.runtime._blockInfo[0].blocks[2].info.blockType, 'command');
    t.equal(vm.runtime._blockInfo[0].blocks[2].info.customContextMenu.length, 1);
    t.type(vm.runtime._blockInfo[0].blocks[2].info.customContextMenu[0], 'function');
```

src/blocks/scratch3_core_example.js
```diff
@@ -31,6 +32,18 @@ class Scratch3CoreExample {
...
                     text: 'example block'
+                },
+                {
+                    opcode: 'exampleDynamicOpcode',
+                    blockType: BlockType.COMMAND,
+                    text: 'example dynamic block',
+                    isDynamic: true,
+                    customContextMenu: [
+                        {
+                            name: 'Custom Context Menu Option',
```
cwillisf: What do you think about `label` instead of `name`?

cwillisf (commented):
Looking good! I made a couple suggestions that we can discuss but they're mostly practical details rather than objections about the approach or spec proposal.

test/integration/internal-extension.js
kchadha: Done!

docs/extensions.md
```diff
@@ -487,3 +487,113 @@ class SomeBlocks {
...
+    // ...
+}
+```
+
+For each context menu item, you can also optionally specify which situations that context menu item
+should appear in.
+
+The options for the context menu item context are as follows:
+
+`ContextMenuContext.ALL` - The context menu item should always appear in the block's context menu. This is the default option.
```
cwillisf, line 577: `ContextMenuContext` :D

test/integration/internal-extension.js
```diff
@@ -81,16 +82,27 @@ test('internal extension', t => {
...
-    t.type(
-        vm.runtime._blockInfo[0].menus[1].json.args0[0].options, 'function');
-
-    t.end();
+    // There should be at least one extension loaded, but there could be more
+    // loaded if we have core extensions defined in src/virtual-machine.
+    t.ok(vm.runtime._blockInfo.length > 0);
+
+    for (const extensionInfo of vm.runtime._blockInfo) {
+        if (extensionInfo.id !== 'testInternalExtension') continue;
```
cwillisf: 👍
You could slightly simplify this using something like `const extensionInfo = vm.runtime._blockInfo.find(x => x.id === 'testInternalExtension');` but the for/continue setup is already much better than the sometimes-broken version we had before :)

cwillisf (approved):
I made one suggestion but it's totally optional... this looks great! Thanks for the docs, too!

test/integration/internal-extension.js
kchadha: That's a great suggestion! Done.

PR #2164, opened 2019-05-02
Do not localize BOOST motor IDs

ericrosenbaum:
We inadvertently were using `formatMessage` for strings that should not be localized: the strings for selecting BOOST motors from a menu, which are "A", "B", "C", "D", "AB" and "ABCD," and should not vary by language (because they are on the BOOST hub).


evhan55 (approved):
lgtm!

PR #2168, opened 2019-05-11
Change BOOST variable names regarding duration and rotation blocks

### LINKED ISSUE #2153: BOOST extension: Rename pendingPromiseFunction variable?
The `pendingPromiseFunction` variable is used to track a Promise that resolves when a motor has reached a desired position after some rotations.  The variable name could be clearer to reflect that fact, perhaps `pendingRotationPromise` ? The comment on lines `#359-#360` also seems inaccurate, saying the variable is used to track duration and not rotation.  This might also require changing the setter/getter names on lines `#464` and `#471`. (This setter/getter pair might not be necessary since this variable is private, anyway).

evhan55:
### Resolves

- Resolves #2153: BOOST extension: Rename pendingPromiseFunction variable?
- Fixes some inaccurate class variable comments.

### Reason for change

- Although this name change diverges slightly from the WeDo2 variable names, it adds clarity to the complex BOOST duration/rotation pending functions, timings and behaviors.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -437,21 +437,21 @@ class BoostMotor {
     }
 
     /**
-     * @return {number} - time, in milliseconds, of when the pending timeout began.
+     * @return {number} - time, in milliseconds, of when the pending duration timeout began.
      */
     get pendingTimeoutStartTime () {
         return this._pendingDurationTimeoutStartTime;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


ericrosenbaum (approved):
LGTM! tested a bunch of motor commands on mac, and tried all the starter projects. It looks like `_pendingRotationDestination` is basically no longer being used (since we are not updating rotation commands anymore), but it's okay to leave it in.

PR #2175, opened 2019-05-20
Do not load speech2text extension

ericrosenbaum:


picklesrus (approved):

AmazingMech2418:
Why did the Speech2Text extension need to be removed? I personally think that it may be ready to be deployed as a featured extension after some testing I have done. Also, some of my projects now don't load since the blocks no longer exist.

Issue #2172, opened 2019-05-15
Generalize scratch-link connection method


paulkaplan:
Right now, the code for managing the connection to scratch-link sessions is tightly coupled to web-sockets, and is not configurable. We need to be able to decouple the `BLE`/`BT` IO objects used by extensions from WebSocket connections, and allow the VM to be configured to use a different way of communicating with a scratch-link service.

We are experimenting with a few different ways of doing this, but just wanted to create a stub issue to track it.

/cc @cwillisf @evhan55

PR #2182, opened 2019-05-24
Initial prototype of configurable scratch link socket

paulkaplan:
/cc @chrisgarrity @cwillisf @evhan55 @knandersen 

I wanted to put up some code that we could talk through as a group to solve https://github.com/LLK/scratch-vm/issues/2172

The objective is to allow someone using the VM to customize how extensions communicate with scratch-link to talk to hardware. I had a couple of priorities:
- Extension code should be minimally impacted
- Configuring the scratch-link connection should not require re-implementing the JSONRPC code on the client-side, since that code already exists and works well. 

The use-cases I used as references were
1. A native JSInterface, where communication with native code takes place over a few global JS functions provided by the WebView.
2. A hypothetical in-browser scratch-link built on web-bluetooth. Mostly as another point to triangulate from, although in-browser scratch-link is a bit ackward because it requires another in-browser JSONRPC to act as the "server", but that seems ok.

I landed for this prototype on removing the `JSONRPCWebSocket` class from the inheritance chain, but keeping BLE /BT inheriting from JSONRPC. I made a new, abstract ScratchLinkSocket, with a websocket implementation. The VM has public methods for configuring the ScratchLinkSocket creation function. BLE/BT request a ScratchLinkSocket from the runtime, and wire up the lifecycle methods (onOpen,onClose,onError) and message handling to work with the JSONRPC functionality. 

---

For the native JSInterface side, here is some pseudocode i imagine could be used for that.
First, the JSInterface would provide a few global functions like `window.createLinkSession `, `window.receiveLinkMessage()`, `window.closeLinkSession()`. The native side would execute a global function like `window.sendLinkMessage` when it wants to send a message.

Then on the JS side, a custom Socket class could be configured on the VM, which bridges the send/receive, so `Socket.sendMessage` => `window.receiveMessage`, and `window.sendMessage` => `Socket.handleMessage`

(this is just by way of example, it is ignoring multiple link sessions, the lifecycle methods, etc.)


src/util/scratch-link-websocket.js
```diff
@@ -0,0 +1,83 @@
...
+    constructor (type) {
+        this._type = type;
+        this._onOpen = null;
+        this._onClose = null;
+        this._onError = null;
+        this._handleMessage = null;
+
+        this._ws = null;
+    }
+
```
chrisgarrity, line 24: Would it make sense to also have a constructor that includes all the functions?

evhan55 (commented):
This looks great to me so far!  Especially since there are no changes to calls within BT/BLE which I think means little to no changes in existing extensions, awesome!

One tiny point of confusion I can see coming up in the future with these changes:
In the GDX_FOR extension from Vernier: in the file `scratch-link-device-adapter.js`, the BLE object is given the name `scratchLinkSocket`.  I've been meaning to rename that back to `ble` or something else to match all other extensions, and this PR would probably make that even more necessary.

src/util/scratch-link-websocket.js
chrisgarrity: Nope, on re-reading I don't see a case where it would be used.

src/util/scratch-link-websocket.js
```diff
@@ -0,0 +1,83 @@
...
+        this._handleMessage = fn;
+    }
+
+    isOpen () {
+        return this._ws && this._ws.readyState === this._ws.OPEN;
+    }
+
+    _onMessage (e) {
+        const json = JSON.parse(e.data);
+        this._handleMessage(json);
```
chrisgarrity, line 80: Need to check for `_handleMessage === null`. I'd lean towards checking in `open`, and just having one error.

src/util/scratch-link-websocket.js
```diff
@@ -0,0 +1,83 @@
...
+        default:
+            throw new Error(`Unknown ScratchLink socket Type: ${this._type}`);
+        }
+
+        if (this._onOpen && this._onClose && this._onError) {
+            this._ws.onopen = this._onOpen;
+            this._ws.onclose = this._onClose;
+            this._ws.onerror = this._onError;
+        } else {
+            throw new Error('Must set open, close and error handlers before calling open on the socket');
```
chrisgarrity: Obviously this error should never been seen by a user, so it's not something we should worry about localizing. Would it be worth having a convention for this type of error to be clear that it's not something the end user would see? e.g. `Error('INTERNAL: Must set...')`

Just started thinking about this after reading the API endpoint guide: https://github.com/LLK/scratch-api/wiki/Endpoint-route-design-guide#responses-should-be-machine-readable

chrisgarrity (changes requested):
Mostly details, but it should probably handle the case where the user of `ScratchLinkWebSocket` forgets to `setHandleMessage`

chrisgarrity (approved):

evhan55:
I have submitted a PR to change some variable names in the existing GDXFOR extension to avoid potential confusions with changes in this PR:
https://github.com/LLK/scratch-vm/pull/2194

evhan55 (approved):
lgtm! Would be nice to sanity-check and test all the hardware extensions whenever this lands.

evhan55:
@ericrosenbaum @BryceLTaylor I think it would be nice to minimally test all the hardware extensions if/whenever this lands.

paulkaplan:
@evhan55 yes absolutely. I'm going to merge this now and make sure it gets up to staging ASAP so we can do both async testing and testing as a group on thursday.

PR #2184, opened 2019-05-26
Fall back to default bitmap if asset loading fails

adroitwhiz:
### Resolves

I'm not sure whether there's a bug report for this, but it resolves one cause of the dreaded "Oops!" screen. Sometimes, bitmap assets will 404, and previously there was no error checking to deal with this like there is for vector assets, so the project would crash upon loading and become completely unviewable. This was causing one forum user [a lot of distress](https://scratch.mit.edu/discuss/topic/352542/).

### Proposed Changes

This PR loads the placeholder question mark if a bitmap asset fails to load, just like it does with vector assets.

### Reason for Changes

The "Oops" screen must be avoided at all costs.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -268,7 +268,14 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
                 return loadVector_(costume, runtime);
             });
     }
-    return loadBitmap_(costume, runtime, rotationCenter, optVersion);
+    return loadBitmap_(costume, runtime, rotationCenter, optVersion)
+        .catch(() => {
+            // Use default asset if original fails to load
+            costume.assetId = runtime.storage.defaultAssetId.ImageBitmap;
+            costume.asset = runtime.storage.get(costume.assetId);
+            costume.md5 = `${costume.assetId}.${AssetType.ImageBitmap.runtimeFormat}`;
+            return loadBitmap_(costume, runtime);
+        });
 };
 
 /**


src/import/load-costume.js
```diff
@@ -268,7 +268,14 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
...
     }
-    return loadBitmap_(costume, runtime, rotationCenter, optVersion);
+    return loadBitmap_(costume, runtime, rotationCenter, optVersion)
+        .catch(() => {
+            // Use default asset if original fails to load
+            costume.assetId = runtime.storage.defaultAssetId.ImageBitmap;
+            costume.asset = runtime.storage.get(costume.assetId);
+            costume.md5 = `${costume.assetId}.${AssetType.ImageBitmap.runtimeFormat}`;
+            return loadBitmap_(costume, runtime);
+        });
```
towerofnix, line 278: Should there still be a console warning in here, so that it's easier for users/mods to investigate why a costume isn't loading?

adroitwhiz: Probably. For consistency, however, that would require adding a warning to the vector loading code too, which makes it seem like a different change for a different PR.

thisandagain:
This is something we've discussed at length internally. The problem here is that the project may get saved in the this state if the asset fails to load even if it is an issue with our servers rather than the asset actually not existing on our servers. We need to carefully distinguish between these two types of error conditions or we run the risk of corrupting many, many assets in the event of a service outage.

PR #2187, opened 2019-05-29
Consistently format text bubbles

### LINKED ISSUE #2186: "Say" and "Think" bubbles have different numeric precision
### Expected Behavior

"Say" and "Think" bubbles should display the same output when given the same input, with the only difference being the shape of the bubble.

### Actual Behavior

~While visions of floating-point precision errors drift through Scratch Cat's thoughts, he is too shy to vocalize them~
"Say" bubbles are limited to 2 decimal digits of numeric precision, but "think" bubbles are not:
image

### Steps to Reproduce

1. Make a sprite "say" a number with more than 2 digits of precision
2. Make the sprite "think" the number

(Or see [this project](https://scratch.mit.edu/projects/313667447/editor/))

adroitwhiz:
### Resolves

Resolves #2186 

### Proposed Changes

This adds a new private function, `_formatBubbleText`, to `Scratch3LooksBlocks`. This function, which formats the bubble's text by clamping decimal precision to 2 digits and limiting its length, is called by `_updateBubble`, which is in turn called by all functions which change text bubble state.

This formatting was previously done by the `say` and `think` functions. However, only the `say` function limited decimal precision, making speech bubbles and thinking bubbles inconsistent.

This does introduce a minor compatibility change from 2.0: "ask" bubbles' decimal precision is now limited as well, whereas it was not limited in 2.0.

Another minor (non-compatibility) change: `say` emitted a `SAY` event, while `think` called `_updateBubble` directly. Now, both emit a `SAY` event.

### Reason for Changes

These changes, in addition to fixing one current bug, reduce code duplication and potential for future bugs.

### Test Coverage

I have modified the `blocks_looks` unit test to check the "think" block's rounding, and to check the string from the bubble as displayed instead of from the `SAY` event.

- Should the `SAY` event's text come already formatted? I don't believe it should, which is why I changed the test, but I want your opinion(s).

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -234,6 +234,26 @@ class Scratch3LooksBlocks {
     }
 
     /**
+     * Properly format text for a text bubble.
+     * @param {string} text The text to be formatted
+     * @return {string} The formatted text
+     * @private
+     */
+    _formatBubbleText (text) {
+        if (text === '') return text;
+
+        // Limit decimal precision to 2 digits.
+        if (typeof text === 'number') {
+            text = parseFloat(text.toFixed(2));
+        }
+
+        // Limit the length of the string.
+        text = String(text).substr(0, Scratch3LooksBlocks.SAY_BUBBLE_LIMIT);
+
+        return text;
+    }
+
+    /**
      * The entry point for say/think blocks. Clears existing bubble if the text is empty.
      * Set the bubble custom state and then call _renderBubble.
      * @param {!Target} target Target that say/think blocks are being called on.
@@ -244,7 +264,7 @@ class Scratch3LooksBlocks {
     _updateBubble (target, type, text) {
         const bubbleState = this._getBubbleState(target);
         bubbleState.type = type;
-        bubbleState.text = text;
+        bubbleState.text = this._formatBubbleText(text);
         bubbleState.usageId = uid();
         this._renderBubble(target);
     }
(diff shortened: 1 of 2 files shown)


adroitwhiz:
@paulkaplan Mind taking a look at this when you get the chance?

src/blocks/scratch3_looks.js
```diff
@@ -325,7 +340,7 @@ class Scratch3LooksBlocks {
     }
 
     think (args, util) {
-        this._updateBubble(util.target, 'think', String(args.MESSAGE).substr(0, Scratch3LooksBlocks.SAY_BUBBLE_LIMIT));
+        this.runtime.emit('SAY', util.target, 'think', args.MESSAGE);
```
fsih: Can we change 'SAY' to a constant variable and rename it to 'SAY_OR_THINK'?

fsih (approved):
Looks good besides one nit!

src/blocks/scratch3_looks.js
adroitwhiz: As a static getter on `Runtime`, on `Scratch3LooksBlocks`, or just as a `const` at the top of the file?

fsih: I think static getter on Scratch3LooksBlocks

adroitwhiz: Hopefully [this](https://github.com/LLK/scratch-vm/pull/2187/commits/ffd2a2a9e8f7a95df6bd81d60fe1457fcb4a1918) is good. There are a lot of places in the codebase that still use the raw event string `'SAY'` as the event name, but the codebase also uses raw event strings for many other event listeners as well, so I opted against changing all of the `'SAY'` listeners over to `Scratch3LooksBlocks.SAY_OR_THINK`.

Even if I did rename all the `'SAY'` events in this codebase to `'SAY_OR_THINK'` (the string), `scratch-render`'s tests use `'SAY'` events as well, so those would need to be changed simultaneously to prevent breakage.

PR #2191, opened 2019-05-31
Add optional 'pin' parameter to BT connections

evhan55:
### Resolves

- This change makes it so that BT devices can send their own pairing pin to the Scratch Link connection request.
- It also updates the EV3 extension to send its own pairing pin to Scratch Link.
- Related to: https://github.com/LLK/scratch-link/pull/130


cwillisf (approved):
LGTM

ericrosenbaum (approved):
LGTM! tested with EV3 on mac and windows with Scratch Link 1.1.46

PR #2192, opened 2019-06-01
Fix Cast.LIST_ALL casting case

apple502j:
### Resolves
Resolves a bug where
`add (item (join [all] []) of [list v]) to [list v]` crashes the editor.
### Proposed Changes
Fix toListIndex and add new `acceptAll` argument.
If false, it will return Cast.LIST_INVALID instead.
### Reason for Changes
No crashing!

### Test Coverage
Added a new test to `unit/cast.js`


kchadha (approved):
LGTM! Thank you for fixing this bug! I apologize for the delayed review.

Issue #1974, opened 2019-02-05
Vernier Extension: rename `scratchLinkSocket` throughout to `ble`


evhan55:
To avoid potential confusions with the upcoming `ScratchLinkWebSocket` proposed in https://github.com/LLK/scratch-vm/pull/2182.

evhan55:
cc: @paulkaplan 
re: #2172 

Tagging you on this issue because I'll probably be closing this issue soon. But, I wanted you to know it had been on my radar to rethink these class names, in case you decide to completely change them during your work on refactoring these for future Scratch Link platform support.

evhan55:
Resolved via https://github.com/LLK/scratch-vm/pull/2194

PR #2194, opened 2019-06-03
GDXFOR: Rename scratchLinkSocket variables to 'socket' and 'ble' 

evhan55:
### Resolves

- #1974 Vernier Extension: rename `scratchLinkSocket` throughout to `ble`

### Reason for Changes

To avoid potential confusions with the upcoming `ScratchLinkWebSocket` class (see: https://github.com/LLK/scratch-vm/pull/2182).  Renamed to `ble` in the extension file to match the other extensions, and renamed to `socket` in the device adapter class.


ericrosenbaum (approved):
Looks good to me!

PR #2196, opened 2019-06-05
add Profiler.increment and Profiler.frame

mzgoddard:
### Resolves

Scratch performance when the profiler is recording.

### Proposed Changes

Let profiled code track frames and arguments called by incrementing a counter for a frame id or frame id and argument. This replaces the same counting by recording the call as part of a history of calls and returns. Updating the array for all calls counted takes enough time to bias the profiled run towards less overall executions.

### Reason for Changes

Scratch performance has improved to the point again that we need to change how the Profiler records performance to spend more time executing blocks than spending time recording that a block executed.

<details>
<summary><b>Benchmark Data</b></summary>

These numbers indicate a large improvement. Take note this improvement is only affecting when profiling is recording. This is a scratch developer facing improvement so we can get a better measure a change will make while developing the change and reviewing the change.

|  device | project id | warm up (s) | before (bps) | after (bps) | difference | change |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
|  chrome mbp 15" 2017 | 130041250 | 0 | 803343 | 1338256 | 534913 | 66.59% |
|  chrome mbp 15" 2017 | 130041250 | 4 | 1038659 | 1715644 | 676985 | 65.18% |
|  chrome mbp 15" 2017 | 14844969 | 0 | 1259861 | 1922094 | 662233 | 52.56% |
|  chrome mbp 15" 2017 | 14844969 | 1 | 1393017 | 2238741 | 845724 | 60.71% |
|  chrome mbp 15" 2017 | 173918262 | 0 | 1120676 | 1853687 | 733011 | 65.41% |
|  chrome mbp 15" 2017 | 173918262 | 5 | 1149243 | 1602950 | 453707 | 39.48% |
|  chrome mbp 15" 2017 | 155128646 | 0 | 1011027 | 2084500 | 1073473 | 106.18% |
|  chrome mbp 15" 2017 | 155128646 | 5 | 856636 | 855513 | -1123 | -0.13% |
|  chrome mbp 15" 2017 | 89811578 | 0 | 1550713 | 2469557 | 918844 | 59.25% |
|  chrome mbp 15" 2017 | 89811578 | 5 | 1711245 | 2640594 | 929349 | 54.31% |
|  chrome mbp 15" 2017 | 139193539 | 0 | 1398716 | 2378486 | 979770 | 70.05% |
|  chrome mbp 15" 2017 | 139193539 | 5 | 1462682 | 2508699 | 1046017 | 71.51% |
|  chrome mbp 15" 2017 | 187694931 | 0 | 863060 | 2118414 | 1255354 | 145.45% |
|  chrome mbp 15" 2017 | 187694931 | 5 | 980702 | 2220644 | 1239942 | 126.43% |
|  chrome mbp 15" 2017 | 219313833 | 0 | 164755 | 176664 | 11909 | 7.23% |
|  chrome mbp 15" 2017 | 219313833 | 5 | 165293 | 182444 | 17151 | 10.38% |
|  chrome mbp 15" 2017 | 236115215 | 0 | 5268 | 5553 | 285 | 5.41% |
|  chrome mbp 15" 2017 | 236115215 | 5 | 5289 | 5526 | 237 | 4.48% |
|  chrome mbp 15" 2017 | 238750909 | 0 | 90887 | 91218 | 331 | 0.36% |
|  chrome mbp 15" 2017 | 238750909 | 5 | 96803 | 94128 | -2675 | -2.76% |

</details>

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -192,6 +192,18 @@ class BlockCached {
         this.mutation = cached.mutation;
 
         /**
+         * The profiler the block is configured with.
+         * @type {?Profiler}
+         */
+        this._profiler = null;
+
+        /**
+         * Profiler information frame.
+         * @type {?ProfilerFrame}
+         */
+        this._profilerFrame = null;
+
+        /**
          * Is the opcode a hat (event responder) block.
          * @type {boolean}
          */
@@ -371,6 +383,25 @@ class BlockCached {
 }
 
 /**
+ * Initialize a BlockCached instance so its command/hat
+ * block and reporters can be profiled during execution.
+ * @param {Profiler} profiler - The profiler that is currently enabled.
+ * @param {BlockCached} blockCached - The blockCached instance to profile.
+ */
+const _prepareBlockProfiling = function (profiler, blockCached) {
+    blockCached._profiler = profiler;
+
+    if (blockFunctionProfilerId === -1) {
+        blockFunctionProfilerId = profiler.idByName(blockFunctionProfilerFrame);
+    }
+
+    const ops = blockCached._ops;
+    for (let i = 0; i < ops.length; i++) {
+        ops[i]._profilerFrame = profiler.frame(blockFunctionProfilerId, ops[i].opcode);
+    }
+};
+
+/**
  * Execute a block.
  * @param {!Sequencer} sequencer Which sequencer is executing.
  * @param {!Thread} thread Thread which to read and execute.
@@ -466,6 +497,8 @@ const execute = function (sequencer, thread) {
         currentStackFrame.reported = null;
     }
 
+    const start = i;
+
     for (; i < length; i++) {
         const lastOperation = i === length - 1;
         const opCached = ops[i];
@@ -487,27 +520,7 @@ const execute = function (sequencer, thread) {
 
         // Inputs are set during previous steps in the loop.
 
-        let primitiveReportedValue = null;
-        if (runtime.profiler === null) {
-            primitiveReportedValue = blockFunction(argValues, blockUtility);
-        } else {
-            const opcode = opCached.opcode;
-            if (blockFunctionProfilerId === -1) {
-                blockFunctionProfilerId = runtime.profiler.idByName(blockFunctionProfilerFrame);
-            }
-            // The method commented below has its code inlined
-            // underneath to reduce the bias recorded for the profiler's
-            // calls in this time sensitive execute function.
-            //
-            // runtime.profiler.start(blockFunctionProfilerId, opcode);
-            runtime.profiler.records.push(
-                runtime.profiler.START, blockFunctionProfilerId, opcode, 0);
-
-            primitiveReportedValue = blockFunction(argValues, blockUtility);
-
-            // runtime.profiler.stop(blockFunctionProfilerId);
-            runtime.profiler.records.push(runtime.profiler.STOP, 0);
-        }
+        const primitiveReportedValue = blockFunction(argValues, blockUtility);
 
         // If it's a promise, wait until promise resolves.
         if (isPromise(primitiveReportedValue)) {
@@ -556,6 +569,20 @@ const execute = function (sequencer, thread) {
                     parentValues[inputName] = primitiveReportedValue;
                 }
             }
+        }
+    }
+
+    if (runtime.profiler !== null) {
+        if (blockCached._profiler !== runtime.profiler) {
+            _prepareBlockProfiling(runtime.profiler, blockCached);
+        }
+        // Determine the index that is after the last executed block. `i` is
+        // currently the block that was just executed. `i + 1` will be the block
+        // after that. `length` with the min call makes sure we don't try to
+        // reference an operation outside of the set of operations.
+        const end = Math.min(i + 1, length);
+        for (let p = start; p < end; p++) {
+            ops[p]._profilerFrame.count += 1;
         }
     }
 };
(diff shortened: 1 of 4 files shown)


src/engine/execute.js
```diff
@@ -191,6 +191,18 @@ class BlockCached {
          */
         this.mutation = cached.mutation;
 
+        /**
+         * Profiler block is configured with.
```
kchadha: I read this wrong and thought it was an incomplete sentence at first.
```suggestion
         * The profiler the block is configured with.
```

src/engine/execute.js
```diff
@@ -370,6 +382,19 @@ class BlockCached {
     }
 }
 
+const prepareBlockProfiling = function (profiler, blockCached) {
```
kchadha: We should probably add jsdoc to this function, and probably also follow the `_` naming convention, implying this is for private use.

src/engine/execute.js
```diff
@@ -558,6 +565,16 @@ const execute = function (sequencer, thread) {
             }
         }
     }
+
+    if (runtime.profiler !== null) {
+        if (blockCached._profiler !== runtime.profiler) {
+            prepareBlockProfiling(runtime.profiler, blockCached);
+        }
+        const end = Math.min(i + 1, length);
```
kchadha, line 583: Can you add a comment about what this calculation is? It's difficult to keep track of units and what `i` and `length` correspond to.

src/engine/profiler.js
```diff
@@ -235,6 +285,20 @@ class Profiler {
             }
         }
 
+        for (let j = 0; j < this.increments.length; j++) {
+            if (this.increments[j] && this.increments[j].count > 0) {
```
kchadha: Is this second condition (`this.increments[j].count > 0`) ever false?

src/engine/profiler.js
```diff
@@ -126,6 +132,12 @@ class Profiler {
          */
         this.records = [];
 
+        this.increments = [];
```
kchadha, line 140: Could you add comments describing what `increments` and `counters` are?

src/engine/profiler.js
```diff
@@ -106,6 +106,12 @@ class ProfilerFrame {
          * @type {number}
          */
         this.depth = depth;
+
+        /**
+         * A summarized count of the number of calls to this frame.
+         * @type {number}
+         */
+        this.count = 1;
```
kchadha: Hmmm why isn't this initialized to `0`? Increments creates a new ProfilerFrame if one doesn't exist, and then increments this count anyways, so that means a newly created frame would get incremented to `2` which seems weird...

src/playground/benchmark.js
```diff
@@ -360,11 +369,12 @@ class Frames {
         this.frames = [];
     }
 
-    update (id, selfTime, totalTime) {
+    update (id, arg, selfTime, totalTime, count) {
```
kchadha, line 372: It doesn't look like `arg` is being used here

src/engine/execute.js
```diff
@@ -558,6 +565,16 @@ const execute = function (sequencer, thread) {
...
+        for (let p = start; p < end; p++) {
+            ops[p]._profilerFrame.count += 1;
```
kchadha, line 585: Why not use the increment function here?

kchadha (changes requested):
A few questions and minor changes requested inline.

src/playground/benchmark.js
mzgoddard: It's not used by this method. Frames.update uses the same arguments as RunningStats.update and Opcodes.update. They're all called together in Profiler.onFrame.

They have sequential arguments instead of deconstructing on there own for performance. The prior version of the update methods would be called per block executed. So it could thousands or more blocks per step of the runtime. This change aggregates those so we could change this API now to deconstruct without the same performance concern.

src/engine/profiler.js
mzgoddard: Hmm. I faintly remember this being 1 due to the branches creating frames not incrementing them. In that case this was an oversight after refactoring.

src/engine/execute.js
mzgoddard: `frame` supports an argument, normally the block opcode. `increment` does not support an argument.

src/engine/profiler.js
mzgoddard: Added.

src/engine/execute.js
mzgoddard: Added.

mzgoddard:
```js
// Determine the index that is after the last executed block. `i` is
// currently the block that was just executed. `i + 1` will be the block
// after that. `length` with the min call makes sure we don't try to
// reference an operation outside of the set of operations.
```

src/engine/execute.js
mzgoddard: Added.

src/engine/execute.js
mzgoddard: Added.

mzgoddard:
- Update code to fulfill requested changes.

kchadha:
LGTM but I think I want to get a second pair of eyes on this (either @cwillisf or @rschamp maybe?). I can confirm that I see a large improvement in the benchmark numbers being reported!

kchadha (approved):

rschamp (approved):

PR #2197, opened 2019-06-05
Update sound duration

### LINKED ISSUE #1780: Sound tile duration display should update on edits that change duration
### Expected Behavior

The displayed duration for a sound in the sound tile should update when an edit causes the duration to change.

### Actual Behavior

The displayed duration does not update.

### Steps to Reproduce

- Note that the duration of "Meow" is 0.85
- Make an edit that changes the duration: trim, faster, slower, echo, or robot
- Duration is unchanged

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/1780

### Proposed Changes

Update the sound object's properties, so that the sound tile in the GUI can display an updated duration when the sound is edited.

update-duration


paulkaplan (approved):
LGTM

PR #2201, opened 2019-06-06
Clean up hardware disconnect behavior

### LINKED ISSUE #1929: Hardware Extensions: separate extension cleanup and disconnect

evhan55:
### Resolves

- Resolves #1929: Hardware Extensions: separate extension cleanup and disconnect

### Proposed Changes

- Instead of relying on a "disconnect callback" from the extension to close the socket on some disconnection cases, always close the socket from within itself instead, and just call a "reset callback" on the extension so it can reset its state.  This requires the hardware extensions to separate out their disconnect behavior into distinct `disconnect` and `reset` functions.

### Reason for Changes

- Clarifies the hardware extension disconnect routines so they are more legible for developers and concerns are more neatly refactored into cleaner functions

### Testing

Would be good to test and verify the following behavior for *all* hardware extensions:
1. Disconnect by turning off bluetooth
2. Disconnect by turning off Scratch Link
3. Disconnect by turning off peripheral
4. Disconnect by clicking the 'disconnect' button in the peripheral connection modal
5. Disconnect by closing the tab
6. Disconnect by closing the browser
7. Disconnect by removing the battery from peripheral (if possible)

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -823,18 +823,22 @@ class Boost {
     }
 
     /**
-     * Disconnects from the current BLE socket.
+     * Disconnects from the current BLE socket and resets state.
      */
     disconnect () {
+        console.log('BOOST DISCONNECT CALLED');
         if (this._ble) {
             this._ble.disconnect();
         }
+
+        this.reset();
     }
 
     /**
      * Reset all the state and timeout/interval ids.
      */
     reset () {
+        console.log('BOOST RESET CALLED');
         this._ports = [];
         this._motors = [];
         this._sensors = {
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -826,8 +826,6 @@ class Boost {
      * Disconnects from the current BLE socket and resets state.
      */
     disconnect () {
-        console.log('BOOST DISCONNECT CALLED');
-
         if (this._ble) {
             this._ble.disconnect();
         }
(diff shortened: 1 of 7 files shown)


evhan55:
If this lands, it would be good to reflect the changes in the current SPIKE Prime prototype

evhan55:
@ericrosenbaum 

I just tested this PR with the EV3, GDXFor, microbit, WeDo2 and Boost on Windows, and all the disconnection cases worked properly. (sockets disconnect, GUI updates, and extension state resets).

I did have to make a small fix to your merge from last week: the GDXFor extension needed one more change from `this._scratchLinkSocket` to `this._ble`, so I made a commit on this PR today, on top of your merge.

ericrosenbaum:
I've now tested all seven disconnection testing scenarios on each of the five hardware extensions, and this is working well. I noticed a couple of issues but they do not seem to be specific to this PR (i.e. they also occurred on production):

- Mac Scratch Link crashed when EV3 battery was removed https://github.com/LLK/scratch-link/issues/138
- On Mac, extension does not show disconnection alert when WeDo battery is removed (added a note to the existing issue https://github.com/LLK/scratch-link/issues/87)

ericrosenbaum (approved):
LGTM!

PR #2202, opened 2019-06-06
LegoBoost: adding support for distance-reporting from vision-sensor

apfeuti:
### Resolves

I couldn't find any feature-request to support "distance" in the LegoBoost-extension.

### Proposed Changes

The LegoBoost-vision-sensor reports the distance (how far is the sensor away from any obstacle).
This PR adds support to the LegoBoost-Extension to read the distance-sensor-value from the MoveHub.

According to https://github.com/JorgePe/BOOSTreveng/blob/master/ColorDistanceSensor.md the color-sensor can be put into a "color and distance (0x08)" mode, where not only the color is reported but also the distance.

This PR adds two new blocks for the LegoBoost-Extension:
1) whenDistance - BlockType.HAT with a menu for the "<, >, =" operators
2) getDistance -    BlockType.REPORTER - reports the distance as a float.
                              the resolution is higher for closer distances

### Reason for Changes

Having a possibility to get the distance-sensor-value from the MoveHub, greatly improves the possibilities what can be done with LegoBoost and Scratch.
Eg:
- Stop a robot (or drive into another direction) before he hits a wall or some other obstacle -> we can write some Scratch-programs for autonomous navigation.

- Count objects passing (slowly) the distance-sensor

### Test Coverage

There's no test-automation for LegoBoost-Extension. I tested manually in this environments:
Ubuntu 18.04.2 LTS
	Browsers: Chrome, Firefox
	ScratchLink running on Windows

Windows 10 Pro
	Browsers: Chrome, Edge

macOS Mojave 10.14.5
	Browsers: Safari, Chrome

See the attached SB3-file for the test-program, which does the following:
- Play a sound (in a forever-loop) when distance-reporter detects a distance < 4
- Play an alarm-sound whenDistance (HAT-block) equals 5
- Show a message when the vision-sensor detects the color "green" -> regression-test for the existing color-feature.
[LegoBoost-Distance-Test.zip](https://github.com/LLK/scratch-vm/files/3263729/LegoBoost-Distance-Test.zip)

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -132,6 +132,16 @@ const BoostColorIndex = {
 };
 
 /**
+ * Enum for comparison-operator
+ * @type {{LESS: string, GREATER: string, EQUAL: string}}
+ */
+const BoostOperator = {
+    LESS: '<',
+    GREATER: '>',
+    EQUAL: '='
+};
+
+/**
  * Enum for Message Types
  * @readonly
  * @enum {number}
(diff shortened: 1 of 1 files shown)


thisandagain:
Please file an issue for discussion prior to submitting a PR.

Issue #2199, opened 2019-06-06
Trig functions lose precision at large angles


adroitwhiz:
### Expected Behavior

Since trigonometric functions are periodic, they should output the same values for angles modulo 360.

### Actual Behavior

At very large angles, trigonometric functions (as implemented in the underlying JavaScript VM) accumulate rounding error.

Here's a [test case](https://scratch.mit.edu/projects/315387057/). Mash the spacebar enough and the problem will manifest. Note that these angles are still less than `Number.MAX_SAFE_INTEGER`. I've confirmed this on Linux Chrome + Firefox and Windows Firefox.

ecc521:
Even though the numbers are less than Number.MAX_SAFE_INTEGER, we are still encountering issues with floating point integers.

For example, try running Math.sin(360/180*Math.PI) -  You will not get zero, but rather a number very close to zero (-2.4492935982947064e-16)

As you double the angle, you can see that the difference from zero is doubling. This happens because when converting from degrees to radians, we multiply by PI, an irrational number that can not be exactly represented. As the angle increases, the exponent bit of the floating point integer must be increased by 1, causing decimal precision to be halved, which eventually leads to large inaccuracies.

This issue can be remedied by taking the degrees modulo 360 before converting to radians. For example:

Old Method: 
```javascript
Math.sin(1509949440/180*Math.PI) //Off by about 1e-9
```

New Method: 
```javascript
Math.sin(1509949440%360/180*Math.PI) //Returns 0, the correct answer
```

Chessing234:
Opened #5473 for this. Since sin/cos are periodic, reducing the angle `% 360` before the radian conversion keeps the argument small and avoids the accumulated float error at large angles — `MathUtil.tan` already does the same thing, which is why `tan` wasn't affected.

ecc521:
@Chessing234 Pretty sure scratch-vm is an old repo that is deprecated now (switch to monorepo). 
I opened this already on the new repo (https://github.com/scratchfoundation/scratch-editor/pull/512

PR #2204, opened 2019-06-10
Fix #2198 and #2199

### LINKED ISSUE #2198: Trig functions stringify then unstringify their inputs
### Expected Behavior

The words "string" and "trigonometric function" should not be anywhere near each other

### Actual Behavior

The current implementations of the `sin`, `cos`, and `tan` functions (but not the `asin`, `acos`, and `atan` functions) stringify inputs to 10 _decimal_ digits then un-stringify them.

This was introduced in #422 so that values which should be 0 but aren't due to rounding error (e.g. sin(180)) are rounded to 0. There's likely a much faster way to do this, though (see [this benchmark](https://jsperf.com/decimal-rounding-tofixed-vs-math-round)).

ecc521:
Fix #2198 and #2199

Take angle modulus 360 before multiplying my Math.PI to avoid issues with decimal accuracy when the angle measure gets large. (#2199) 
Round to 10 digits using Math.round and division instead of stringifying (#2198). This should result in some performance improvements.

### DIFF
--- a/src/blocks/scratch3_operators.js
+++ b/src/blocks/scratch3_operators.js
@@ -136,8 +136,8 @@ class Scratch3OperatorsBlocks {
         case 'floor': return Math.floor(n);
         case 'ceiling': return Math.ceil(n);
         case 'sqrt': return Math.sqrt(n);
-        case 'sin': return parseFloat(Math.sin((Math.PI * n) / 180).toFixed(10));
-        case 'cos': return parseFloat(Math.cos((Math.PI * n) / 180).toFixed(10));
+        case 'sin': return Math.round(Math.sin(n % 360 / 180 * Math.PI) * 1e10) / 1e10;
+        case 'cos': return Math.round(Math.cos(n % 360 / 180 * Math.PI) * 1e10) / 1e10;
         case 'tan': return MathUtil.tan(n);
         case 'asin': return (Math.asin(n) * 180) / Math.PI;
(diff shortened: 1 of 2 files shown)


adroitwhiz:
Thanks for fixing this! MathUtil.tan still stringifies + unstringifies, however--might want to get that fixed in this PR as well.

ecc521:
Done.

AmazingMech2418:
This seems good, I think. Only one thing that might be good is adding the modulus to the tangent as well, since that seems to have the same sorts of issues. Otherwise, I think this is ready to merge! ~~Although it's apparently been ready to merge for almost 2 years now...~~

ecc521:
@AmazingMech2418: Unless I'm missing something, tangent already does that (see MathUtil.tan - expand the second file being changed).

AmazingMech2418:
> @AmazingMech2418: Unless I'm missing something, tangent already does that (see MathUtil.tan - expand the second file being changed).

Yeah, I just saw that. You are correct.

ecc521:
@cwillisf 
Any reason this hasn't been merged yet?
This should be a very easy PR to review and merge, as it is 3 lines of code changes, and doesn't change any behaviors (beyond improving performance trivially and reducing calculation errors with large numbers)

redstone-dev:
i love how this hasn't been merged for 5 years 😭

ecc521:
The funniest part is that I’m pretty sure there are still no merge
conflicts. Still, this really should’ve been merged within like a month at
most.

PR #2206, opened 2019-06-12
Allow selecting the last two piano keys

aeons:
### Proposed Changes

Currently, the last two piano keys in the `note_field` type are not selectable. This PR changes the max note for the field validation, so that you can select the last two piano keys.

### Reason for Changes

Either the last two blocks should be selectable or the last octave should be removed (IMO :))

See https://github.com/LLK/scratch-blocks/pull/1950 as well.


ericrosenbaum:
Thanks- see note on https://github.com/LLK/scratch-blocks/pull/1950

PR #2207, opened 2019-06-12
Removing comment in WeDo2 extension

evhan55:
Removing a comment referencing possibly renaming `startBraking` to `coastAfter`.  After comparing those two functions across the EV3 and WeDo2 extensions, it was decided that they were distinct enough, so the comment is no longer needed.

The EV3 and WeDo2 command structures that are sent over bluetooth are different in their makeup, so complicated motor commands are sent in slightly different ways for the two peripheral types.


ericrosenbaum (approved):
LGTM!

PR #2227, opened 2019-07-09
Add API for attaching a block updater callback function to the VM

kchadha:
### Resolves

Towards #2120

### Proposed Changes

Add api for attaching a callback to update a block on the workspace. This PR works with LLK/scratch-gui#4973.

### Reason for Changes

Needed to allow extensions to be able to dynamically update dynamic extension blocks.
This will be used for variable renaming, editing a custom proc, changing the shape of the control stop block, and more.

### Test Coverage

Added a unit test for the callback attaching functionality.

### Related PRs
- Related to LLK/scratch-gui#4973. This PR should be merged first.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -654,12 +654,18 @@ class Runtime extends EventEmitter {
         return 'RUNTIME_DISPOSED';
     }
 
+    // TODO remove this (it will be replaced by BLOCK_UPDATE below)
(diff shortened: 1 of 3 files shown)


cwillisf (commented):
This code looks reasonable but I wonder: what do you think about making this an event emitted by the Runtime, or emitted by the VM and echoed by the Runtime?

cwillisf (approved):
👍

Issue #2215, opened 2019-06-24
LEGO Boost firmware update 2.0.00.0017 breaks extension functionality


apple502j:
https://twitter.com/ooba/status/1142967014761254912
According to this tweet, a recent firmware update changed Boost port IDs:
A 0, B 1, C 2, D 3
Therefore old IDs (defined at https://github.com/LLK/scratch-vm/blob/develop/src/extensions/scratch3_boost/index.js#L96 ) won't work. However, if the firmware is not updated, old IDs must be used - so it's hard to define it as constant.

evhan55:
@apple502j Thank you for alerting us to this tweet! Do you have a LEGO Boost with the latest firmware?

apple502j:
Nope sorry @evhan55

evhan55:
@apple502j Okay, thank you for letting us know!

evhan55:
@ericrosenbaum 

I have tested the latest LEGO Boost Firmware `2.0.00.0017` and it seems that the motor ports did indeed change from:

```
    A: 55,
    B: 56,
    C: 1,
    D: 2
```

to:

```
    A: 0,
    B: 1,
    C: 2,
    D: 3
```

It looks like the Boost connect behavior has changed as well and is not incompatible with the Scratch 3 / Scratch Link connection flow as currently designed for Boost, and other miscellaneous sensor values seem to give errors in the console.  It might be best to open a new issue to track all of these changes together (or as separate issues?)...

evhan55:
@apple502j Thank you!  It seems you are correct about this being a critical bug in the Scratch 3 LEGO Boost extension!  Thanks for bringing this issue to our attention.

evhan55:
Some issues I noticed so far in preliminary testing:

1. Connection flow seems broken and behaves like this, in a reproducible way, now:
- Click connect within the extension to connect to Boost
- Press button on Boost
- Scratch says 'connected' but Boost connection light keeps blinking, blocks don't work
- Scratch alert comes up saying 'Boost has disconnected'
- Click connect within the extension to connect to Boost
- Press button on Boost
- Scratch says 'connected' and this time it connects properly
- Blocks now work

2. Motor blocks are not working since internal port IDs got remapped:
- `motor A` option affects nothing
- `motor B` option affects nothing
- `motor C` option affects port B
- `motor D` option affects port C

3. Setting LED light with `set color` block works, but it gives a warning in the console about a sensor value not being correct

evhan55:
[...]

PR #2229, opened 2019-07-11
Fix 2215 by properly remapping ports to fit new firmware changes

knandersen:
### Resolves

Resolves the part of #2215 that relates to block behavior, NOT connection issues.

### Proposed Changes

This PR changes port mapping as such:

Port `A`: `55` is now `0`
Port `B`: `56` is now `1`
Port `C`: `1` is now `2`
Port `D`: `2` is now `3`

Additionally it adds a `case` in `onMessage()` that catches responses from the LED to avoid warnings from appearing in the browser console. 

### Reason for Changes

LEGO has released a new firmware version 2.0.00.0017 for the LEGO BOOST Move Hub. The release included changes to the IDs that physical ports map to, and adds responses to LED commands.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -1012,6 +1012,7 @@ class Boost {
                 break;
             case BoostIO.CURRENT:
             case BoostIO.VOLTAGE:
+            case BoostIO.LED:
                 break;
             default:
                 log.warn(`Unknown sensor value! Type: ${type}`);
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -166,6 +166,45 @@ const BoostMessage = {
 };
 
 /**
+ * Enum for Hub Property Types
+ * @readonly
+ * @enum {number}
+ */
+
+const BoostHubProperty = {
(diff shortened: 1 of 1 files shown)


knandersen:
Blocked by https://github.com/LLK/scratch-link/pull/120

apple502j:
This breaks old-firmware Boosts...

thisandagain:
/cc @ericrosenbaum @cwillisf @BryceLTaylor 

We should discuss how to handle communication and support for Boost hubs that are running the older firmware as @apple502j points out above.

thisandagain:
🎉 Thanks @knandersen!

ericrosenbaum:
Looks like ports C and D are now set incorrectly for the old firmware

ericrosenbaum:
Motor position reporter for port A is not working- I think it's because the `portID` of zero for port A is false-y at this line (gotta check for not null I guess!): https://github.com/LLK/scratch-vm/blob/f6f97ef3a7b562374ab9d52108c717827fe19cde/src/extensions/scratch3_boost/index.js#L1855

ericrosenbaum (approved):
LGTM! Tested on mac chrome with Scratch Link 1.1.46, with both old and new BOOST firmware versions.

PR #2231, opened 2019-07-11
Implements BLE manufacturer data-feature in LEGO BOOST extension

### LINKED ISSUE #2230: LEGO BOOST Move Hub firmware 2.0.00.0017 breaks connection flow
### Expected Behavior

LEGO BOOST Move Hubs should appear with advertisement data and be connectable.

### Actual Behavior

As mentioned in #2215, with the recently released firmware version 2.0.00.0017, BOOST Move Hubs appear in a seemingly improperly initialised state for a few seconds before properly advertising. If Scratch Link on the host device (e.g. MacBook, PC) detects it before it's properly initialised it will attempt to connect to it and fail since the hub does not have a valid MAC-address.

### Steps to Reproduce

1. Install LEGO BOOST Move Hub firmware version 2.0.00.0017
2. Attempt to connect to the hub through Scratch 3.0 using latest Scratch Link (version 1.1.46 at time of writing)
3. Scratch 3.0 will most likely remain in a state of 'Connecting...' forever

It is possible, if the host device has a slow BLE stack that it won't see the Move Hub before it's properly initialised. In that case it will connect.

knandersen:
### Resolves

#2230 

### Proposed Changes

Adds a filter to the `scan()`-function so that only properly initialised LEGO BOOST Move Hubs are discovered and connected to.

### Reason for Changes

In Move Hub firmware 2.0.00.0017 the Move Hub briefly appears with a non-connectable MAC-address. Additionally, while non-connectable, no advertisement data is found. By implementing this filter, the non-connectable device will not pass through it.

### DIFF
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -802,9 +802,9 @@ class Boost {
             filters: [{
                 services: [BoostBLE.service],
                 manufacturerData: {
-                    0: {
-                        dataPrefix: [0x97, 0x03, 0x00, 0x40],
-                        mask: [0xFF, 0xFF, 0, 0xFF]
+                    0x0397: {
+                        dataPrefix: [0x00, 0x40],
+                        mask: [0x00, 0xFF]
                     }
                 }
             }],
--- a/src/extensions/scratch3_boost/index.js
+++ b/src/extensions/scratch3_boost/index.js
@@ -800,13 +800,13 @@ class Boost {
         }
         this._ble = new BLE(this._runtime, this._extensionId, {
             filters: [{
-                services: [BoostBLE.service]/* ,
+                services: [BoostBLE.service],
                 manufacturerData: {
                     0: {
                         dataPrefix: [0x97, 0x03, 0x00, 0x40],
                         mask: [0xFF, 0xFF, 0, 0xFF]
                     }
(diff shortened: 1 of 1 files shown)


knandersen:
Reason for blocked label: We discovered during a test that while the Mac-version of Scratch Link doesn't seem to mind being passed a filter it doesn't know, the Windows-version will not discover anything.

knandersen:
With the release of Scratch Link 1.2, this PR technically isn't blocked anymore. 

PSA: If this PR gets merged and a user is using Scratch Link 1.1 on Windows and use the LEGO BOOST extension, they will get an error when trying to connect.

thisandagain:
@cwillisf @ericrosenbaum Would you mind giving this a re-review?

cwillisf (approved):
👍 looks good!
I've tested connecting to my BOOST hub with these values and it seems to work reliably for me. Most of my tests were with Windows but I've done a few on Mac as well.

knandersen:
Thanks @cwillisf ! After discussing with @ericrosenbaum, we'll hold off on merging until we've done some more testing.

ericrosenbaum (approved):
LGMT! we tested on mac and win, with old and new boost firmware versions, with old and new scratch link versions, and each case behaved as expected.

PR #2239, opened 2019-07-20
replaceItemOfList() by assigning value to array index instead of Array.splice

### LINKED ISSUE #1934: Replace-Item-of-List Block is very slow on large lists (Scratch 3.0 regression)
### Expected Behavior

Replacing an item in a list should always perform at the same speed, no matter the size of the list (expected runtime complexity O(1)). This was the case in Scratch 2.0

### Actual Behavior

The larger the list, the longer it takes to execute a Replace-Item-of-List block on it. If I read the sourcecode correctly, JS splice() is used. This might be the reason for O(N) runtime complexity. Why not just replace the item via an indexer? Actually that is why many Scratchers might apply Replace - they expect it to be much faster than deleting / inserting. See also: https://stackoverflow.com/a/32543865 and https://jsperf.com/array-index-vs-splice

### Steps to Reproduce

Create a list, fill it with 10 items, and replace some values. Then do the same with 10000 pre-filled items. Invoking Replace Item blocks takes much longer when there are more items in the list.

When I profiled my Scratch Chess engine (https://scratch.mit.edu/projects/148769358/), I noticed that Scratch 3.0 was significantly slower because of replacing values in the transposition table. I had to shrink the transposition table by a huge factor. Also my Game of Life project (https://scratch.mit.edu/projects/96326891/) seems to spend 95% of the time in Replace Item blocks. A list size of 1000 items is already sufficient to really hurt performance, if invoked often enough.

Windows 10, Chrome 71.0.3578.98

adroitwhiz:
### Resolves

Resolves #1934

### Proposed Changes

This PR changes the `replaceItemOfList` method to change a list item's value by `list[index] = value` instead of `list.splice(index, 1, value)`.

### Reason for Changes

`Array.splice()` is much slower than setting an array element by index.

### DIFF
--- a/src/blocks/scratch3_data.js
+++ b/src/blocks/scratch3_data.js
@@ -180,7 +180,7 @@ class Scratch3DataBlocks {
         if (index === Cast.LIST_INVALID) {
             return;
         }
-        list.value.splice(index - 1, 1, item);
+        list.value[index - 1] = item;
         list._monitorUpToDate = false;
     }
 


fsih (approved):
Thanks!

PR #2242, opened 2019-07-23
Variable Renaming Context Menu Item

kchadha:
### Resolves

Towards #2120 (specifically towards extension-ifying the Variables category).

### Proposed Changes

Add a custom context menu item to the variable and list reporter blocks for renaming a variable/list.
More comments inline.

### Related PRs
This PR is related to LLK/scratch-gui#5021.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -531,7 +531,10 @@ In order to specify custom context menu options, you can provide a list of conte
 which contain the text label for the menu item as well as the name of the function in the extension that should be run when
 the block is selected.
 
-The context menu callback function will have access to an object with the `blockInfo` for the block instance whose context menu was triggered.
+The context menu callback function will have access to an object with at least two properties:
+
+* `blockInfo` - The extension block descriptor for the block instance whose context menu was triggered.
+* `blockId` - An identifier for the specific block instance whose context menu was triggered.
 
 ```js
 class SomeBlocks {
@@ -560,11 +563,11 @@ class SomeBlocks {
         };
     }
 
-    myContextMenuFunction ({blockInfo}) {
+    myContextMenuFunction ({blockInfo, blockId}) {
         // ...
     }
 
-    anotherContextMenuFunction ({blockInfo}) {
+    anotherContextMenuFunction ({blockInfo, blockId}) {
         // ...
     }
     // ...
@@ -639,10 +642,10 @@ class SomeBlocks {
         };
     }
 
-    changeDropdown ({blockInfo}) {
+    changeDropdown ({blockInfo, blockId}) {
         blockInfo.arguments.MENU.selectedValue = 'value 2';
         // Tell the runtime to update the block with the new info.
-        this.runtime.updateBlock(blockInfo.id, blockInfo);
+        this.runtime.updateBlock(blockId, blockInfo);
     }
     // ...
 }
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -599,3 +599,51 @@ customContextMenu: [
     }
 ]
 ```
+
+#### Specifying/Changing the selected value of a dropdown menu
+Specifying a dynamic block allows dynamically properties that
+define what the block looks like (block text, how many arguments it has, etc.).
+One new block property that can be set or changed in a dynamic block is
+the currently `selectedValue` of a menu. E.g. you may have a custom context menu item that changes the selected value of a drop down menu on a block:
+
+```js
+class SomeBlocks {
+    // ...
+    getInfo () {
+        return {
+            // ...
+            menus: {
+                myMenu: ['value 1', 'value 2'];
+            }
+            blocks: [
+                {
+                    isDynamic: true
+                    opcode: 'dynamicReporter',
+                    blockType: BlockType.REPORTER,
+                    text: 'my dynamic reporter block [MENU]',
+                    arguments: {
+                        MENU: {
+                            type: ArgumentType.STRING,
+                            menu: 'myMenu',
+                            defaultValue: 'value 1'
+                        }
+                    }
+                    customContextMenu: [
+                        {
+                            text: 'Change Dropdown Menu',
+                            callback: 'changeDropdown'
+                        }
+                    ]
+                }
+            ]
+        };
+    }
+
+    changeDropdown ({blockInfo}) {
+        blockInfo.arguments.MENU.selectedValue = 'value 2';
+        // Tell the runtime to update the block with the new info.
+        this.runtime.updateBlock(blockInfo.id, blockInfo);
+    }
+    // ...
+}
+```
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -1,6 +1,7 @@
 const ArgumentType = require('../extension-support/argument-type');
 const BlockType = require('../extension-support/block-type');
 const Cast = require('../util/cast');
+const log = require('../util/log');
 const Variable = require('../engine/variable');
 const formatMessage = require('format-message');
 
@@ -378,11 +379,23 @@ class Scratch3DataBlocks {
             opcode: 'variable',
             func: 'getVariableValue',
             blockType: BlockType.REPORTER,
-            text: v.name
+            text: v.name,
+            customContextMenu: [
+                {
+                    text: formatMessage({
+                        id: 'data.renameVariableContextMenuItem',
+                        default: 'Rename Variable',
+                        description: 'Context menu item on a variable reporter block for renaming the variable'
+                    }),
+                    builtInCallback: 'RENAME_A_VARIABLE',
+                    callback: 'renameCallback'
+                }
+            ]
         }));
         blocks.push(
             '---',
             {
+                isDynamic: true,
                 opcode: 'setvariableto',
                 func: 'setVariable',
                 blockType: BlockType.COMMAND,
@@ -400,6 +413,7 @@ class Scratch3DataBlocks {
                 }
             },
             {
+                isDynamic: true,
                 opcode: 'changevariableby',
                 func: 'changeVariableBy',
                 blockType: BlockType.COMMAND,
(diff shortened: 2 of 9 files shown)


docs/extensions.md
```diff
@@ -599,3 +599,51 @@ customContextMenu: [
...
+                    ]
+                }
+            ]
+        };
+    }
+
+    changeDropdown ({blockInfo}) {
+        blockInfo.arguments.MENU.selectedValue = 'value 2';
+        // Tell the runtime to update the block with the new info.
+        this.runtime.updateBlock(blockInfo.id, blockInfo);
```
kchadha: The corresponding GUI PR makes it so that the block ID is included in the blockInfo in the block instance's mutation. While I didn't end up using this functionality directly in the variables code (since I had access to the VM blocks), I think this feature is more generally useful for extension developers (you can't actually update a block instance's blockInfo without knowing it's ID, and if there are many blocks on the workspace with the same opcode, it would not be possible to differentiate the different block instances without this information).

src/blocks/data.js
```diff
@@ -377,16 +368,35 @@ class Scratch3DataBlocks {
...
+            text: v.name,
+            customContextMenu: [
+                {
+                    text: formatMessage({
+                        id: 'data.renameVariableContextMenuItem',
+                        default: 'Rename Variable',
+                        description: 'Context menu item on a variable reporter block for renaming the variable'
+                    }),
+                    builtInCallback: 'RENAME_A_VARIABLE',
+                    callback: 'renameCallback'
```
kchadha, line 391: These callback are the same for both variable and list blocks.

src/blocks/data.js
```diff
@@ -680,6 +699,67 @@ class Scratch3DataBlocks {
             }
         );
     }
+
+    renameCallback ({blockInfo, newName, varType}) {
```
kchadha, line 693: I wasn't sure about whether to pass in varType here or determine it based on the opcode. See related code in the GUI pr which mentions having access to the VM's `Variable` class.

src/blocks/data.js
```diff
@@ -680,6 +699,67 @@ class Scratch3DataBlocks {
...
+        }
+
+        const oldName = blockInfo.text;
+        const variable = editingTarget.lookupVariableByNameAndType(oldName, varType);
+
+        const fieldName = varType === Variable.SCALAR_TYPE ? 'VARIABLE' : 'LIST';
+
+        // Accumulate blocks that pertain to this extension
+        let allRelevantBlocks;
+        const visibleBlocks = editingTarget.blocks.getAllReferencesForVariable(variable);
```
kchadha, line 719: This is a call to a new function on the blocks container, which will return a list of all the variable/list blocks in that block container that reference a variable of the same name and type. There are other functions that are doing similar things for the regular core variables category, and I expect those to be removed post-extensionification of the variables category.

src/engine/blocks.js
```diff
@@ -379,6 +379,9 @@ class Blocks {
                 if (!editingTarget.lookupVariableById(e.varId)) {
                     editingTarget.createVariable(e.varId, e.varName, e.varType);
                     this.emitProjectChanged();
+                    // TODO this should probably be batched
+                    // (esp. if we receive multiple new var_creates in a row).
+                    this.runtime.requestToolboxExtensionsUpdate();
```
kchadha, line 384: If we end up changing the gui handling of the variable create button so that it no longer goes through scratch-blocks to display the variable creation modal (and triggering the actual variable creation) this code should be removed from here and moved into the new extension function that will be responsible for triggering variable creation.

src/engine/blocks.js
```diff
@@ -578,8 +584,10 @@ class Blocks {
 
             // Update block value
             if (!block.fields[args.name]) return;
-            if (args.name === 'VARIABLE' || args.name === 'LIST' ||
-                args.name === 'BROADCAST_OPTION') {
+            if ((args.name === 'VARIABLE' || args.name === 'LIST' ||
+                // Can get rid of the checks above and the && clause
+                // after extensionification of variables is complete
+                args.name === 'BROADCAST_OPTION') && block.fields[args.name].id) {
```
kchadha, line 590: The original condition here is for handling the old variable/list blocks. I added an extra check to make sure that the old blocks still work, and that this doesn't break the new extension-ified variable blocks. I expect this whole check to be removed when we are done extensionifying the Variables category.

src/engine/blocks.js
```diff
@@ -589,6 +597,9 @@ class Blocks {
             } else {
                 // Changing the value in a dropdown
                 block.fields[args.name].value = args.value;
+                if (block.mutation && block.mutation.blockInfo) {
+                    block.mutation.blockInfo.arguments[args.name].selectedValue = args.value;
```
kchadha, line 601: The fundamental issue that this code is solving is keeping the VM in sync with field and mutation info changes.

I update the actual selected value in blockInfo mutation here so that the block still has the correct variable name selected after the following steps: 

1) Create new variable foo
2) Drag out `set [my variable] to (0)` block
3) Change the dropdown menu selection in that block to `foo`
4) Switch to the costumes tab and switch back

Without this change, switching back to the code tab would change the selected variable option in that block back to `my variable`. See discussion below for more info about why this is.

src/engine/blocks.js
```diff
@@ -1080,25 +1128,33 @@ class Blocks {
...
-            }
-            let value = blockField.value;
-            if (typeof value === 'string') {
-                value = xmlEscape(blockField.value);
+
+        // Add fields to the block, but only if the block itself
+        // doesn't have a mutation with a blockinfo.
+        // This is a dynamic extension block which will be laid out using
+        // it's blockInfo mutation
+        if (!(block.mutation && block.mutation.blockInfo)) {
```
kchadha, line 1136: I am putting the code for generating the XML for a field inside of this check that the block is not dynamic so that we don't emit XML that has both a dynamic extension block with a mutation determining how to lay it out, and also conflicting field information.

There's a trade off here, because this is working towards making the field value more obsolete (if we accomplish our goal of switching all blocks to be dynamic extension blocks, I think the `fields` value of the VM JSON representation of blocks will be completely obsolete).

However, if we think that is too risky, the alternative is to keep this code dependent on scratch-blocks, and react to the scratch-blocks `var_rename` event emitted by the variables modal. I have an alternative set of changes stashed away that handle keeping the block field values in sync in the VM (as we do currently with the regular variable blocks right now) by adding code that updates the blockInfo to the pre-existing `updateBlocksAfterVarRename` function that updates all the field values and ids for all the relevant blocks referencing the renamed variable. In an earlier comment, I mentioned that this code would be obsolete post-extensionification, taking this alternative would make that statement false.

I think the latter alternative is a little more confusing, because it makes the blocks unnecessarily dependent on both the blockInfo mutation and the "fields" part of the VM JSON representation. I think this will ultimately lead to confusion down the line (post-extensionification)....

kchadha:
@cwillisf, this PR has many different components, so I commented inline where I thought things might need a clarification as to why they're in this PR or why the code is written the way that it is.

src/engine/blocks.js
cwillisf: Does this cause a significant performance problem when loading a project which uses a lot of variables?

src/engine/blocks.js
cwillisf: Thanks for the thorough explanation here! I agree with you -- unless we run into some sort of Big Problem I think it makes sense to work toward obsoleting the `fields` value in the VM JSON.

docs/extensions.md
```diff
@@ -531,7 +531,10 @@ In order to specify custom context menu options, you can provide a list of conte
 which contain the text label for the menu item as well as the name of the function in the extension that should be run when
 the block is selected.
 
-The context menu callback function will have access to an object with the `blockInfo` for the block instance whose context menu was triggered.
+The context menu callback function will have access to an object with at least two properties:
```
cwillisf, line 534: "at least two properties" :satisfied: :shipit:

src/engine/blocks.js
```diff
@@ -1080,25 +1128,33 @@ class Blocks {
...
-                xmlString += ` variabletype="${varType}"`;
-            }
-            let value = blockField.value;
-            if (typeof value === 'string') {
-                value = xmlEscape(blockField.value);
+
+        // Add fields to the block, but only if the block itself
+        // doesn't have a mutation with a blockinfo.
+        // This is a dynamic extension block which will be laid out using
+        // it's blockInfo mutation
```
cwillisf:
```suggestion
        // its blockInfo mutation
```

src/virtual-machine.js
```diff
@@ -28,6 +28,7 @@ require('canvas-toBlob');
 const RESERVED_NAMES = ['_mouse_', '_stage_', '_edge_', '_myself_', '_random_'];
 
 const CORE_EXTENSIONS = [
+    'data'
```
cwillisf, line 31: 😎

src/blocks/data.js
cwillisf: I'd rather pass it in, especially since you already have it here. Determining it from the opcode seems less future-proof though I admit I don't have a specific case in mind.

cwillisf (approved):
Good stuff! I have one minor suggestion for a comment typo but other than that LGTM!

src/engine/blocks.js
kchadha: We talked about this one offline. I think this code path should not be called during project load, but I will confirm. I'll also look at flame charts with @picklesrus.

docs/extensions.md
kchadha: After discussion offline, we decided to pass in the blockId to the custom context menu callback directly instead of embedding it in the blockInfo. We can always change this in the future if needed.

src/engine/blocks.js
kchadha: I locally removed the old variables category (so that only the extension was loaded), and I loaded a new project containing variables. I was able to confirm that this code path is not called during project load, and the variable blocks still show up in the toolbox correctly.

PR #2243, opened 2019-07-25
E16N: move extension manager from VM to Runtime

cwillisf:
### Proposed Changes

Move the Extension Manager class from being owned by the VM to being owned by the Runtime.

### Reason for Changes

For one thing, the extension manager has more to do with the runtime than the VM, so it seems to make sense that it would be more closely associated with the runtime. I could even see the argument that the extension manager's functionality should be merged into the runtime.

The real reason, though, is that I'm in a situation where it would be really handy to make calls into the extension manager from a place which has access to the runtime but not to the VM. Specifically, in `src/engine/blocks.js`. I think it's right that `blocks.js` doesn't have access to the VM, but it needs access to information that comes from the extension manager, so... here's my proposed change :)

~Interestingly, the GUI doesn't care about the extension manager moving. All of the interaction between the GUI and the extension manager happens through events, so as long as those events get registered correctly everyone's happy. It's a bit unfortunate that the VM registers the runtime for events externally, but that's a separate problem.~

EDIT: I was wrong... there are a couple of lines in `scratch-gui/containers/extension-library.jsx` that would need to change but it'd just be a matter of replacing `vm.extensionManager` with `vm.runtime.extensionManager`.

### Test Coverage

This change is covered by existing tests, a few of which needed minor changes to access the extension manager in its new home.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -10,7 +10,6 @@ const JSZip = require('jszip');
 
 const Buffer = require('buffer').Buffer;
 const centralDispatch = require('./dispatch/central-dispatch');
-const ExtensionManager = require('./extension-support/extension-manager');
 const log = require('./util/log');
 const MathUtil = require('./util/math-util');
 const Runtime = require('./engine/runtime');
@@ -129,7 +128,7 @@ class VirtualMachine extends EventEmitter {
             this.emit(Runtime.BLOCK_UPDATE, blockId, blockInfo);
         });
         this.runtime.on(Runtime.TOOLBOX_EXTENSIONS_NEED_UPDATE, () => {
-            this.extensionManager.refreshBlocks();
+            this.runtime.extensionManager.refreshBlocks();
         });
         this.runtime.on(Runtime.PERIPHERAL_LIST_UPDATE, info => {
             this.emit(Runtime.PERIPHERAL_LIST_UPDATE, info);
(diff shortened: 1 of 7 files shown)


rschamp:
Would love to chat about this — I think a walk through would be the easiest way, so let's do that sometime soon.

src/virtual-machine.js
```diff
@@ -169,6 +169,14 @@ class VirtualMachine extends EventEmitter {
         this.variableListener = this.variableListener.bind(this);
     }
 
+    /**
+     * @returns {ExtensionManager} the extension manager, now owned by the runtime.
+     * @deprecated Please access the extension manager through the runtime instead.
```
rschamp, line 174: Could also log this warning if it's accessed this way so we can find places we've missed in our own code.  Although I wonder if we care/want that? Seems like accessing via VM from the outside could be seen as the _right_ thing to do — the VM is the external interface. But internally we refer to the actual location?

rschamp (approved):
Thanks for making it a non-breaking change!

src/virtual-machine.js
```diff
@@ -129,7 +128,7 @@ class VirtualMachine extends EventEmitter {
             this.emit(Runtime.BLOCK_UPDATE, blockId, blockInfo);
         });
         this.runtime.on(Runtime.TOOLBOX_EXTENSIONS_NEED_UPDATE, () => {
-            this.extensionManager.refreshBlocks();
+            this.runtime.extensionManager.refreshBlocks();
```
kchadha, line 131: I think there's no longer to bubble this event all the way up to the VM anymore. We could either remove the event all together or just have the runtime listen and handle its own event since it also owns the extensionManager.

PR #2245, opened 2019-07-30
E16n serialization

cwillisf:
### Proposed Changes

This change allows an extension to specify custom serialization and/or deserialization functions for specific blocks. As a supporting change, the extension manager has moved from the VM to the Runtime.

### Reason for Changes

The current SB3 format handles certain blocks, including variable references, as special cases. This change allows such special cases to be implemented as part of an extension.

### Test Coverage

*tests needed*

### DIFF
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -16,7 +16,6 @@ const uid = require('../util/uid');
 const MathUtil = require('../util/math-util');
 const StringUtil = require('../util/string-util');
 const VariableUtil = require('../util/variable-util');
-const dispatch = require('../dispatch/central-dispatch');
 
 const {loadCostume} = require('../import/load-costume.js');
 const {loadSound} = require('../import/load-sound.js');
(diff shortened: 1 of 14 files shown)


src/serialization/sb3.js
```diff
@@ -102,9 +103,23 @@ const primitiveOpcodeInfoMap = {
...
-    if (hasOwnProperty.call(primitiveOpcodeInfoMap, block.opcode)) {
+    const extensionInfo = block.extensionInfo;
+    if (extensionInfo) {
+        try {
+            return dispatch.callSync(extensionInfo.serviceName, 'serializeBlock', block);
+        } catch (e) {
+            // if the extension doesn't implement `serializeBlock` that's OK
+            // if anything else went wrong, rethrow the exception
+            // TODO: implement a better way to recognize the extension service not implementing `serializeBlock`
+            if (e.message.indexOf('Method serializeBlock not found on service') !== 0) {
```
adroitwhiz: should this be -1 and not 0?

cwillisf:
Note to self: try to give the runtime only the loaded extensions set instead of the whole extension manager

kchadha:
Unassigning myself from this PR (see discussion in #2279).

PR #2247, opened 2019-07-31
Add enter key to list of keyboard events detected.

spl237:
A centre push on the joystick on the Raspberry Pi SenseHAT generates the equivalent of pressing the "enter" key on a keyboard. This PR adds detection for that event to the keyboard handling for Scratch.

### DIFF
--- a/src/io/keyboard.js
+++ b/src/io/keyboard.js
@@ -9,7 +9,8 @@ const KEY_NAME = {
     LEFT: 'left arrow',
     UP: 'up arrow',
     RIGHT: 'right arrow',
-    DOWN: 'down arrow'
+    DOWN: 'down arrow',
+    ENTER: 'enter'
 };
 
 /**
@@ -56,6 +57,7 @@ class Keyboard {
         case 'ArrowRight': return KEY_NAME.RIGHT;
         case 'Down':
         case 'ArrowDown': return KEY_NAME.DOWN;
+        case 'Enter': return KEY_NAME.ENTER;
         }
         // Ignore modifier keys
         if (keyString.length > 1) {


thisandagain:
@carljbowman @rschamp @ericrosenbaum @kchadha @BryceLTaylor Any thoughts on this? I'd like to make sure we discuss the potential side effects of supporting "enter" as a keyboard event before we move forward with this change.

ericrosenbaum:
I'm noticing two side effects of this change:

- The "any" key option in both the key pressed hat and boolean blocks will trigger for the enter key.
- The key pressed boolean can be made to detect the enter key, by dropping the string "enter" onto its input. 

As far as I can tell, the keyboard event for the enter key is pretty cross platform compatible (except for old versions of IE, which we don't support). 

Some keyboards do have the word "return", or both "return" and "enter", or just an icon on the key. I don't think we need to worry about that.

But given that especially the "any" key change is a change in core Scratch behavior (albeit small), we should consider if it introduces any new problems.

We have discussed non-alphanumeric key detection a few times in the past (e.g. https://github.com/LLK/scratch-vm/issues/980). At that point @thisandagain raised the question of how these keys interact with browser accessibility features and screen readers. Does anybody know if the enter key event is commonly used by these systems?

thisandagain:
Thanks @ericrosenbaum. Regarding a11y: I'm much less worried about the `enter` key than many others. Here is a helpful overview:
https://www.xfinity.com/support/articles/screen-reader-hot-keys

rschamp:
I'm not sure how common it is to use tabbing + enter to activate different parts of the GUI, but it's possible that "enter" could activate the hat when the user is only attempting to use the UI e.g. change tabs or activate a button.

I don't think this is a reason not to listen for the enter key, just something that might require some cleanup on the GUI side (using `preventDefault()` (or is it `stopPropagation()`?) on anything that responds to the enter keypress).

picklesrus:
I think the blocks use this for the kind of thing @rschamp  mentioned a lot. e.g. when editing a number/text field, hitting enter closes the mini editor and unfocuses the field. Also, you can select items from a menu, etc this way. Probably worth playing with this in editor mode with a project that has a when enter key pressed hat.

thisandagain:
Agreed! Block interaction was my biggest worry as well. Perhaps we should stage this in a branch and bring it to one of our upcoming smoke tests?

ericrosenbaum (approved):
Looks good to me!

ericrosenbaum:
I approved this but lets test a bit before we merge it.

Here's a branch where you can try it:
https://ericrosenbaum.github.io/scratch-gui/enter-key/

Note that there are no visible changes. Here's a way to see the two effects I mentioned above ("when any key pressed" hat triggers for enter key, and dropping "enter" onto the boolean will work):
Screen Shot 2019-08-08 at 4 27 39 PM

ericrosenbaum:
Unblocked following a design discussion, so I'm going to merge

PR #2250, opened 2019-08-01
Sort variable reporter blocks and variables dropdown by name.

kchadha:
### Resolves

- Towards #2120 

### Proposed Changes

Refactor variables extension code a bit (no longer using forEachVariable because we don't actually use anything other than the variable name, and there is already a function in `target` which returns a list of variable names. 

Sort variables by name in set of variable/list reporters in toolbox as well as in variable/list dropdown menus.

### Reason for Changes

Consistency with regular (non-extensionified) variable category.

### DIFF
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -1,6 +1,7 @@
 const ArgumentType = require('../extension-support/argument-type');
 const BlockType = require('../extension-support/block-type');
 const Cast = require('../util/cast');
+const StringUtil = require('../util/string-util');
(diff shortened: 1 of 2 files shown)


kchadha:
Will probably end up closing this PR in favor of a different one which does not refactor the code to remove the `forEachVariable` function because we will need it for a different set of bugs.

kchadha:
Closing this PR as #2262 addresses the main changes this PR was making.

PR #2259, opened 2019-08-13
Image as value support

asnoba:
### Proposed Changes
When a `iconURI` is present use `field_image` instead of `field_value`.

### Reason for Changes
When a value need a visual representation in form of an image.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1219,10 +1219,24 @@ class Runtime extends EventEmitter {
         // Sanitize the placeholder to ensure valid XML
         placeholder = placeholder.replace(/[<"&]/, '_');
 
-        const argJSON = {
-            type: 'input_value',
-            name: placeholder
-        };
+        const iconURI = context.blockInfo.arguments[placeholder].iconURI;
+        let argJSON;
+
+        if (iconURI) {
+            argJSON = {
+                type: 'field_image',
+                name: placeholder,
+                src: iconURI,
+                width: 28,
+                height: 28,
+                alt: context.blockInfo.arguments[placeholder].defaultValue
+            };
+        } else {
+            argJSON = {
+                type: 'input_value',
+                name: placeholder
+            };
+        }
 
         const argInfo = context.blockInfo.arguments[placeholder] || {};
         let argTypeInfo = ArgumentTypeMap[argInfo.type] || {};


thisandagain:
@kchadha Could you review this PR this week?

src/engine/runtime.js
```diff
@@ -1219,10 +1219,24 @@ class Runtime extends EventEmitter {
         // Sanitize the placeholder to ensure valid XML
         placeholder = placeholder.replace(/[<"&]/, '_');
 
-        const argJSON = {
-            type: 'input_value',
-            name: placeholder
-        };
+        const iconURI = context.blockInfo.arguments[placeholder].iconURI;
```
kchadha, line 1222: This code should probably go after the code below (https://github.com/LLK/scratch-vm/pull/2259/files#diff-38d875e1d50b3158dc96404b65275a54R1241) that declares `argInfo` because it handles the case where `context.blockInfo.arguments[placeholder]` does not exist. As is right now, this line of code will throw an error if that doesn't exist.

kchadha: I am also a little concerned about this approach to specifying an image field from an extension spec design perspective in general. Whether or not `field_image` is used in place of `input_value` should be determined by the argument type as opposed to whether a regular argument has an additional `iconURI` property provided.

src/engine/runtime.js
```diff
@@ -1219,10 +1219,24 @@ class Runtime extends EventEmitter {
...
+        let argJSON;
+
+        if (iconURI) {
+            argJSON = {
+                type: 'field_image',
+                name: placeholder,
+                src: iconURI,
+                width: 28,
+                height: 28,
+                alt: context.blockInfo.arguments[placeholder].defaultValue
```
kchadha, line 1232: Similar to my comment above about the general approach from an extension spec design perspective, I don't think defaultValue is the appropriate analogy to use for the image alt text because it has a different meaning for actual value input arguments.

kchadha:
Hi @asnoba! Thanks for the pull request. I added some comments inline, but I think in general we would like to add support for inline images to the extension specification in a more careful way. I don't think adding an `iconURI` property to a regular string or number argument type is necessarily the way we want extension developers to be able to specify an inline image in a block.

In future work we plan to support inline images in extension blocks (similar to the Scratch `turn -> 15 degrees` blocks) in a simpler way. I am imaging something where the user will be able to specify an argument of type `ArgumentType.Image` and specify the icon specific information there (e.g. `iconURI` and alt text and whether or not the image should be flipped in right to left languages, etc.).

asnoba:
Hi @kchadha! Thanks for your comments and review. I think your approach sounds like the proper way to go forward with this. Are there any plans to implement this in the near future? :) We also need to make sure that other fields can't be dropped on top of these inline images.

PR #2262, opened 2019-08-15
Variables Extension Monitors

kchadha:
### Proposed Changes

- Make variable/list monitors work with the new variables extension.
    - Variable reporter blocks in the toolbox were getting random IDs assigned to them, and these ids were getting newly generated every time we refreshed the toolbox. This makes it impossible to have the block monitor stay associated with the toolbox block. Add `paletteKey` property to dynamic extension block info to keep a block id consistent.
     - When a variable or list is created from `make a variable` button, it should have its monitor checkbox checked. This was previously handled automatically by scratch-blocks, but with the new extension code paths, the variable block does not exist at the time when we have submitted the information in the modal. The VM needs to hold on to the variable creation information when it receives a `var_create` event from scratch-blocks and keep track that there is a monitor pending for a block with the same id as the variable. This pending monitor is used to later add the check to the monitor checkbox for the newly created variable reporter block.
    - Monitor label information (e.g. label text and monitor color) for extension monitors are now compatible with dynamic extension blocks as well as core extensions. The information returned about the monitor is now its label and color instead of its label and a string representing a category from which to derive color information from. The block color of the block being monitored is used directly instead.
- Sorts the variable reporters and the variables/lists (in the block dropdowns) alphabetically.

### Related PRs
This PR is related to LLK/scratch-gui#5102. This PR should go in first.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -597,9 +597,6 @@ class Blocks {
             } else {
                 // Changing the value in a dropdown
                 block.fields[args.name].value = args.value;
-                if (block.mutation && block.mutation.blockInfo) {
-                    block.mutation.blockInfo.arguments[args.name].selectedValue = args.value;
-                }
 
                 // The selected item in the sensing of block menu needs to change based on the
                 // selected target.  Set it to the first item in the menu list.
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -378,10 +378,13 @@ class Blocks {
             if (e.isLocal && editingTarget && !editingTarget.isStage && !e.isCloud) {
                 if (!editingTarget.lookupVariableById(e.varId)) {
                     editingTarget.createVariable(e.varId, e.varName, e.varType);
-                    this.emitProjectChanged();
+
+                    this.runtime.addPendingMonitor(e.varId);
+
                     // TODO this should probably be batched
                     // (esp. if we receive multiple new var_creates in a row).
                     this.runtime.requestToolboxExtensionsUpdate();
+                    this.emitProjectChanged();
                 }
             } else {
                 if (stage.lookupVariableById(e.varId)) {
@@ -396,10 +399,14 @@ class Blocks {
                     }
                 }
                 stage.createVariable(e.varId, e.varName, e.varType, e.isCloud);
-                this.emitProjectChanged();
+
+                this.runtime.addPendingMonitor(e.varId);
+
                 // TODO same as above, this should probably be batched
                 // (esp. if we receive multiple new var_creates in a row).
                 this.runtime.requestToolboxExtensionsUpdate();
+                this.emitProjectChanged();
+
             }
             break;
         case 'var_rename':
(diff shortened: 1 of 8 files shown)


src/engine/blocks.js
```diff
@@ -396,10 +399,14 @@ class Blocks {
                     }
                 }
                 stage.createVariable(e.varId, e.varName, e.varType, e.isCloud);
-                this.emitProjectChanged();
+
+                this.runtime.addPendingMonitor(e.varId);
```
cwillisf, line 403: Is it possible to get here without a paired `removePendingMonitor` call? If so, would that cause problems with a future project load or anything like that?

EDIT: Ah, looks like you take care of it in the runtime's `dispose` method. I'll leave this here in case you have further comment but it looks OK to me.

src/engine/runtime.js
```diff
@@ -2414,30 +2424,97 @@ class Runtime extends EventEmitter {
...
+    }
+
+    /**
+     * Get the monitor label and color for a given toolbox block id.
+     * @param {string} id - the block id for a block with a monitor being requested
+     * @return {object} - object with label and color
+     * @property {string} color - the color for the monitor for the given block id
+     * @property {string} label - the label for the monitor for the given block id
+     */
+    getMonitorLabelForBlock (id) {
```
cwillisf, line 2443: 👍 on the new name

src/engine/runtime.js
```diff
@@ -2414,30 +2424,97 @@ class Runtime extends EventEmitter {
...
+                color: extensionBlockInfo.color1 || categoryInfo.color1,
+                label: extensionBlockInfo.text
+            };
+        }
 
         // TODO: we may want to format the label in a locale-specific way.
         return {
-            category: 'extension', // This assumes that all extensions have the same monitor color.
-            label: `${categoryInfo.name}: ${block.info.text}`
+            color: categoryInfo.color1, // TODO This assumes that all extensions have the same monitor color.
```
cwillisf: This TODO is now incorrect

src/engine/runtime.js
```diff
@@ -2414,30 +2424,97 @@ class Runtime extends EventEmitter {
...
     }
 
+    /**
+     * Keep track of the fact that a monitor has been requested
+     * for the given blockId, even though the block with the given id
+     * does not exist in the flyout yet.
+     * @param {string} blockId The id of the block with a pending monitor
+     */
+    addPendingMonitor (blockId) {
+        this._pendingMonitors[blockId] = true;
```
cwillisf: If you're feeling fancy you could use [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) instead

src/util/string-util.js
```diff
@@ -89,6 +89,21 @@ class StringUtil {
             }
         });
     }
+
+    // Port over the string comparison utiliy from scratch-blocks
```
cwillisf: typo: utility

cwillisf (changes requested):
Looks good overall! There are a couple minor questions and things but generally 👍

kchadha:
@cwillisf, addressed the PR review comments!

cwillisf (approved):
👍

PR #2264, opened 2019-08-22
Add md5 for default asset

fsih:
### Resolves

This was causing a crash when opening the paint editor on a sprite that had the default question mark costume


paulkaplan (approved):
LG. I wish there was a way to condense the different properties needed by costumes

PR #2269, opened 2019-08-29
Revert "Revert "Raise params to the next frame when pushing""

kchadha:
Reverts LLK/scratch-vm#2240

Corresponds to the original PR #2145 which was reverted because of a regression.


kchadha:
cc/ @mzgoddard

kchadha:
@mzgoddard assigning this one to you based on conversations offline

Issue #2273, opened 2019-09-08
Pen Colors repeat each 101 numbers rather than each 100 numbers


helderman:
### Expected Behavior

Pen colors 0, 100, 200, 300 ... should all be red.

### Actual Behavior

Instead, pen colors 0, 101, 202, 303 ... are red.
Full explanation (by Scratch-Minion) can be found in the Scratch forum:
https://scratch.mit.edu/discuss/topic/369220/

### Steps to Reproduce

Draw a line with a pen color of 5000. Notice it is not red.
Demonstrated in this project (by Scratch-Minion):
https://scratch.mit.edu/projects/327014325

### Operating System and Browser

Windows NT 10.0, Chrome 76.0.3809.132

benjiwheeler:
I checked out Scratch-Minion's project ( https://scratch.mit.edu/projects/327014325 ) and verified this. If you continuously increment pen color by 101, it's always red. If you use 100, the color cycles through the rainbow.

benjiwheeler:
The current 3.0 behavior also appears to be inconsistent with Scratch 2.0. 

Note that from 2.0 to 3.0, we changed the pen color range size from 200 to 100; still, you can see that continually incrementing by 200 used to keep the color red:

![image](https://user-images.githubusercontent.com/3431616/64712154-bcd18800-d488-11e9-8083-8be30306e157.png)

So in 2.0, the period size for pen colors to cycle back to equivalent values was 200 -- not 201.

But currently in 3.0, the period size is 101. 

Seems like we should change 3.0 to use a period size of 100.

benjiwheeler:
Looking into `MathUtil.wrapClamp`, it seems to me like we _might_ want to reconsider how it works.

Its current code is:

```
    static wrapClamp (n, min, max) {
        const range = (max - min) + 1;
        return n - (Math.floor((n - min) / range) * range);
    }
```

I think that's more understandable by rewriting it like this: (note I'm not proposing this change, just understanding it)
```
    static wrapClamp (n, min, max) {
        const range = (max - min) + 1;
        const timesOverRange = Math.floor((n - min) / range);
        const adjustedN = n - timesOverRange * range
        return adjustedN;
    }
```
I think this is built around the assumption that if you kept incrementing n by 1, you'd want `max` and `min` to be distinct stopping points.

E.g.:

[...]

PR #2274, opened 2019-09-08
Pen color range [0..99]

helderman:
Attempted bugfix for "Pen Colors repeat each 101 numbers rather than each 100 numbers" as reported in the Scratch forum:
https://scratch.mit.edu/discuss/topic/369220/

### Resolves

https://github.com/LLK/scratch-vm/issues/2273

### Proposed Changes

Apply modulo 100 on the pen color, instead of modulo 101.

### Reason for Changes

Current behavior is confusing, inconsistent with Scratch 2.0, and (most likely) unintended.

### Test Coverage

No test coverage yet. My apologies.

### DIFF
--- a/src/extensions/scratch3_pen/index.js
+++ b/src/extensions/scratch3_pen/index.js
@@ -191,13 +191,13 @@ class Scratch3PenBlocks {
     }
 
     /**
-     * Wrap a color input into the range (0,100).
+     * Wrap a color input into the range (0,99).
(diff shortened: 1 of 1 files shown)


benjiwheeler:
I looked at this briefly and while I think there are deeper questions that #2273 raises that we should discuss, I think @helderman's fix is an improvement to the current situation.

PR #2278, opened 2019-09-13
Move error log to right place

fsih:


src/import/load-costume.js
```diff
@@ -261,7 +260,8 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
     }
     if (costume.asset.assetType.runtimeFormat === AssetType.ImageVector.runtimeFormat) {
         return loadVector_(costume, runtime, rotationCenter, optVersion)
-            .catch(() => {
+            .catch(error => {
+                log.warn(`${error.name}: ${error.message}`);
```
kchadha: Minor comment: maybe we should add an error message prefix so it's clear what part of the code caught the error
```suggestion
                log.warn(`Error loading vector image: ${error.name}: ${error.message}`);
```

kchadha (commented):
LGTM, but minor suggestion included in comments.

PR #2279, opened 2019-09-16
E16n serialization (slightly more conservative version)

cwillisf:
This is the same as #2245 but does not move the Extension Manager into the Runtime. Instead, only the information about currently-loaded extensions is moved to the Runtime.

### DIFF
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -9,7 +9,6 @@ const LIST_ITEM_LIMIT = 200000;
 
 // serialization primitives
 const VAR_PRIMITIVE = 12;
-const LIST_PRIMITIVE = 13;
 
 /**
  * This Scratch 3.0 extension implements Scratch "data" blocks for variables and lists, including cloud variables.
@@ -787,7 +786,7 @@ class Scratch3DataBlocks {
             throw new Error('bad primitive ID on variable!');
         }
         const result = {};
-        result.opcode = 'data_variable';
+        result.opcode = 'variable';
         result.fields = {
             VARIABLE: {
                 name: 'VARIABLE',
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -790,17 +790,40 @@ class Scratch3DataBlocks {
     }
 
     deserializeVariable (primitiveArray) {
-        if (primitiveArray[0] !== VAR_PRIMITIVE) {
+        const [primitiveTag, variableName, variableId] = primitiveArray;
+        if (primitiveTag !== VAR_PRIMITIVE) {
             throw new Error('bad primitive ID on variable!');
         }
-        const result = {};
-        result.opcode = 'variable';
-        result.fields = {
-            VARIABLE: {
-                name: 'VARIABLE',
-                value: primitiveArray[1],
-                id: primitiveArray[2],
-                variableType: Variable.SCALAR_TYPE
+        // @todo reduce duplication here.
+        // This effectively duplicates some of the content in `getInfo()` as well as some of the logic in
+        // `defineDynamicBlock`. Maybe it should return only the mutation blockInfo, or maybe it should return
+        // something that looks like the block's entry in `getInfo()`. If we do either of those things we'll need to
+        // decide how we want to handle the block's (x,y) position and `topLevel` flag.
+        const result = {
+            opcode: 'variable',
+            fields: {
+                VARIABLE: {
+                    name: 'VARIABLE',
+                    id: variableId,
+                    variableType: Variable.SCALAR_TYPE
+                }
+            },
+            inputs: {},
+            next: null,
+            parent: null,
+            shadow: false,
+            mutation: {
+                blockInfo: {
+                    blockType: BlockType.REPORTER,
+                    isDynamic: true,
+                    opcode: 'variable',
+                    text: variableName,
+                    paletteKey: variableId,
+                    customContextMenu: [{
+                        text: 'Rename Variable',
+                        builtInCallback: 'RENAME_A_VARIABLE'
+                    }]
+                }
             }
         };
         if (primitiveArray.length > 3) {
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -2,6 +2,7 @@ const ArgumentType = require('../extension-support/argument-type');
 const BlockType = require('../extension-support/block-type');
 const Cast = require('../util/cast');
 const log = require('../util/log');
+const StringUtil = require('../util/string-util');
 const Variable = require('../engine/variable');
 const formatMessage = require('format-message');
 
@@ -56,6 +57,8 @@ class Scratch3DataBlocks {
                 menuItems.push(v.name);
             }
         });
+        menuItems.sort(StringUtil.compareStrings);
+
         const selectedVariableText = menuState && menuState.selectedValue;
         if (selectedVariableText) {
             menuItems.push({
@@ -131,7 +134,7 @@ class Scratch3DataBlocks {
                 listNames.push(v.name);
             }
         });
-        return listNames;
+        return listNames.sort(StringUtil.stringCompare);
     }
 
     getListsMenuItems (editingTargetID, menuState) {
@@ -160,8 +163,8 @@ class Scratch3DataBlocks {
         return menuItems;
     }
 
-    getListContents ({LIST}, util) {
-        const list = this.getVariableObject(util, LIST, Variable.LIST_TYPE);
+    getListContents (args, util, blockInfo) {
+        const list = util.target.lookupOrCreateVariableByNameAndType(blockInfo.text, Variable.LIST_TYPE);
 
         // If block is running for monitors, return copy of list as an array if changed.
         if (util.thread.updateMonitor) {
@@ -337,6 +340,10 @@ class Scratch3DataBlocks {
             }
         });
 
+        // Sort the variables and lists alphabetically
+        variables.sort((a, b) => StringUtil.compareStrings(a.name, b.name));
+        lists.sort((a, b) => StringUtil.compareStrings(a.name, b.name));
+
         const blocks = [];
         this.addBlocksForVariables(blocks, variables);
         this.addBlocksForLists(blocks, lists);
@@ -389,6 +396,7 @@ class Scratch3DataBlocks {
             func: 'getVariableValue',
             blockType: BlockType.REPORTER,
             text: v.name,
+            paletteKey: v.id,
             customContextMenu: [
                 {
                     text: formatMessage({
@@ -513,6 +521,7 @@ class Scratch3DataBlocks {
             func: 'getListContents',
             blockType: BlockType.REPORTER,
             text: list.name,
+            paletteKey: list.id,
             customContextMenu: [
                 {
                     text: formatMessage({
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -779,7 +779,6 @@ class Scratch3DataBlocks {
     }
 
     serializeVariable (block, target) {
-        target = this.runtime.getTargetById(target.id);
         const variableName = block.mutation.blockInfo.text;
         const variable = target.lookupVariableByNameAndType(variableName, Variable.SCALAR_TYPE);
         const result = [VAR_PRIMITIVE, variableName, variable.id];
--- a/src/blocks/data.js
+++ b/src/blocks/data.js
@@ -7,6 +7,10 @@ const formatMessage = require('format-message');
 
 const LIST_ITEM_LIMIT = 200000;
 
+// serialization primitives
+const VAR_PRIMITIVE = 12;
+const LIST_PRIMITIVE = 13;
+
 /**
  * This Scratch 3.0 extension implements Scratch "data" blocks for variables and lists, including cloud variables.
  */
@@ -350,6 +354,12 @@ class Scratch3DataBlocks {
             menus: {
                 variables: 'getVariablesMenuItems',
                 lists: 'getListsMenuItems'
+            },
+            serialization: {
+                variable: {
+                    serialize: 'serializeVariable',
+                    deserialize: 'deserializeVariable'
+                }
             }
         };
     }
@@ -758,6 +768,40 @@ class Scratch3DataBlocks {
         });
 
         this.runtime.requestToolboxExtensionsUpdate();
+    }
+
+    serializeVariable (block, target) {
+        target = this.runtime.getTargetById(target.id);
+        const variableName = block.mutation.blockInfo.text;
+        const variable = target.lookupVariableByNameAndType(variableName, Variable.SCALAR_TYPE);
+        const result = [VAR_PRIMITIVE, variableName, variable.id];
+        if (block.topLevel) {
+            result.push(block.x ? Math.round(block.x) : 0);
+            result.push(block.y ? Math.round(block.y) : 0);
+        }
+        return result;
+    }
+
+    deserializeVariable (primitiveArray) {
+        if (primitiveArray[0] !== VAR_PRIMITIVE) {
+            throw new Error('bad primitive ID on variable!');
+        }
+        const result = {};
+        result.opcode = 'data_variable';
+        result.fields = {
+            VARIABLE: {
+                name: 'VARIABLE',
+                value: primitiveArray[1],
+                id: primitiveArray[2],
+                variableType: Variable.SCALAR_TYPE
+            }
+        };
+        if (primitiveArray.length > 3) {
+            result.topLevel = true;
+            result.x = primitiveArray[3];
+            result.y = primitiveArray[4];
+        }
+        return result;
     }
 }
 
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -378,10 +378,13 @@ class Blocks {
             if (e.isLocal && editingTarget && !editingTarget.isStage && !e.isCloud) {
                 if (!editingTarget.lookupVariableById(e.varId)) {
                     editingTarget.createVariable(e.varId, e.varName, e.varType);
-                    this.emitProjectChanged();
+
+                    this.runtime.addPendingMonitor(e.varId);
+
                     // TODO this should probably be batched
                     // (esp. if we receive multiple new var_creates in a row).
                     this.runtime.requestToolboxExtensionsUpdate();
+                    this.emitProjectChanged();
                 }
             } else {
                 if (stage.lookupVariableById(e.varId)) {
@@ -396,10 +399,14 @@ class Blocks {
                     }
                 }
                 stage.createVariable(e.varId, e.varName, e.varType, e.isCloud);
-                this.emitProjectChanged();
+
+                this.runtime.addPendingMonitor(e.varId);
+
                 // TODO same as above, this should probably be batched
                 // (esp. if we receive multiple new var_creates in a row).
                 this.runtime.requestToolboxExtensionsUpdate();
+                this.emitProjectChanged();
+
             }
             break;
         case 'var_rename':
@@ -561,6 +568,20 @@ class Blocks {
             this._addScript(block.id);
         }
 
+        // A block was just created, see if it had an associated pending monitor,
+        // if so, keep track of the monitor state change by mimicing the checkbox
+        // event from the flyout. Clear record of this block from
+        // the pending monitors list.
+        if (this === this.runtime.monitorBlocks &&
+            this.runtime.getPendingMonitor(block.id)) {
+            this.changeBlock({
+                id: block.id, // Monitor blocks for variables are the variable ID.
+                element: 'checkbox', // Mimic checkbox event from flyout.
+                value: true
+            }, this.runtime);
+            this.runtime.removePendingMonitor(block.id);
+        }
+
         this.resetCache();
 
         // A new block was actually added to the block container,
@@ -604,9 +625,6 @@ class Blocks {
             } else {
                 // Changing the value in a dropdown
                 block.fields[args.name].value = args.value;
-                if (block.mutation && block.mutation.blockInfo) {
-                    block.mutation.blockInfo.arguments[args.name].selectedValue = args.value;
-                }
 
                 // The selected item in the sensing of block menu needs to change based on the
                 // selected target.  Set it to the first item in the menu list.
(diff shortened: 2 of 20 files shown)


src/extension-support/extension-manager.js
```diff
@@ -171,16 +155,25 @@ class ExtensionManager {
...
+        const loadedExtensions = this.runtime.getLoadedExtensions();
+        const allPromises = [];
+        loadedExtensions.forEach((extensionInfo, oldId) => {
+            const serviceName = extensionInfo.serviceName;
+            allPromises.push(dispatch.call(serviceName, 'getInfo')
                 .then(info => {
                     info = this._prepareExtensionInfo(serviceName, info);
+                    loadedExtensions.set(info.id, info);
+                    if (oldId !== info.id) {
+                        log.warn(`Extension changed ID from ${oldId} to ${info.id}!`);
```
kchadha: Hmmm is this still relevant now that the entire extension manager is no longer being moved to the runtime in this PR? There shouldn't really be a way to change the extension ID right?

src/blocks/data.js
```diff
@@ -7,6 +7,10 @@ const formatMessage = require('format-message');
 
 const LIST_ITEM_LIMIT = 200000;
 
+// serialization primitives
+const VAR_PRIMITIVE = 12;
+const LIST_PRIMITIVE = 13;
```
kchadha: This isn't getting used yet and is causing a lint error.

src/engine/blocks.js
```diff
@@ -547,6 +547,13 @@ class Blocks {
         }
         // Create new block.
         this._blocks[block.id] = block;
+
+        const categoryId = block.categoryId ||
+            (block.mutation && block.mutation.blockInfo && block.mutation.blockInfo.categoryId);
+        if (categoryId) {
+            block.extensionInfo = this.runtime.getExtensionInfo(categoryId);
```
kchadha: I'm worried about this approach of including the entire extension info in the block vm representation. I feel like I would rather have the block just know about what category it belongs to (e.g. it already has this info in its opcode, but I could also see us adding another property called `categoryId`). 

I think instead of having the entire `extensionInfo` embedded in the block VM representation (which I believe is currently just plain JSON) we should just have the serialization code look up the extension info in the runtime when checking for whether a block has custom serialization code. See corresponding comment in the `serialization/sb3.js` code.

src/serialization/sb3.js
```diff
@@ -95,16 +95,42 @@ const primitiveOpcodeInfoMap = {
...
  * @return {array} An array representing the information in the block,
  * or null if the given block is not one of the primitives described above.
  */
-const serializePrimitiveBlock = function (block) {
-    // Returns an array represeting a primitive block or null if not one of
-    // the primitive types above
-    if (hasOwnProperty.call(primitiveOpcodeInfoMap, block.opcode)) {
+const serializePrimitiveBlock = function (block, target) {
+    const customSerialize = getSerializationInfo(block.extensionInfo, block.opcode).serialize;
+    if (customSerialize) {
```
kchadha, line 152: I would rather have this code look up whether the given block belongs to a loaded extension with extensionInfo here via `runtime.getExtensionInfo(block.categoryId)` instead of the VM block JSON representation owning the `extensionInfo` which ultimately has a serialization function nested inside it.

kchadha: As far as how to get access to the runtime.... the top level serialize function has it (which ultimately calls this function through a series of nested calls). Maybe we can do something clever and figure out which blocks have custom serializers ahead of time (or we can just pass the runtime object through to this function).

I'm happy to talk more about all of this offline.

src/serialization/sb3.js
```diff
@@ -288,17 +315,18 @@ const getExtensionIdForOpcode = function (opcode) {
...
  * Serialize the given blocks object (representing all the blocks for the target
  * currently being serialized.)
- * @param {object} blocks The blocks to be serialized
+ * @param {object} target The target containing the blocks to be serialized.
  * @return {Array} An array of the serialized blocks with compressed inputs and
  * compressed primitives and the list of all extension IDs present
  * in the serialized blocks.
  */
-const serializeBlocks = function (blocks) {
+const serializeBlocks = function (target) {
```
kchadha: I think this function was set up on purpose to take just the VM block representation, so that one could have a way of serializing/deserializing just the blocks without the context of the target (to be used for other tools etc.). That's why this file exports `serializeBlocks` and `deserializeBlocks` separately alongside the full `serialize` and `deserialize` functions.

kchadha: See https://github.com/LLK/scratch-vm/pull/1468

kchadha: I guess this in general become more difficult to do with our custom serialization approach...

src/serialization/sb3.js
```diff
@@ -791,30 +819,58 @@ const deserializeFields = function (fields) {
     return obj;
 };
 
+const decodeBlock = function (block) {
```
kchadha: Maybe this can be called `decodeBlockOpcode` to be a little more descriptive (and set it apart from `deserializeBlock`. There's also a function exported by this module that has somewhat similar functionality (`getExtensionIdForOpcode`) with one key difference -- it checks that the `extensionId` is in the `CORE_EXTENSIONS` list defined in this file.

We might want to modify `getExtensionIdForOpcode` to use this new function, or somehow combine/consolidate the overlapping parts of the two functions.

src/serialization/sb3.js
```diff
@@ -791,30 +819,58 @@ const deserializeFields = function (fields) {
...
-const deserializeBlocks = function (blocks) {
+const deserializeBlocks = function (runtime, blocks) {
     for (const blockId in blocks) {
-        if (!Object.prototype.hasOwnProperty.call(blocks, blockId)) {
+        if (!blocks.hasOwnProperty(blockId)) {
             continue;
         }
         const block = blocks[blockId];
+        const {extensionId, extendedOpcode} = decodeBlock(block);
+        const extensionInfo = runtime.getExtensionInfo(extensionId);
```
kchadha, line 885: Hmmmm I don't think this will work if the runtime doesn't already have the appropriate extension loaded.

In the future if we have non-core extensions with blocks with custom serialization, we'll have to figure something else out...

This isn't a problem for core extensions because we load them when we initialize the VM.

kchadha: I remember we talked about this before but I forget if we had a solution for the future problem or not.

src/serialization/sb3.js
```diff
@@ -791,30 +819,58 @@ const deserializeFields = function (fields) {
...
+        const customDeserialize = getSerializationInfo(extensionInfo, extendedOpcode).deserialize;
+        if (customDeserialize) {
+            const newBlock = customDeserialize(block);
+            newBlock.id = uid();
+            delete blocks[blockId];
```
kchadha: I think we should talk about whether this line is needed or not. There is an analogous line below (https://github.com/LLK/scratch-vm/pull/2279/files?file-filters%5B%5D=.js#diff-35528b8c179a308ffa56ce087ed40052R868) because in the old way of serializing primitives, we are not serializing block id information. I don't know that custom serialization/deserialization will necessarily be similar in that way where we won't be serializing an ID as a part of the custom serialization.

src/blocks/data.js
```diff
@@ -759,6 +769,40 @@ class Scratch3DataBlocks {
...
+        if (primitiveArray[0] !== VAR_PRIMITIVE) {
+            throw new Error('bad primitive ID on variable!');
+        }
+        const result = {};
+        result.opcode = 'data_variable';
+        result.fields = {
+            VARIABLE: {
+                name: 'VARIABLE',
+                value: primitiveArray[1],
+                id: primitiveArray[2],
```
kchadha: Hmmm I think this `deserializeVariable` code should be producing the new dynamic block json as opposed to the old variable code block json... (same with the hardcoded line `result.opcode = 'data_variable';` above).

I believe we want the `deserializeVariable` code to be creating a block with a mutation with the blockInfo object in it instead...

kchadha (changes requested):
I did a preliminary review of this code. I tried to be pretty thorough in this pass, but did not look carefully at the documentation or the new `_prepareSerializationInfo` function yet. I think there's a lot to talk about here, so it might be good to discuss some of this in person.

src/extension-support/extension-manager.js
cwillisf: Note from conversation with @kchadha: it's probably best to keep `oldId` instead of switching to `info.id` here.

src/serialization/sb3.js
cwillisf: Notes from discussion: In the long run it seems likely that we'll need to do a two-pass approach to loading. In the first pass we'll initiate loading each extension that the project uses, then we'll wait for them all to finish loading, then we'll do a second pass to actually load the blocks. We don't need that just yet since internal extensions can load synchronously, but we'll need it once we start loading extensions asynchronously.

src/serialization/sb3.js
cwillisf: The problem I ran into is that variable serialization wants to call `target.lookupVariableByNameAndType(...)` in order to get the variable's name and type, so I needed to feed `target` all the way down the line. That's a bit awkward since the "target" being passed into `serializeBlocks` here is the flattened target, not the "real" target, so variable serialization does `target = this.runtime.getTargetById(target.id);` to get the real target before calling `target.lookupVariableByNameAndType`.

Given that, I started thinking we should just pass down the target ID instead of the flattened target object. Then I started thinking about passing a "context" object in case we need more context info in the future instead of just the target ID. Then I was like, "well, the flattened target object is kind of the whole context" -- which is how I got here. Maybe it would be better to just pass in the target ID instead, if for no other reason than to avoid hassling extension authors with flattened targets.

src/serialization/sb3.js
cwillisf: I wasn't thinking of the block "owning" an extensionInfo -- just "borrowing" it from the extension system. That said, I do agree that keeping the VM block JSON simple (as you mentioned in the [`blocks.js` comment above](https://github.com/LLK/scratch-vm/pull/2279#discussion_r325419366)) might be a good goal. I'll try removing `block.extensionInfo` and see how it goes.

src/serialization/sb3.js
cwillisf: Notes from discussion: In the long run it seems likely that we'll need to do a two-pass approach to loading. In the first pass we'll initiate loading each extension that the project uses, then we'll wait for them all to finish loading, then we'll do a second pass to actually load the blocks. We don't need that just yet since internal extensions can load synchronously, but we'll need it once we start loading extensions asynchronously.

src/serialization/sb3.js
kchadha: Agreed about hassling extension authors with flattened targets. I'm okay with passing down a target ID but that also assumes that everywhere the target will be used (currently only in the custom serialization code) there will be access to the runtime. Maybe we can just pass around the hydrated target and call it `runtimeTarget` or something to indicate that it's not flattened? 

Related to a different suggestion I made, however, it seems like we might also need to start passing around the `runtime` as well, in which case `runtime + targetID` is okay too.

src/serialization/sb3.js
kchadha: To clarify part of the note above:

> We don't need that just yet since internal extensions can load synchronously, but we'll need it once we start loading extensions asynchronously.

We don't need that just yet since _the extensions for core block packages get loaded synchronously when the VM is initialized_, but we'll need it _for non-core extensions that start having custom serialization_.

src/blocks/data.js
```diff
@@ -768,6 +777,62 @@ class Scratch3DataBlocks {
...
+        }
+        // @todo reduce duplication here.
+        // This effectively duplicates some of the content in `getInfo()` as well as some of the logic in
+        // `defineDynamicBlock`. Maybe it should return only the mutation blockInfo, or maybe it should return
+        // something that looks like the block's entry in `getInfo()`. If we do either of those things we'll need to
+        // decide how we want to handle the block's (x,y) position and `topLevel` flag.
+        const result = {
+            opcode: 'variable',
+            fields: {
+                VARIABLE: {
```
kchadha, line 805: I don't think we need this field descriptor here. This block should only need the mutation.

src/blocks/data.js
```diff
@@ -768,6 +777,62 @@ class Scratch3DataBlocks {
...
+        if (primitiveTag !== VAR_PRIMITIVE) {
+            throw new Error('bad primitive ID on variable!');
+        }
+        // @todo reduce duplication here.
+        // This effectively duplicates some of the content in `getInfo()` as well as some of the logic in
+        // `defineDynamicBlock`. Maybe it should return only the mutation blockInfo, or maybe it should return
+        // something that looks like the block's entry in `getInfo()`. If we do either of those things we'll need to
+        // decide how we want to handle the block's (x,y) position and `topLevel` flag.
+        const result = {
+            opcode: 'variable',
```
kchadha, line 803: This opcode is incorrect, the top level block opcode is the "extended" opcode.

src/blocks/data.js
```diff
@@ -768,6 +777,62 @@ class Scratch3DataBlocks {
...
+                    id: variableId,
+                    variableType: Variable.SCALAR_TYPE
+                }
+            },
+            inputs: {},
+            next: null,
+            parent: null,
+            shadow: false,
+            mutation: {
+                blockInfo: {
```
kchadha, line 816: Similar to the comment above, I think it would be better for this function to just return this `blockInfo` object and have the rest of this top-level object constructed in either a helper function or by the sb3 deserialize code itself.

src/blocks/data.js
```diff
@@ -768,6 +777,62 @@ class Scratch3DataBlocks {
 
         this.runtime.requestToolboxExtensionsUpdate();
     }
+
+    serializeVariable (block, target) {
+        const variableName = block.mutation.blockInfo.text;
```
kchadha, line 782: Probably this function should just take `blockInfo` as an argument, and then some helper function outside of the extension definition should handle high level block things like setting the x and y.

src/engine/mutation-adapter.js
```diff
@@ -34,7 +34,7 @@ const mutatorTagToObject = function (dom) {
  * @param {(object|string)} mutation Mutation XML string or DOM.
  * @return {object} Object representing the mutation.
  */
-const mutationAdpater = function (mutation) {
+const mutationAdapter = function (mutation) {
```
kchadha, line 37: 😅

src/extension-support/extension-manager.js
```diff
@@ -171,16 +155,33 @@ class ExtensionManager {
      * @returns {Promise} resolved once all the extensions have been reinitialized
      */
     refreshBlocks () {
-        const allPromises = Array.from(this._loadedExtensions.values()).map(serviceName =>
-            dispatch.call(serviceName, 'getInfo')
+        const loadedExtensions = this.runtime.getLoadedExtensions();
+        const allPromises = [];
+        loadedExtensions.forEach((extensionInfo, oldId) => {
+            const serviceName = extensionInfo.serviceName;
+            allPromises.push(dispatch.call(serviceName, 'getInfo')
```
kchadha: I think the `dispatch.call` here should be on its own line for a little bit more clarity when reading.

src/extension-support/extension-manager.js
```diff
@@ -389,6 +393,34 @@ class ExtensionManager {
...
+     */
+    _prepareSerializationInfo (extensionId, serviceName, rawSerializationInfo) {
+        rawSerializationInfo = rawSerializationInfo || {};
+        const cookedSerializationInfo = {};
+        for (const opcode in rawSerializationInfo) {
+            if (!rawSerializationInfo.hasOwnProperty(opcode)) continue;
+            const functionNames = rawSerializationInfo[opcode];
+            const extendedOpcode = `${extensionId}_${opcode}`;
+            const boundFunctions = {};
+            if (functionNames.serialize) {
```
kchadha, line 418: Do we want to throw an error or log a warning if `serialize` or `deserialize` are not specified in the serialize descriptor for the opcode?

src/serialization/sb3.js
```diff
@@ -95,16 +97,59 @@ const primitiveOpcodeInfoMap = {
     data_listcontents: [LIST_PRIMITIVE, 'LIST']
 };
 
+const primitiveIdMap = (() => {
```
kchadha, line 102: Could you add a comment for this function?

src/serialization/sb3.js
```diff
@@ -95,16 +97,59 @@ const primitiveOpcodeInfoMap = {
...
+ * @param {Runtime} runtime - the VM runtime.
+ * @returns {ExtensionInfo} - metadata for the extension providing the block, or `undefined` if absent / not loaded.
+ */
+const getExtensionInfoForBlock = function (block, runtime) {
+    const categoryId = block.categoryId ||
+        (block.mutation && block.mutation.blockInfo && block.mutation.blockInfo.categoryId);
+    if (categoryId) {
+        return runtime.getExtensionInfo(categoryId);
+    }
+    // undefined
```
kchadha: cruft?

kchadha: Should we log a warning here if the categoryId was not found?

src/serialization/sb3.js
```diff
@@ -272,6 +319,28 @@ const compressInputTree = function (block, blocks) {
...
+ * @property {string} extensionId - the extension/category for this block, like 'myExtension'. May be an empty string.
+ */
+const getExtensionAndOpcode = function (blockJSON) {
+    const extendedOpcode = Array.isArray(blockJSON) ? primitiveIdMap[blockJSON[0]] : blockJSON.opcode;
+    const firstSplit = extendedOpcode.indexOf('_');
+
+    // note: if we ever need the simple / non-extended opcode, it's just extendedOpcode.substring(firstSplit + 1)
+    return {
+        extendedOpcode,
+        extensionId: extendedOpcode.substring(0, firstSplit)
```
kchadha: Elsewhere we use `StringUtil.splitFirst` to get this info, might be good to use that here for consistency.

src/serialization/sb3.js
```diff
@@ -288,20 +357,20 @@ const getExtensionIdForOpcode = function (opcode) {
 /**
  * Serialize the given blocks object (representing all the blocks for the target
  * currently being serialized.)
- * @param {object} blocks The blocks to be serialized
- * @return {Array} An array of the serialized blocks with compressed inputs and
- * compressed primitives and the list of all extension IDs present
- * in the serialized blocks.
+ * @param {RenderedTarget} target The target containing the blocks to be serialized.
+ * @param {Runtime} runtime The VM runtime
+ * @param {Set} extensions A set of extensions to add extension IDs to
```
kchadha: This param comment doesn't really explain which extension IDs are being tracked or why

src/serialization/sb3.js
```diff
@@ -434,33 +503,33 @@ const serializeComments = function (comments) {
 /**
  * Serialize the given target. Only serialize properties that are necessary
  * for saving and loading this target.
- * @param {object} target The target to be serialized.
+ * @param {RenderedTarget} target The target to be serialized.
+ * @param {Runtime} runtime The VM runtime
  * @param {Set} extensions A set of extensions to add extension IDs to
```
kchadha: I guess we have the same problem here.. so you can probably ignore the comment above.

Alternatively we can make it more descriptive -- the set of extension IDs from the blocks being used by this target (or some better phrasing)

src/serialization/sb3.js
```diff
@@ -792,28 +855,57 @@ const deserializeFields = function (fields) {
 };
 
 /**
- * Covnert serialized INPUT and FIELD primitives back to hydrated block templates.
+ * Convert dynamic blockInfo JSON from a custom deserialization function into an XML element ready for runtime use.
+ * @param {object} dynamicBlockInfo - a BlockInfo-like object with properties like `opcode` and `isDynamic`.
+ * @returns {Element} - an XML Element with tag name 'mutation' and an attribute called 'blockInfo'
```
kchadha: The `mutationAdapter` function does not return an XML Element, it takes an XML element (or string) and returns a json object w/ properties relevant to the VM.

src/serialization/sb3.js
```diff
@@ -792,28 +855,57 @@ const deserializeFields = function (fields) {
...
-            // delete the old entry in object.blocks and replace it w/the
-            // deserialized object
+        let block = blocks[blockId];
+        const {extensionId, extendedOpcode} = getExtensionAndOpcode(block);
+        const extensionInfo = runtime.getExtensionInfo(extensionId);
+        const customDeserialize = getSerializationInfo(extensionInfo, extendedOpcode).deserialize;
+        if (customDeserialize) {
+            block = customDeserialize(block);
+            blocks[blockId] = block; // customDeserialize might have made a new object; make sure to keep it!
+            // the deserialize function should return opcode 'foo' and we'll fix it to 'ext_foo'
```
kchadha, line 890: Related to the comment earlier in this PR, if we're doing fix-up here anyways of what the customDeserialize function returns, I think we might as well construct the block JSON here instead of the customDeserialize function and have that function just return the blockInfo.

kchadha (changes requested):
Left comments in-line in the PR.

src/blocks/data.js
cwillisf: The problem is that variables have their own custom way of serializing x and y, so it kind of needs to be here. That's also why the custom deserialize function needs to return the whole block instead of just the blockInfo guts :/

src/serialization/sb3.js
cwillisf: This can happen legitimately for now (until 100% extensionification) so I added some more comment here to explain that.

src/serialization/sb3.js
cwillisf: oops - yeah, I forgot to update the comment after settling on how this should work...

src/blocks/data.js
cwillisf: I was trying to avoid having the extension every mention the extended opcode, since in theory we might sometimes post-process the extension ID so that it isn't what the extension author expects. Instead I fix up the ID after the fact, in the function which calls this one.

kchadha:
Unassigning myself from this PR since we discussed offline (a few weeks ago) that maybe custom serialization in the extension spec is not the right approach since we only want it in this specific case of the core variables extension. We can handle this special case in the "backend" (e.g. vm runtime or extension manager) instead of making this a developer facing feature. We also discussed removing this from the immediate scope of this "first phase" of the e16n project (wrapping up by the end of this year).

towerofnix:
>We can handle this special case in the "backend" (e.g. vm runtime or extension manager) instead of making this a developer facing feature. We also discussed removing this from the immediate scope of this "first phase" of the e16n project (wrapping up by the end of this year).

does this mean custom variable types are out of the question for developer extensions? not that I actually know what's going on in this code, nor that it'd mean that wouldn't be supported *ever*, but this was one of the most exciting things about extensions IMO (localstorage-based variables, etc).

PR #2280, opened 2019-09-21
Inline Images in Extensions

kchadha:
### Resolves

Adds support for inline images in extension blocks. Towards supporting the goals of the extension-ification project as well, however additional work will be needed to support inline images in *dynamic extension blocks* as introduced by the extensionification effort.

Closes #2259 as this PR provides the same functionality.

### Proposed Changes

- Add a new argument type "IMAGE". Add support in the `convertPlaceholders` function for creating an inline image on a block. This function is responsible for translating the extension specification for block arguments into the JSON and XML representations scratch blocks uses to define a block opcode and layout a block.

- Make the COLOR argument type more consistent with the others: the extension developer can now specify a defaultValue for the color picker to be used when the extension is first loaded. If the extension developer leaves out the `defaultValue` property in the argument description, the default scratch-blocks behavior of choosing a random color will be used.

Things that need discussion:
- We are adding something that does not actually represent a block "argument" to the set of "argument types" because inline images are currently the only type of addition to a block other than block text that does not represent a block argument of some sort.
- The re-structuring of the convertPlaceholders can be up for debate.
- Is `false` the right default for the `flipRTL` property for the inline image? A value of false means that the image will not be flipped horizontally when the block is laid out for an RTL language.
@chrisgarrity Any thoughts about this? I chose false as the default because that's how it is for our turn blocks (clockwise and counterclockwise should not be flipped in RTL), and the extension developer can be explicit about making the argument flip in RTL by specifying `flipRTL: true` in the argument specification in the extension info.
- Do we want to do anything else when the `dataURI` property is not provided in the specified IMAGE argument specification? Currently we log a warning in the console and the block gets laid out with space allocated for the image.
image
- It's unclear what the `alt` property in the scratch-blocks/blockly `field_image` descriptor is used for. I had expected the alt text to be either a tool-tip or alt text if the image was not provided or could not be loaded, but the above image indicates that's not true. Blockly documentation indicates that the alt text is used when the block is collapsed, see description of `opt_alt` property here:
https://developers.google.com/blockly/reference/js/Blockly.FieldImage
@picklesrus any ideas about this?

### Reason for Changes

Towards extension-ification, and also adding more features to the extension spec.

### Test Coverage

Added unit test for new inline image support.

You can test out these changes by bringing in the code from this PR and enabling the coreExample extension by adding `'coreExample'` to this list of core extensions:
https://github.com/LLK/scratch-vm/blob/develop/src/virtual-machine.js#L30

cc/ @thisandagain @knandersen @asnoba

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1250,7 +1250,9 @@ class Runtime extends EventEmitter {
                     src: argInfo.dataURI,
                     // TODO these probably shouldn't be hardcoded...?
                     width: 24,
-                    height: 24
+                    height: 24,
+                    // TODO is false a good default here?
+                    flip_rtl: argInfo.flipRTL || false
                 };
             }
         } else {
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -51,38 +51,58 @@ const defaultExtensionColors = ['#0FBD8C', '#0DA57A', '#0B8E69'];
 const ArgumentTypeMap = (() => {
     const map = {};
     map[ArgumentType.ANGLE] = {
-        shadowType: 'math_angle',
-        fieldType: 'NUM'
+        shadow: {
+            type: 'math_angle',
+            // We specify fieldNames here so that we can pick
+            // create and populate a field with the defaultValue
+            // specified in the extension.
+            // When the `fieldName` property is not specified,
+            // the <field></field> will be left out of the XML and
+            // the scratch-blocks defaults for that field will be
+            // used instead (e.g. default of 0 for number fields)
+            fieldName: 'NUM'
+        }
     };
     map[ArgumentType.COLOR] = {
-        shadowType: 'colour_picker'
+        shadow: {
+            type: 'colour_picker',
+            fieldName: 'COLOUR'
+        }
     };
     map[ArgumentType.NUMBER] = {
-        shadowType: 'math_number',
-        fieldType: 'NUM'
+        shadow: {
+            type: 'math_number',
+            fieldName: 'NUM'
+        }
     };
     map[ArgumentType.STRING] = {
-        shadowType: 'text',
-        fieldType: 'TEXT'
+        shadow: {
+            type: 'text',
+            fieldName: 'TEXT'
+        }
     };
     map[ArgumentType.BOOLEAN] = {
         check: 'Boolean'
     };
     map[ArgumentType.MATRIX] = {
-        shadowType: 'matrix',
-        fieldType: 'MATRIX'
+        shadow: {
+            type: 'matrix',
+            fieldName: 'MATRIX'
+        }
     };
     map[ArgumentType.NOTE] = {
-        shadowType: 'note',
-        fieldType: 'NOTE'
+        shadow: {
+            type: 'note',
+            fieldName: 'NOTE'
+        }
     };
     map[ArgumentType.IMAGE] = {
-        // TODO was going to use fieldType here, but I didn't want to
-        // confuse it with how fieldType is being used in all these
+        // TODO was going to use shadowFieldName here, but I didn't want to
+        // confuse it with how shadowFieldName is being used in all these
         // other argument type cases
         // Inline images are weird because they're not actually "arguments".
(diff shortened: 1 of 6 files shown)


chrisgarrity:
@kchadha `flipRTL` false is probably the correct default. In general they will be a representation of something physical that won't change when the language changes. Developers can always override it for any exceptions.

kchadha:
> It's unclear what the alt property in the scratch-blocks/blockly field_image descriptor is used for. I had expected the alt text to be either a tool-tip or alt text if the image was not provided or could not be loaded, but the above image indicates that's not true. Blockly documentation indicates that the alt text is used when the block is collapsed, see description of opt_alt property here:
https://developers.google.com/blockly/reference/js/Blockly.FieldImage
@picklesrus any ideas about this?

Based on an offline conversation w/ @picklesrus, it seems like this 'alt' property is being used for a feature from Blockly that is not being used in scratch. See the [Blockly playground](https://blockly-demo.appspot.com/static/tests/playground.html) for collapsing a block (drag out a block from the toolbox, open the context menu of the block and click "Collapse Block").

Based on this conversation, I took out the 'alt' property from the inline image argument descriptor in the extension spec.

src/engine/runtime.js
```diff
@@ -51,30 +51,58 @@ const defaultExtensionColors = ['#0FBD8C', '#0DA57A', '#0B8E69'];
 const ArgumentTypeMap = (() => {
     const map = {};
     map[ArgumentType.ANGLE] = {
-        shadowType: 'math_angle',
-        fieldType: 'NUM'
+        shadow: {
+            type: 'math_angle',
+            // We specify fieldNames here so that we can pick
+            // create and populate a field with the defaultValue
+            // specified in the extension.
```
cwillisf, line 58: I'm not sure I understand this comment... specifically the "so that we can pick create and populate" part. Is there a typo there or am I just failing to parse it right?

src/engine/runtime.js
```diff
@@ -1232,63 +1256,102 @@ class Runtime extends EventEmitter {
...
+        case 'field_image': {
+            if (!argInfo.dataURI) {
+                log.warn('Missing data URI in extension block with argument type IMAGE');
+            }
+            argJSON = {
+                type: 'field_image',
+                src: argInfo.dataURI || '',
+                // TODO these probably shouldn't be hardcoded...?
+                width: 24,
+                height: 24,
```
cwillisf: Hmm... I think it makes sense to specify the height so that it fits the block correctly. If you leave out the width will it auto-scale appropriately, maintaining the aspect ratio?

src/engine/runtime.js
```diff
@@ -1232,63 +1256,102 @@ class Runtime extends EventEmitter {
...
+                // Whether or not the inline image should be flipped horizontally
+                // in RTL languages. Defaults to false, indicating that the
+                // image will not be flipped.
+                flip_rtl: argInfo.flipRTL || false
+            };
+            break;
         }
+        default: {
```
cwillisf: I'm a little uncomfortable with how much code is in this `default: { ... }` here. I see that `'field_image'` needs very different treatment from the rest but... hmm. Does the `default` section use too many of the surrounding variables for it to be pulled out into its own function reasonably? If this really is the best organization then I can probably cope ;)

cwillisf (commented):
This is great! I have a couple of questions but I'm not actually certain that they'll lead to changes. At any rate I think it's either ready or very close :)

src/engine/runtime.js
kchadha: Ah, I was differentiating "creating a field" e.g. laying out space on the block for the correct field, vs. "populating a field" as in filling the empty text input (e.g. for number fields) with the number `10` for `move 10 steps`

src/engine/runtime.js
kchadha: Oh mostly I had an issue with the numbers being hardcoded here instead of constants we pull from somewhere (esp. if we use the VM with a version of scratch-blocks that has some other settings). The number `24` has no context here.

cwillisf (approved):
👍

PR #2289, opened 2019-10-03
Localize the Text to Speech extension "set language" menu

### LINKED ISSUE #1673: "Set language" menu in Text to Speech extension should accept dropped language names

### LINKED ISSUE #1678: Translate languages for Set language block (TTS Extension)

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1678
Resolves https://github.com/LLK/scratch-vm/issues/1673

Depends on https://github.com/LLK/scratch-translate-extension-languages/pull/12

### Proposed Changes

In the Text to Speech extension, use the translated language names from the `scratch-translate-extension-languages` module to display a localized menu of language names in the "set language" block.  

Note that this change depends on an update to the `scratch-translate-extension-languages` module that includes translations of some spoken language names.

Screen Shot 2019-10-03 at 4 26 09 PM
Screen Shot 2019-10-03 at 4 26 30 PM

Also, support dropped arguments in the "set language" block, so that a language could be selected using a string that is a language name (in any language). This works with the translate extension's "language" reporter block, so you can now create projects that translate into and then speak in whatever language the editor is currently set to:

Screen Shot 2019-10-03 at 4 31 24 PM
Screen Shot 2019-10-03 at 4 31 07 PM

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -519,6 +519,15 @@ class Scratch3Text2SpeechBlocks {
             stage.textToSpeechLanguage = this._getExtensionLocaleForSupportedLocale(locale);
         }
 
+        // Support language names dropped onto the menu via reporter block
+        // such as a variable containing a language name (in any language),
+        // or the translate extension's language reporter.
+        const localeForDroppedName = languageNames.nameMap[locale.toLowerCase()];
+        if (localeForDroppedName && this.isSupportedLanguage(localeForDroppedName)) {
+            stage.textToSpeechLanguage =
+                this._getExtensionLocaleForSupportedLocale(localeForDroppedName);
+        }
+
         // If the language is null, set it to the default language.
         // This can occur e.g. if the extension was loaded with the editor
         // set to a language that is not in the list.
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -628,17 +628,19 @@ class Scratch3Text2SpeechBlocks {
         };
 
         // Get the array of localized language names
-        let nameArray = languageNames.menuMap[this.getEditorLanguage()];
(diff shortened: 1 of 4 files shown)


src/extensions/scratch3_text2speech/index.js
```diff
@@ -584,14 +595,44 @@ class Scratch3Text2SpeechBlocks {
...
+        let nameArray = languageNames.menuMap[editorLanguage];
+        if (nameArray) {
+            // Also get any localized names of spoken languages
+            let spokenNameArray = [];
+            if (languageNames.spokenLanguages) {
+                spokenNameArray = languageNames.spokenLanguages[editorLanguage];
+                nameArray = nameArray.concat(spokenNameArray);
+            }
+            // Create a map of language code to localized name
+            nameArray.forEach(lang => {
```
picklesrus, line 623: Maybe add a note about how we don't end up with duplicates for Chinese and Mandarin in the array.

picklesrus: And a test for it :)

ericrosenbaum: all set!

ericrosenbaum:
(tests will fail until we update the scratch-translate-extension-languages module)

PR #2291, opened 2019-10-11
Fix video transparency and deeply explain the hack

### LINKED ISSUE #4888: Video transparency in video sensing doesn't go all the way invisible
### Expected Behavior

Setting the video transparency to 100 using the video sensing block `set video transparency to 100` block should cause the video to be totally transparent, and not visible.

### Actual Behavior

`set video transparency block to 100` seems to make the parts of the video that are white be white and the things that are dark to be clear.  It seems to be using the brigthness channel to affect the transparency in some way.  

Transparency100videoSensingPurple

The color that you set the background to seems to matter.  This issue is not observable if you set the backdrop to white, but the contrast changes depending on what color the backdrop is.

White/Clear backdrop at 100% transparency:
transparency100videoSensingWhite

Purple backdrop at 100% transparency:
Transparency100videoSensingPurple

Here is an example project made by @ericrosenbaum to illustrate that. scratch.mit.edu/projects/314377995

It is also worth noting that less transparent video sensing makes the colors look strange.

50% transparent on white backdrop:
Screen Shot 2019-06-04 at 11 35 38 AM

50% transparent on purple backdrop:
Transparency50videoSensingPurple

### Steps to Reproduce

Create a new project
Add the video sensing extension
Turn video on
Note: the colors in the video look a little desaturated
Drag out a `Set video transparency block to ` block out
Set the value to 100 and run the block
Note: the video appears to disappear
Add a backdrop, it's easy to see if it is a solid color
Note: the video on the stage is still visible.

Mac Chrome, Mac Safari

paulkaplan:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4888

### Proposed Changes

The only functional line of code here is to override the skin used for video to return `hasPremultipliedAlpha` as false, so that the ghost effect can be applied correctly.

I also put in a thorough explanation comment of why video works the way it does, using a PenSkin and private drawStamp method instead of a BitmapSkin and the public updateBitmapSkin method, and why the above override is needed for correct compositing. I also indicated what kind of future work could improve this part of the code (lazy updating bitmap skin silhouettes, see bottom).

The other code added makes asserts about the exact shape of the private APIs we are using, so that if they change in the future we will be greeted with an alert message that tells us exactly ~why we are bad and should feel bad~ what happened and how to fix it.

### Reason for Changes

@cwillisf this wasn't exactly the fix we talked about, basically because I felt that I couldn't make the fix that I really want to in limited time (specifically making bitmap skins lazy update silhouettes) and I didn't want to introduce a new public API that would introduce more maintenance cost and more confusion about its capabilities (a "VideoSkin" that is an implicitly opaque PenSkin would surely come back to haunt us). 

As to why this bug happened, @kchadha was very helpful thinking this through. I believe it comes down to the VM using a highly specialized API (PenSkin) in a way that it wasn't intended (which is obvious because it required using private methods), and then being burned when an assumption was changed that was only considered in the narrow scope (pen extension) without realizing the broader implications. 

To be clear this only deepens the hack of using the PenSkin for video, but hopefully improves the situation somewhat by clearly documenting the situation and adding at least a tiny bit of guard rails against future changes. 

---

_Further info about `updateBitmapSkin`_

I realize now that mentioning lazy updating a skins silhouette may be confusing/out of left field as the solution to the problem, so let me explain a bit more down here.

The bitmap skin update does two things. It consumes an image in many possible formats (canvas, video, image, imagedata), then updates the skins texture and also updates the skin silhouette right away. The silhouette update demands ImageData, so whatever is passed to the update must be converted right away. the texture is less demanding, it can be updated from any source. Timing wise this looks like (very rough, from my computer)

1. Draw video to canvas (4ms)
2. Update texture from canvas (very fast)
3. Convert canvas to ImageData (4ms) 
4. Update silhouette from image data (very fast)

The pen skin update only updates the texture, the silhouette update is done when the silhouette is needed, which for the video skin is never, so the conversion to ImageData is never done. So switching to BitmapSkin now would double the amount of time per frame. But updating the silhouette on demand would allow us to bring it back down to current levels.

A further performance boost might be achieved by reading the video data directly from the video element, skipping to 4ms drawImage call to copy the video to a canvas. The `gl.texImage2D` function allows this, but I haven't tried it. That would require a new format in the VideoProvider as well.

### DIFF
--- a/src/io/video.js
+++ b/src/io/video.js
@@ -152,13 +152,53 @@ class Video {
         const {renderer} = this.runtime;
         if (!renderer) return;
 
+        /**
+         * Use a PenSkin for drawing the video, and a Drawable with the ghost effect for transparency.
+         * This requires several hacks, however.
+         *
+         * 1. Need to access private renderer APIs to get the Skin directly and to call drawStamp,
+         *    since the public renderer API only allow stamping other drawables onto the skin.
+         * 2. Override the hasPremultipliedAlpha getter on the skin to return false. Without this,
+         *    the ghost effect does not work correctly because the shader modifies the alpha channel
+         *    assuming straight alpha compositing will be used.
+         *
+         * We considered using a BitmapSkin instead, updating with the public renderer#updateBitmapSkin,
+         * but the PenSkin#drawStamp is faster because of lazy silhouette updating. Future work that
+         * makes all skins update silhouette lazily would allow us to use a BitmapSkin here.
+         *
+         * To try to make future work here easier, I've added asserts around the private APIs we are using
+         * so that internal renderer changes impacting video are more easily caught.
+         */
+
         if (this._skinId === -1 && this._skin === null && this._drawable === -1) {
             this._skinId = renderer.createPenSkin();
-            this._skin = renderer._allSkins[this._skinId];
             this._drawable = renderer.createDrawable(StageLayering.VIDEO_LAYER);
             renderer.updateDrawableProperties(this._drawable, {
                 skinId: this._skinId
             });
+
+            try {
+                this._skin = renderer._allSkins[this._skinId];
+
+                // Make sure the private APIs relied on are still valid to help surface changes faster.
+                if (!this._skin) throw new Error('Could not get skin from _allSkins');
+                if (!this._skin.clear) throw new Error('PenSkin#clear method does not exist');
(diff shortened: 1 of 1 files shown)


paulkaplan:
Welp, this isn't needed anymore. See https://github.com/LLK/scratch-vm/pull/2292

PR #2293, opened 2019-10-15
Fix video sensing error with stale matrices

paulkaplan:
### Resolves

An issue where video sensing on a sprite was broken. 

An alternative to https://github.com/LLK/scratch-render/pull/516
### Proposed Changes

https://github.com/LLK/scratch-render/pull/468 introduced a performance optimization that removed inverse matrix updating from `getFastBounds`. It was observed that everywhere that needed the inverse matrix was updating it as needed, but that was only true looking within the scratch-render repo. Video sensing was using that private API and depending on getFastBounds to update the inverse matrix, because it uses `drawable.isTouching`. 

Since the public APIs on the renderer manage updating the inverse matrix and also doing lazy silhouette updates, it doesn't seem to be the renderers fault. instead, just make those updates before using the dangerous private APIs on this side.

This seems like the classic result of using private APIs, and I'd really like to move away from it here, but that will take more work. 

/cc @mzgoddard

### DIFF
--- a/src/extensions/scratch3_video_sensing/library.js
+++ b/src/extensions/scratch3_video_sensing/library.js
@@ -289,6 +289,11 @@ class VideoMotion {
                 _curr: curr
             } = this;
 
+            // The public APIs for Renderer#isTouching manage keeping the matrix and
+            // silhouette up-to-date, which is needed for drawable#isTouching to work
+            drawable.updateMatrix();
+            if (drawable.skin) drawable.skin.updateSilhouette();
+
             // Restrict the region the amount and direction are built from to
             // the area of the current frame overlapped by the given drawable's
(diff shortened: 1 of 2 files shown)


src/extensions/scratch3_video_sensing/library.js
```diff
@@ -289,6 +289,11 @@ class VideoMotion {
                 _curr: curr
             } = this;
 
+            // The public APIs for Renderer#isTouching manage keeping the matrix and
+            // silhouette up-to-date, which is needed for drawable#isTouching to work
```
kchadha:
```suggestion
            // silhouette up-to-date, which is needed for drawable#isTouching to work (used below)
```

PR #2298, opened 2019-11-03
Allow extension to force to BLE write without response as well as with response

terado0618:
### Resolves

https://github.com/LLK/scratch-link/issues/150

### Proposed Changes

* Avoid missing ```withResponse``` is ```false``` in BLE write parameter

### Reason for Changes

This PR allows extension to force to write without response as well as with response (#1471).


cwillisf (approved):
👍
This is probably how it should have always been 😅
Thanks!

PR #2302, opened 2019-11-13
Fix custom field types

ErikMejerHansen:
### Resolves

Handling of fieldTypes seems to have changed since 0.2.0-prerelease.20190619042313 and the expected structure of argumentTypes have changed

### Proposed Changes

Fix handling of custom field types

### Reason for Changes

Fix handling of custom field types

### Test Coverage

None

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -981,8 +981,10 @@ class Runtime extends EventEmitter {
             fieldName: fieldName,
             extendedName: extendedName,
             argumentTypeInfo: {
-                shadowType: extendedName,
-                fieldType: `field_${extendedName}`
+            shadow: {
+                type: extendedName,
+                fieldName: `field_${extendedName}`
+            }
             },
             scratchBlocksDefinition: this._buildCustomFieldTypeForScratchBlocks(
                 extendedName,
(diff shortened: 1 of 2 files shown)


kchadha:
@ErikMejerHansen good catch! I didn't realize this use case existed when I pushed up the change for inline images in extension blocks (e.g. https://github.com/LLK/scratch-vm/pull/2280). The code looks good but I hope to test this change before putting it in...

kchadha:
@ErikMejerHansen good catch, thank you for submitting this PR!

I missed this use case when I updated the argument type structure when pushing the change for inline images in extension blocks (#2280). Since this code path isn't really used by the main Scratch code base, do you think it'd be possible to add a test to this repo that could help prevent something like this from happening again?

ErikMejerHansen:
@kchadha Yeah, that sounds like a plan! I'll get it added.

ErikMejerHansen:
Hi @kchadha 
I've added a somewhat simplistic test. 
It would probably have been better to parse the block XML and verify it in a bit more stringent way. 

Hopefully I can extend the test when I get to working on enabling non-droppable custom fields.

kchadha (approved):
LGTM! Thank you for adding the test!

PR #1793, opened 2018-11-29
Remove vestigial device manager

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1545

### Proposed Changes

Remove the code related to the legacy Scratch Device Manager, which is now replaced by Scratch Link, supported through the BLE and BT io devices.

Farewell, Scratch Device Manager! You served us well!

PR #549, opened 2017-04-28
Device manager client

cwillisf:
### Resolves

Progress toward #518 

### Proposed Changes

This change adds a new type of I/O device, called `DeviceManager`, which acts as a client for the Scratch Device Manager application. An instance of this new type is now available in the runtime's list of I/O devices and is called `deviceManager`.

### Test Coverage

A few basic tests for the new `DeviceManager` I/O device have been added as part of this change.

Testing the actual functionality requires a running instance of the Scratch Device Manager application. You can test this in the browser console with commands like this:

Retrieve a list of currently-available WeDo 2.0 devices (and trigger a scan if there isn't one already going):
```js
Scratch.vm.runtime.ioDevices.deviceManager.list('wedo2', 'test').then(x => console.log(`List result: ${x}`))
```

Create a `DeviceFinder` to handle searching for and connecting to a WeDo 2.0:
```js
var finder = Scratch.vm.runtime.ioDevices.deviceManager.searchAndConnect('test', 'wedo2');
finder.promise.then(x => console.log(`Result: ${x}`), reason => console.log(`Failed: ${reason}`));
```
You should start seeing list requests once per second in your browser's network panel.

If you connect a WeDo 2.0 through the Device Manager, you should see that the result is a socket. If you instead cancel the search:
```js
finder.cancel();
```
...you should see `Failed: canceled` in your console.

PR #2303, opened 2019-11-14
Remove unused socket.io-client dependency

adroitwhiz:
### Proposed Changes

This PR removes the `socket.io-client` NPM package.

It also adds the `callsite` package as a dev dependency.

### Reason for Changes

The `socket.io-client` package dates back to #549, where it was used solely by the `DeviceManager`.

The `DeviceManager` was removed in #1793, but `socket.io-client` was not. Despite being unused, it's still listed as a dependency and pulling in a lot of other dependencies.

The `callsite` package is used by `dispatch-test-worker-shim.js`. It was erroneously missing as a dependency, but was incidentally provided somewhere in `socket.io-client`'s dependency tree.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1453,11 +1453,6 @@
       "integrity": "sha512-L8vcjDTCOIJk7wFvmlEUN7AsSb8T+2JrdP7KINBjzr24TJ5Mwj590sLu3BC7zNZowvJWa/JtPmD8eJCzdtDWjA==",
       "dev": true
     },
-    "after": {
-      "version": "0.8.2",
-      "resolved": "https://registry.npmjs.org/after/-/after-0.8.2.tgz",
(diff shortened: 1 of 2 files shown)


ericrosenbaum:
thanks @adroitwhiz! @cwillisf: it looks like this is failing on travis due to a unit test failing, which is maybe related to a change to the way the `callsite` package is being pulled in? Can you please have a look?

adroitwhiz:
That was my bad--when I re-re-rebased this I started over from scratch, and forgot to `npm install --save-dev callsite` again.

cwillisf (approved):
> The callsite package is used by dispatch-test-worker-shim.js. It was erroneously missing as a dependency, but was incidentally provided somewhere in socket.io-client's dependency tree.

Oops! Thanks for tracking this down!

Issue #1445, opened 2018-08-12
EV3 Feature request: Enable IR sensors


nekketsuuu:
### Expected Behavior

We can use [45509 EV3 Infrared Sensor](https://shop.lego.com/en-US/EV3-Infrared-Sensor-45509) in Scratch 3.0 EV3 extension.

Although [5003400 LEGO MINDSTORMS Education EV3 Core Set](https://education.lego.com/en-us/products/lego-mindstorms-education-ev3-core-set-/5003400) does not contain the IR sensor, [31313 MINDSTORMS EV3](https://www.lego.com/en-us/mindstorms/products/mindstorms-ev3-31313) for toys contains it. Moreover, the latter does not have the ultrasonic sensor. So I supposed the IR sensor can be used from the EV3 extension.

### Actual Behavior

We cannot. The IR sensor can measure distance (or proximity), but `distance` block always shows zero.

### Steps to Reproduce

1. Connect one IR sensor to a EV3.
2. Connect the EV3 to a computer using Bluetooth and Scratch Link on <https://beta.scratch.mit.edu/>.
3. Try to read from the sensor using `distance` block.

### Operating System and Browser

Windows 10 Home 1803, Google Chrome 68.0.3440.106 (Official build) (64 bit), 45509 EV3 Infrared Sensor
I tested on https://beta.scratch.mit.edu/, on Aug 12, 2018.

jokuha:
New to scratch (and still fairly new to EV3) – thank you for the efforts, my daughter and me love it.

Now, is there any chance to get the IR distance sensor working? It indeed still seems to be putting out zero distance no matter what(?).

PR #2306, opened 2019-12-02
EV3 populate distance from either IR or UltraSonic sensor.

SirTechington:
### Resolves
https://github.com/LLK/scratch-vm/issues/1445

### Proposed Changes
In the EV3 Scratch extension, allow the distance to be populated from an IR sensor if installed.

### Reason for Changes
This is useful since the EV3 commercial kit (Lego 31313) contains only an IR sensor for distance sensing whereas educational kits (Lego 45544) contains the already supported UltraSonic sensor. This allows those with the commercial kit to use all Scratch EV3 features.

### Note
The range of values from the IR sensor matches what comes from the UltraSonic sensor, but the units are different. The UltraSonic returns values between 0 and 100 inches (the extension hard-codes it to inches mode), and the IR sensor returns "proximity" in the range of 0 to 100%, where 100% corresponds to "approximately 50-70cm". This is no different from how the sensors report in native EV3 programming.

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -111,6 +111,7 @@ const Ev3Device = {
     29: 'color',
     30: 'ultrasonic',
     32: 'gyro',
+    33: 'ir',
     16: 'touch',
     8: 'mediumMotor',
     7: 'largeMotor',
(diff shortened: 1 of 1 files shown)


Kristina-Winterland:
Yes, my daughter and I are as well hoping that the IR sensor can be read by Scratch 3.0. After all, on their webpage, Lego officially had announced cooperation with Scratch 3.0 on behalf of EV3 but practice is not that jolly, apparently. Thank you for setting this request up for all of us who are waiting.

PR #2309, opened 2019-12-24
touching mouse-pointer returns false when outside stage

### LINKED ISSUE #5243: Touching Mouse Pointer returns true even if mouse is not on the stage
### Expected Behavior

If the mouse isn't on the stage the "touching mouse pointer" block won't return true.

### Actual Behavior
If a sprite is partially of the edge of the screen the "touching mouse pointer" block will return true if the mouse is over the part of the sprite that is off the stage.
image4

### Steps to Reproduce

Create a new project
drag the cat partially off the stage
Add the following stack
`When green flag clicked` `forever` `if touching mouse pointer` `change color effect by 15`
place the mouse off the stage but over where the cat would be
Note: the cat changes color

Mac Chrome

apple502j:
### Resolves
Resolves LLK/scratch-gui#5243

### Proposed Changes
Adds `getIsOutside` function in io/mouse and uses it inside rendered-target `isTouchingObject`. If the param is `__mouse__` and getIsOutside is true, it returns false.

### Reason for Changes
See LLK/scratch-gui#5243

### Test Coverage
Added one test.

### DIFF
--- a/test/unit/sprites_rendered-target.js
+++ b/test/unit/sprites_rendered-target.js
@@ -295,6 +295,20 @@ test('isTouchingObjectOutsideStage', t => {
     t.end();
 });
 
+test('isTouchingMouse', t => {
+    const r = new Runtime();
+    const s = new Sprite(null, r);
+    const renderer = new FakeRenderer();
+    r.attachRenderer(renderer);
+    const a = new RenderedTarget(s, r);
+    a.renderer = renderer;
+    r.ioDevices.mouse.postData({
+        x: 0,
+        y: 0
+    });
+    t.equals(a.isTouchingObject('__mouse__'), true);
+    t.end();
+});
 
 test('isTouchingEdge', t => {
     const r = new Runtime();
--- a/test/unit/sprites_rendered-target.js
+++ b/test/unit/sprites_rendered-target.js
@@ -291,7 +291,7 @@ test('isTouchingObjectOutsideStage', t => {
         x: 1000,
         y: -300
     });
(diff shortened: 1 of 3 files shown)


test/unit/sprites_rendered-target.js
```diff
@@ -280,6 +280,22 @@ test('isTouchingPoint', t => {
     t.end();
 });
```
benjiwheeler, line 282: Thanks for adding this test. Could you also please add a test where the sprite _is_ touching the mouse?

apple502j: Not sure how to add that test?

adroitwhiz:
Why are you clamping the existing mouse position in `_scratchX` and `_scratchY` and adding additional `_rawScratchX` and `_rawScratchY` properties?

test/unit/sprites_rendered-target.js
adroitwhiz:
```js
test('isTouchingMouse', t => {
    const r = new Runtime();
    const s = new Sprite(null, r);
    const renderer = new FakeRenderer();
    r.attachRenderer(renderer);
    const a = new RenderedTarget(s, r);
    a.renderer = renderer;
    r.ioDevices.mouse.postData({
        x: 0,
        y: 0
    });
    t.equals(a.isTouchingObject('__mouse__'), true);
    t.end();
});
```

I think this will work because `FakeRenderer.drawableTouching` always returns `true` (haven't tried it though)

apple502j:
It seems like my Python knowledge is causing issues with underscored special names (see also #2427)

That should fix the test

PR #2314, opened 2020-01-30
Remove calls to updateDrawableSkinIdRotationCenter

adroitwhiz:
### Proposed Changes

This PR removes calls to `Renderer.updateDrawableSkinIdRotationCenter`, replacing them with calls to `Renderer.updateDrawableSkinId`.

It also fixes `loadBitmap_` to properly set bitmap costumes' rotation centers.

### Reason for Changes

Everywhere in the current codebase, updating a drawable's rotation center should do nothing.

It appears to be a vestige of the time when rotation centers belonged to `Drawable`s and not `Skin`s (the code to update drawables' rotation centers [was added Oct 9, 2016](https://github.com/LLK/scratch-vm/commit/9af9a87cb6d5dbc4d9408252f26d0b55650a396d), `Skin`s [were introduced Dec 28, 2016](https://github.com/LLK/scratch-render/commit/64c59914861a47162aee48a8e9f60854118617e8)).

Fixing a long-standing bug with vector costumes in `scratch-render` (see https://github.com/LLK/scratch-render/issues/550 for a detailed writeup) involves changes to the way vector costumes' rotation centers are set; notably, changing the rotation center requires re-loading the SVG. Supporting the ability to set a skin's rotation center after it is created makes the required `scratch-render` fix much harder to implement.

Furthermore, since the functionality to set a `Drawable`'s rotation center is currently effectively unused, changing the parts of `scratch-render` that involve rotation centers could break it unknowingly, causing future headaches if others try to use the API.

Finally, as I mentioned earlier, rotation centers now belong to `Skin`s (which correspond to costumes), so it doesn't make sense to consider them a "`Drawable` property".

There was one case in which updating a costume's rotation center actually did something: when first loading a bitmap, the costume rotation center was not divided by 2. The call to `updateDrawableSkinIdRotationCenter` when setting a target's costume, which did properly set the rotation center for bitmap costumes, hid this oversight.

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -212,7 +212,7 @@ const loadBitmap_ = function (costume, runtime, _rotationCenter) {
             return fetched;
         })
         .then(({canvas, mergeCanvas, rotationCenter}) => {
-            // createBitmapSkin does the right thing if costume.rotationCenter are undefined.
+            // createBitmapSkin does the right thing if costume.rotationCenter is undefined.
             // That will be the case if you upload a bitmap asset or create one by taking a photo.
             let center;
             if (rotationCenter) {
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -212,8 +212,17 @@ const loadBitmap_ = function (costume, runtime, _rotationCenter) {
             return fetched;
         })
         .then(({canvas, mergeCanvas, rotationCenter}) => {
-            // createBitmapSkin does the right thing if costume.bitmapResolution or rotationCenter are undefined...
-            costume.skinId = runtime.renderer.createBitmapSkin(canvas, costume.bitmapResolution, rotationCenter);
+            // createBitmapSkin does the right thing if costume.bitmapResolution is undefined...
+            // TODO: is rotationCenter ever undefined?
+            let center;
+            if (rotationCenter) {
+                center = [
+                    rotationCenter[0] / 2,
+                    rotationCenter[1] / 2
+                ];
+            }
+
+            costume.skinId = runtime.renderer.createBitmapSkin(canvas, costume.bitmapResolution, center);
             canvasPool.release(mergeCanvas);
             const renderSize = runtime.renderer.getSkinSize(costume.skinId);
             costume.size = [renderSize[0] * 2, renderSize[1] * 2]; // Actual size, since all bitmaps are resolution 2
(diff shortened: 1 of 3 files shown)


src/import/load-costume.js
```diff
@@ -212,8 +212,17 @@ const loadBitmap_ = function (costume, runtime, _rotationCenter) {
...
         .then(({canvas, mergeCanvas, rotationCenter}) => {
-            // createBitmapSkin does the right thing if costume.bitmapResolution or rotationCenter are undefined...
-            costume.skinId = runtime.renderer.createBitmapSkin(canvas, costume.bitmapResolution, rotationCenter);
+            // createBitmapSkin does the right thing if costume.bitmapResolution is undefined...
+            // TODO: is rotationCenter ever undefined?
+            let center;
+            if (rotationCenter) {
+                center = [
+                    rotationCenter[0] / 2,
+                    rotationCenter[1] / 2
```
adroitwhiz, line 223: Note that it's okay to always divide by 2 here-- [`_fetchBitmapCanvas` always ensures the rotation center is scaled up](https://github.com/LLK/scratch-vm/blob/8faad674396c906596e6a678c0c11fa4aa9f1333/src/import/load-costume.js#L154).

adroitwhiz:
Make sure you test this *after* merging https://github.com/LLK/scratch-render/pull/561 -- there are some existing rotation-center-related issues in the codebase that may interfere with testing.

src/import/load-costume.js
```diff
@@ -212,8 +212,17 @@ const loadBitmap_ = function (costume, runtime, _rotationCenter) {
             return fetched;
         })
         .then(({canvas, mergeCanvas, rotationCenter}) => {
-            // createBitmapSkin does the right thing if costume.bitmapResolution or rotationCenter are undefined...
-            costume.skinId = runtime.renderer.createBitmapSkin(canvas, costume.bitmapResolution, rotationCenter);
+            // createBitmapSkin does the right thing if costume.bitmapResolution is undefined...
+            // TODO: is rotationCenter ever undefined?
```
fsih: Yes, the rotation center is undefined if you upload your own bitmap asset, or if you make an asset by taking a photo.

src/import/load-costume.js
cwillisf: Could you add this as a comment in the code?

cwillisf (commented):
The code looks good to me... assigned to @fsih pending testing

fsih (approved):
Everything looks good!

PR #2321, opened 2020-02-10
Use newName argument when renaming list

### LINKED ISSUE #1415: Renaming list does not update monitor name
Title says it all.

### Steps to reproduce
1. Create a new list
2. Make sure monitor is enabled.
3. Rename list to new name.
4. Observe that monitor does not display correct name.

### LINKED ISSUE #4098: When a list's name is modified, the list monitor does not show the modified name
### Expected Behavior

After the list name is modified, the stage synchronizes the modified name

### Actual Behavior

When the list name is modified, the stage cannot synchronize the modified name

chrome

### LINKED ISSUE #5256: Change List Variable name not updated on stage.
### Expected Behavior

I defined a list called "MyLst",it showed in stage,then I changed this list to "MyList",but the name in variable monitor for this list didn't change, even if I refresh the page.

### Actual Behavior

The variable name in variable monitor should change to the new one.

### Steps to Reproduce
1. Define a list variable, name it "MyLst"
2. Change this variable name to "MyList"
3. Look at the variable name in list monitor on stage.

 Mac OS 10.15 Chrome 77.0.3865.90

apple502j:
### Resolves
Resolves LLK/scratch-gui#4098
Resolves LLK/scratch-gui#5256
Resolves LLK/scratch-blocks#1415

### Proposed Changes
In renameVariable function (in engine/target.js), if the params from `blocks._getBlockParams` has LIST item and it is different from the `newName` argument, it will use the `newName` argument. `blocks._getBlockParams` makes a new object every time it is called, so pollution (if any) should not be a problem.

### Reason for Changes
The previous code caused the monitor data passed to GUI to have old name, so the list monitor kept the old name. This bug is well-known and several topics have been made on BaG forums.

### Test Coverage
Manually tested:
1) Renaming normal variable should change the name on the monitor
2) Renaming list should change the name on the monitor

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -331,9 +331,13 @@ class Target extends EventEmitter {
                     }, this.runtime);
                     const monitorBlock = blocks.getBlock(variable.id);
                     if (monitorBlock) {
+                        const params = blocks._getBlockParams(monitorBlock);
+                        // _getBlockParams will return old name when the variable type is list.
+                        // Patch it so that the new name is passed.
+                        if (params.hasOwnProperty('LIST') && params.LIST !== newName) params.LIST = newName;
                         this.runtime.requestUpdateMonitor(Map({
                             id: id,
-                            params: blocks._getBlockParams(monitorBlock)
+                            params
                         }));
                     }
                 }


apple502j:
Closing because I found legitimate way to fix this :P

PR #2325, opened 2020-02-17
Fix old/incorrect list monitor names during deserialization

adroitwhiz:
### Resolves

This PR is the other half of #2322, and so it resolves the same issues.

### Proposed Changes

#2322 deals with renaming lists properly. This PR deals with deserializing lists that were previously not renamed properly.

This PR changes `deserializeMonitor` to set the `LIST` parameter of all `data_listcontents` monitors to the name of the list in the monitor's corresponding target.

### Reason for Changes

A previous bug resulted in the `LIST` parameter not being properly set for monitors.

### Test Coverage

An integration test has been added that loads a project with a previously-incorrectly-named list monitor and ensures that it is properly renamed.


fsih (changes requested):
Thanks!

Could you also add a test for a renamed for-this-sprite-only list?

fsih (approved):
LG, thanks!

fsih:
Note: I'm holding off on merging this until the next release cycle

PR #2203, opened 2019-06-08
Use new updateDrawable* methods

mzgoddard:
### Resolves

Depends on https://github.com/LLK/scratch-vm/pull/2196.
Depends on https://github.com/LLK/scratch-render/pull/469.

Make VM and Render faster.

### Proposed Changes

- Use updateDrawable* methods added in render PR

### Reason for Changes

> - Use updateDrawable* methods added in render PR

As mentioned in the other PR, using an active tense API in place of a passive tense API allows changes to propagate to Drawables much faster.

### Benchmark Data

Refer to tables in https://github.com/LLK/scratch-render/pull/469.

PR #2329, opened 2020-02-27
Remove remaining updateDrawableProperties calls

adroitwhiz:
### Resolves

Resolves the rest of https://github.com/LLK/scratch-render/issues/364

### Proposed Changes

This PR builds on #2203 and removes the rest of the `updateDrawableProperties` calls from the VM codebase.

### Reason for Changes

As explained in https://github.com/LLK/scratch-render/pull/469, the new, more specific `updateDrawable` methods do less unnecessary work, possibly improving performance.

Since the old deprecated `updateDrawableProperties` method is now unused, this PR allows it to be fully removed from the renderer, reducing the API surface and cleaning up the codebase.

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -178,23 +178,21 @@ class Scratch3LooksBlocks {
             bubbleState.onSpriteRight = true;
             this._renderBubble(target);
         } else {
-            this.runtime.renderer.updateDrawableProperties(bubbleState.drawableId, {
-                position: [
-                    bubbleState.onSpriteRight ? (
-                        Math.max(
-                            stageBounds.left, // Bubble should not extend past left edge of stage
-                            Math.min(stageBounds.right - bubbleWidth, targetBounds.right)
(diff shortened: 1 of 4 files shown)


fsih (approved):
LG, thanks!

cwillisf (approved):
😎

fsih:
Testing steps: There should be no changes from production
- Make a say or think bubble and move the sprite around and check that it exists and is drawing in the right position
- Draw something with the pen extension, it should appear
- Add the video extension and change video transparency. It should work
- Take a picture of yourself for a new sprite, and make sure the video turns off after you finish taking a picture, assuming you aren't running the video extension or a virtual meeting

adroitwhiz:
A note about the first step ("Make a say or think bubble and move the sprite around and check that it exists and is drawing in the right position"): bubbles will not move when you drag a sprite and will only update when you put it back down. This is already filed as https://github.com/LLK/scratch-gui/issues/1977 and does not need to be filed again.

PR #2330, opened 2020-02-29
Rename variables in sensing_of dropdown

### LINKED ISSUE #4392: Renaming a variable doesn't update existing "(var) of (sprite)" sensing blocks
### Expected Behavior

### Actual Behavior

### Steps to Reproduce

3. Rename the variable.

apple502j:
### Resolves
Resolves LLK/scratch-gui#4392

### Proposed Changes
Adds a new function, `updateVariableName`, similiar to existing `updateAssetName`. This checks for every block that
1) has opcode `sensing_of`
2) has old name in PROPERTY
3) has the same OBJECT
and changes the property. 

### Reason for Changes
Variable blocks will change their variable name by itself, but sensing_of does not. This will make sensing_of blocks referencing the variable return 0 when they are renamed.

### Test Coverage
Added type test and 3 tests:
1) updateVariableName normal behavior
2) updateVariableName different variable
3) updateVariableName different target

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -915,6 +915,33 @@ class Blocks {
     }
 
     /**
+     * Update blocks after a variable gets renamed.
+     * Any block referring to the old name of the asset should get updated
+     * to refer to the new name.
+     * @param {string} oldName The old name of the asset that was renamed.
+     * @param {string} newName The new name of the asset that was renamed.
+     * @param {string} targetName The name of the target the variable belongs to.
+     */
+    updateVariableName (oldName, newName, targetName) {
+        const blocks = this._blocks;
+        for (const blockId in blocks) {
+            const block = this.getBlock(blockId);
+            if (block &&
+                block.fields.hasOwnProperty('PROPERTY') &&
+                block.opcode === 'sensing_of' &&
+                block.fields.PROPERTY.value === oldName &&
+                block.inputs.hasOwnProperty('OBJECT')) {
+                const menuFields = this.getBlock(block.inputs.OBJECT.shadow).fields;
+                if (
+                    menuFields.hasOwnProperty('OBJECT') &&
+                    menuFields.OBJECT.value === targetName) {
+                    block.fields.PROPERTY.value = newName;
+                }
+            }
+        }
+    }
+
+    /**
      * Update blocks after a sound, costume, or backdrop gets renamed.
      * Any block referring to the old name of the asset should get updated
      * to refer to the new name.
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -322,6 +322,12 @@ class Target extends EventEmitter {
                         this.runtime.ioDevices.cloud.requestRenameVariable(oldName, newName);
                     }
 
+                    if (variable.type === '') {
+                        this.runtime.targets.forEach(t => t.blocks.updateVariableName(
+                            oldName, newName, this.isStage ? '_stage_' : this.getName()));
+                        this.runtime.requestBlocksUpdate();
+                    }
+
                     const blocks = this.runtime.monitorBlocks;
                     blocks.changeBlock({
                         id: id,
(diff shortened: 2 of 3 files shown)


src/engine/blocks.js
```diff
@@ -914,6 +914,33 @@ class Blocks {
...
+     * @param {string} oldName The old name of the asset that was renamed.
+     * @param {string} newName The new name of the asset that was renamed.
+     * @param {string} targetName The name of the target the variable belongs to.
+     */
+    updateVariableName (oldName, newName, targetName) {
+        const blocks = this._blocks;
+        for (const blockId in blocks) {
+            const block = this.getBlock(blockId);
+            if (block &&
+                block.fields.hasOwnProperty('PROPERTY') &&
```
adroitwhiz, line 930:
```suggestion
            if (!Object.prototype.hasOwnProperty.call(blocks, blockId)) continue;
            const block = blocks[blockId];
            if (block.fields.hasOwnProperty('PROPERTY') &&
```

I think this is a bit easier to follow:
- `getBlock` is just a lookup on `this._blocks` which we're already iterating over.
- Instead of checking that `block` exists, check if `this._blocks` has the block ID.

Unless there's a scenario in which `block` exists and is null?

src/engine/blocks.js
```diff
@@ -914,6 +914,33 @@ class Blocks {
...
+                block.opcode === 'sensing_of' &&
```
adroitwhiz, line 931: Also may be clearer to check the opcode first?

src/engine/target.js
```diff
@@ -322,6 +322,12 @@ class Target extends EventEmitter {
                         this.runtime.ioDevices.cloud.requestRenameVariable(oldName, newName);
                     }
 
+                    if (variable.type === '') {
```
adroitwhiz, line 325:
```suggestion
                    if (variable.type === Variable.SCALAR_TYPE) {
```

adroitwhiz: Also add a note that we're only doing this for variables because lists don't exist in the "sensing_of" dropdown.

Although `updateVariableName` could *theoretically* be expanded in the future to blocks which could reference lists...  And currently, there's nothing in `updateVariableName`'s documentation that indicates that it only does things to `sensing_of`blocks.

src/engine/blocks.js
fsih, line 931: Can we move this one line up?

src/engine/blocks.js
```diff
@@ -914,6 +914,33 @@ class Blocks {
...
+                block.fields.PROPERTY.value === oldName &&
+                block.inputs.hasOwnProperty('OBJECT')) {
```
adroitwhiz, line 933: When do you expect the `PROPERTY` and `OBJECT` fields/inputs to not exist?

src/engine/blocks.js
```diff
@@ -914,6 +914,33 @@ class Blocks {
...
+                const menuFields = this.getBlock(block.inputs.OBJECT.shadow).fields;
+                if (
+                    menuFields.hasOwnProperty('OBJECT') &&
+                    menuFields.OBJECT.value === targetName) {
```
adroitwhiz, line 937: It looks like this will still change the variable name even if a block is dropped into the selector menu:
Peek 2020-05-21 15-56

src/engine/target.js
```diff
@@ -322,6 +322,12 @@ class Target extends EventEmitter {
...
+                        this.runtime.targets.forEach(t => t.blocks.updateVariableName(
+                            oldName, newName, this.isStage ? '_stage_' : this.getName()));
```
adroitwhiz, line 327: Is this even further coupling together `updateVariableName` and `sensing_of`?

IIRC, `_stage_` is a "magic value" that only makes sense in dropdowns. If `updateVariableName` is really a "black-box" function meant to simply "Update blocks after a variable gets renamed" then does it make sense to pass in a "dropdown-ified" target name?

apple502j:
Gonna open this again once I've fixed everything

PR #2335, opened 2020-03-14
WeDo 2.0 isTilted-any function bug fix

### LINKED ISSUE #2334: WeDo 2.0 isTilted-any function fails the threshold comparison
### Expected Behavior

Only when the sensor steeply tilts (which means that the tilt sensor value is more than TILT_THRESHOLD=15), the following behaviors are expected:
- _isTilted function returns true
- "when tilted any"-block is triggered
- "tilted any ?"-block returns true

### Actual Behavior

Even when the tilt sensor value is less than TILT_THRESHOLD=15, the following behaviors happen:
- _isTilted function incorrectly returns true
- "when tilted any"-block is incorrectly triggered
- "tilted any ?"-block incorrectly returns true

### Steps to Reproduce

1. Add WeDo 2.0 extension on Scratch
2. Connect WeDo 2.0 Hub and tilt sensor
3. Put "when tilted any"-block and "tilted any ?"-block on Scratch
4. Change the tilt sensor's angle slightly like 1-5 degrees
5. See whether the "when tilted any"-block is triggered or not, and "tilted any ?"-block returns true or not

All Operating System and Browser with WeDo 2.0 tilt sensor

daikifukumori:
### Resolves

Resolves #2334

### Proposed Changes

Adds the value conversion before the threshold comparison

### Reason for Changes

The values of tiltX and tiltY should not be compared with the threshold(TILT_THRESHOLD) directly.
Because the tiltX and tiltY have the value 255 for 1 degree tilt, 254 for 2 degrees tilt for example. Therefore, the value should be converted like (256 - tiltX) before the comparison.

### Test Coverage

Manually tested:
1. "when tilted any"-block is triggered only when the tilt angle is more than the threshold(TILT_THRESHOLD=15)
2. "tilted any ?"-block returns true only when the tilt angle is more than the threshold(TILT_THRESHOLD=15)

### DIFF
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1543,10 +1543,10 @@ class Scratch3WeDo2Blocks {
     _isTilted (direction) {
         switch (direction) {
         case WeDo2TiltDirection.ANY:
-            return (this._peripheral.tiltX > 45 ? 256 - this._peripheral.tiltX : this._peripheral.tiltX) >=
-                Scratch3WeDo2Blocks.TILT_THRESHOLD ||
-                (this._peripheral.tiltY > 45 ? 256 - this._peripheral.tiltY : this._peripheral.tiltY) >=
-                Scratch3WeDo2Blocks.TILT_THRESHOLD;
+            return this._getTiltAngle(WeDo2TiltDirection.UP) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
+                this._getTiltAngle(WeDo2TiltDirection.DOWN) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
+                this._getTiltAngle(WeDo2TiltDirection.LEFT) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
+                this._getTiltAngle(WeDo2TiltDirection.RIGHT) >= Scratch3WeDo2Blocks.TILT_THRESHOLD;
         default:
             return this._getTiltAngle(direction) >= Scratch3WeDo2Blocks.TILT_THRESHOLD;
         }
--- a/src/extensions/scratch3_wedo2/index.js
+++ b/src/extensions/scratch3_wedo2/index.js
@@ -1543,8 +1543,10 @@ class Scratch3WeDo2Blocks {
     _isTilted (direction) {
         switch (direction) {
         case WeDo2TiltDirection.ANY:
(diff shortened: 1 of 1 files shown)


apple502j:
Lint is falling (max-len)

src/extensions/scratch3_wedo2/index.js
```diff
@@ -1543,8 +1543,10 @@ class Scratch3WeDo2Blocks {
     _isTilted (direction) {
         switch (direction) {
         case WeDo2TiltDirection.ANY:
-            return (Math.abs(this._peripheral.tiltX) >= Scratch3WeDo2Blocks.TILT_THRESHOLD) ||
-                (Math.abs(this._peripheral.tiltY) >= Scratch3WeDo2Blocks.TILT_THRESHOLD);
+            return (this._peripheral.tiltX > 45 ? 256 - this._peripheral.tiltX : this._peripheral.tiltX) >=
```
fsih: Can we make a named constant for what 45 and 256 are?
Also, can you add comments explaining the bug to this code?

Pulling `this._peripheral.tiltX > 45 ? 256` into a boolean variable might also make this more readable

fsih (changes requested):

src/extensions/scratch3_wedo2/index.js
daikifukumori: I agree with you, it's not easy-to-read. On second thought, as the two constants(45, 256) come from other function [_getTiltAngle()](https://github.com/LLK/scratch-vm/blob/7af161f1a2e4cadd02d747d77f6cea0299937581/src/extensions/scratch3_wedo2/index.js#L1559), I think it would be better the tilt value conversion is calculated in the single point where is the function _getTiltAngle(). So can I suggest changing these lines as follows so that the complicated conversion is not needed in _isTilted() anymore?

``` JavaScript
return this._getTiltAngle(WeDo2TiltDirection.UP) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
    this._getTiltAngle(WeDo2TiltDirection.DOWN) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
    this._getTiltAngle(WeDo2TiltDirection.LEFT) >= Scratch3WeDo2Blocks.TILT_THRESHOLD ||
    this._getTiltAngle(WeDo2TiltDirection.RIGHT) >= Scratch3WeDo2Blocks.TILT_THRESHOLD;

fsih: Ah, I see, that function doesn't explain what those numbers are either. Yes, I think that would make a lot of sense!

fsih (approved):
This looks great, thanks!

PR #2353, opened 2020-04-22
Decrement clone counter only if it's clone

### LINKED ISSUE #505: Clone limit not being enforced properly
### Expected Behavior

- VM should limit number of clones for an individual sprite to 301

### Actual Behavior

- VM limits number of clones for an individual sprite to 302

### Steps to Reproduce

- https://scratch.mit.edu/projects/151234984/
- https://llk.github.io/scratch-gui/#151234984

MacOS 10.12.3  |  Chrome Version 57.0.2987.98 (64-bit)

### LINKED ISSUE #2352: clone counter bug: become minus value when delete a sprite from stage
### Expected Behavior

As we know, the max clones number is 300 and the value is hardcoded in runtime.js
static get MAX_CLONES () { 
        return 300;                                                                                               
}

So the expected behavior is that the max clones number of all sprites in stage should be 300.

### Actual Behavior
Actually, sometimes the max clones number exceeds 300.  The root cause is that when add anohter sprite into stage, then delete that sprite, the cloneCounter becomes minus value. In below code, dispose function is called when removing a sprite from stage, then cloneCounter reduced by calling runtime.changeCloneCounter. Thus changing the value below zero, and then total clones number exceed 300.
/**  
     * Dispose, destroying any run-time properties.
     */
    dispose () { 
        this.runtime.changeCloneCounter(-1);  
        ......
   }

### Steps to Reproduce
1. Create a global variable to record the clones number, say num.
2. Add a sprite named apple. Then add code blocks (pseudocode below):
when click green flag
      loop forever {
            make clone apple sprite
      }

when start as clone body
      set num = num + 1

Run program by clicking green flag, Tick on num and in stage we can see variable num shows 300.
Then stop the program.
3. Add another sprite into stage, say balloon, then delete balloon from stage. 
4. Run program  again, the variable num show 301.
5. Repeat step 3 by adding another sprite into stage and delete it. Then run program again, the max clones num become 302.
6. Repeat again step 3, step 4,  num will become bigger and bigger.

Any.

This bug is not harmful, but make people confused.

apple502j:
### Resolves
Resolves #505
Resolves #2352 

### Proposed Changes
Decrement clone counter only if it's clone

### Reason for Changes
`RenderedTarget.dispose` is called when the target gets disposed somehow, including deleting a sprite or its clone. This makes the clone counter go below 0.

### DIFF
--- a/src/sprites/rendered-target.js
+++ b/src/sprites/rendered-target.js
@@ -1114,7 +1114,9 @@ class RenderedTarget extends Target {
      * Dispose, destroying any run-time properties.
      */
     dispose () {
-        this.runtime.changeCloneCounter(-1);
+        if (!this.isOriginal) {
+            this.runtime.changeCloneCounter(-1);
+        }
         this.runtime.stopForTarget(this);
         this.runtime.removeExecutable(this);
         this.sprite.removeClone(this);


adroitwhiz:
Any reason not to remove the decrement inside `Sprite.removeClone` (specifically, inside the `if (cloneIndex >= 0)` block, since `removeClone` is called regardless of whether sprites are clones)?

PR #2356, opened 2020-05-01
Cast malformed color hex strings to RGBA 0,0,0,255 instead of null

### LINKED ISSUE #2355: Placing a string beginning with "#" in a block's color input causes the VM to stop running scripts
### Expected Behavior

You should never be able to cause the VM to error out and stop executing scripts simply by executing normal blocks

### Actual Behavior

If you drop any reporter returning a string beginning with the "#" character into a block's color input, and then execute that block, an error appears in the browser console and all scripts will stop executing until the green flag is pressed again.

### Steps to Reproduce

1. Drag a "touching color ()" block into the workspace
2. Drag a "join () ()" block into the "touching color" block's color input
3. Set the first text field of the "join" block to "#"
4. Set the second text field of the "join" block to some other string, e.g. "test" or "0"
5. Click the "touching color ()" block to execute it
6. Observe that any other running scripts no longer take effect, and an error appears in the browser console.

You can do this with any block that has a color input-- for instance, "touching color ()", "color () is touching color ()", and "set pen color to ()".

image

All

adroitwhiz:
### Resolves

Resolves #2355

### Proposed Changes

This PR changes the behavior of `Cast.toRgbColorObject` to cast strings beginning with `#` to RGBA [0, 0, 0, 255] (opaque black), matching the behavior of casting other malformed strings to a color object.

Previously, those strings were cast to `null`.

### Reason for Changes

This prevents the VM from erroring out when attempting to cast a string beginning with `#` to a color, as it expects an RGB color object to always be returned.

### Test Coverage

A "cast malformed hex string to color" unit test has been added.

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -93,6 +93,9 @@ class Cast {
         let color;
         if (typeof value === 'string' && value.substring(0, 1) === '#') {
             color = Color.hexToRgb(value);
+
+            // If the color wasn't *actually* a hex color, cast to black
+            if (!color) color = {r: 0, g: 0, b: 0, a: 255};
         } else {
             color = Color.decimalToRgb(Cast.toNumber(value));
         }
--- a/test/unit/util_cast.js
+++ b/test/unit/util_cast.js
@@ -73,7 +73,7 @@ test('toString', t => {
     t.end();
 });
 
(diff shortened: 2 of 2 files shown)


fsih (approved):
Thanks!

PR #2400, opened 2020-05-12
Remove BlockCached._shadowOps

### LINKED ISSUE #2398: What is BlockCached._shadowOps?
### Expected Behavior

Code in the codebase should... do something

### Actual Behavior

The `BlockCached` class has a `_shadowOps` property, which is an array that is pushed to in 3 different places [across these lines](https://github.com/LLK/scratch-vm/blob/develop/src/engine/execute.js#L275-L378), but, as far as I can tell, never read from.

What is this supposed to do? How was it intended to be used? Does it actually do something that I'm missing?

adroitwhiz:
### Resolves

Resolves #2398

### Proposed Changes

This PR removes the `_shadowOps` property from `BlockCached`, along with all code that pushes to it.

### Reason for Changes

`_shadowOps` was never read from, only pushed to, and [shadow values are stored elsewhere anyways](https://github.com/LLK/scratch-vm/blob/f5ce73ae51bcc634a66b3924f639e06b626fc16e/src/engine/execute.js#L367-L371).

### Test Coverage

All existing tests pass

### DIFF
--- a/src/engine/execute.js
+++ b/src/engine/execute.js
@@ -268,13 +268,6 @@ class BlockCached {
         this._parentValues = null;
 
         /**
-         * A sequence of shadow value operations that can be performed in any
-         * order and are easier to perform given that they are static.
-         * @type {Array<BlockCached>}
-         */
-        this._shadowOps = [];
-
-        /**
          * A sequence of non-shadow operations that can must be performed. This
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
👍 approved pending CI pass

PR #2407, opened 2020-05-13
Revert "Set canvas and costume size to zero if bitmap is empty"

### LINKED ISSUE #2307: Remove 0x0 hack in updateBitmap
`ImageData` type of objects cannot have a width/height of zero, so together with https://github.com/LLK/scratch-paint/pull/932 we can remove this hack https://github.com/LLK/scratch-vm/blob/f31c630f9d195aafdcfdfb47542ad931e184e00a/src/virtual-machine.js#L881-L883

/cc @kchadha @fsih

adroitwhiz:
### Resolves

Resolves #2307

### Proposed Changes

This PR removes the `sourceWidth`/`sourceHeight` hack from `VirtualMachine.updateBitmap`.

### Reason for Changes

Since https://github.com/LLK/scratch-paint/pull/932 was merged, it is no longer necessary to check whether the bitmap is 0x0.

`sourceWidth` and `sourceHeight` aren't actual properties of `ImageData` anyways-- they were being set by https://github.com/LLK/scratch-paint/pull/826/ and that code has since been removed.

### DIFF
--- a/src/virtual-machine.js
+++ b/src/virtual-machine.js
@@ -866,13 +866,10 @@ class VirtualMachine extends EventEmitter {
         costume.rotationCenterX = rotationCenterX;
         costume.rotationCenterY = rotationCenterY;
 
-        // If the bitmap originally had a zero width or height, use that value
-        const bitmapWidth = bitmap.sourceWidth === 0 ? 0 : bitmap.width;
-        const bitmapHeight = bitmap.sourceHeight === 0 ? 0 : bitmap.height;
         // @todo: updateBitmapSkin does not take ImageData
         const canvas = document.createElement('canvas');
-        canvas.width = bitmapWidth;
-        canvas.height = bitmapHeight;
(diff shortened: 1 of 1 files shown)


src/virtual-machine.js
```diff
@@ -904,10 +901,7 @@ class VirtualMachine extends EventEmitter {
                 costume.md5 = `${costume.assetId}.${costume.dataFormat}`;
                 this.emitTargetsUpdate();
             });
-            // Bitmaps with a zero width or height return null for their blob
-            if (blob){
-                reader.readAsArrayBuffer(blob);
-            }
```
fsih, line 910: Assert that bitmap width and height should not be 0 in case this regresses?

adroitwhiz:
@fsih I reverted the assertion because `bitmap` will always be an `ImageData`, and you can't construct an `ImageData` object with a width/height of 0 (it'll throw an `IndexSizeError`).

PR #2409, opened 2020-05-15
broadcastAndWait now always executes if there are threads to wait

### LINKED ISSUE #2109: "broadcast (expr) and wait" ends if value of (expr) changes
### Expected Behavior

Broadcast and wait block should always wait until its receiver scripts finish.

### Actual Behavior

If the broadcast block contained an expression, changing the value of that expression (e.g. by changing a var it contains) causes the broadcast-and-wait block to end even if its receivers have not yet finished.

### Steps to Reproduce

Create a script along these lines:
`when GF clicked`
`say "hello"`
`set myvar to "test"`
`broadcast (myvar) and wait`
`say "goodbye"`

And a receiver script:
`when I receive "test"`
`wait 1 sec`
`set myvar to "changed"`
`forever {nothing}`

Click GF and note that it says "hello", then 1 sec later says "goodbye" – and yet the receiver script continues to run the forever loop.

Note if you add more work after the "b-a-w" in the first script, it means that *both* scripts are now running simultaneously. Interesting new feature, that...  :D

Presumably N/A

apple502j:
### Resolves
Resolves #2109

### Proposed Changes
broadcastAndWait now always executes if there are threads to wait.

### Reason for Changes
The `args` could change during the execution of this block, and and `broadcastVar` may become null. Previously, if it was null, it's a instant return - even if the block already has threads to wait. 

### Test Coverage
manually tested


adroitwhiz:
Couldn't you store `broadcastVar` on the stack frame instead?

aeons:
Is there anything we can do to help this get merged?

apple502j:
@aeons Then please review this PR!

aeons (approved):
I have tested this in our app that uses Scratch.

It fixes the issues we have seen and performs as expected.

aeons:
Any update on this?

apple502j:
@aeons Sadly Scratch team now requires us to only submit Pull Requests for issues with priority label. after @BryceLTaylor adds priority label devs will review it.

aeons:
Alright, thanks.

ericrosenbaum:
@apple502j we would like to have tests for this change before a full review. Thanks for your patience on this - Let us know if you aren't able to write the tests yourself, and we can work out another strategy. if @aeons is interested he might be able to help (I don't think he can push to this branch but could suggest test code in a comment here maybe?). @kchadha suggested to look here: https://github.com/LLK/scratch-vm/tree/develop/test/fixtures/execute

apple502j:
@benjiwheeler  Okay I think I can add a test.

apple502j:
@benjiwheeler Test added. CI failed but I don't think it's related (different file anyway) - maybe retry?

ericrosenbaum:
restarted CI and it passed!

ericrosenbaum:
Hi folks- I've now done some investigation along with @kchadha, and here's where we are so far:
- We decided to investigate the approach suggested by @adroitwhiz (storing broadcastVar on the stackFrame), and found that the resulting code changes were less complex and felt less risky. 
- Execute test fixture projects can only be sb2 files, it turns out, so we had to recreate the one from @apple502j as an sb2
- Also, these test projects have a short execution time limit, but I found I could get it to pass by reducing the wait times to under 1 second (I made a few other small changes as well)
- I'm going to open a new PR (with credit to those who have come before!).
- I plan to do some more manual testing, and also learn more about the lifecycle and behavior of stackFrame to make sure this strategy makes sense

ericrosenbaum:
replacing this with https://github.com/LLK/scratch-vm/pull/2557

PR #2416, opened 2020-05-18
Remove Timer.relativeTime

### LINKED ISSUE #2401: Timer.relativeTime is unused
[Timer.relativeTime](https://github.com/LLK/scratch-vm/blob/develop/src/util/timer.js#L78) is not used [anywhere](https://github.com/LLK/scratch-vm/search?q=relativetime&unscoped_q=relativetime) (except for test mocks). Also, this code is the same as `Timer.time` 😛 

We should remove unnecessary code to make files smaller.

adroitwhiz:
### Resolves

Resolves #2401

### Proposed Changes

This PR removes the `Timer.relativeTime` function.

### Reason for Changes

`Timer.relativeTime` contains the exact same code as `Timer.time`, is not used anywhere in the current codebase, and as such is dead code.

### Test Coverage

Also removed it from the `MockTimer` class.

### DIFF
--- a/src/util/timer.js
+++ b/src/util/timer.js
@@ -8,7 +8,7 @@
  * ... pass some time ...
  * var timeDifference = timer.timeElapsed();
  * ---
- * Or, you can use the `time` and `relativeTime`
+ * Or, you can use the `time` function
  * to do some measurement yourself.
  */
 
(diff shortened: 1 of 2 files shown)


fsih (approved):
The commit that took this out says "Disable use of more accurate timer temporarilly due to performance." I guess we never felt we needed the extra accuracy again?

adroitwhiz:
> I guess we never felt we needed the extra accuracy again?

That and I'm not sure if `relativeTime` was ever used (it's currently not), meaning that even if `relativeTime` was more accurate, there doesn't appear to be any place in the codebase that needs the extra accuracy?

[According to MDN, performance.now's precision was limited as a Spectre mitigation anyway](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now), which might also affect things.

PR #2419, opened 2020-05-20
Don't emit PROJECT_RUN_START for edge activation check thread

### LINKED ISSUE #1948: Edge-triggered(?) hat blocks make the green flag always look active
### Expected Behavior

Edge-triggered (extension?) hat blocks shouldn't behave any differently from ordinary hat blocks.

### Actual Behavior

Dragging out any hat block from extensions makes it so that the green flag looks active, as though the project is constantly running, even if you click the stop button. (The buttons do what they're supposed to when clicked, but the flag constantly appears active.)

### Steps to Reproduce

Drag out a hat block from an extension, like "when button (_) pressed" from LEGO EV3, or "when (...) pressed in order" from Makey Makey.

I can also repro this with "when (loudness/timer) > (N)", so I think it might be a more general issue related to edge-triggered blocks (if that's the right term...I've never worked with hat blocks much).

Firefox 65.0b11, Linux (Debian). I can repro this on scratch.mit.edu as well as llk.github.io/develop.

---

I think this might be what the user here is seeing: https://scratch.mit.edu/discuss/topic/337846/

apple502j:
### Resolves
Resolves #1948

### Proposed Changes
- Counts number of threads that are edge-activated
- Exclude the threads from PROJECT_RUN_START checks

### Reason for Changes
All edge-activated hats will be executed once per tick at `_step`, triggering PROJECT_RUN_START, which makes the green flag GUI "stuck" at active state. This is also compatibility issue; Scratch 2.0 didn't have this problem.

### Test Coverage
tested in vivo(aka in my browser)

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -254,6 +254,12 @@ class Runtime extends EventEmitter {
         this._nonMonitorThreadCount = 0;
 
         /**
+         * Number of threads of edge-activated blocks that executes condition check.
+         * @type {number}
+        */
+        this._edgeActivatedThreadCount = 0;
+
+        /**
          * All threads that finished running and were removed from this.threads
          * by behaviour in Sequencer.stepThreads.
          * @type {Array<Thread>}
@@ -1815,6 +1821,10 @@ class Runtime extends EventEmitter {
         // For compatibility with Scratch 2, edge triggered hats need to be processed before
         // threads are stepped. See ScratchRuntime.as for original implementation
         newThreads.forEach(thread => {
+            // This is used to count how many edge-activated block condition check threads will exist
+            if (hatMeta.edgeActivated) {
+                this._edgeActivatedThreadCount++;
+            }
             execute(this.sequencer, thread);
             thread.goToNextBlock();
         });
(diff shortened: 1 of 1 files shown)


Wowfunhappy:
@apple502j One issue I noticed—while this pull request fixes https://github.com/LLK/scratch-vm/issues/1948 in the general case, the problem remains if the program was stopped via a `stop all` block.

Screen Shot 2023-03-08 at 12 41 23 PM

Test project attached:
[Pull 2419 Issue.sb3.zip](https://github.com/LLK/scratch-vm/files/10923332/Pull.2419.Issue.sb3.zip)

adroitwhiz:
I think this also just doesn't count *any* scripts started by edge-activated hat blocks as "running".

Wowfunhappy:
> I think this also just doesn't count _any_ scripts started by edge-activated hat blocks as "running".

No, that part seems to work for me, at least in terms of the active state of the green flag. When an edge-activated hat is triggered, the green flag becomes active until the script finishes, at which point the flag becomes inactive. It's just `stop all` that seems to screw things up.

apple502j:
Well, updating 2.5 years old PR definitely sounds like a good idea... I'll try

Wowfunhappy:
@apple502j Before you do, note that @adroitwhiz just made a separate fix: https://github.com/LLK/scratch-vm/pull/3912

PR #2425, opened 2020-05-22
Update sensing_of fields if variable gets renamed

### LINKED ISSUE #4392: Renaming a variable doesn't update existing "(var) of (sprite)" sensing blocks

apple502j:
Thanks @adroitwhiz and @fsih for some advice! (This is basically re-work of #2330)

### Resolves
Resolves LLK/scratch-gui#4392

### Proposed Changes
Adds a new function, `updateSensingOfReference`, similiar to existing `updateAssetName`. This checks for every block that
1) has opcode `sensing_of`
2) has old name in PROPERTY
3) has the same OBJECT
4) doesn't have blocks inserted to OBJECT
and changes the property. 

### Reason for Changes
Variable blocks will change their variable name by itself, but sensing_of does not. This will make sensing_of blocks referencing the variable return 0 when they are renamed.

### Test Coverage
Added type test and 4 tests:
1) updateSensingOfReference normal behavior
2) updateSensingOfReference block inserted
3) updateSensingOfReference different variable
4) updateSensingOfReference different target

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -947,6 +947,32 @@ class Blocks {
     }
 
     /**
+     * Update sensing_of blocks after a variable gets renamed.
+     * @param {string} oldName The old name of the variable that was renamed.
+     * @param {string} newName The new name of the variable that was renamed.
+     * @param {string} targetName The name of the target the variable belongs to.
+     * @return {boolean} Returns true if any of the blocks were updated.
+     */
+    updateSensingOfReference (oldName, newName, targetName) {
+        const blocks = this._blocks;
+        let blockUpdated = false;
+        for (const blockId in blocks) {
+            const block = blocks[blockId];
+            if (block.opcode === 'sensing_of' &&
+                block.fields.PROPERTY.value === oldName &&
+                // If block and shadow are different, it means a block is inserted to OBJECT, and should be ignored.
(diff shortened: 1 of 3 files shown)


src/engine/blocks.js
```diff
@@ -946,6 +946,32 @@ class Blocks {
...
+     * Update sensing_of blocks after a variable gets renamed.
+     * @param {string} oldName The old name of the variable that was renamed.
+     * @param {string} newName The new name of the variable that was renamed.
+     * @param {string} targetName The name of the target the variable belongs to.
+     * @return {boolean} Returns true if any of the blocks were updated.
+     */
+    updateSensingOfReference (oldName, newName, targetName) {
+        const blocks = this._blocks;
+        let blockUpdated = false;
+        for (const blockId in blocks) {
```
adroitwhiz, line 959: I think this is missing a `hasOwnProperty` check

cwillisf (approved):
These changes look good! In the future, it would help us out if you could just update the existing PR instead of making a new one when changes are requested. Thanks!

apple502j:
@cwillisf I tried to do that (and TIL force-push won't work well with closed PR!)

PR #2427, opened 2020-05-24
Fix sensing_of when renaming global variables

### LINKED ISSUE #2435: Global variable names are not updated in '(var) of (sprite)' block when renamed
### Expected Behavior
Subset of https://github.com/LLK/scratch-gui/issues/4392
When a variable is renamed, all occurrences of that variable in the code on the workspace should update

### Actual Behavior
When a global variable is renamed, the variable name is not updated in occurrences of the (var) of (sprite) block, which will cause the block to start reporting '0' instead of the variable's value

### Steps to Reproduce
1. Create a global variable (for all sprites) named A
2. Change the value of A
3. Drag out a (var) of (sprite) block, and set the sprite to stage, and var to A. The block should report the value of A.
4. Rename A
5. The block does not update, and now reports 0 for the nonexistant variable name

all

apple502j:
Resolves: https://github.com/LLK/scratch-vm/issues/2435

Yikes, I did the mistake :P

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -330,7 +330,7 @@ class Target extends EventEmitter {
                             blockUpdated = t.blocks.updateSensingOfReference(
                                 oldName,
                                 newName,
-                                this.isStage ? '__stage__' : this.getName()
+                                this.isStage ? '_stage_' : this.getName()
                             ) || blockUpdated;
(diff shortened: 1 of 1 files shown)


fsih (approved):
Thanks!

PR #2437, opened 2020-05-28
Set graphic effect to zero if it's Infinity

### LINKED ISSUE #2317: Graphic Effects can have Infinity as argument
### Expected Behavior
Infinity should be handled as 0 or max value of the effect
### Actual Behavior
Giving Infinity as argument causes unexpected graphic effect such as grayscale sprite (when color effect is used)
### Steps to Reproduce
Run `set color effect to (1/0)`.

Any

apple502j:
### Resolves
Resolves #2317 

### Proposed Changes
If the graphic value is not finite (for example `Infinity`), cast it to zero, if it's not already clamped

### Reason for Changes
Some effects become strange if the value is Infinity. I chose zero because 1) all effects should accept 0 as a value and 2) not all effects have maximum/minimum values. (For those that have it (currently ghost and brightness), this change will NOT apply)


fsih (approved):
LG! Thank you for your help!

PR #2441, opened 2020-05-28
Reset blocks cache when sensing_of blocks are updated

### LINKED ISSUE #4392: Renaming a variable doesn't update existing "(var) of (sprite)" sensing blocks

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-gui/issues/4392

### Proposed Changes

This PR adds code to `updateSensingOfReference` to clear the block container's cache if `sensing_of` blocks are updated.

### Reason for Changes

Previously, the block cache wasn't getting cleared, which meant that the old `sensing_of` blocks were being cached when they shouldn't have been, causing them to report the wrong values until they were updated by clicking on the sprites they belonged to.

### Test Coverage

Tested manually

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -969,6 +969,7 @@ class Blocks {
                 }
             }
         }
+        if (blockUpdated) this.resetCache();
         return blockUpdated;
     }
 


fsih:
@fsih Todo: look at https://github.com/LLK/scratch-vm/pull/770

fsih (approved):
LGTM! Thanks!

PR #2442, opened 2020-05-28
Support extension block filtering per-target

adroitwhiz:
### Resolves

A step towards https://github.com/LLK/scratch-gui/issues/2174

### Proposed Changes

This PR implements per-target filtering of extension blocks to display in the blocks palette.

It does this by adding an optional parameter, the current editing target, to `Runtime.getBlocksXML`. If this parameter is passed, `getBlocksXML` will filter out extension blocks using the block info's `filter` property, per the [extension spec](https://github.com/LLK/scratch-vm/blob/develop/docs/extensions.md).

### Reason for Changes

This is necessary to change which blocks get displayed when you switch targets (e.g. to hide certain pen blocks when you select the stage).

### Test Coverage

Tested manually

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1365,13 +1365,27 @@ class Runtime extends EventEmitter {
 
     /**
      * @returns {Array.<object>} scratch-blocks XML for each category of extension blocks, in category order.
+     * @param {?Target} [target] - the active editing target (optional)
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
👍
I'm glad this turned out to be relatively easy to implement -- thanks for figuring out the missing piece!

PR #2450, opened 2020-05-31
Use toUpperCase at setVoice

### LINKED ISSUE #2448: Set Voice To input is droppable but expected inputs don't match expectations
### Expected Behavior

If you drop a variable reporter into the input for the block "Set Voice To" then if you input the names of the voices as they appear on the blocks, then it should change the voice to that name.

Alternatively accepting non-case sensitive strings of the names would make sense.

### Actual Behavior

Since it accepts the EXACT id string used by the vm, which is case sensitive, it accepts the value "GIANT" and "TENOR" but not "giant" and "tenor" despite the fact that the dropdown uses lower case names for the voices.  

This came from a forum post: https://scratch.mit.edu/discuss/topic/408699/

### Steps to Reproduce

Add the text to speech extension to a project.  
Create this stack:
Screen Shot 2020-05-29 at 10 36 01 AM
Note: the dropdown menu shows all names of the voices in lower case:
Screen Shot 2020-05-29 at 10 35 49 AM
But only if you put the names in all caps does it accept the inputs.

Mac Chrome (almost certainly others.)

BoomerScratch:
### Resolves

Resolves #2448

### Proposed Changes

Adds ".toUpperCase()" to the text to speech setVoice(), so lowercase works.

### Reason for Changes

So, for example, this will work:
image

### Test Coverage

Manually tested with a clone of the GUI.

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -648,7 +648,7 @@ class Scratch3Text2SpeechBlocks {
     setVoice (args, util) {
         const state = this._getState(util.target);
 
-        let voice = args.VOICE;
+        let voice = args.VOICE.toUpperCase();
 
         // If the arg is a dropped number, treat it as a voice index
         let voiceNum = parseInt(voice, 10);


adroitwhiz:
Note @ericrosenbaum's [comment](https://github.com/LLK/scratch-vm/issues/2448#issuecomment-636014802) in the corresponding issue:

> If we wanted to turn it into a feature, we would want it to be case-insensitive, but it would also need to work every language. I don't think we have an example yet of using the localized menu item strings as data in the VM. I'd be curious to learn if we can easily do that.

src/extensions/scratch3_text2speech/index.js
```diff
@@ -648,7 +648,7 @@ class Scratch3Text2SpeechBlocks {
     setVoice (args, util) {
         const state = this._getState(util.target);
 
-        let voice = args.VOICE;
+        let voice = args.VOICE.toUpperCase();
```
apple502j, line 651: This breaks when VOICE is not string (and such case is mentioned just below!)

apple502j (changes requested):

src/extensions/scratch3_text2speech/index.js
BoomerScratch: Oh, you are right. Maybe use `String(args.VOICE).toUpperCase();`?

ericrosenbaum:
Thanks for your contribution @BoomerScratch. Unfortunately, I am going to have to close this PR. As @adroitwhiz mentioned above as well, this is not our preferred fix for the problem, because it works only in English. Sorry about that. 

We really appreciate help from people like you, and we don't want to discourage it- instead, we are working on better communicating our process so that people can contribute more effectively. We have updated our contributor guidelines here: https://github.com/LLK/scratch-vm/blob/develop/.github/CONTRIBUTING.md

In this case, for example, it's worth noting that the relevant [issue](https://github.com/LLK/scratch-vm/issues/2448) was not marked `help-wanted`. According to our new process, for issues without the help-wanted label, we'd like you to request to work on it in a comment on the issue before attempting a fix.

Issue #2457, opened 2020-06-04
VM recording/replay tests


adroitwhiz:
Currently, compatibility testing is error-prone--writing compatibility tests requires a priori knowledge of which VM behavior is prone to breakage, leading to gaps in test coverage. This means that changes to the VM could potentially break compatibility of real-world projects without causing tests to fail.

There should be a way to deterministically record a series of inputs into the VM and play them back to get the same results every time. Once that's in place, "snapshots" of the project state could be taken and compared against future VM versions to ensure that they match. This would allow automated testing to be done on a wide variety of projects simply by playing through them while recording your inputs.

### Tasks
- Make the VM deterministic enough to reliably test
  - Centralize usage of timer utilities
  - Provide a way to swap out all timers with mock timers
  - Add a VM mode which runs all projects as fast as possible, faking all timers
- Add recording and playback functionality to the VM
  - Wrap IO classes with some sort of recorder?
  - Capture which inputs happen in which order in between VM steps
  - Replay those inputs in the proper order and during the proper steps in "fast mode"
- Implement a way to compare project states
  - Traverse the VM and save things like active costumes, sprite positions, etc.
  - Either through existing SB3 serialization or through a custom state-serialization step
  - Compare these serialized states with each other, and error out if they differ

PR #2458, opened 2020-06-04
Use stack timer utility for time-based blocks

adroitwhiz:
### Resolves

Progress on #2457

### Proposed Changes

This PR changes the music, say/think, and glide blocks to use the stack timer utility instead of returning promises or instantiating their own timers.

### Reason for Changes

Consolidating time-based blocks to all use the stack timer provided by `BlockUtility` will make it easier to swap that timer out for something else, allowing projects to be played deterministically at arbitrary speeds.

### Test Coverage

Tested manually

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -331,25 +331,40 @@ class Scratch3LooksBlocks {
         };
     }
 
+    /**
+     * Create a text bubble and yield on this thread until it's done. Used for "say/think for () secs".
+     * @param {number} args Block arguments
+     * @param {BlockUtility} util Utility object provided by the runtime.
+     * @param {string} type The type of bubble (say/think)
+     * @private
+     */
+    _bubbleForSecs (args, util, type) {
+        if (util.stackTimerNeedsInit()) {
+            this.runtime.emit(Scratch3LooksBlocks.SAY_OR_THINK, util.target, type, args.MESSAGE);
+
+            const duration = Math.max(0, 1000 * Cast.toNumber(args.SECS));
+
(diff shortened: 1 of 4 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -298,6 +298,33 @@ class Scratch3LooksBlocks {
...
+                this._updateBubble(util.target, type, '');
+            }
+        } else {
+            util.yield();
+        }
+    }
+
     say (args, util) {
         // @TODO in 2.0 calling say/think resets the right/left bias of the bubble
         let message = args.MESSAGE;
```
fsih: Blocked on the change that makes say and think bubbles round the same, so that we don't change the rounding behavior

fsih (commented):
Let's merge these changes into their own branch on LLK/ to keep them separate from develop while the VM rewrite is in-progress

cwillisf (approved):
👍

ericrosenbaum:
I'd like to see this go in on its own merits, since it seems like it will improve the timing behavior of say/think and glide. 

Looks like there's some unrelated code that snuck in there though? @adroitwhiz

adroitwhiz:
@ericrosenbaum I assume by the unrelated code you mean the text bubble stuff? The change I made to "say () for () secs" depends on that. It's already been merged into develop, but this is supposed to be merged into the "determinism" branch, which was supposed to be for a project that I never really got into, and that branch is a bit out-of-date.

I could change this PR so that these changes are merged to develop. I think we made the choice to merge into a different branch to avoid unnecessary risks, but since the "determinism" thing never really got off the ground it would probably be best to figure out what to do with this change.

PR #2465, opened 2020-06-08
Consistently format text bubbles (this time, like 2.0 does)

### LINKED ISSUE #2186: "Say" and "Think" bubbles have different numeric precision

adroitwhiz:
### Resolves

Resolves #2186

### Proposed Changes

This PR does mostly the same thing as https://github.com/LLK/scratch-vm/pull/2187, but matches 2.0's behavior of always rounding to 2 decimal places even if the number only needs 1 decimal place.

### Reason for Changes

These changes, in addition to fixing one current bug, reduce code duplication and potential for future bugs.

### Test Coverage

Yes

### DIFF
--- a/src/blocks/scratch3_looks.js
+++ b/src/blocks/scratch3_looks.js
@@ -234,6 +234,29 @@ class Scratch3LooksBlocks {
     }
 
     /**
+     * Properly format text for a text bubble.
+     * @param {string} text The text to be formatted
+     * @return {string} The formatted text
+     * @private
+     */
+    _formatBubbleText (text) {
+        if (text === '') return text;
+
+        // Non-integers should be rounded to 2 decimal places (no more, no less), unless they're small enough that
+        // rounding would display them as 0.00. This matches 2.0's behavior:
+        // https://github.com/LLK/scratch-flash/blob/2e4a402ceb205a042887f54b26eebe1c2e6da6c0/src/scratch/ScratchSprite.as#L579-L585
+        if (typeof text === 'number' &&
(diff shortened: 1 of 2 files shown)


src/blocks/scratch3_looks.js
```diff
@@ -233,6 +243,30 @@ class Scratch3LooksBlocks {
...
+    _formatBubbleText (text) {
+        if (text === '') return text;
+
+        // Non-integers should be rounded to 2 decimal places (no more, no less), unless they're small enough that
+        // rounding would display them as 0.00. This matches 2.0's behavior:
+        // https://github.com/LLK/scratch-flash/blob/2e4a402ceb205a042887f54b26eebe1c2e6da6c0/src/scratch/ScratchSprite.as#L579-L585
+        if (typeof text === 'number' &&
+            Math.abs(text) >= 0.01 && text % 1 !== 0) {
+            text = text.toFixed(2);
+
```
fsih: Extra newline?

fsih (approved):
Thanks for your patience with us figuring out what behavior we wanted. LG!

PR #2476, opened 2020-06-16
sanitize extension ID in getExtensionIdForOpcode

cwillisf:
### Proposed Changes

When loading an SB3, we determine whether a block is part of an extension by inspecting its "extended" opcode -- for example, `pen_clear` requires the `pen` extension. This change introduces an extra sanitizing step to ensure that the extension ID doesn't contain any characters that might cause problems. Also, this change defines the characters allowed in extension IDs as alphanumerics plus hyphen.

### Reason for Changes

We should always validate user input, and project files count. We validate most of the project file but we weren't validating the extension ID in this case. See also support ticket 229791.

Special thanks to @apple502j for bringing this to our attention!

### Test Coverage

A new test has been added to verify that characters are replaced appropriately.

### DIFF
--- a/docs/extensions.md
+++ b/docs/extensions.md
@@ -310,6 +310,7 @@ class SomeBlocks {
         return {
             // Required: the machine-readable name of this extension.
             // Will be used as the extension's namespace.
+            // Allowed characters are those matching the regular expression [\w-]: A-Z, a-z, 0-9, and hyphen ("-").
             id: 'someBlocks',
 
             // Core extensions only: override the default extension block colors.
--- a/src/serialization/sb3.js
+++ b/src/serialization/sb3.js
@@ -273,13 +273,17 @@ const compressInputTree = function (block, blocks) {
 };
 
 /**
- * Get non-core extension ID for a given sb3 opcode.
+ * Get sanitized non-core extension ID for a given sb3 opcode.
+ * Note that this should never return a URL. If in the future the SB3 loader supports loading extensions by URL, this
+ * ID should be used to (for example) look up the extension's full URL from a table in the SB3's JSON.
  * @param {!string} opcode The opcode to examine for extension.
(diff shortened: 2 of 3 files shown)


chrisgarrity (approved):
Code looks fine to me. I also checked that a selection of core extensions still work as expected. I didn't have a bad extension file to test with. @ericrosenbaum perhaps you can check whether this interferes with other allowed extensions.

ericrosenbaum:
So if I understand correctly, the extension manager has its own similar check [here](https://github.com/LLK/scratch-vm/blob/da68c4faf68014bfdfdb1af263910940d96ddd3a/src/extension-support/extension-manager.js#L273), which prevents loading a built-in extension that has an id with non-alphanumeric characters (it's more strict than the changes here, which allow uppercase letters and hyphens). I was playing around with making built-in extensions with weird id values, but none of them would load because of that existing code. So I wasn't able to test this code path in that way. Do you have a suggestion for how I might test it directly, or should we not worry about it?

apple502j:
@ericrosenbaum sent an email to contact us, with PoC attached

ericrosenbaum:
Thanks for your help @apple502j- I was able to use your info to replicate the issue and confirm that this change prevents it. nice!

ericrosenbaum (approved):
LGTM!

cwillisf:
filed #2517 to track the extension ID filter mismatch mentioned in https://github.com/LLK/scratch-vm/pull/2476#issuecomment-652555570

apple502j:
NOTE: CVE-2020-14000 assigned for this vulnerability. Affected Versions are scratch-vm from 0.1.0-prerelease.1524239808 to 0.2.0-prerelease.20200709173451 (both inclusive). This is patched in 0.2.0-prerelease.20200714185213.

PR #2341, opened 2020-03-25
Clear some stats when disposing runtime

apple502j:
### Resolves
Resolves #2340
Resolves #2346
Resolves #2347

### Proposed Changes
Answer, counter and timer gets cleared when runtime is disposed.

### Reason for Changes
Previous behavior (answer does not get cleared when runtime is disposed, but gets cleared when green flag is pressed) did not make sense. Also, this matches Scratch 1.4 behavior.

### Test Coverage
Added 1 test

Issue #2346, opened 2020-04-04
Counter is not cleared by making new project


apple502j:
### Expected Behavior
Counter (a hidden control block)'s value should be cleared when runtime is disposed.

### Actual Behavior
Counter is not cleared.

### Steps to Reproduce
You can get counter blocks by exporting the sprite on https://scratch.mit.edu/projects/45834258/editor/. After exporting it,
1) Import it.
2) Click increase counter block.
3) Click File->New.
4) Click counter block and see it says 1.

### Operating System and Browser
any

Related to #2340

Kenny2github:
TIL Scratch 3.0 actually still implements those blocks

PR #2492, opened 2020-06-22
Add a unit test for the clock being reset on runtime dispose

ericrosenbaum:
This just adds a test for the clock io module, to verify that it is reset when the runtime is disposed. 

This is a follow-up to @apple502j 's [PR](https://github.com/LLK/scratch-vm/pull/2341) that added that functionality.  

I've used a "simulated" timer in this case (though other similar tests use timeouts), assuming that this helps prevent flaky tests.

It might be reasonable to put this test in the `blocks_sensing` file instead, since it is really a test of the timer reporter block, but because of the indirection we use for io queries (`util.ioQuery('clock', 'projectTimer')`) I couldn't figure out how to do that.

### DIFF
--- a/test/unit/engine_runtime.js
+++ b/test/unit/engine_runtime.js
@@ -250,3 +250,24 @@ test('Disposing the runtime emits an event', t => {
     t.equal(disposed, true);
     t.end();
 });
+
+test('Clock is reset on runtime dispose', t => {
+    const rt = new Runtime();
+    const c = rt.ioDevices.clock;
+    let simulatedTime = 0;
+
+    c._projectTimer = {
+        timeElapsed: () => simulatedTime,
+        start: () => {
+            simulatedTime = 0;
+        }
+    };
+
+    t.ok(c.projectTimer() === 0);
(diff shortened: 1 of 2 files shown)


apple502j:
Thanks! 👍

picklesrus:
Yay tests!

I was actually thinking this particular unittest belongs where the runtime is being tested because IIUC, what you are verifying is that when the runtime is disposed that the clock gets reset. That doesn't seem to involve the sensing block at all despite changing the block's behavior.

If we wanted to have tests verify that the counter block's counter is reset in certain situations, we could add tests for that too, but I think they'd be system or integration tests. e.g. we'd force the VM through the steps that were causing https://github.com/LLK/scratch-vm/issues/2346 and check the value of the counter block is what we want it to be. 

That was a very long winded way of saying, the test looks fine, but maybe belongs in runtime's test :)

ericrosenbaum:
good point @picklesrus - I moved it!

picklesrus (approved):
LGTM

PR #2502, opened 2020-06-30
Refactor comments

### LINKED ISSUE #1893: Comments attached to variable blocks are altogether lost upon saving
### Expected Behavior

All comments are saved/restored when a project is saved/loaded.

### Actual Behavior

Comments attached to variable blocks disappear.

### Steps to Reproduce

Create a project that contains a variable block and attach a comment to it. Save it, then load it. Note that the comment is gone.

Firefox 65.0b7, Linux (Debian).

### LINKED ISSUE #2501: Comments are not deleted when blocks it's attached to get deleted
### Expected Behavior
If a comment is visually gone, it's logically gone.

### Actual Behavior
There is no code deleting the comment when the block gets deleted. However, Blockly does not render comments with blockId if there is no block associated with it, so you cannot see this on the workspace.

### Steps to Reproduce
- Add a block
- Right-click it and add comment
- Delete the block
- Export project
- The comment still exists in the JSON

any

### LINKED ISSUE #3687: Duplicated sprites lose their comments
_Thanks [stmsmarty30](https://scratch.mit.edu/users/stmsmarty30) for [finding this issue](https://scratch.mit.edu/discuss/post/3305071/)!_

### Expected Behavior
When duplicating sprites, comments should be kept

### Actual Behavior
Sprites lose their comments

### Steps to Reproduce
Add a comment on a sprite
Duplicate the sprite
The new sprite doesn't have a comment

Windows 10 Firefox 63

apple502j:
### Resolves
Resolves LLK/scratch-gui#3687
Resolves #2501 

note: #1893 will no longer be resolved

### Proposed Changes
First commits move comment attribute to `Sprite`, similar to `blocks` attribute. This was necessary to fix gui#3687. It also changes newBlockId so that it returns the mapping; the mapping is used inside `Sprite.duplicate` to fix gui#3687. RenderedTarget and Target still has comments attribute, so it should not cause problems.

Second commit is to fix #2501 - comments attached to blocks will now be deleted if the block is deleted, by deleting the comment item in `deleteBlock`.

Third commit is for #1893 . It has one breaking change; `deserializeBlocks` now return `{blocks, oldToNew}` instead of `blocks`. The mapping is used to sync `Comment.blockId` and `block.comment` later, in `parseScratchObject`. It also checks if the block actually exists - so any project that was saved with "invisible comment" (by #2501) will lose them.

Fourth commit is mostly tests. One new integration test is added to test #1893 and two unit tests are added to `engine_blocks` and `sprites_rendered-targets` to test #2501 and gui#3687, and they all pass for me.

### Reason for Changes
- moving comments to `Sprite`: first, it does not make sense that a Target - which includes clones - is controlling the sprite-wide comments. This also fixes consistency between blocks and comments. The major reason is, though, we needed to handle duplication inside `Sprite.duplicate`, not `RenderedTarget.duplicte` - because we needed the mapping. I could use `this.clones[0]` but it will make messy code.
- `newBlockIds` return value - it was necessary to implement duplicating comments attached to blocks
- Deleting comments when blocks are deleted - it was visually gone, but not on VM. They still exist on the memory and in the project file, and the only way to remove it is removing the sprite (or modifying it)
- deserializeBlocks change - deserializeInputDesc generates random uid for primitives (including variables/lists), which broke comments because the id is different. It now returns oldToNew mapping which can be used to correct comment's block id.
- parseScratchObject was changed to use the returned value from deserializeBlocks to attach correct blocks to comments.

### Test Coverage
Added 1 integration test (variable-comment) and 2 unit tests (see above).

Manual testing:
1) Make a variable and a list
2) Grab a variable reporter and a list reporter on a sprite
3) Add comments to those blocks
4) Add workspace comment
5) Duplicate sprite
6) The comments exist on the duplicated sprite
7) Export sprite
8) Upload again
9) All comments still exist
10) Delete blocks by dragging them to toolbox (do not touch comments)
11) Export sprite again
12) See the sprite.json of the exported sprite
13) There should be only one comment (workspace comment)

### DIFF
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -20,9 +20,10 @@ class Target extends EventEmitter {
     /**
      * @param {Runtime} runtime Reference to the runtime.
      * @param {?Blocks} blocks Blocks instance for the blocks owned by this target.
+     * @param {?Object.<string, *>} comments Array of comments owned by this target.
      * @constructor
      */
-    constructor (runtime, blocks) {
+    constructor (runtime, blocks, comments) {
         super();
 
         if (!blocks) {
@@ -55,7 +56,7 @@ class Target extends EventEmitter {
          * Key is the comment id.
          * @type {Object.<string,*>}
          */
-        this.comments = {};
+        this.comments = comments || {};
         /**
          * Dictionary of custom state for this target.
          * This can be used to store target-specific custom state for blocks which need it.
--- a/src/engine/target.js
+++ b/src/engine/target.js
@@ -20,7 +20,7 @@ class Target extends EventEmitter {
     /**
      * @param {Runtime} runtime Reference to the runtime.
      * @param {?Blocks} blocks Blocks instance for the blocks owned by this target.
-     * @param {?Object.<string, *>} comments Array of comments owned by this target.
+     * @param {?Object.<string, Comment>} comments Array of comments owned by this target.
      * @constructor
      */
     constructor (runtime, blocks, comments) {
--- a/src/sprites/sprite.js
+++ b/src/sprites/sprite.js
@@ -1,10 +1,12 @@
 const RenderedTarget = require('./rendered-target');
 const Blocks = require('../engine/blocks');
+const Comment = require('../engine/comment');
 const {loadSoundFromAsset} = require('../import/load-sound');
 const {loadCostumeFromAsset} = require('../import/load-costume');
 const newBlockIds = require('../util/new-block-ids');
 const StringUtil = require('../util/string-util');
 const StageLayering = require('../engine/stage-layering');
+const log = require('../util/log');
 
 class Sprite {
     /**
@@ -54,6 +56,13 @@ class Sprite {
         if (this.runtime && this.runtime.audioEngine) {
             this.soundBank = this.runtime.audioEngine.createBank();
         }
+
+        /**
+         * Dictionary of comments for this target.
+         * Key is the comment id.
+         * @type {Object.<string,*>}
+         */
+        this.comments = {};
     }
 
     /**
@@ -140,9 +149,39 @@ class Sprite {
         const blocksContainer = this.blocks._blocks;
         const originalBlocks = Object.keys(blocksContainer).map(key => blocksContainer[key]);
         const copiedBlocks = JSON.parse(JSON.stringify(originalBlocks));
-        newBlockIds(copiedBlocks);
+        const oldToNew = newBlockIds(copiedBlocks);
         copiedBlocks.forEach(block => {
             newSprite.blocks.createBlock(block);
+        });
+
+        Object.keys(this.comments).forEach(commentId => {
+            const comment = this.comments[commentId];
+            const newComment = new Comment(
+                null, // use new comment id
+                comment.text,
+                comment.x,
+                comment.y,
+                comment.width,
+                comment.height,
+                comment.minimized
+            );
+            if (comment.blockId) {
+                const newBlockId = oldToNew[comment.blockId];
+                if (newBlockId) {
+                    newComment.blockId = newBlockId;
+                    const newBlock = newSprite.blocks.getBlock(newBlockId);
+                    if (newBlock) {
+                        newBlock.comment = newComment.id;
+                    } else {
+                        log.warn(`Could not find block with id ${newBlockId
+                        } associated with comment ${newComment.id}`);
+                    }
+                } else {
+                    log.warn(`Could not find block with id ${comment.blockId
+                    } associated with comment ${comment.id} in the block ID mapping`);
+                }
+            }
+            newSprite.comments[newComment.id] = newComment;
         });
 
 
--- a/src/sprites/sprite.js
+++ b/src/sprites/sprite.js
@@ -60,7 +60,7 @@ class Sprite {
         /**
          * Dictionary of comments for this target.
          * Key is the comment id.
-         * @type {Object.<string,*>}
+         * @type {Object.<string, Comment>}
          */
         this.comments = {};
     }
@@ -157,7 +157,7 @@ class Sprite {
         Object.keys(this.comments).forEach(commentId => {
             const comment = this.comments[commentId];
             const newComment = new Comment(
-                null, // use new comment id
+                null, // generate new comment id
                 comment.text,
                 comment.x,
                 comment.y,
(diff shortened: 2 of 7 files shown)


src/sprites/sprite.js
```diff
@@ -54,6 +56,13 @@ class Sprite {
         if (this.runtime && this.runtime.audioEngine) {
             this.soundBank = this.runtime.audioEngine.createBank();
         }
+
+        /**
+         * Dictionary of comments for this target.
+         * Key is the comment id.
+         * @type {Object.<string,*>}
```
cwillisf:
```suggestion
         * @type {Object.<string,Comment>}
```

src/sprites/sprite.js
```diff
@@ -140,11 +149,41 @@ class Sprite {
...
-        newBlockIds(copiedBlocks);
+        const oldToNew = newBlockIds(copiedBlocks);
         copiedBlocks.forEach(block => {
             newSprite.blocks.createBlock(block);
         });
 
+        Object.keys(this.comments).forEach(commentId => {
+            const comment = this.comments[commentId];
+            const newComment = new Comment(
+                null, // use new comment id
```
cwillisf:
```suggestion
                null, // generate new comment id
```

src/sprites/sprite.js
```diff
@@ -140,11 +149,41 @@ class Sprite {
...
+            if (comment.blockId) {
+                const newBlockId = oldToNew[comment.blockId];
+                if (newBlockId) {
+                    newComment.blockId = newBlockId;
+                    const newBlock = newSprite.blocks.getBlock(newBlockId);
+                    if (newBlock) {
+                        newBlock.comment = newComment.id;
+                    } else {
+                        log.warn(`Could not find block with id ${newBlockId
+                        } associated with comment ${newComment.id}`);
```
adroitwhiz:
```suggestion
                        log.warn(`Could not find block with id ${newBlockId} associated with comment ${newComment.id}`);
```
Exactly 120 characters long, so it barely passes lint

src/sprites/sprite.js
```diff
@@ -140,11 +149,41 @@ class Sprite {
...
+            const newComment = new Comment(
+                null, // use new comment id
+                comment.text,
+                comment.x,
+                comment.y,
+                comment.width,
+                comment.height,
+                comment.minimized
+            );
+            if (comment.blockId) {
```
adroitwhiz, line 171:
```suggestion
            // If the comment is attached to a block, it has a blockId property for the block it's attached to.
            // Because we generate new block IDs for all the duplicated blocks, change all the comments' attached-block
            // IDs from the old block IDs to the new ones.
            if (comment.blockId) {
```

src/sprites/sprite.js
```diff
@@ -140,11 +149,41 @@ class Sprite {
...
+                    const newBlock = newSprite.blocks.getBlock(newBlockId);
+                    if (newBlock) {
+                        newBlock.comment = newComment.id;
+                    } else {
+                        log.warn(`Could not find block with id ${newBlockId
+                        } associated with comment ${newComment.id}`);
+                    }
+                } else {
+                    log.warn(`Could not find block with id ${comment.blockId
+                    } associated with comment ${comment.id} in the block ID mapping`);
```
adroitwhiz, line 188: Will either this code path or the one above be hit because of #2501, or is this a "just in case" thing?

src/engine/target.js
```diff
@@ -20,9 +20,10 @@ class Target extends EventEmitter {
     /**
      * @param {Runtime} runtime Reference to the runtime.
      * @param {?Blocks} blocks Blocks instance for the blocks owned by this target.
+     * @param {?Object.<string, *>} comments Array of comments owned by this target.
```
adroitwhiz:
```suggestion
     * @param {?Object.<string, Comment>} comments Array of comments owned by this target.
```

src/serialization/sb3.js
```diff
@@ -811,14 +813,14 @@ const deserializeBlocks = function (blocks) {
...
-            deserializeInputDesc(block, null, false, blocks);
+            oldToNew[blockId] = deserializeInputDesc(block, null, false, blocks);
             continue;
         }
         block.id = blockId; // add id back to block since it wasn't serialized
         block.inputs = deserializeInputs(block.inputs, blockId, blocks);
         block.fields = deserializeFields(block.fields);
     }
-    return blocks;
+    return ({blocks, oldToNew});
```
adroitwhiz: It seems a bit weird to complicate the signature of this function when the only places that use the old `blocks` return value are the tests.

cwillisf (changes requested):
Thanks for the fixes! I made a couple of suggestions (and I think @adroitwhiz's suggestions are pretty good too!). Also I think we should reconsider the call signatures of `deserializeInputDesc` and `deserializeBlocks` to make it easier to understand what they're doing. For example, I think we can avoid a lot of the complication in comment attachment fixup for deserialized primitives by reusing the existing ID instead of mapping old ID to new ID. I want to talk to some others on the Scratch team and maybe we'll have more suggestions after that.

cwillisf:
If you split the fix for #1893 into a separate PR then we can probably take the rest of the fixes while we're still debating the details of the deserialized primitive block IDs and such.

src/sprites/sprite.js
apple502j: It's probably "just in case" thing - deserialization should fix 2501 case

src/engine/target.js
apple502j: There are other places where we used that. Should we fix those too?

adroitwhiz: Yes

src/sprites/sprite.js
apple502j: But because we removed deserialization changes from this PR, this handles 2501 case now.

apple502j:
I'm doing that @cwillisf

Wowfunhappy:
One thing I noticed with the current patch is that if you:

1. Create a comment attached to a variable reporter (ie, to trigger #1893)
2. Save and reload the page (comment will disappear per #1893)
3. Duplicate the sprite

Then the disappeared comment will appear in the duplicated sprite (but not the original sprite).

This is just an observation, I couldn't tell if it's intended or not! I know that this pull request is no longer supposed to fix #1893 per cwillisf but it seems it got half fixed anyway?

Issue #1386, opened 2018-07-22
Highlight the item loaded by list blocks


apple502j:
### Expected Behavior
The item on the stage monitor loaded by list blocks should be highlighted.

### Actual Behavior
Not implemented.

### Steps to Reproduce
1. Open https://scratch.mit.edu/projects/230003247/
2. Open https://llk.github.io/scratch-gui/develop/#230003247
On 3.0, highlighting doesn't work

NOTE: The project told me two issues: other one #1385 

### Operating System and Browser
Windows 7 HomePremium 64bit SP1
Firefox 61.0.1

joker314:
Note to anyone trying to do this:
- list monitors are rendered in LLK/scratch-gui
- the behaviour of 'flashing' is implemented in LLK/scratch-gui#3585 (though it deals with boxes around sprites rather than list indices), so it might be a useful reference for you

towerofnix:
Note that you the highlighted item is also automatically scrolled to.

Also, while working with the linked project, **do not** click the list items; this prevents the "when key pressed" hat block from detecting input.

Issue #1385, opened 2018-07-22
Scroll to item loaded by list blocks


apple502j:
### Expected Behavior
The list monitor should scroll to where the item loaded by list blocks places.

### Actual Behavior
Not implemented so far.

### Steps to Reproduce
1. Open https://scratch.mit.edu/projects/230003247/
2. Open https://llk.github.io/scratch-gui/develop/#230003247
3.0 version doesn't work

### Operating System and Browser
Windows 7 HomePremium 64bit SP1
Firefox 61.0.1

PR #2507, opened 2020-07-02
Highlight and scroll to list index used by list blocks

apple502j:
GUI part: LLK/scratch-gui#5961

### Resolves
VM part for fixing #1385 and #1386 - **VM must be merged first**, and then GUI can be merged.

### Proposed Changes
This adds `util.highlightListItem` (used by "index related blocks", has some checks) and `Runtime.highlightListItem`. This also adds `highlightItem` to MonitorRecord.

This also adds highlight handling to `Runtime._step`. It can only highlight one item per list per frame, and the index passed from last "index related block" executed in that frame is used. This is handled just before sending monitor update, so amounts of events emitted from runtime should be equal, if they only write.

"index related block" are:
- add to list
- delete index of list
- insert at list
- replace item of list
- get item of list

Note: if the monitor is hidden, `util.highlightListItem` will not call `runtime.highlightListItem`, meaning it will not emit unnecessary events.

### Reason for Changes
Compatibility with Scratch 2.0.

### Test Coverage
Manually tested, on modified scratch-gui (PR coming soon)

### DIFF
--- a/src/blocks/scratch3_data.js
+++ b/src/blocks/scratch3_data.js
@@ -128,6 +128,7 @@ class Scratch3DataBlocks {
             args.LIST.id, args.LIST.name);
         if (list.value.length < Scratch3DataBlocks.LIST_ITEM_LIMIT) {
             list.value.push(args.ITEM);
+            util.highlightListItem(list, list.value.length - 1);
             list._monitorUpToDate = false;
         }
     }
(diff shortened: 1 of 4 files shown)


ericrosenbaum:
Closing for now, see comment in https://github.com/LLK/scratch-gui/pull/5961

PR #2557, opened 2020-08-12
Prevent "broadcast and wait" being interrupted by a change to its input 

### LINKED ISSUE #2109: "broadcast (expr) and wait" ends if value of (expr) changes

ericrosenbaum:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/2109

### Proposed Changes

Store the input to the "broadcast and wait" block on the stackFrame. While waiting, use the stored value, instead of looking up the argument again. 

### Reason for Changes

Previously, if you used a variable in a "broadcast and wait", and you change the variable's value while it was waiting for the resulting threads to run, the wait would end immediately. This change causes the block to wait until all threads started by it have completed, as in Scratch 2.

### Test Coverage

Added a fixture file to the execute tests

### Credits

Thanks to @apple502j and @adroitwhiz for prior work and investigation
This PR replaces https://github.com/LLK/scratch-vm/pull/2409


src/blocks/scratch3_event.js
```diff
@@ -90,10 +90,12 @@ class Scratch3EventBlocks {
...
     broadcastAndWait (args, util) {
-        const broadcastVar = util.runtime.getTargetForStage().lookupBroadcastMsg(
-            args.BROADCAST_OPTION.id, args.BROADCAST_OPTION.name);
-        if (broadcastVar) {
-            const broadcastOption = broadcastVar.name;
+        if (!util.stackFrame.broadcastVar) {
+            util.stackFrame.broadcastVar = util.runtime.getTargetForStage().lookupBroadcastMsg(
+                args.BROADCAST_OPTION.id, args.BROADCAST_OPTION.name);
+        }
+        if (util.stackFrame.broadcastVar) {
```
kchadha, line 97: This could just be an `else` to the previous `if`

adroitwhiz: The first `if` block modifies `util.stackFrame.broadcastVar`, so this condition could evaluate to something different though, right?

kchadha: Ah yes, you're right! Good catch. I guess it's written in a confusing way. It could be clearer to write the two checks as:

```
if (!util.stackFrame.broadcastVar) {
    // attempt to set it
}

// If we failed to find the broadcast message, exit
if (!util.stackFrame.broadcastVar) return;
```

But for the sake of simplicity of the code change from what we had before, this is fine too.

kchadha (approved):
LGTM! Thanks @ericrosenbaum, @apple502j, and @adroitwhiz!

PR #2632, opened 2020-10-15
Pass on event type from CDM

fsih:
### Resolves
Toward fixing LLK/scratch-android#308

### Proposed Changes
Handle a new event type for CDM interface

### Reason for Changes
The GUI list is redundant, so the new event type will indicate that the user doesn't need to pick a device from the list a second time

### Test Coverage
Tested locally


cwillisf (approved):
LGTM! Sorry I missed this before.

PR #2710, opened 2020-11-23
Handle user did not pick peripheral

fsih:
Part of https://github.com/LLK/scratch-android/pull/340
Handle actions on the CDM interface

### DIFF
--- a/src/io/bt.js
+++ b/src/io/bt.js
@@ -128,6 +128,16 @@ class BT extends JSONRPC {
                 window.clearTimeout(this._discoverTimeoutID);
             }
             break;
+        //TODO why isn't 'user did pick peripheral' here?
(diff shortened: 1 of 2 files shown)


src/io/bt.js
```diff
@@ -128,6 +128,16 @@ class BT extends JSONRPC {
                 window.clearTimeout(this._discoverTimeoutID);
             }
             break;
+        //TODO why isn't 'user did pick peripheral' here?
```
cwillisf: Oops - there definitely should be a 'userDidPickPeripheral' handler here! Connecting on BT using CDM doesn't work for me currently.

cwillisf (approved):
:+1: thanks for handling this!

PR #2800, opened 2021-01-08
Call into V2SVGAdapter instead of SVGRenderer

### LINKED ISSUE #211: Split up the SvgRenderer monolith

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-svg-renderer/issues/211

### Proposed Changes

This PR changes the costume loading API to use the new `V2SVGAdapter` API (a function) instead of the `SvgRenderer` API (a class).

### Reason for Changes

By depending only on the parts of the SVG renderer API that are actually used here, this will allow the monolithic `SvgRenderer` class to be removed.

### Test Coverage

Tested manually

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -8,8 +8,7 @@ const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
         if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
             log.error('No V2 SVG adapter present; SVGs may not render correctly.');
         } else if (optVersion && optVersion === 2 && runtime.v2SvgAdapter) {
-            runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
-            svgString = runtime.v2SvgAdapter.toString();
+            svgString = runtime.v2SvgAdapter(svgString);
             // Put back into storage
             const storage = runtime.storage;
             costume.asset.encodeTextData(svgString, storage.DataFormat.SVG, true);
--- a/src/playground/benchmark.js
+++ b/src/playground/benchmark.js
@@ -650,7 +650,7 @@ const runBenchmark = function () {
     vm.attachRenderer(renderer);
     const audioEngine = new AudioEngine();
     vm.attachAudioEngine(audioEngine);
-    vm.attachV2SVGAdapter(new ScratchSVGRenderer.SVGRenderer());
+    vm.attachV2SVGAdapter(ScratchSVGRenderer.V2SVGAdapter);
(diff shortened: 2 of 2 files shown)


src/import/load-costume.js
```diff
@@ -8,8 +8,7 @@ const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
         if (optVersion && optVersion === 2 && !runtime.v2SvgAdapter) {
             log.error('No V2 SVG adapter present; SVGs may not render correctly.');
         } else if (optVersion && optVersion === 2 && runtime.v2SvgAdapter) {
-            runtime.v2SvgAdapter.loadString(svgString, true /* fromVersion2 */);
-            svgString = runtime.v2SvgAdapter.toString();
+            svgString = runtime.v2SvgAdapter(svgString);
```
fsih, line 11: I'm planning to change this to always run, regardless of whether optVersion === 2 or not, since we want to run fixup on all SVGs. After that, we shouldn't need to export V2SVGAdapter from ScratchSVGRenderer, and we shouldn't need to run fixup in paint (since SVGs will pass through loadCostume before being loaded in paint)

adroitwhiz: Let me know once that change goes in--I can rebase this after it does, to instead call:
```js
serializeSvgToString(loadSvgString(svgString, optVersion === 2 /* fromVersion2 */));
```

One issue is that once we start exporting multiple individual functions, passing them *all* in from the GUI (whereas before, we only had to pass in a single `SVGRenderer` object) could start to get cumbersome. IIRC this change was made to make updating the SVG renderer dependency less painful--is there some way to encode "this package works with any version of `scratch-svg-renderer`, but use the parent package's version"?

fsih: I think peer dependencies? I would have to research them more.

adroitwhiz:
Closing in favor of #2954

PR #2891, opened 2021-02-18
Replace nets with Fetch API

### LINKED ISSUE #2890: Remove nets and use native fetch
scratch-vm uses `nets` to fetch remote resources. However, all supported browsers support `fetch`, and there is no reason to add `nets` (besides [causing issues with Scratch Desktop](https://github.com/LLK/scratch-desktop/issues/157).)

Just like scratch-storage, nets should be removed and native `fetch()` should be used. If Node support is required for unittests, we can probably add `node-fetch` as devdeps.

adroitwhiz:
### Resolves

Resolves #2890

### Proposed Changes

This PR removes the `nets` package and replaces calls to it with `fetch` calls.

### Reason for Changes

The `nets` package is unmaintained, and depends on the deprecated `request` package. In addition, as described in https://github.com/LLK/scratch-desktop/issues/157, its behavioral differences between Node and the browser can cause issues.

### Test Coverage

Must be tested manually

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -1,5 +1,4 @@
 const formatMessage = require('format-message');
-const nets = require('nets');
 const languageNames = require('scratch-translate-extension-languages');
 
 const ArgumentType = require('../../extension-support/argument-type');
@@ -8,6 +7,7 @@ const Cast = require('../../util/cast');
 const MathUtil = require('../../util/math-util');
 const Clone = require('../../util/clone');
 const log = require('../../util/log');
+const fetchWithTimeout = require('../../util/fetch-with-timeout');
 
 /**
  * Icon svg to be displayed in the blocks category menu, encoded as a data URI.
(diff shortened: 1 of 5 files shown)


src/util/fetch-with-timeout.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Fetch a remote resource like `fetch` does, but with a time limit.
+ * @param {Request|string} resource Remote resource to fetch.
+ * @param {?object} init An options object containing any custom settings that you want to apply to the request.
+ * @param {number} timeout The amount of time before the request is canceled, in milliseconds
+ * @returns {Promise<Response>} The response from the server.
+ */
+const fetchWithTimeout = (resource, init, timeout) => {
```
cwillisf, line 8: I was looking at the docs for `fetch` timeouts, and I see that different browsers have different built-in timeouts. Does this function react gracefully if the `timeout` argument is greater than the browser's built-in timeout duration and the fetch times out due to the browser's timeout rather than the caller's timeout?

src/extensions/scratch3_text2speech/index.js
```diff
@@ -722,46 +722,46 @@ class Scratch3Text2SpeechBlocks {
...
+                return new Promise(resolve => {
                     soundPlayer.on('stop', () => {
                         this._soundPlayers.delete(soundPlayer.id);
                         resolve();
                     });
                 });
+            })
+            .catch(err => {
+                log.warn(err);
+                return;
```
cwillisf: This `return;` seems unnecessary... is it there for lint purposes?

src/extensions/scratch3_text2speech/index.js
```diff
@@ -722,46 +722,46 @@ class Scratch3Text2SpeechBlocks {
...
                 // Play the sound
                 const sound = {
                     data: {
-                        buffer: body.buffer
+                        buffer: buffer
                     }
                 };
-                this.runtime.audioEngine.decodeSoundPlayer(sound).then(soundPlayer => {
-                    this._soundPlayers.set(soundPlayer.id, soundPlayer);
+                return this.runtime.audioEngine.decodeSoundPlayer(sound);
```
cwillisf, line 740: 👍

webpack.config.js
```diff
@@ -77,7 +77,6 @@ module.exports = [
             'immutable': true,
             'jszip': true,
             'minilog': true,
-            'nets': true,
```
cwillisf, line 80: 😎

cwillisf (commented):
This is great! I have a couple questions but nothing major. Thanks!

cwillisf:
@ericrosenbaum or @picklesrus, do you have any concerns with this change?

ericrosenbaum (approved):
I tested on Scratch Desktop on Mac and verified that this fixes Text to Speech. Yay! Translate also works fine.

src/extensions/scratch3_text2speech/index.js
adroitwhiz: I think it was, but it looks like linting still passes without it!

src/util/fetch-with-timeout.js
adroitwhiz: Yes--on Chrome at least, the timeout results in a `TypeError` of all things being thrown, so while it's different from the error thrown here, at the very least, an error is thrown in both cases.

`Promise.race` [takes the earliest result](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/race) and ignores everything else, so even if we didn't clear the timeout or abort the request, any further happenings would be ignored.

adroitwhiz:
@cwillisf The `fetchWithTimeout` function could be [much simpler](https://stackoverflow.com/a/57888548) if support for Safari <11 is not needed, with the caveat that `AbortController` [does nothing on Safari <12](https://bugs.webkit.org/show_bug.cgi?id=174980) (requests will still work, but the timeout will not). What's the current policy on supporting older Safari versions? Is this a worthwhile tradeoff?

apple502j:
@adroitwhiz There was a recent discussion about whether to support Safari<**14**, and how much support is required: LLK/scratch-www#4895

Here is the commit message:

> Safari < 14 does not support Intl.RelativeTimeFormat which is used in the moderation messaging. Adding this polyfill to support it. RelativeTimeFormat depends on Locale, which is also not available in Safari <14, and PluralRules, not available in Safari < 13.
There is a trade off between adding locale-data for all users that load the page regardless of what browser they’re using and whether they get muted, and only localizing the most common languages.
In the last 30 days only ~3% of users on scratch use a version of Safari (13.x) that doesn’t support Intl. Of that 3%, these 7 locales would cover over 60% of the users. English will be used by default if the locale data is not available.

IMO, they should also consider dropping Legacy Edge, but that depends on the numbers.

cwillisf (approved):
👍 thanks for the changes, and the discussion!

Regarding Safari < 11: I think it's OK to leave the extra code there for a "belt and suspenders" approach, considering that (if I understand correctly) it doesn't complicate the caller and only complicates the guts of `fetchWithTimeout`. I'm also happy to consider a followup PR simplifying the function if you feel strongly about it ;)

PR #2920, opened 2021-03-01
Do not translate strings containing only digits

ericrosenbaum:
In the translate extension, before making a request, check if the input string contains only digits 0-9. If it does, just return it without making a request.

We don't currently have a unit test file for the translate extension. I couldn't quickly add one without restructuring the extension a little bit (we use the window property `navigator` to get the browser language setting, which would need to be stubbed out).

### DIFF
--- a/src/extensions/scratch3_translate/index.js
+++ b/src/extensions/scratch3_translate/index.js
@@ -247,6 +247,10 @@ class Scratch3TranslateBlocks {
      * @return {Promise} - a promise that resolves after the response from the translate server.
      */
     getTranslate (args) {
+        // If the text contains only digits 0-9 and nothing else, return it without
+        // making a request.
+        if (/^\d+$/.test(args.WORDS)) return Promise.resolve(args.WORDS);
+
         // Don't remake the request if we already have the value.
         if (this._lastTextTranslated === args.WORDS &&
             this._lastLangTranslated === args.LANGUAGE) {


src/extensions/scratch3_translate/index.js
```diff
@@ -247,6 +247,10 @@ class Scratch3TranslateBlocks {
      * @return {Promise} - a promise that resolves after the response from the translate server.
      */
     getTranslate (args) {
+        // If the text contains only digits 0-9 and nothing else, return it without
+        // making a request.
+        if (/^\d+$/.test(args.WORDS)) return Promise.resolve(args.WORDS);
```
picklesrus, line 252: Do we need the promise.resolve? Below (line ~253) we don't have one, though I expect it doesn't matter, but asking because I've forgotten how this fits in with the block execution :-/

apple502j: I don't think it matters too much - returing just args could break projects that rely on Promise block execution inside warp mode procedures but in general you shouldn't use these blocks inside warp procedure anyway.

apple502j:
Question: this changes the block so that it could return Number instead of String if Number is passed. is it intended? (AKA, should we call Cast.toString? again, Scratch is strangely-typed language, so it probably doesn't matter too much)
image

src/extensions/scratch3_translate/index.js
ericrosenbaum: Yeah, looks like this will be okay, since we were already inconsistent. In normal use, where the API returns a result, the opcode returns a promise. If we hit the local cache for the same exact request, we return the value. This change adds returning a promise for the numbers-only case.

ericrosenbaum:
@apple502j good catch on that detail about the type of the result. I believe the implications are sufficiently minor that we do not need to worry about them.

PR #2926, opened 2021-03-04
Simplify RuntimeScriptCache

### LINKED ISSUE #3031: Remove dead code path from `RuntimeScriptCache`
The `RuntimeScriptCache` is used by `Runtime.startHats` to filter only the hat blocks matching the specified opcode and fields, aka non-droppable menus on blocks (e.g. 'event_whenkeypressed', where {'KEY': 'SPACE'}).

Each RuntimeScriptCache object holds data related to one hat block: its opcode and uppercase versions of all its fields (so that field matching is case-insensitive: you can broadcast 'MeSsAgE1' and 'when I receive message1' will still fire).

Currently, this cache also holds droppable menus, aka inputs. These are actually blocks in their own right, known as "shadow blocks". They are essentially reporter blocks that store their reported value in a specific field. The RuntimeScriptCache is copying that value out of the field and into the `fieldsOfInputs` property.

However, this breaks if you drop a different reporter block into that input--you'll have no way of knowing what it evaluates to. Thus, all hat blocks which use droppable inputs are "edge-activated"--their inputs are instead evaluated once per frame by the runtime. If you try to pass `Runtime.startHats` a list of fields to filter by, and the hat block you want to start contains any droppable menus, it will error out because the code path never worked in the first place.

adroitwhiz:
### Resolves

Resolves #3031

### Proposed Changes

This PR simplifies the logic in `RuntimeScriptCache`, removing a code path that copied the values of hat blocks' droppable inputs into the cache.

### Reason for Changes

Hat blocks which use droppable inputs utilize a different code path which does not interact with the `RuntimeScriptCache`. The removed code path was incomplete and would never have properly worked.

I go into more detail in my commit message:

> The RuntimeScriptCache is used by Runtime.startHats to filter only the
> hat blocks matching the specified opcode and *fields*, aka non-droppable
> menus on blocks
> (e.g. 'event_whenkeypressed', where {'KEY': 'SPACE'}).
> Each RuntimeScriptCache object holds data related to one hat block: its
> opcode and uppercase versions of all its fields (so that field matching
> is case-insensitive: you can broadcast 'MeSsAgE1' and 'when I receive
> message1' will still fire).
> 
> Previously, this cache also held droppable menus, aka *inputs*. These
> are actually blocks in their own right, known as "shadow blocks". They
> are essentially reporter blocks that store their reported value in a
> specific field. The RuntimeScriptCache was copying this value out of the
> field and into the `fieldsOfInputs` property.
> 
> However, this broke if you dropped a different reporter block into that
> input--you'd have no way of knowing what it would evaluate to. Thus, all
> hat blocks which use droppable inputs are "edge-activated"--they are
> evaluated once per frame by the runtime as opposed to being activated by
> the `startHats` function. Thus, storing inputs is not only unnecessary
> but would never have worked anyway.

### Test Coverage

Tested manually

### DIFF
--- a/src/engine/blocks-runtime-cache.js
+++ b/src/engine/blocks-runtime-cache.js
@@ -30,8 +30,7 @@ class RuntimeScriptCache {
         const fields = container.getFields(block);
 
         /**
-         * Formatted fields or fields of input blocks ready for comparison in
-         * runtime.
+         * Formatted fields ready for comparison in runtime.
          *
          * This is a clone of parts of the targeted blocks. Changes to these
          * clones are limited to copies under RuntimeScriptCache and will not
@@ -41,16 +40,6 @@ class RuntimeScriptCache {
          * @type {object}
          */
         this.fieldsOfInputs = Object.assign({}, fields);
-        if (Object.keys(fields).length === 0) {
-            const inputs = container.getInputs(block);
-            for (const input in inputs) {
-                if (!inputs.hasOwnProperty(input)) continue;
-                const id = inputs[input].block;
-                const inputBlock = container.getBlock(id);
-                const inputFields = container.getFields(inputBlock);
-                Object.assign(this.fieldsOfInputs, inputFields);
-            }
-        }
         for (const key in this.fieldsOfInputs) {
             const field = this.fieldsOfInputs[key] = Object.assign({}, this.fieldsOfInputs[key]);
             if (field.value.toUpperCase) {
--- a/src/engine/blocks-runtime-cache.js
+++ b/src/engine/blocks-runtime-cache.js
@@ -26,9 +26,6 @@ class RuntimeScriptCache {
          */
         this.blockId = blockId;
 
-        const block = container.getBlock(blockId);
-        const fields = container.getFields(block);
-
         /**
          * Formatted fields ready for comparison in runtime.
          *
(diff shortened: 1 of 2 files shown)


apple502j:
This PR should be fine, although there are some notes:

All hat blocks are activated by startHats. However, the commit message should still be true; edge hats are activated inside `_step` without arguments to specify conditions. These conditions are evaluated inside execute.js `handleReport`; the code reads, "if the hat is edge, and condition is falsy or previous condition was truthy, retire." I am sure this process is done in the same way as other reporter blocks do.

adroitwhiz:
> All hat blocks are activated by startHats.

I'm not sure what you mean by this. `startHats` takes an argument for which opcode of hat block to call, and only starts threads for hat blocks with that specific opcode.

Theoretically, you could start edge-activated hats with `startHats`, but that would probably break, and there isn't any place in the codebase that does so.

apple502j:
`_step` calls `startHats` for all edge hat blocks registered. The thing is, the problem is probably not whether it's called via startHats or not, but it's more of "whether the second argument is used". It is possible (and _step does) to call startHats with no `optMatchFields` against edge hats; and it would probably break if optMatchFields is used for those (see adroitwhiz's posts for reasons). Just calling startHats without optMatchFields is safe since it doesn't trigger the problematic code, and instead uses normal processing routes including `handleReport` function that handles thread retiring for edge hats.

Issue #2935, opened 2021-03-07
Spurious log output during tests


adroitwhiz:
### Expected Behavior

Logging during tests should be kept minimal to make it easy to spot where and how tests fail.

### Actual Behavior

There are ~~four~~ five different kinds of spurious log warnings that flood the console, making it nigh impossible to look through:

`storage warn Deprecation: Storage.addWebSource has been replaced by addWebStore.`

`[DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.`

`vm error No rendering module present; cannot load costume:`

`vm error No audio engine present; cannot load sound asset:`

`Translation for "..." in "en" is missing`

These are logged several times per test, completely outnumbering any sort of useful output

### Steps to Reproduce

1. Clone the `scratch-vm` repository (if you haven't already)
2. Run `npm install` (if you haven't already)
3. Run `npm run test`
4. Observe the output

### Operating System and Browser

All

AlexisGoodfellow:
It seems like this could be fixed with proper environment separation using a mechanism like `NODE_ENV`. 

Depending on how the logging is already done (and if there are numerous tests which consume log output), it might be worth considering disabling `minilog` entirely in testing environments. That might look something like changing `src/util/log.js` to do the following:

~~~~
if (process.env.NODE_ENV !== 'test') {
    minilog.enable();
}
~~~~

If that can't be done at global scale, then the same environment check could be used to silence those specific superfluous log messages (they do cause *quite* a lot of visual clutter).

I tried to look into whether or not `tap` automatically sets `NODE_ENV` to `test` during unit tests, but my cursory investigation there was inconclusive. If `tap` doesn't set `NODE_ENV` explicitly, there may be some small necessary alterations to how the `tap` commands are set up in the `package.json` file.

PR #2936, opened 2021-03-07
Replace `new Buffer()` with `Buffer.from()`

adroitwhiz:
### Resolves

Part of #2935

### Proposed Changes

This PR replaces the two usages of `new Buffer(...)` with `Buffer.from(...)`, which should function identically.

### Reason for Changes

The `Buffer` constructor has long been deprecated in Node due to security concerns, and outputs an annoying message to the console every time it is used.

Side note: `fs.readFileSync` appears to always return a `Buffer` anyway, so I'm not really sure why it was ever wrapped in `new Buffer`. In the interest of not accidentally messing anything up, I'm applying the minimum changes necessary.

### Test Coverage

Yes

### DIFF
--- a/test/fixtures/readProjectFile.js
+++ b/test/fixtures/readProjectFile.js
@@ -3,7 +3,7 @@ const fs = require('fs');
 
 module.exports = {
     readFileToBuffer: function (path) {
-        return new Buffer(fs.readFileSync(path));
+        return Buffer.from(fs.readFileSync(path));
     },
     extractProjectJson: function (path) {
         const zip = new AdmZip(path);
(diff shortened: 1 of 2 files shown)


AlexisGoodfellow (commented):
I don't see any reason _not_ to do this. `grep` reveals that no new instances of `new Buffer` have been introduced anywhere in the codebase. Given that these function identically and the changed code gets rid of a `DeprecationWarning` that annoyingly fills up the console logs when running tests, I think this is a worthwhile quality-of-life change.

kchadha (approved):
This is great, thanks!

PR #2948, opened 2021-03-09
Raise params redo

adroitwhiz:
### Resolves

N/A

### Proposed Changes

Puts #2145 back in, since #2953 fixes the issue that was introduced.

### Reason for Changes

See #2145

### Test Coverage

Tests have been added to catch the bug that caused #2145 to be reverted


adroitwhiz:
#2953 should make this easier to fix and should go in first

PR #2953, opened 2021-03-09
Thread always done redo

adroitwhiz:
### Resolves

Closes #2148

### Proposed Changes

This PR includes the changes from #2148, but includes a fix for block glow. Script stacks which only run for one frame should now glow properly.

### Reason for Changes

See #2148

My fix (the last commit) has this to say:

> retireThread sets requestScriptGlowInFrame to false, which results in
> stacks that only execute for one frame never glowing. It also resets
> stack and stackFrames, which is redundant because we were only calling
> it if the stack was empty (and thus did not need to be emptied again).

### Test Coverage

Tested manually

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1621,8 +1621,6 @@ class Runtime extends EventEmitter {
      * @param {!Thread} thread Thread object to remove from actives
      */
     _stopThread (thread) {
-        // Mark the thread for later removal
-        thread.isKilled = true;
         // Inform sequencer to stop executing that thread.
         this.sequencer.retireThread(thread);
     }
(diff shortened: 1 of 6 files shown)


src/engine/runtime.js
```diff
@@ -2043,9 +2041,6 @@ class Runtime extends EventEmitter {
             this.profiler.start(stepProfilerId);
         }
 
-        // Clean up threads that were told to stop during or since the last step
-        this.threads = this.threads.filter(thread => !thread.isKilled);
-
```
adroitwhiz, line 2038: https://github.com/LLK/scratch-vm/pull/1683 is what makes this safe to remove, I think

PR #2954, opened 2021-03-09
Remove v2SvgAdapter

fsih:
### Resolves

A step toward https://github.com/LLK/scratch-svg-renderer/issues/219 and https://github.com/LLK/scratch-svg-renderer/issues/220

### Proposed Changes

- We no longer need gui to send us a copy of svg renderer since we can access the methods statically now
- Moves svg renderer to peer dependencies since instead of receiving an object from gui, we can now call methods from svg renderer directly
- call fixup on all svgs regardless of if version is 2. This fixes issues such as removing svg: from tags, which was breaking costume uploads from inkscape

### Reason for Changes
- calls to SVG renderer are now parallelizable
- now you can upload more types of svgs to scratch without them turning into question marks

### Test Coverage


cwillisf:
I tried to run the playground and got this. Looks like `benchmark.js` needs to be adjusted to not use `attachV2SVGAdapter` and instead use the bare functions?
image

cwillisf (approved):
👍

PR #2963, opened 2021-03-14
Don't deserialize hidden extension monitors

### LINKED ISSUE #2331: Extensions cannot be removed if its stage monitor had been shown
### Expected Behavior
Extensions without its blocks should not be serialized, or should be skipped in deserialization.

### Actual Behavior
The monitor block for it will be kept during serialization and deserialization, making it impossible to remove without modifying JSON. 

### Related Code
https://github.com/LLK/scratch-vm/blob/develop/src/serialization/sb3.js#L1168 - Deserialization of monitors; extensions are re-added there.

### Steps to Reproduce
Add an extension with a stage monitor (such as Translate), show it then hide it. Save the project locally or on the website, and reload it. The extension is not removed, despite having no translation blocks.

any

Related: #712

apple502j:
### Resolves
Resolves #2331 

### Proposed Changes
Hidden extension monitors are no longer deserialized.

### Reason for Changes
This allows unused extensions to not be loaded. Previously, if extension monitors had been shown, those extensions were always added (even if they were hidden) - as those monitors keep monitor state in runtime. This feature itself is good (for example, hidden variables need to keep their positions), but causes problems when it comes to extensions.

Since no extensions track extra data (like slider min), the only data that could be lost as part of this change are the positions of hidden extension monitors. However, they are hidden with no ways to show with blocks (unlike variables), and it's unlikely that someone is using such behavior.

### Test Coverage
Added one integration test


towerofnix:
> Since no extensions track extra data (like slider min), the only data that could be lost as part of this change are the positions of hidden extension monitors. However, they are hidden with no ways to show with blocks (unlike variables), and it's unlikely that someone is using such behavior.

Definitely agreed with this! A solution which prevents that data loss *could* be made, but it would involve reworking quite a lot of the existing monitor code. Not really worth it for such a low-severity fix, at the moment anyhow.

Also +kudos for the very nice integration test :3

PR #3026, opened 2021-04-11
Go back to only using svgrenderer to process vectors that we think are coming from scratch 2

fsih:
### Resolves
Processing all SVGs was causing a spike of saved assets on the server

### Proposed Changes
Reverts the part of https://github.com/LLK/scratch-vm/pull/2954/files which changed which assets are run through the scratch-svg-renderer repo's fixup code. The rest of pull 2954 switches vm over from using deprecated functions to the replacement functions in scratch-svg-renderer, which is part of solving https://github.com/LLK/scratch-svg-renderer/issues/220

### Reason for Changes
Be nice to the servers

### Test Coverage

Tested manually
- Before the change, fixupSvgString IS called when loading an sb3 project
- After the change, it IS NOT called on sb3 project
- After the change, it IS called on sb2 project

### DIFF
--- a/src/import/load-costume.js
+++ b/src/import/load-costume.js
@@ -4,22 +4,26 @@ const {loadSvgString, serializeSvgToString} = require('scratch-svg-renderer');
 
 const loadVector_ = function (costume, runtime, rotationCenter, optVersion) {
     return new Promise(resolve => {
-        const svgString = costume.asset.decodeText();
-        // scratch-svg-renderer fixes syntax that causes loading issues,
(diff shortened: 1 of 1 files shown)


cwillisf (approved):
This looks good as an urgent fix!

I think in the longer run it might make sense to do something like "process all vector costumes but don't mark them dirty" but we should probably talk/think that through more thoroughly. We may also want to implement some way to avoid saving question-mark assets as a prerequisite.

PR #3067, opened 2021-04-30
Save and load origin metadata

### LINKED ISSUE #2851: Preserve project json metadata: `meta.origin`
A project saved with an `origin` key in the `meta` section of the JSON should preserve the value when saved. Currently it's removed.

Don't create a default for this value. It should just be preserved if it's present.

(For CSFirst version of Scratch)

ericrosenbaum:
### Resolves

resolves https://github.com/LLK/scratch-vm/issues/2851

### Proposed Changes

- Add a field to the runtime to store an `origin` string, to represent the location of the editor where the project was created, if it was not the Scratch community editor (e.g. the CSFirst editor).
- On load, if the json's meta object contains an origin field, store it in the runtime.
- On save, if the runtime has an origin value, write it in the json


rschamp (changes requested):
For tests, we could add to serialization_sb3 to check:
* when serializing/deserializing a project with the field set, things work properly
* when deserializing a project without the field after one with the field, it does _not_ persist

rschamp (approved):
Looks good! When we test this we should make sure everything also looks good in the full save/load process.

PR #3101, opened 2021-05-16
Cast: Easier to understand grammar

nonusx:
* value.substring(0,1) => value.charAt(0)
* Increase color value length judgment

### Resolves

### Proposed Changes

### Reason for Changes

### Test Coverage

### DIFF
--- a/src/util/cast.js
+++ b/src/util/cast.js
@@ -91,7 +91,7 @@ class Cast {
      */
     static toRgbColorObject (value) {
         let color;
-        if (typeof value === 'string' && value.substring(0, 1) === '#') {
+        if (typeof value === 'string' && (value.length === 7 && value.length === 4 /*"#" takes up one character*/) && value.charAt(0) === '#') {
             color = Color.hexToRgb(value);
 
             // If the color wasn't *actually* a hex color, cast to black


apple502j:
I'm sure value.length cannot be 7 and 4 at the same time?

joker314:
Thanks for taking the time to file the PR.

I think these checks are actually already made in `colour.js`, using [a regular expression](https://github.com/LLK/scratch-vm/blob/develop/src/util/color.js#L63). `colour.js` performs further checks, such as making sure that all the characters in the hexstring (bar the initial `#`) are valid hexadecimal digits, in addition to making sure that there are either 4 or 7 characters in total.

As such, the length check here isn't quite needed, because if `colour.js` reported an invalid colour, then it is treated as black (see #2356). This would be true with or without this PR, I think.

nonusx:
Thanks

nonusx:
> I'm sure value.length cannot be 7 and 4 at the same time?

Sorry, I wanted to write it as "||".

ericrosenbaum:
I'm going to close this PR, because it does not address a "help wanted" issue. Please have a look at our [contributor guidelines](https://github.com/LLK/scratch-vm/blob/develop/.github/CONTRIBUTING.md). Thanks!

PR #3152, opened 2021-06-09
Fix Editor Hangs By Using Queue for Threads to Start on Clone Creation

### LINKED ISSUE #2282: Creating and deleting a clone in the same When I start as a clone stack crashes Scratch
### Expected Behavior

If you build this stack:
Screen Shot 2019-09-23 at 11 40 22 AM
And click the green flag it should not crash Scratch.

### Actual Behavior

Scratch freezes and nothing you click is responsive.  After a minute or so it pops up with a "page not responsive" dialog:
Screen Shot 2019-09-23 at 11 46 53 AM
It may require attempting to interact with Scratch to get the dialog.

### Steps to Reproduce

Build this stack:
`when I start as a clone`
`create a clone of myself`
`delete this clone`

Run a 'create a clone of myself' block
Note: Scratch becomes unresponsive and after a minute a dialog appears.

Interestingly, if you put the `create a clone of myself` block in a forever loop and then have the stack `when I start as a clone` `delete this clone` the application does not crash.  The create and delete blocks must be in the same stack.  
Screen Shot 2019-09-23 at 11 51 55 AM

Similarly if you build two stacks one that is `When I start as a clone` `create a clone of myself` and a second stack that is `When I start as a clone` `delete this clone` it doesn't crash either.  Further evidence that the two must be in the same stack.

Screen Shot 2019-09-23 at 11 53 41 AM

Here is my repro project: https://scratch.mit.edu/projects/330894575/editor
It is a simplified version of one that came in from the community: https://scratch.mit.edu/projects/314578144

Mac Chrome

AmazingMech2418:
### Resolves

Fixes #2282

### Proposed Changes

This pull request creates a queue system for hat blocks that is used to trigger scripts starting with `when I start as a clone`  after sequencer step completion, in order to prevent the hanging of the editor from an infinite loop involving clone creation and deletion.

### Reason for Changes

Without this change, a script that follows the following script would create an infinite loop and hang the editor, crashing the tab.
scratchblocks (9)
*Image created with [s3blocks](https://s3blocks.github.io/)*

This change prevents the `when I start as a clone` hat blocks from firing until the next iteration to prevent the editor hang and page crash.

There are other potential solutions to this issue, such as using an event with the `EventEmitter` or making clone creation act as a control modifier, but the `EventEmitter` would be less efficient for the CPU and making clone creation act as a control modifier would harm project execution performance drastically, so this is the most efficient method for fixing this bug. 

### Test Coverage

I created a local copy of `scratch-gui` and my fork of `scratch-vm` to test project ID 541133287 (a testing project I used to analyze this bug's execution steps to identify the source) and the execution order of the Scratch project remained the same while not hanging the editor. Additionally, I tested with stopping the project with the stop button, and it worked. I also ran a few tests to ensure the same previous functionality was kept, including with stopping scripts programmatically and a test to make sure I didn't accidentally cause an equivalent of a forever loop on all clone scripts (which I accidentally did at first until I fixed it), and all tests succeeded after edits, ensuring the same execution as without this update, but with the bug fixed.

There were two things I noticed during tests though:
1. `stop [other scripts in this sprite]` does not stop scripts in clones both before and after my changes
2. `stop [other scripts in this sprite]` will not cancel the creation of a clone after my changes (since testing this before changes is not possible due to the bug in question)

I am not sure if these are intended since the intended results are quite ambiguous with this, but the issue seems to be that clones are described as separate targets than their parent sprites if they are unintended.

However, the tests I ran do seem to show that this PR solves the problem in #2282 and do not cause any other unintended bugs (other than potentially the second thing I noticed in tests).

**Note:** I know that this issue was not labelled `help wanted`, but since I did not get a response after a few days after saying I was willing to submit a PR, I decided to go ahead and submit this PR. This is a severe bug that could potentially even cause devices to overheat if this is run on devices with lower CPU power on a browser that doesn't have systems in place to prevent the page from hanging, so a fix coming sooner rather than later would be ideal.


ericrosenbaum:
Thanks for your work on this @AmazingMech2418. As you noted, this is not a help-wanted issue. One reason is that we consider changes in this area of the code to be quite risky, requiring a lot of time from the Scratch team. We don't currently have that time. When we do return to look at the issue it will be useful to have your fix to look at for reference, but I'm going to close this for now. Thanks!

PR #3194, opened 2021-07-04
Music extension: chords

PauRosello97:
### Proposed Changes

This Pull Requests creates a new block to play chords. 

### Reason for Changes

Scratch is a great platform to create animations, games or explain stories, and it has a great potential for composing music. Being able to work with chords is a really basic functionality that will help users to learn music with Scratch. 

https://user-images.githubusercontent.com/49373982/124401581-766aae80-dd2a-11eb-8931-35eb468ee020.mp4

### DIFF
--- a/src/extensions/scratch3_music/index.js
+++ b/src/extensions/scratch3_music/index.js
@@ -539,6 +539,33 @@ class Scratch3MusicBlocks {
     }
 
     /**
+     * An array of info about each chord.
+     * @type {object[]}
+     * @param {string} name - the translatable name to display in the chords menu.
+     * @param {int[]} notes - the notes of the chord.
+     */
+    get CHORD_INFO () {
+        return [
+            {
+                name: formatMessage({
+                    id: 'music.chordMajor',
+                    default: '(1) Major',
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
Thanks for your work on this, @PauRosello97. We're not going to merge it, unfortunately. 

As you are probably aware, you can already play chords in several ways, but I'll describe some of those here just in case:

- You can trigger multiple stacks at once, like in this example using a keypress event: 
image
- You can also use broadcasts, like this: 
image
- A totally different way is to use the "start sound" blocks (as opposed to "play sound until done"): 
image

I like the simplicity of having a single block to play chords, of course. There are other improvements to that which we might make (to allow for other chord qualities and variations). More generally there are lots of improvements we've imagined over the years to the music extension (e.g. to make it easy to use musical scales, simpler representations of rhythm values, or ways to improve synchronization, etc.). Changes like that need to be carefully considered by our design team and iteratively tested with children and refined before we consider releasing them to the Scratch community.

PauRosello97:
Thank you!

PR #3214, opened 2021-07-20
Add support for layers in the EV3 extension.

kadhonn:
This works by adding a new block which allows you to choose a layer + port as a motor. Under the hood pretty much everything was implemented already, i just needed to change some loops to include a bigger range and map the indexes correctly.

### DIFF
--- a/src/extensions/scratch3_ev3/index.js
+++ b/src/extensions/scratch3_ev3/index.js
@@ -92,7 +92,6 @@ const Ev3Opcode = {
  * @enum {number}
  */
 const Ev3Args = {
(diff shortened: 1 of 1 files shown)


ericrosenbaum:
Thanks for your work on this @kadhonn, but we're not going to merge this change. We make careful tradeoffs in the design of Scratch extensions, especially focusing on making them simple and understandable for beginners, and keeping the total number of blocks relatively small. This often means omitting features, especially if they might mostly be used for advanced projects.

kadhonn:
Ahh, sad for me but understandable, thanks for explaining @ericrosenbaum !

PR #3601, opened 2022-04-13
Added function to stop thread if a running script is removed from or attached to another script to resolve #2263

### LINKED ISSUE #2263: Disconnecting a running stack leads to broken behavior
### Expected Behavior

Clicking a running (highlighted) stack should always stop it.

### Actual Behavior

If you disconnect a running stack, so that e.g. a forever loop is running in the detached part, and then click the still-running fragment, it does not stop. Instead, there are now two "copies" of the stack running in the VM (for example, if the stack causes a turn of 10 degrees per frame, it will now turn 20 degrees per frame). Clicking again stops and re-starts this second version, without affecting the first.

disconnect-bug2

### Steps to Reproduce

- Run a stack including a forever loop in it
- Detach the part containing the forever loop, which stays highlighted and running
- Click the detached stack. Note that it stays highlighted, and is running "twice", so e.g. its side effect is occurring twice as fast.
- Click the detached stack again. Still highlighted, running "once".

cblesy:
### Resolves

Resolves #2263

### Reason for Changes

Attaching and detaching a running loop from a script both lead to broken behavior. As described in #2263 originally, a second "copy" of the running script seems to be created when the script is toggled. This behavior exists because the toggleScript() function in runtime.js pushes a new thread if the top block of a script stack of blocks doesn't already exist as a running thread. The issue occurs when a running script is attached to or removed from another script because the top block with which the original script was run does not match the top block of the script after the running loop block has been moved. Thus, two "copies" of the loop can be run as two different threads with two different top blocks at the same time.

### Proposed Changes

To address this issue, I defined a new function, stopMovingThread(), in runtime.js to stop the thread currently running a script when that script is either removed from or attached to another script. This function is called by moveBlock() in blocks.js each time a block is moved. Thus, this change allows us to prevent similar threads from running the same scripts and resolves #2263.

### DIFF
--- a/src/engine/blocks.js
+++ b/src/engine/blocks.js
@@ -213,7 +213,7 @@ class Blocks {
     getTopLevelScript (id) {
         let block = this._blocks[id];
         if (typeof block === 'undefined') return null;
-        while (block.parent !== null) {
+        while (block.parent !== null && typeof this._blocks[block.parent] !== 'undefined') {
             block = this._blocks[block.parent];
         }
         return block.id;
@@ -710,6 +710,11 @@ class Blocks {
             block.x = e.newCoordinate.x;
             block.y = e.newCoordinate.y;
         }
+
+        // Stops a thread if a running script is removed from or attached to another script
+        const destBlockId = (typeof e.newParent === 'undefined') ? e.id : this.getTopLevelScript(e.newParent);
+        const originBlockId = (typeof e.oldParent === 'undefined') ? e.id : this.getTopLevelScript(e.oldParent);
+        this.runtime.stopMovingThread(destBlockId, originBlockId);
 
         // Remove from any old parent.
         if (typeof e.oldParent !== 'undefined') {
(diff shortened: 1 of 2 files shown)


cblesy:
ScratchDemo

This is a demo that shows program behavior after the changes I implemented to resolve the issue as discussed above.

ericrosenbaum:
Thanks for your contribution! Our team will have to look at the issue again to determine exactly what behavior we want to resolve it (which may not be the same as your change). It's hard to predict when we will get to that. I just did a quick test and found that the same exact bug exists in Scratch 2 and Scratch 1! So that's pretty funny, and perhaps means it will be hard to fully fix. Also, this change is in a risky area of our code (i.e. changes have complex implications), and we usually only accept PRs for issues marked "help wanted". Please see the [contributor guidelines](https://github.com/LLK/scratch-vm/blob/develop/.github/CONTRIBUTING.md).

PR #3619, opened 2022-04-29
correctly handle JSON-RPC requests with ID=0

cwillisf:
### Proposed Changes

Check the JSON-RPC request ID against `null` and `undefined` rather than just checking if it's truthy.

### Reason for Changes

The old code would skip responding to a request with ID=0, which is in violation of the JSON-RPC spec.

### Test Coverage

Tested locally:
* in the Scratch Link playground with Scratch Link 2.0
* in scratch-gui with Scratch Link 1.4.3


kchadha (approved):
LGTM

PR #3637, opened 2022-05-13
Handle errors while loading vector costumes or sounds

### LINKED ISSUE #6539: Try to prevent you from messing up your project by saving over your assets with question mark assets
Sometimes we have bugs that cause users' projects to be broken because costumes have turned into question marks. Often these projects are salvageable because the server still has all the data from when the project was working, and you can just wait for the client to get fixed. But if the user changes anything in their project, there's no way for them to recover the working version of the project on their own.

When we load a project where we've converted costumes into question marks, the client could do something to try to prevent this. Maybe when you try to save it can say hey, some of your assets seem to be broken. Are you sure you want to save? Maybe when we load the question mark asset, we can save the original version of the asset as metadata, in case it becomes readable in the future and we can switch back to it.

kchadha:
### Resolves

- Resolves LLK/scratch-gui#6539
- Closes #3025 as they are addressing the same issue.

### Proposed Changes

While loading a vector costume, handle any errors in the process by displaying a gray question mark (same existing behavior) but with this PR, we now we keep track of the original costume data that failed to load.

When serializing the project out, we save the original (potentially broken) costume data instead of the default costume. This prevents data loss from the gray question mark costume overwriting potentially recoverable data (e.g. if the asset server was down when we first tried to load the costume).

Do not save any info about the costume being broken so that we make sure to try and load the costume again if we reload the project.

While the original costume contents are preserved, costume metadata may be fixed in the process of attempting to load the costume (e.g. if the data in the project.json like the costume md5 does not match the actual costume contents).

#### Update after review cycle and release of storage v2.0.0 & v2.0.1
- Storage was previously returning an HTML 404 page instead of null when the requested asset was missing
- Storage was also swallowing errors that it's not forwarding up the chain

This PR pulls in the fixes for these above issues via storage v2.0.1 and addresses handling storage now possibly returning null. This resulted in adding error handling while loading sounds as well. The same behavior described above about using the default sound but serializing what we know about the original sound is used here.

### Reason for Changes

This fixes a critical issue that can lead to user data loss.

### Test Coverage

- Added integration tests for .sb3 projects and .sprite3 files with missing or corrupted vector costumes.
- Added integration tests for .sb3 projects with missing or corrupted sounds


src/serialization/sb3.js
```diff
@@ -345,18 +345,23 @@ const serializeBlocks = function (blocks) {
...
     // what it's actually referring to. TODO runtime objects need to be
     // updated to actually refer to this as 'md5ext' instead of 'md5'
     // but that change should be made carefully since it is very
     // pervasive
-    obj.md5ext = costume.md5;
+    obj.md5ext = (costume.broken && costume.broken.md5) || costume.md5;
     obj.dataFormat = costume.dataFormat.toLowerCase();
-    obj.rotationCenterX = costume.rotationCenterX;
-    obj.rotationCenterY = costume.rotationCenterY;
+    if (costume.broken) {
```
cwillisf: Minor: seems like moving the `(costume.broken && foo) || bar` items from above into this `if` could make this function easier to read

src/serialization/serialize-assets.js
```diff
@@ -16,10 +16,18 @@ const serializeAssets = function (runtime, assetType, optTargetId) {
...
             const currAsset = currAssets[j];
-            const asset = currAsset.asset;
+            let asset = currAsset.asset;
+            if (currAsset.broken) {
+                if (currAsset.broken.asset) {
+                    asset = currAsset.broken.asset;
+                } else {
+                    continue;
+                }
+            }
```
cwillisf: How about this as a simpler version? It also covers the case where `currAsset.broken` and `currAsset.asset` are both falsy. I think with your new code that would never happen but if it does, this suggestion would just skip the asset instead of breaking serialization altogether.
```suggestion
            const asset = currAsset.broken ? curAsset.broken.asset : currAsset.asset;
            if (!asset) {
                continue;
            }
```

src/serialization/serialize-assets.js
```diff
@@ -16,10 +16,18 @@ const serializeAssets = function (runtime, assetType, optTargetId) {
...
             assetDescs.push({
                 fileName: `${asset.assetId}.${asset.dataFormat}`,
-                fileContent: asset.data});
+                fileContent: asset.data
+            });
```
cwillisf: 😄

src/virtual-machine.js
```diff
@@ -448,12 +446,13 @@ class VirtualMachine extends EventEmitter {
...
 
     /**
-     * Export project as a Scratch 3.0 JSON representation.
+     * Export project or sprite as a Scratch 3.0 JSON representation.
+     * @param {string=} optTargetId - Optional id of a sprite to serialize
      * @return {string} Serialized state of the runtime.
      */
-    toJSON () {
+    toJSON (optTargetId) {
         const sb3 = require('./serialization/sb3');
```
cwillisf, line 458: Shouldn't this be at the top level? Why isn't lint complaining about this?

src/virtual-machine.js
```diff
@@ -932,6 +931,7 @@ class VirtualMachine extends EventEmitter {
      */
     updateSvg (costumeIndex, svg, rotationCenterX, rotationCenterY) {
         const costume = this.editingTarget.getCostumes()[costumeIndex];
+        if (costume && costume.broken) costume.broken = null;
```
cwillisf:
```suggestion
        if (costume && costume.broken) delete costume.broken;
```

test/fixtures/readProjectFile.js
```diff
@@ -12,5 +12,10 @@ module.exports = {
             return JSON.parse(zip.readAsText(projectEntry.entryName, 'utf8'));
         }
         return null;
+    },
+    extractAsset: function (path, assetFileName) {
+        const zip = new AdmZip(path);
+        const assetEntry = zip.getEntries().filter(item => item.entryName.match(assetFileName))[0];
```
cwillisf:
```suggestion
        const assetEntry = zip.getEntries().find(item => item.entryName.match(assetFileName));
```
(I know this is just following the pattern above but probably both should be using `find` instead of `filter`...)

test/integration/sb3_corrupted_svg.js
```diff
@@ -0,0 +1,107 @@
...
+
+let vm;
+let defaultVectorAssetId;
+
+tap.beforeEach(() => {
+    const storage = makeTestStorage();
+
+    vm = new VirtualMachine();
+    vm.attachStorage(storage);
+    defaultVectorAssetId = vm.runtime.storage.defaultAssetId.ImageVector;
```
cwillisf, line 35: Mmmm... that fresh VM smell!

test/integration/sb3_corrupted_svg.js
```diff
@@ -0,0 +1,107 @@
...
+    
+    const corruptedCostume = blueGuySprite.getCostumes()[0];
+    t.equal(corruptedCostume.name, 'costume1');
+    t.equal(corruptedCostume.assetId, defaultVectorAssetId);
+    t.equal(corruptedCostume.dataFormat, 'svg');
+    // Runtime should have info about broken asset
+    t.ok(corruptedCostume.broken);
+    t.equal(corruptedCostume.broken.assetId, brokenCostumeMd5);
+    // Verify that we saved the original asset data
+    t.equal(md5(corruptedCostume.broken.asset.data), brokenCostumeMd5);
```
cwillisf, line 71: 👍

test/integration/sb3_corrupted_svg.js
```diff
@@ -0,0 +1,107 @@
...
+    t.equal(blueGuySprite.name, 'Blue Square Guy');
+    t.equal(blueGuySprite.costumes.length, 1);
+    
+    const corruptedCostume = blueGuySprite.costumes[0];
+    t.equal(corruptedCostume.name, 'costume1');
+    // Resaved project costume should have the metadata that corresponds to the original broken costume
+    t.equal(corruptedCostume.assetId, brokenCostumeMd5);
+    t.equal(corruptedCostume.dataFormat, 'svg');
+    // Test that we didn't save any data about the costume being broken
+    t.notOk(corruptedCostume.broken);
```
cwillisf, line 94: Is it valid to say "nice catch" for a detail in a big block of new code? 😅

Seems like it could have been easy to forget to check for `broken` being removed. Nice catch!

test/integration/sb3_missing_svg.js
```diff
@@ -0,0 +1,87 @@
...
+
+tap.beforeEach(() => {
+    const storage = makeTestStorage();
+    
+    // This line removes the webhelper from the list of available helpers.
+    // W/o the following line, this fails because storage doesn't handle the case
+    // where none of the tools have isGetSupported: true
+    // TODO: Remove this line when the related storage bug is resolved so that
+    // storage gracefully handles non-browser situations where assets are missing.
+    storage._helpers = [storage._helpers[0]];
```
cwillisf: Is this still necessary with storage @ 1.3.6?

test/integration/sb3_missing_svg.js
```diff
@@ -0,0 +1,87 @@
...
+    t.equal(blueGuySprite.costumes.length, 1);
+    
+    const missingCostume = blueGuySprite.costumes[0];
+    t.equal(missingCostume.name, 'costume1');
+    // Costume should have both default cosutme (e.g. Gray Question Mark) data and original data
+    t.equal(missingCostume.assetId, missingCostumeAssetId);
+    t.equal(missingCostume.dataFormat, 'svg');
+    // Test that we didn't save any data about the costume being broken
+    t.notOk(missingCostume.broken);
+
```
cwillisf, line 78: Would it make sense to add a line verifying that there is no file content associated with the missing costume?

test/integration/sprite3_missing_svg.js
```diff
@@ -0,0 +1,85 @@
...
+    t.equal(resavedSprite.costumes.length, 1);
+    
+    const missingCostume = resavedSprite.costumes[0];
+    t.equal(missingCostume.name, 'costume1');
+    // Costume should have both default cosutme (e.g. Gray Question Mark) data and original data
+    t.equal(missingCostume.assetId, missingCostumeAssetId);
+    t.equal(missingCostume.dataFormat, 'svg');
+    // Test that we didn't save any data about the costume being broken
+    t.notOk(missingCostume.broken);
+
```
cwillisf, line 76: Same comment about verifying the absence of file content

cwillisf (commented):
Looking good -- I especially like the extensive tests!
I marked several items for discussion but it's all little stuff.

src/virtual-machine.js
kchadha: What do you mean?

kchadha: oh the import line... we don't have a lint rule for that and explicitly do requires elsewhere in the file sometimes. Mostly it's been for async require purposes (like loading the different extensions). I assume this particular one was a performance optimization that was made..?

test/integration/sb3_corrupted_svg.js
kchadha: ❤️

test/integration/sb3_corrupted_svg.js
kchadha: 😄

test/integration/sb3_missing_svg.js
kchadha: Yeah unfortunately this is still necessary, I checked the tests without it. I think we need to take a closer look into what's going on with the web helper...

test/integration/sb3_missing_svg.js
kchadha: Hmmm.. what did you have in mind?

kchadha: I might be able to have a separate test in here that runs serializeCostume (like I do in the corrupt svg tests), but test that it doesn't serialize anything for the missing costume

src/serialization/serialize-assets.js
kchadha: Yes I think that makes sense! Thanks!

src/serialization/sb3.js
kchadha: Done!

test/integration/sb3_missing_svg.js
kchadha: Added a new test that `serializeCostumes` correctly skips the costume that was missing.

test/integration/sprite3_missing_svg.js
kchadha: Added a new test that `serializeCostumes` correctly skips the costume that was missing.

src/serialization/serialize-assets.js
```diff
@@ -16,10 +16,14 @@ const serializeAssets = function (runtime, assetType, optTargetId) {
         const currAssets = currTarget.sprite[assetType];
         for (let j = 0; j < currAssets.length; j++) {
             const currAsset = currAssets[j];
-            const asset = currAsset.asset;
+            const asset = currAsset.broken ? currAsset.broken.asset : currAsset.asset;
+            if (!asset) {
```
kchadha: Oh I think I have a bug here, this should actually be:
```suggestion
            if (!(asset && asset.data)) {
```
so that we don't serialize zero byte files... (e.g. if the original costume was missing, it should still be missing, it shouldn't have an empty file in the re-saved .sb3 in its place)

kchadha: Though actually I'll just write this in the affirmative so that it's not confusing

e.g.
```
if (asset && asset.data) {
    assetDescs.push(....);
}
```

src/virtual-machine.js
apple502j: Could be. I'm not sure how well it could act as a performance optimization since the VM has to deserialize SB3 files during project loading (which is handled by the same file) anyway.

test/integration/sb3_missing_svg.js
apple502j: The behavior described above is implemented at https://github.com/LLK/scratch-storage/blob/develop/src/ProxyTool.js#L49

What we probably need is a Node.js-only, last-resort tool that returns `Promise.resolve(null)`.

src/virtual-machine.js
kchadha: Yeah, FWIW, it was introduced as part of #1956 in this commit: https://github.com/LLK/scratch-vm/commit/bc5605cf807a2b8b9c1b41253e8e89baad0fa266

I don't think I'll update this line of code in this PR since it's not relevant to the gray question mark work.

src/serialization/serialize-assets.js
kchadha: Okay! Instead of the above change, I made LLK/scratch-storage#388 to handle some inconsistencies and bugs in the way storage resolves assets (especially missing assets). That should eliminate the need for the above change I am talking about to add `&& asset.data`

kchadha: Storage 2.0.0 and 2.0.1 bring in these fixes

src/import/load-costume.js
```diff
@@ -298,7 +306,7 @@ const loadCostumeFromAsset = function (costume, runtime, optVersion) {
     costume.assetId = costume.asset.assetId;
     const renderer = runtime.renderer;
     if (!renderer) {
-        log.error('No rendering module present; cannot load costume: ', costume.name);
+        log.warn('No rendering module present; cannot load costume: ', costume.name);
```
cwillisf, line 309: 😌
<sub>(that's allegedly "relieved" 🤷)</sub>

src/import/load-sound.js
```diff
@@ -61,6 +95,12 @@ const loadSound = function (sound, runtime, soundBank) {
         runtime.storage.load(runtime.storage.AssetType.Sound, md5, ext)
     ).then(soundAsset => {
```
cwillisf: Does this need a `catch` that ends up at `handleSoundLoadError`?

src/serialization/sb3.js
```diff
@@ -345,23 +345,25 @@ const serializeBlocks = function (blocks) {
  */
 const serializeCostume = function (costume) {
     const obj = Object.create(null);
-    obj.assetId = (costume.broken && costume.broken.assetId) || costume.assetId;
     obj.name = costume.name;
-    obj.bitmapResolution = costume.bitmapResolution;
+
+    const costumeToSerialize = costume.broken || costume;
```
cwillisf, line 350: 😎

src/import/load-sound.js
cwillisf: Actually, thinking about this further: we should probably have tests that verify `loadCostume` and `loadSound` return default assets (with original data in `.broken`) if `storage.load` rejects. That might be out of scope for the GQM project, though, as long as what happens now is anything other than a data-loss-by-default-asset situation. (In this case I think it might be an unhandled exception => BSOD?)

src/virtual-machine.js
```diff
@@ -366,7 +366,11 @@ class VirtualMachine extends EventEmitter {
         const vm = this;
         const promise = storage.load(storage.AssetType.Project, id);
         promise.then(projectAsset => {
-            vm.loadProject(projectAsset.data);
+            if (!projectAsset) {
+                log.error("Failed to fetch project with id: ", id);
```
cwillisf: lint error here with the double-quotes
```suggestion
                log.error(`Failed to fetch project with id: ${id}`);
```

kchadha: oh oops! I thought I ran the tests again but I think I probably added that change later

test/integration/sb3_missing_svg.js
kchadha: The errors were unfortunately not being handled correctly in `WebHelper.load` but that has now been fixed in scratch-storage v2.0.0

src/import/load-sound.js
kchadha: I decided to leave the error case out so as not to complicate this PR (which was supposed to be focused on vector costumes) too much more. This is similar to how I left off adding a `.catch` to `loadBitmap_` for a later PR as well.

I think there should be a separate PR for broken bitmap costumes and a separate PR for corrupted sounds (or general sound error handling). I just wanted to make sure that we are handling storage.load resolving null in the same PR where we're bringing in the update to storage 2.0.1.

I am hoping to get those other 2 PRs in soon (barring complications I run into testing bitmaps..)

src/serialization/sb3.js
kchadha: yeah I had been itching to do this for a while but it didn't seem worth it for just 1 or 2 properties but now it's basically all of them! =D

src/import/load-sound.js
kchadha: Nvm! I *will* address corrupted sound stuff in this PR, since an error thrown here could result in the editor crashing, which might be more likely now that storage is swallowing less errors.

Also I assume you're talking about having unit tests for loadCostume and loadSound in addition to the integration tests? I might add those after this PR.

kchadha: Added https://github.com/LLK/scratch-vm/pull/3637/commits/8ab21dd701a6c2b20260c30cd4994a2e4eb75900 to address the sound error handling stuff

cwillisf (approved):
LGTM! Thanks for your persistence!

kchadha:
🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉  Wohooo!!! 🥳 

Thanks for all your help with this work @cwillisf!!! The ambient (and real) pairing was a lot of fun and very much appreciated! ❤️

PR #3660, opened 2022-05-23
Handle errors from loading missing or broken bitmap costumes

### LINKED ISSUE #6838: Don't replace costumes that have failed to load with ? sprites
### Expected Behavior

Costumes that fail to load should not replace the asset in your project.  When it fails to load or display, it should temporarily put a placeholder asset (the ? ) onto the stage, but not actually replace the content of the costume.

When the project saves, since the ? asset is not actually part of the project it would not overwrite the original asset in the project.json or the asset on the server.  Design will be needed to figure out what to do when the project is saved locally.

There should be some indication of the problem in the costumes tab when the costume is selected, including the costume tile and something in the paint editor itself when it is selected.

This should be the case for both vector and bitmap costumes that fail to load/display.  

### Actual Behavior

Currently, when an svg fails to load it is replaced with a question mark (?) placeholder asset.   When we do this it actually replaces the asset in the scratch project and saving the project changes the asset id in the project.json to the ? image.  This is destructive to people's work, especially if the failure to load is temporary or a regression that is fixed after a short time.

Also, when a bitmap fails to load it causes a BSOD (Blue Screen of Death), preventing the whole project from loading.

kchadha:
### Resolves

Resolves LLK/scratch-gui#6838 and possibly others

### Proposed Changes

Similar to #3637 but for bitmap costumes.

Tried to change loading flow as minimally as possible while fixing the issue where the editor BSODs when a bitmap costume fails to load. This should now result in the project loading with a gray question mark displaying in place of the broken costume. The gray question mark should not get saved in place of the original broken or missing costume.

### Reason for Changes

Fixing a BSOD and finishing up the gray question mark work.

### Test Coverage

Added tests for broken and missing bitmap costumes in .sb2 and .sprite2 files. Scratch 3 related tests coming soon.


src/import/load-costume.js
```diff
@@ -176,7 +176,7 @@ const fetchBitmapCanvas_ = function (costume, runtime, rotationCenter) {
                 assetMatchesBase: scale === 1 && !textImageElement
             };
         })
-        .catch(() => {
+        .finally(() => {
```
kchadha, line 179: Errors will get caught below, but this change does mean that we will no longer be swallowing errors with loading the text layer... the costume will turn into a gray question mark if the text layer fails to load, but separating out the intermingled code paths was seeming a little more risky than displaying a gray question mark if the text layer failed to load...

src/virtual-machine.js
```diff
@@ -877,6 +877,7 @@ class VirtualMachine extends EventEmitter {
     updateBitmap (costumeIndex, bitmap, rotationCenterX, rotationCenterY, bitmapResolution) {
         const costume = this.editingTarget.getCostumes()[costumeIndex];
         if (!(costume && this.runtime && this.runtime.renderer)) return;
+        if (costume && costume.broken) delete costume.broken;
```
cwillisf, line 880: ooh, good catch!

cwillisf (approved):
So much test! Congratulations on surviving all that ;)

I would almost suggest making a `doGqmTest('sb2' / 'sb3', 'svg' / 'png' / 'wav', 'missing' / 'corrupted');` or whatever but we're here now, and I suspect the small differences would make that more trouble than it's worth.

src/import/load-costume.js
cwillisf: Hmm... looks like `finally` isn't being accepted by `tap`. Maybe it's only supported by newer Node? I'm having trouble looking up info on that.

src/virtual-machine.js
kchadha: We had a similar thing for `updateSvg` too

kchadha:
Rebased off of develop which pulls in the `node @ lts` change, so the tests are no longer failing because of the `.finally`

cwillisf (approved):
👍 Glad that `.finally` worked ;)

PR #3661, opened 2022-05-23
ci(node): update CI config to use Node LTS (was 8 & 10)

cwillisf:
### Proposed Changes

For the Node version number in the Travis CI config, specify `lts/*` instead of 8 and 10.

### Reason for Changes

We should keep up to date with LTS versions of Node, which are currently 14 & 16. Ideally I'd like to test both, but IMO keeping up to date is more important than testing more than one version of Node... especially with how stable Node is these days.

### Test Coverage

See CI build of this PR :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 node_js:
-- 8
-- 10
+- lts/*
 env:
   global:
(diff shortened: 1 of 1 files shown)


kchadha (approved):
Nice!! LGTM

PR #3667, opened 2022-05-25
Delete sound.broken when updating the sound buffer

kchadha:
We don't want to save info about the sound being broken if the user managed to make an update to the sound. This is just for completeness of the sound loading error handling work from #3637


cwillisf (approved):
👍

PR #3679, opened 2022-06-07
Migrate to CircleCI & use semantic release

cwillisf:
### Proposed Changes

Main changes:

* Migrate from Travis CI to Circle CI
* Use `semantic-release`

Supporting changes:

* Use Node LTS for CI builds
* Upgrade `package-lock.json` from version 1 to version 2
* Add `commitlint` (with `husky`)
* Add `commitizen` settings
* Upgrade `tap`
* Fix tests which are incompatible or unreliable with new `tap`

### Reason for Changes

* As an organization we're migrating to CircleCI
* `semantic-release` will help us follow `semver` instead of continuing with `0.x.x.prerelease-${DATE}`
* Node LTS will help us avoid deprecated versions of Node
* All recent versions of Node use `package-lock.json` version 2, so upgrading the lockfile eliminates a warning. In theory it also speeds up `npm ci`.
* New versions of `tap` parallelize tests more aggressively and seem to run tests "more asynchronously" so several tests which relied on careful timing began to fail after the upgrade. The individual fixes are described in the related commits.

### Test Coverage

I simulated both a normal release and a hotfix release with `semantic-release --dry-run`:

* Hotfix: https://app.circleci.com/pipelines/github/LLK/scratch-vm/24/workflows/92fde949-dcb0-4f4a-b89b-eb8edbf994e4/jobs/26
* Normal: https://app.circleci.com/pipelines/github/LLK/scratch-vm/25/workflows/8a7d3929-6ceb-400d-aec8-4f9ede4cfd31/jobs/27

See also CI build results


.husky/.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
kchadha, line 1: Is this OS / Platform specific?

cwillisf: If there's no `.gitattributes` file, then the default git configuration on Windows will give CRLF line endings to these files. Unfortunately, that causes the hooks to fail because `sh` only accepts LF (even on Windows).

Explicitly specifying LF is only necessary for Windows, but it won't hurt other platforms.

package.json
```diff
@@ -7,27 +7,39 @@
   "homepage": "https://github.com/LLK/scratch-vm#readme",
   "repository": {
     "type": "git",
-    "url": "git+ssh://git@github.com/LLK/scratch-vm.git"
+    "url": "https://github.com/LLK/scratch-vm.git"
```
kchadha, line 10: ?

test/integration/addSprite.js
```diff
@@ -27,8 +27,8 @@ test('default cat', t => {
     vm.on('playgroundData', e => {
         const threads = JSON.parse(e.threads);
         t.ok(threads.length === 0);
+        vm.quit();
```
kchadha, line 30: Can this be run in a tap test clean up function instead? E.g. `t.afterEach()` or `t.teardown()`?

test/integration/clone-cleanup.js
```diff
@@ -4,7 +4,7 @@ const makeTestStorage = require('../fixtures/make-test-storage');
 const readFileToBuffer = require('../fixtures/readProjectFile').readFileToBuffer;
 const VirtualMachine = require('../../src/index');
 
-const projectUri = path.resolve(__dirname, '../fixtures/clone-cleanup.sb2');
+const projectUri = path.resolve(__dirname, '../fixtures/clone-cleanup.sb3');
```
kchadha: I'm a little concerned about this change with clone cleanup becoming an sb3 project instead of an sb2. Doesn't this test exist in part to check compatibility with 2.0 behavior?

test/integration/monitors_sb2_to_sb3.js
```diff
@@ -140,7 +140,6 @@ test('saving and loading sb2 project with monitors preserves sliderMin and slide
         t.equal(monitorRecord.targetId, null);
 
         t.end();
-        process.nextTick(process.exit);
```
kchadha, line 143: Does this one not need a `vm.quit()`?

test/unit/blocks_control.js
```diff
@@ -261,7 +261,8 @@ test('wait', t => {
     const args = {DURATION: .01};
     const waitTime = args.DURATION * 1000;
     const startTest = Date.now();
-    const threshold = 1000 / 60; // 60 hz
+    const thresholdSmall = 1000 / 60; // only allow the wait to end one 60Hz frame early
+    const thresholdLarge = 1000 / 3; // be less picky about when the wait ends, in case CPU load makes the VM run slowly
```
kchadha, line 265: Maybe these would be better named `thresholdMin` and `thresholdMax`?

test/unit/serialization_sb3.js
```diff
@@ -4,7 +4,7 @@ const VirtualMachine = require('../../src/index');
 const Runtime = require('../../src/engine/runtime');
 const sb3 = require('../../src/serialization/sb3');
 const readFileToBuffer = require('../fixtures/readProjectFile').readFileToBuffer;
-const exampleProjectPath = path.resolve(__dirname, '../fixtures/clone-cleanup.sb2');
+const exampleProjectPath = path.resolve(__dirname, '../fixtures/complex.sb2');
```
kchadha: 👀

test/integration/clone-cleanup.js
```diff
@@ -60,17 +54,19 @@ test('clone-cleanup', t => {
             break;
 
         case 3:
-            // The second batch of clones has been created and the main thread has ended
-            verifyCounts(10, 0);
+            // The second batch of clones has been created and the main thread is about to end
+            verifyCounts(10, 1);
```
kchadha, line 58: Why is this different from before?

kchadha (commented):
Did a general / cursory review of this PR. Overall looks good and there's a lot happening here, so thank you for making these updates!!

There's questions I have about the clone-cleanup test, and about optimizing the changes to add `vm.quit()` everywhere.

package.json
cwillisf: This is needed for `semantic-release`: it errors out otherwise. I don't remember what the error is exactly but I suspect it's related to the fact that, at least for a public repo, anyone (or any script) can access the repo over `https` but only authenticated users can access it through `git+ssh`.

test/integration/addSprite.js
cwillisf: In this case it could be, since `const vm` is at the outermost scope, but in several of the other tests the `vm` is created as a local variable so an `afterEach` or `teardown` won't have access to it.

In this particular file, only this one test starts the VM so I figured it should only `quit` in the one test. That way other tests will timeout & fail if they accidentally cause the VM to start or otherwise open a handle unexpectedly.

test/integration/clone-cleanup.js
cwillisf: Hmm... yeah, that's legit. I'll redo this with an SB2.

test/integration/clone-cleanup.js
cwillisf: Previously, case 3 ran after the "main" thread of the project had finished, so the expected thread count was zero. Now, case 3 runs as part of the last `say` block in that script, so the main thread is still running => one instead of zero.

Case 4 previously ran after the main thread finished, and still does, so it stays at zero.

test/integration/monitors_sb2_to_sb3.js
cwillisf: Nope! It never actually starts the VM. I should probably correct the comment below.

test/unit/blocks_control.js
cwillisf: That was my initial impulse, but `thresholdSmall` is used in two different ways: once as the amount to add to `timeEleapsed` as the failsafe to "emergency exit" the busy-wait in case `mockUtil.stackFrame.timer.timeElapsed()` isn't working correctly, and the other is as an amount to subtract from `timeElapsed` to establish the minimum allowed wait time. IMO, `thresholdMin` only fits the latter.

test/unit/serialization_sb3.js
cwillisf: I wanted to keep this as an SB2, so I swapped it when `clone-cleanup` turned into an SB3. I'll undo this change when I make `clone-cleanup` an SB2 again.

adroitwhiz:
Not sure if this should be part of this PR or fixed later, but it seems that none of the .sb3 `execute` tests are running, since `test/integration/execute.js` only searches for files ending in `.sb2`:
image

package.json
delasare: I'm wondering if you could still use SSH here and pass the credentials as the bot user - similar to the way we are doing commits with SJR and GUI (I think).

delasare (approved):
This looks good to me, given my limited knowledge. I've never used `tap` before. I'd love to take some time to understand that a bit more

The CI/CD stuff looks good. One comment regarding SSH vs HTTPS but I'm ok with merging and addressing at a later time.

I'm interested in using `commitlint` in SJR. :)

package.json
cwillisf: There probably is a way to get this to work, but I'm afraid it's not quite that simple. The credentials are already in place so that `semantic-release` can push the release back to GitHub, but even with that it still errors out with a `git+ssh` URL here.

I would argue that it's more correct to list this as `https` anyway. A contributor is going to be using `git clone` with whatever URL they prefer, probably on their own fork instead of LLK, regardless of the value of this field. Since contributors won't be using this field, that implies it's for people (or bots / scripts) which just want to read the repository, and (IMO) `https` is a better choice for that.

PR #3685, opened 2022-07-06
Fix "execute" integration tests

cwillisf:
### Proposed Changes

* Adjust `test/integration/execute.js` to look for files ending in `/\.sb[23]?$/i` instead of only `'.sb2'`.
* Alter test to use messages instead of timing: `order-changes-backwards-2-broadcast-and-wait-repeat-message.sb2`.

Code change in `order-changes-backwards-2-broadcast-and-wait-repeat-message.sb2`:

Before:
image

After:
image
(Plus `broadcast("sprite done")` in the other sprites)

### Reason for Changes

* We already had "execute" tests in SB3 format, but they weren't being run because the test only looked for SB2 files. This change means we now run all the projects with valid extensions.
* The `order-changes-backwards-2-broadcast-and-wait-repeat-message.sb2` test has been failing on CI recently. The test relied on `wait` blocks to cause enough delay for some work to finish, but it seems that the timing of that work is less consistent than it used to be. Adjusting the test to work based on messages instead of timing means that the test should be reliable in any environment under any amount of load.

### Test Coverage

It's hard to prove that a test is no longer flaky, but I'll try running the tests on CI several times.


cwillisf:
Thanks to @adroitwhiz for noticing that this test was only looking for `'.sb2'`!

benjiwheeler (changes requested):
This LGTM -- what should we look for after merging it?

benjiwheeler (approved):
LGTM

cwillisf:
> This LGTM -- what should we look for after merging it?

It looks like that particular test magically started being reliable again, but it's still better to not rely on a delay.

With the "execute" tests running SB3s now, we should see some of these numbers go up:
```
Suites:   ​123 passed​, ​123 of 123 completed​
Asserts:  ​​​3447 passed​, ​of 3447​
```

If I made an important mistake, we might see "execute" tests failing after this is merged.

cwillisf:
(Thanks, @adroitwhiz, for reminding me about this old PR!)

cwillisf:
```
Suites:   ​123 passed​, ​123 of 123 completed​
Asserts:  ​​​3455 passed​, ​of 3455​
```
:D

PR #3712, opened 2022-08-16
feat(extensions): support Scratch Link 2.0 browser extension

cwillisf:
### Proposed Changes

Add support for the upcoming Safari browser extension for Scratch Link 2.0:
- In the default Scratch Link socket factory, check for `ScratchLinkSafariSocket` first and create that if it's supported. Otherwise, create a `ScratchLinkWebSocket` as before.
- Add a step for initializing the Scratch Link 2.0 extension during Runtime construction.

### Reason for Changes

Safari is not currently compatible with Scratch Link 1.4, and I plan to fix that in Scratch Link 2.0 by adding a browser extension. This change adds support for that extension if it's present.

### Test Coverage

Tested locally with the in-development browser extension.
Existing VM tests also verify that these changes don't break linting or behavior in Node.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -399,6 +399,8 @@ class Runtime extends EventEmitter {
          * @type {?string}
          */
         this.origin = null;
+
+        this._initScratchLink();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


deepankarmalhan (approved):
LGTM!

Issue #2418, opened 2020-05-18
ENA-196: TaskQueue is unused


apple502j:
`TaskQueue` is unused. It was previously used as a rate limiter for extensions, but `RateLimiter` is used now.

cwillisf:
We're taking another look at rate limiting soon, and might decide to use this `TaskQueue` for this upcoming work. If we don't, we'll reopen #3815 to remove the unused code. Thanks!

PR #3815, opened 2022-12-01
ENA-196: Remove TaskQueue and related unused code

travissanderson:
### Resolves

https://github.com/LLK/scratch-vm/issues/2418

### Proposed Changes

Simply deletes TaskQueue as it was unused, outside of its own tests.

### Reason for Changes

Shrink the codebase, minimize cognitive overhead, etc.

### Test Coverage

Only removing code, no new tests to add


aoneill01:
Created ENA-196 to track this.

cwillisf:
We're taking another look at rate limiting soon, and might decide to use this `TaskQueue` for this upcoming work. If we don't, we'll reopen this to remove the unused code. Thanks!

PR #3819, opened 2022-12-12
feat: ability to clear flyout blocks when language changes

aoneill01:
### Resolves

- Resolves [ENA-146](https://scratchfoundation.atlassian.net/browse/ENA-146)

### Proposed Changes

- Add a `clearFlyoutBlocks()` to `VirtualMachine` that removes all of the blocks from `runtime.flyoutBlocks`
- Use Node 16 for CircleCI builds

### Reason for Changes

- See LLK/scratch-gui#8705 for details of the problem this is attempting to solve. `scratch-gui` needs a way to clear out the flyout blocks so that the virtual machine and `scratch-blocks` stays in sync during a language change.
- The `scratch-vm` CircleCI builds are currently broken. There may be a different [long-term fix](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported), but switching to Node 16 will allow this fix to proceed.

### Test Coverage

- Added unit test for `clearFlyoutBlocks` functionalty
- Tested integration with `scratch-gui` in LLK/scratch-gui#8705

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,7 +21,9 @@ commands:
             fi
 jobs:
   build_no_release:
-    executor: node/default # defaults to LTS
+    executor:
+      name: node/default
+      tag: '16.18'
     steps:
       - checkout
       - node/install-packages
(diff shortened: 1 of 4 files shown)


cwillisf (approved):
Looks good! Thanks!

hyperobject (approved):
lgtm!

PR #3912, opened 2023-03-08
Use script glow status to track whether the project is running

### LINKED ISSUE #1948: Edge-triggered(?) hat blocks make the green flag always look active

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/1948
Closes https://github.com/LLK/scratch-vm/pull/2419

### Proposed Changes

- Refactor _emitProjectRunStatus to take a simple Boolean "project running/not running" argument
- Count threads as "project running" only if they request a block glow
- Allow flyout blocks to request a block glow

### Reason for Changes

This fixes edge-activated hat blocks causing the project to always count as "running" and making the green flag appear active even if no scripts are running.

The system for tracking script glows is much more robust and accurate than the system that was previously used for determining if the project was running, so we should make use of that.

The rationale for each change is described in the corresponding commit message.

### Test Coverage

Tested manually, and existing tests pass.

Here's what you should look for when testing this:
- Regular event hat blocks (when key pressed, etc) should cause the green flag button to appear activated.
- Edge-activated hat blocks (when loudness, etc) with *no blocks attached* should never cause the green flag button to appear activated.
- Edge-activated hat blocks *with blocks attached* should cause the green flag button to appear activated when those attached blocks are running.
- All of the above should hold true even if another sprite is selected.
- Clicking a flyout block should cause the green flag button to appear activated. Blocks like "wait" and "glide" should make it appear activated for as long as the block is running.
- Block monitors (aka variable readouts) should never cause the green flag button to appear activated.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -248,10 +248,10 @@ class Runtime extends EventEmitter {
         this._scriptGlowsPreviousFrame = [];
 
         /**
-         * Number of non-monitor threads running during the previous frame.
-         * @type {number}
+         * Whether any non-monitor threads ran during the previous frame.
+         * @type {boolean}
          */
-        this._nonMonitorThreadCount = 0;
+        this._projectRanLastFrame = false;
 
         /**
          * All threads that finished running and were removed from this.threads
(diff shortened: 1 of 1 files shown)


adroitwhiz:
@wowfunhappy See if this fixes the issue you found with #2419.

Wowfunhappy:
@adroitwhiz It does, thank you!

PR #3924, opened 2023-03-14
All my sequencer refactorings, with commentary

### LINKED ISSUE #3031: Remove dead code path from `RuntimeScriptCache`

### LINKED ISSUE #3032: Remove dead code path from `BlocksExecuteCache.getCached`
The `BlocksExecuteCache.getCached` function returns a "cached" version of a block given a block container, a block ID, and an optional constructor function for the "cached block" type. Because all three call sites pass in that constructor function, we can remove the logic that handles the case where it's *not* passed in.

### LINKED ISSUE #3033: `BlockCached` contains many unnecessary member variables/properties
The `BlockCached` class is an internal representation of a block which precomputes a lot of execution data in order to make repeated execution faster (e.g. evaluating "shadow blocks", unrolling recursion). Currently, it contains many member variables which are never read outside of its constructor function. Those member variables could be turned into regular variables inside the constructor function in order to make the code clearer.

### LINKED ISSUE #3034: Block inputs are cached in two different places
PR #770 introduced caching of block inputs in the blocks container--each container caches the result of a `getInputs` call for a given block, and then returns the cached result.

However, a later PR (#917) added the `BlockCached` class, which stores its own `inputs` and removes the need to call the block container's `getInputs` method in the first place.

This makes the block container's `getInputs` caching unnecessary.

adroitwhiz:
### Resolves

Resolves https://github.com/LLK/scratch-vm/issues/3031
Resolves https://github.com/LLK/scratch-vm/issues/3032
Resolves https://github.com/LLK/scratch-vm/issues/3033
Resolves https://github.com/LLK/scratch-vm/issues/3034

Also removes a lot of tech debt and improves performance.

### Proposed Changes

This PR contains all the refactoring work I've done on the block sequencing and execution code, including the work I did to re-land mzgoddard's performance PRs #2145 and #2148. All commits, including his, contain my (hopefully) detailed commentary on what each change does.

### Reason for Changes

The VM is currently a massive pile of tech debt. Many of the people who originally worked on huge portions of the sequencer are no longer at Scratch, and the current code is hard to understand. My hope is that these refactors make it clearer how the sequencer works, both to anyone on the current team and to any third parties who want to replicate its behavior when implementing their own versions of Scratch.

These changes will also make it much easier to add future functionality to the sequencer--[event blocks with droppable inputs](https://github.com/LLK/scratch-vm/pull/2945), custom reporter blocks, etc. are all very hard to implement given the current state of the codebase.

Reviewing these changes would also be a good way to learn about the sequencer and how it works. I've done my best to make each change atomic, and I explain the previous spaghetti code that I'm removing at each step, so it might explain the sequencer better than just reading the current code.

### Test Coverage

This PR includes tests that mzgoddard added in #2145, as well as my own tests to ensure that https://github.com/LLK/scratch-gui/issues/5011 doesn't reoccur.


src/engine/execute.js
```diff
@@ -360,7 +353,7 @@ class BlockCached {
 
         // The final operation is this block itself. At the top most block is a
         // command block or a block that is being run as a monitor.
-        if (this._definedBlockFunction) {
+        if (typeof this._blockFunction !== 'undefined') {
```
towerofnix: Note that although the commentary describes using a local variable, here that would-be variable's value is just inlined. That means moving it down from being defined *ages* earlier, and instead in-place, immediately where it's used!

The `this._definedBlockFunction = ...` was out of place to begin with - the nearby comment describes `_isHat` and `_blockFunction`, but not `_definedBlockFunction` (so `_definedBlockFunction` feels tacked on), and the point of the earlier calls was to avoid a dynamic lookup. Since `this._blockFunction` is, by definition, a static value, it's just fine to bring this `typeof _blockFunction !== 'undefined'` call further down - it'll have the same result as being evaluated earlier.

src/engine/execute.js
```diff
@@ -215,12 +209,6 @@ class BlockCached {
          */
         this._shadowValue = null;
 
-        /**
-         * A copy of the block's inputs that may be modified.
-         * @type {object}
-         */
-        this._inputs = Object.assign({}, this.inputs);
-
```
towerofnix, line 247: I'm noting that `this.inputs` and `this._inputs` are treated as the same for the later code. This seems to be at odds with the original behavior, where `this.inputs` was literally the original block's `inputs` dictionary, and `this._inputs` was a safely mutable copy. Is it really okay to mutate, with `delete` statements, `cached.inputs`?

Well, yes. The follow-up commits reveal:

bcd1530870b69b73358988afb817e44b21a807af: `getNonBranchInputs` (née `getInputs`) *was* a cached function (note how `Blocks` tries to hit `this._cache.inputs[block.id]`)...

192e499807f577a0cec00353af69c5eead0239af: But in fact there's no need to cache it - `BlockCached` is already in charge of that, more appropriately. No caching means it's a fresh object each time `BlockCached` is called and is never accessed outside of `BlockCached`, and so is safe for `BlockCached` to mutate.

b316ab8450aae5e7ff960019534550005bbd43f5: Except surprise! Then we pass the whole `block` in, directly. Now `BlockCached`, rather than `Blocks`, is responsible for calling `getNonBranchInputs`... meaning it's the one making the assumption that this function returns a new (safely mutable) object each time. [And this assumption is laid out bare.](https://github.com/scratchfoundation/scratch-vm/pull/3924/commits/b316ab8450aae5e7ff960019534550005bbd43f5#diff-891cfb1d04f97b904176b7f772916ea88c39fb030ebadb7a60e0b64e46d5aed4R247-R248)

src/engine/sequencer.js
```diff
@@ -77,7 +77,7 @@ class Sequencer {
         // Start counting toward WORK_TIME.
         this.timer.start();
         // Count of active threads.
-        let numActiveThreads = Infinity;
+        let activeThreads = true;
```
towerofnix, line 78: `activeThreads` is *initially* set to `true`, but only so that the `while` loop below runs at all.

src/engine/sequencer.js
```diff
@@ -96,48 +96,46 @@ class Sequencer {
                 this.runtime.profiler.start(stepThreadsInnerProfilerId);
             }
 
-            numActiveThreads = 0;
+            activeThreads = false;
```
towerofnix, line 97: `activeThreads` is set to false here, at the start of each iteration of the step loop. This is also the *only* place where `activeThreads` is set to false. So by definition, it'll only come *out* of the current iteration as `false` if nothing sets it to `true` (before the iteration is completed).

src/engine/sequencer.js
```diff
@@ -96,48 +96,46 @@ class Sequencer {
...
+                // it completed while it just stepped to make sure remove it
+                // before the next iteration of all threads.
                 if (activeThread.status === Thread.STATUS_RUNNING) {
-                    numActiveThreads++;
-                }
-                // Check if the thread completed while it just stepped to make
-                // sure we remove it before the next iteration of all threads.
-                if (activeThread.stack.length === 0 ||
-                    activeThread.status === Thread.STATUS_DONE) {
+                    activeThreads = true;
```
towerofnix, line 148: And here's the only place where `activeThreads` is set to true. Keep in mind this is inside a nested `for` loop which iterates over all the threads that were active as of the last iteration of the step loop.

At a high level this means the call to `stepThread(activeThread)` is responsible for setting the current thread's `status` to `Thread.STATUS_DONE`, and all `activeThreads` does is reflect if that result was *not* `STATUS_DONE` for any of the threads, during this iteration of the main `stepThreads` loop.

src/engine/sequencer.js
```diff
@@ -155,12 +153,11 @@ class Sequencer {
...
+                    if (thread.status === Thread.STATUS_DONE) {
+                        doneThreads.push(thread);
+                    } else {
                         this.runtime.threads[nextActiveThread] = thread;
                         nextActiveThread++;
-                    } else {
-                        doneThreads.push(thread);
                     }
                 }
                 this.runtime.threads.length = nextActiveThread;
```
towerofnix, line 174: Now here's the interesting part. This is an "old fashioned" filter which overwrites the array in-place as it iterates over it (it'll always be done with the current item before that item gets overwritten), setting `.length` at the end to "chop off" the garbage items. It's semantically the same as:

```js
doneThreads.push(
  ...this.runtime.threads.filter(({status}) => status === Thread.STATUS_DONE));

this.runtime.threads =
     this.runtime.threads.filter(({status}) => status !== Thread.STATUS_DONE);
```

(Try running through the loop in your head if you need to convince yourself of this!)

So let's rewind to `activeThreads`. Remember that it's set to true if any thread's status *isn't* `STATUS_DONE`.

Now, here's the `while` loop at the very top level of `stepThreads`:

```js
while (this.runtime.threads.length > 0 &&
       activeThreads &&
       /* extra conditions */) {...}
```

What I've been wanting to figure out in the review of this commit is, is it necessary to check `this.runtime.threads.length > 0` here *at all?* Or does `activeThreads` already contain that information for us? Assuming that this loop always iterates at least once (for the moment).

Well the `this.runtime.threads.length = nextActiveThread` here kind of gives it away. It and `activeThreads` sure *sound* synonymous. Is that so? Yes — `threads.length` will be non-zero iff any thread is not `STATUS_DONE`; just the same for `activeThreads`.

Note that this filter runs *after* we've iterated over all active threads - it's the last step in the top-level `stepThreads` loop. If it were somehow possible for a thread's status to be changed to `STATUS_DONE`, *after* it was the one responsible for `activeThreads` becoming true, then there would be trouble — this filter could resolve to zero (no active threads) while `activeThreads` remains true (because nothing sets it to false!). But there just isn't any opportunity for a thread's status to become `STATUS_DONE` — the only thing setting its status to `STATUS_DONE` is inside `stepThread(activeThread)`.

So yes, `this.runtime.threads.length > 0` really does become synonymous with `activeThreads` after the first iteration.

<hr>

Now for the takeaway! To simplify this logic, how about we only check `this.runtime.threads.length > 0` on the first iteration? A simple way to do this would be to just change the initial definition for `activeThreads`:

```diff
-let activeThreads = true;
+let activeThreads = this.runtime.threads.length > 0;
```

Et voila: the `while` loop will still neatly not perform its first iteration, because `activeThreads` was appropriately false, and we skip ahead.

But what exactly are we skipping *to?* Well, we set `this.activeThread = null` and return `doneThreads` below - that's everything. If it's possible, it would be cleaner to simply use an early return. Here's the naive example:

```js
if (this.runtime.threads.length === 0) {
    this.activeThread = null;
    return doneThreads;
}
```

That would work, but we'd like to determine the results of those two steps if the `while` loop was outright skipped - if we do, we might be able to simplify this early return, or move it higher up in the function (instead of right before the main `while` loop).

So for the first line: `this.activeThread = null` is purely a clean-up line. In other words it guarantees that `activeThread` is null once `stepThreads` exits, and so only has a value for the duration of `stepThreads`.

But `stepThreads` is also the only code *setting* `activeThread`... so we know that if it *was* null *before* any code evaluated, it doesn't need to be reset to null during the early return either.

```js
if (this.runtime.threads.length === 0) {
    return doneThreads;
}
```

*(Note: `activeThread` was outright made into a local variable later, with the mentioned cleanup removed. Of course, cleanup remains unneeded in the early return.)*

What's that `doneThreads` value going to be? Not much surprise here: it's just going to be an empty array! The only code which pushes to it is inside the main `while` loop, so if the `while` loop is skipped, nothing gets put in it.

```js
if (this.runtime.threads.length === 0) {
    return [];
}
```

This can be moved *all the way to the top* of `stepThreads`. (Or past the side effects `this.timer.start()` and `this.runtime.updateCurrentMSecs()`, if those have any impact outside of `stepThreads`. And only if!)

An early return for the "no active threads" case means we *know* we're going to run the thread loop at least once. So the final clean-up I'd do (while keeping the general structure of `stepThreads` the same) is to flat-out replace the `while` loop with a `do`-`while`:

```js
let activeThreads;
do {
    activeThreads = false;
    /* all the usual logic */
} while (activeThreads &&
         this.timer.timeElapsed() < WORK_TIME &&
         (this.runtime.turboMode || !this.runtime.redrawRequested))
```

Yes, it's true — `do`-`while` loops *are* spooky scary skeletons. But it really does make sense to move this conditional to the end, even beyond the semantic fact that it's "fine" to skip it the very first time — see, the only case where the conditional could *possibly* be true is *after* the some an iteration is complete! That's clear as much when you actually read the conditional parts:

* There only might *not* be `activeThreads` after we've evaluated all the were-active threads and they turned out to become done.
* The work timer could only possibly have elapsed if, you know, we've actually done any work.
* A redraw could only have been requested if some block requested it... as a result of... stepping a thread!

I haven't read ahead so who knows how much `stepThreads` is going to be restructured further. But those are my suggestions based on its current state, and thinking of how I'd refactor if I were the one doing the PR!

towerofnix (commented):
This is truly fantastic work. I've left a few comments as I'm going through this - since I left a longer commentary on `Sequencer.stepThreads` suggesting further refactoring, I figured I'd share that now before continuing through more.

src/engine/execute.js
```diff
@@ -500,65 +399,49 @@ const execute = function (sequencer, thread) {
...
+            // Resume thread after the promise resolves
+            const generation = thread.generation;
+            primitiveReportedValue
+                .catch(rejectionReason => {
+                    // Promise rejected: the primitive had some error.
+                    // Log it and proceed.
+                    log.warn('Primitive rejected promise: ', rejectionReason);
+                    // Return an empty string
+                    return '';
+                }).then(resolvedValue => {
```
FurryR: `catch` on PromiseLike seems wrong.

p.s.: Turbowarp also had this issue before.

towerofnix: Nothing new in this pull request, it's code that exists in the existing repo: https://github.com/scratchfoundation/scratch-vm/blob/16630ee9dd11b8b939c1a5db1c7ee991cb21b058/src/engine/execute.js#L138-L144

Almost certainly out of scope for this pull request to address.

FurryR: It's not.

PromiseLike (`isPromise`) only provides you `then` with `onfulfilled` and `onrejected` parameter, if you are using `catch` method, you are assuming that the PromiseLike will only be `Promise` without assertions.

Original code didn't use `catch`.

towerofnix: Oh, thank you for the correction, that's a good point.

Note [`isPromise`](https://github.com/scratchfoundation/scratch-vm/blob/16630ee9dd11b8b939c1a5db1c7ee991cb21b058/src/engine/execute.js#L31-L37) is defined in this file, but it probably shouldn't be altered, because it is the definition for how block modules are expected to provide promises. Changing `isPromise` could break compatibility with existing extensions (whether part of official scratch-vm or not) and it probably isn't worth it.

adroitwhiz: Fixed this by replacing it with the 2-argument `then` syntax.

PR #3933, opened 2023-03-17
feat: use scratchFetch

cwillisf:
**DO NOT MERGE THIS YET**

These changes require server-side support and should only be merged after verifying that support is ready.

### Resolves

Implements part of ENA-245

- [x] Depends on LLK/scratch-storage#404. ~This branch will need a `package.json` change once that PR is merged. Until then, CI builds of this branch will fail.~

### Proposed Changes

- Use the new `scratchFetch` from `scratch-storage`.
- Set the `RunId` metadata every time the project starts, stops, or loads

### Reason for Changes

The concept of a "run" is a little messy in Scratch. In many cases it could just mean that the user pressed the green flag: we could catch that by regenerating the run ID in `greenFlag()` and be done with it.

It's possible for projects to run code in other ways, though. One way is with a `when timer >` block, especially `when timer > 0`. That can cause the project to run immediately on load, without the need to press the green flag. It's similarly possible to restart the project immediately after the "stop" button is pressed.

In the end, I made a judgment call regarding when we want to regenerate the run ID. We can adjust in the future if necessary. One other good candidate is to regenerate the run ID any time we go from "no user threads" to "some user threads." We'd need to be careful to exclude "system" threads from that check, like threads responsible for monitors.

### Test Coverage

Added `test/integration/runId.js` to verify that the run ID changes when expected and never repeats.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -18,6 +18,7 @@ const StageLayering = require('./stage-layering');
 const Variable = require('./variable');
 const xmlEscape = require('../util/xml-escape');
 const ScratchLinkWebSocket = require('../util/scratch-link-websocket');
+const fetchWithTimeout = require('../util/fetch-with-timeout');
 
 // Virtual I/O devices.
 const Clock = require('../io/clock');
@@ -1627,6 +1628,7 @@ class Runtime extends EventEmitter {
      */
     attachStorage (storage) {
         this.storage = storage;
+        fetchWithTimeout.setFetch(storage.scratchFetch.scratchFetch);
         this.resetRunId();
     }
 
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -2,8 +2,6 @@ const EventEmitter = require('events');
 const {OrderedMap} = require('immutable');
 const uuid = require('uuid');
 
-const {setMetadata, RequestMetadata} = require('scratch-storage/src/scratchFetch.js');
-
 const ArgumentType = require('../extension-support/argument-type');
 const Blocks = require('./blocks');
 const BlocksRuntimeCache = require('./blocks-runtime-cache');
@@ -1629,6 +1627,7 @@ class Runtime extends EventEmitter {
      */
     attachStorage (storage) {
         this.storage = storage;
+        this.resetRunId();
     }
 
     // -----------------------------------------------------------------------------
@@ -2024,8 +2023,13 @@ class Runtime extends EventEmitter {
      * Reset the Run ID. Call this any time the project logically starts, stops, or changes identity.
      */
     resetRunId () {
+        if (!this.storage) {
+            // see also: attachStorage
+            return;
+        }
+
         const newRunId = uuid.v1();
-        setMetadata(RequestMetadata.RunId, newRunId);
+        this.storage.scratchFetch.setMetadata(this.storage.scratchFetch.RequestMetadata.RunId, newRunId);
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/engine/runtime.js
```diff
@@ -2015,6 +2020,14 @@ class Runtime extends EventEmitter {
         }
     }
 
+    /**
+     * Reset the Run ID. Call this any time the project logically starts, stops, or changes identity.
+     */
+    resetRunId () {
+        const newRunId = uuid.v1();
```
adroitwhiz, line 2035: Instead of adding a dependency on `uuid`, why not just use `util/uid`?

cwillisf: I actually used `util/uid` in an earlier version :)

I made the change based on a request from another member of the Scratch team. I think using a standard data format gives us a little more flexibility. Also, the `uuid` module was already in our dependency tree, just not at the top level.

deepankarmalhan: Did we pull it into the top level to make it explicit then? It might be worth doing that if not so we can lock versions of it properly and it's good code health thing to do anyways for explicit documentation and maintenance. @cwillisf

deepankarmalhan: Nvm, just saw the code change that makes it explicit in `package.json`, ignore me!

src/engine/runtime.js
```diff
@@ -1624,6 +1628,8 @@ class Runtime extends EventEmitter {
      */
     attachStorage (storage) {
         this.storage = storage;
+        fetchWithTimeout.setFetch(storage.scratchFetch.scratchFetch);
```
deepankarmalhan, line 1633: Where is `storage.scratchFetch.scratchFetch` defined and why are we doing this?

src/util/fetch-with-timeout.js
```diff
@@ -1,3 +1,25 @@
...
+ */
+let myFetch = global.fetch;
+
+/**
+ * Tell `fetchWithTimeout` to use a specific `fetch` function.
+ * By default, `fetchWithTimeout` will use the global `fetch` function.
+ * If there is no global `fetch`, then `fetchWithTimeout` will fail unless provided with an alternative.
+ * @param {FetchFunction} newFetch The new `fetch` function to use within fetchWithTimeout.
+ */
+const setFetch = newFetch => {
```
deepankarmalhan, line 19: Why do we need to set a fetch function? What cases are we not using the original fetch method for? And should it be a consistent way to make fetch requests across the codebase?

deepankarmalhan (approved):
A few random questions but other than that looks good to me!

src/engine/runtime.js
cwillisf: `scratchFetch` is a new export from `scratch-storage`
- defined here: https://github.com/LLK/scratch-storage/blob/develop/src/scratchFetch.js
- exported here: https://github.com/LLK/scratch-storage/blob/develop/src/ScratchStorage.js#L54-L61

This line tells `fetchWithTimeout` to use `scratchFetch` instead of the global `fetch`. They have the same API, but `scratchFetch` has new features related to metadata headers.

More info about `scratchFetch`:
- LLK/scratch-storage#404
- LLK/scratch-storage#411

src/util/fetch-with-timeout.js
cwillisf: See above, but also...

> And should it be a consistent way to make fetch requests across the codebase?

Yep! We're starting with a few specific things, but we intend to eventually use this for all HTTP requests in the Scratch editor.

deepankarmalhan: That makes sense, thanks for clarifying!

src/engine/runtime.js
deepankarmalhan: Ohhh ok yeah that makes sense, thanks for the context!

PR #4026, opened 2023-06-01
Elliotbroe/bxc 114 changegraphiceffectby color bug

ElliotRoe:
### Resolves

[BXC-114](https://linear.app/bxcoding/issue/BXC-114/changegraphiceffectby-color-bug)


ElliotRoe:
Accidental sorry

PR #4052, opened 2023-08-07
fix: remove use of 'globalThis' keyword

cwillisf:
### Proposed Changes

Replaces checks that use `globalThis` with slightly more careful checks using `self`.

### Reason for Changes

Some of our supported configurations don't support `globalThis`

### Test Coverage

Tested locally in Chrome on Windows and Safari on macOS, plus the existing tests.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1443,17 +1443,19 @@ class Runtime extends EventEmitter {
      * One-time initialization for Scratch Link support.
      */
     _initScratchLink () {
-        /* global globalThis */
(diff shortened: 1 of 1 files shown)


aoneill01 (approved):
I like the clarifying comments.

cwillisf:
Sorry, I re-read the code after your review and noticed that I accidentally removed the check for a null/undefined origin in the first round. The `commitlint` error is spurious - it doesn't like [force-pushes](https://i.imgur.com/FWJtwXm.mp4).

PR #4085, opened 2023-09-27
Migrate CI/CD to GitHub Actions

cwillisf:
### Resolves

Resolves ENG-14

### Proposed Changes

- Migrate our automated builds to GitHub Actions
- Merge the `build_no_release` and `build_and_release` workflows
- Replace the `gh-pages` tool with the `peaceiris/actions-gh-pages@v3` action
- Replace the `conventional-changelog/commitlint` orb with the `wagoid/commitlint-github-action@v5` action

### Reason for Changes

The only difference between the `build_no_release` and `build_and_release` workflows was whether or not `semantic-release` was called with the `--dry-run` parameter. It turns out that `semantic-release` can decide correctly on its own, based on its configuration file, whether or not to release from the branch in question, so there's no need for this split.

CCI checks out the repository using SSH, but GHA checks out the repository using HTTPS. It turns out that complicates how we used `gh-pages` to publish the build to GH Pages. I tried a bunch of things and couldn't get `gh-pages` to work: it always failed to authenticate. I replaced it with `peaceiris/actions-gh-pages@v3` and suddenly life was peaceful. This also let me remove `gh-pages` from our dependency list.

### Test Coverage

See the "Actions" tab of this repository :)

Note that I temporarily enabled the `update-18n` workflow on `push` for testing purposes. It seems that `workflow_dispatch` doesn't work until the workflow is present in a default branch, or... something.


delasare (approved):
This looks good to me! :shipit:

PR #4092, opened 2023-10-02
fix: disable metadata for now

cwillisf:
### Proposed Changes

Don't set metadata for now.

### Reason for Changes

Effectively backs out the storage changes without relying on `scratch-storage` version 2.3.0 or higher. Those newer versions are causing build problems with `scratch-gui` so this is a very temporary workaround.

### Test Coverage

Metadata tests have been disabled since they don't make sense when we're not setting metadata. Otherwise, coverage remains the same.

### DIFF
--- a/src/engine/runtime.js
+++ b/src/engine/runtime.js
@@ -1,6 +1,6 @@
 const EventEmitter = require('events');
 const {OrderedMap} = require('immutable');
-const uuid = require('uuid');
+// const uuid = require('uuid');
 
 const ArgumentType = require('../extension-support/argument-type');
 const Blocks = require('./blocks');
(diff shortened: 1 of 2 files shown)


cwillisf:
I'm not sure why tests are failing here. I'll try to figure it out later today.

EDIT: and by "I'm not sure why," of course I mean that the reason is obvious but I was in too much of a hurry to read the test output correctly. :/

Anyway, it's "fixed" now in that the metadata tests are now skipped since they don't make sense if we're not setting any metadata.

AlexisGoodfellow (approved):
MAKE IT SO :D

colbygk (approved):
LGTM

PR #4135, opened 2023-12-02
Update index.js

LeoKids:
Adding several voices that's yet to be implemented.

Sorry if it looks poor,i'm bad at coding ._.

### Resolves

### Proposed Changes

Adds more voices than the launch voices can offer

### Reason for Changes

There are no new voice since its launch,By adding new voices it opens more possibilities and can offer more if they do not like the voices,for example,Squeaky people don't speak like female,Cats are deeper than Kitten's voice and Giants should not be limited to male voices.

In the future,we can do the squeak split which uses the child voices Amazon offers,but this will only be in english as others languages do not have child voices.

### Test Coverage

N/A

### DIFF
--- a/src/extensions/scratch3_text2speech/index.js
+++ b/src/extensions/scratch3_text2speech/index.js
@@ -147,7 +147,7 @@ class Scratch3Text2SpeechBlocks {
                 name: formatMessage({
                     id: 'text2speech.alto',
                     default: 'alto',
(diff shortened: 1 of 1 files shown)


LilyMakesThings:
The ambiguous gender comments were probably there for a reason.

LeoKids:
Any news?

PR #5426, opened 2025-05-29
Fix crash when tts language is not a string

husqwc:
### Resolves

Setting the tts language to something other than a string causes a crash

For example, in this script, the sprite never says meow:
block_5_29_2025-6_07_37 PM

### Proposed Changes

a .toString() has been added to fix the crash

### Reason for Changes

.toLowerCase() no longer crashes, and there is always an output for .toString() in JS

### Test Coverage

No tests so far


husqwc:
I have read the CLA Document and I hereby sign the CLA

PR #5427, opened 2025-05-31
Fix ifOnEdgeBounce buggy behavior

TheJupiterDev:
### Proposed Changes

Fixes an issue where `ifOneEdgeBounce` clamps very small direction vectors to `0.2`. This causes major issues in many Scratch programs.

### Reason for Changes

Allow scratch projects to properly use If On Edge Bounce blocks without it bugging out.

### DIFF
--- a/src/blocks/scratch3_motion.js
+++ b/src/blocks/scratch3_motion.js
@@ -183,7 +183,7 @@ class Scratch3MotionBlocks {
         }
     }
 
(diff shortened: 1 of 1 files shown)


TheJupiterDev:
I have read the CLA Document and I hereby sign the CLA

husqwc:
This is extremely unlikely to be merged as it breaks compatibility with previous projects that used the block

